PR #66, opened 2024-04-15
Employee Avatar

dani-moreno:


lib/components/employee-avatar.tsx
```diff
@@ -0,0 +1,67 @@
...
+      </HoverCardTrigger>
+      <HoverCardContent
+        side="top"
+        sideOffset={12}
+        align="start"
+        className="w-80 rounded-2xl slide-in-from-right-2"
+      >
+        <div className="flex flex-col gap-4">
+          <div className="flex items-center gap-4">
+            <div className="flex-none">
```
Rogermax, line 47: Just a general question here: Should we use in this basic components Split/Stack instead of (div_flex_col/div_flex_row) ?

PR #175, opened 2024-05-31
Does not expose all input properties and add form as xray type

edgarlatorre:
## 🚪 Why?

Right now the input component is just forwarded from shadcn

## 🔑 What?
In this first PR I'm changing it to just expose the Input with some props. Apart from that I removed the ring styles for the component and I also added `form` as a new component type in order to xray have an specific color for forms components.

https://github.com/factorialco/factorial-one/assets/52708/9e521a27-dea7-4c64-9ef1-a9c746058b53

### DIFF
--- a/lib/components/Forms/Input/index.tsx
+++ b/lib/components/Forms/Input/index.tsx
@@ -1,1 +1,19 @@
 export * from "@/ui/input"
+import { Component } from "@/lib/component"
+import { Input as ShadcnInput } from "@/ui/input"
+import { ComponentProps } from "react"
+
+const Input: React.FC<
+  Pick<
+    ComponentProps<typeof ShadcnInput>,
+    "type" | "disabled" | "size" | "onChange" | "value" | "placeholder"
+  >
+> = Component(
+  {
+    name: "Input",
+    type: "form",
+  },
+  ShadcnInput
+)
+
+export { Input }
--- a/lib/lib/component.tsx
+++ b/lib/lib/component.tsx
@@ -3,7 +3,7 @@ import { cn } from "@/lib/utils"
 import { forwardRef } from "react"
 import { useComponentXRay } from "./xray"
 
-export const componentTypes = ["layout", "info", "action"] as const
+export const componentTypes = ["layout", "info", "action", "form"] as const
 export type ComponentTypes = (typeof componentTypes)[number]
 
 export interface ComponentMetadata {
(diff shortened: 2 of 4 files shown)


lib/ui/input.tsx
```diff
@@ -10,7 +10,7 @@ const Input = React.forwardRef<HTMLInputElement, InputProps>(
       <input
         type={type}
         className={cn(
-          "flex h-10 w-full rounded-lg border-2 border-solid border-input bg-background px-3 py-2 text-sm ring-offset-background transition-colors file:border-0 file:bg-transparent file:text-sm file:font-medium placeholder:text-muted-foreground/60 hover:border-input-hover focus-visible:border-input-hover focus-visible:outline-none focus-visible:ring-2 focus-visible:ring-ring focus-visible:ring-offset-2 disabled:cursor-not-allowed disabled:bg-muted disabled:opacity-50",
```
josepjaume, line 13: I think `focus-visible` and all of this should stay for accessibility purposes. We need this for keyboard navigation. I think there's workarounds to prevent this behavior when using a mouse, but we would need to find a general solution to this.

lib/lib/component.tsx
```diff
@@ -3,7 +3,7 @@ import { cn } from "@/lib/utils"
 import { forwardRef } from "react"
 import { useComponentXRay } from "./xray"
 
-export const componentTypes = ["layout", "info", "action"] as const
+export const componentTypes = ["layout", "info", "action", "form"] as const
```
josepjaume, line 6: Awesome! X-Ray should eventually create a tiny interface that allows us to toggle and filter what we want to see, so this makes sense.

lib/components/Forms/Input/index.tsx
```diff
@@ -1 +1,19 @@
 export * from "@/ui/input"
+import { Component } from "@/lib/component"
+import { Input as ShadcnInput } from "@/ui/input"
+import { ComponentProps } from "react"
+
+const Input: React.FC<
+  Pick<
+    ComponentProps<typeof ShadcnInput>,
+    "type" | "disabled" | "size" | "onChange" | "value" | "placeholder"
```
josepjaume, line 9: Nice!

lib/ui/input.tsx
edgarlatorre: Okay I just reverted this change.

PR #227, opened 2024-06-17
Add text area

edgarlatorre:
## 🚪 Why?

In order to use all components from factorial-one in the changelogger we are missing the text area.

## 🔑 What?

Add Textarea field.

## 📸 Screenshots

image


josepjaume (approved):
Nice!

PR #296, opened 2024-07-15
Add insights boilerplate

josepjaume:
Adds the boilerplate for insights following [ShadCN's incredibly timely implementation](https://ui.shadcn.com/charts).


lib/ui/chart.tsx
```diff
@@ -0,0 +1,362 @@
...
+  HTMLDivElement,
+  React.ComponentProps<"div"> & {
+    config: ChartConfig
+    children: React.ComponentProps<
+      typeof RechartsPrimitive.ResponsiveContainer
+    >["children"]
+  }
+>(({ id, className, children, config, ...props }, ref) => {
+  const uniqueId = React.useId()
+  const chartId = `chart-${id || uniqueId.replace(/:/g, "")}`
```
rodrigo-picanco, line 45: This might be super silly, but a substring is supposed faster here:

```suggestion
  const chartId = `chart-${id || uniqueId.replace(":", "")}`
```

rodrigo-picanco:
Like the usage of the boilerplate components. Pretty straightforward.

I guess the Area chart here is more of an example than an implementation, right?

josepjaume:
@rodrigo-picanco totally, it's just an example I want to keep refining. More like a proof of concept. Actually, I might just go ahead and add a bit more typing to it, maybe even a generic?

I'm starting to like having "low-level" internal components, and then exposing "higher order" ones. Makes for a flexible design system with strong boundaries.

josepjaume:
Turned this into a draft!

josepjaume:
@rodrigo-picanco improved this to use generics! Not 100% happy, but I'd wait until we have more use cases :)

PR #309, opened 2024-07-17
Add implicit loading state to buttons

josepjaume:
We often want a button to get disabled while we wait for an action to complete. Often times, we implement this by using a hook's "loading" state and passing it to the "disabled" property of the button.

This is error prone (we can forget about it) and inconsistent (people can implement this many different ways).

This introduces an extra behavior to buttons where they will handle this internally. If you pass a promise to their `onClick` handler, they will get disabled in the meanwhile. Extra useful and one less piece of UI state we need to care about!

### DIFF
--- a/lib/components/Actions/Button/index.stories.tsx
+++ b/lib/components/Actions/Button/index.stories.tsx
@@ -61,7 +61,16 @@ export const WithIcon: Story = {
   },
 }
 
+export const Disabled: Story = {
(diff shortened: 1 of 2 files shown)


Rogermax (approved):
Ohhh never seen that! Nice to know!

rubenmoya (approved):
😍

PR #313, opened 2024-07-18
Splits charts into their own section

josepjaume:
This splits charts into their own section for composability, maintainability and clarity.


dani-moreno (approved):
Yaaay 🚀

PR #316, opened 2024-07-19
Dialog gammafied

dani-moreno:
### Context
In order to use One components in Factorial, we're making them mirror (or as close as possible) our current Gamma components.

This PR changes a few things about how the Dialog looks, to make it look like the Gamma Modal component.

### Screenshot
image
_Note: The font is not the same one, so anyone with a keen design eye might spot a few discrepancies. We're rolling out Inter (the font we're using in Factorial) in One for a uniform look, but that tweak's not reflected in this PR just yet._

### DIFF
--- a/lib/components/Overlays/Dialog/index.tsx
+++ b/lib/components/Overlays/Dialog/index.tsx
@@ -59,7 +59,7 @@ const OneDialog = forwardRef<HTMLDivElement, DialogProps>(
             <DialogHeader>
               {header.icon && (
                 <DialogIcon>
-                  <header.icon size={24} />
+                  <header.icon size={32} strokeWidth={1.5} />
                 </DialogIcon>
               )}
               <DialogTitle>{header.title}</DialogTitle>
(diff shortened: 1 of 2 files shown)


lib/components/Overlays/Dialog/index.tsx
```diff
@@ -59,7 +59,7 @@ const OneDialog = forwardRef<HTMLDivElement, DialogProps>(
             <DialogHeader>
               {header.icon && (
                 <DialogIcon>
-                  <header.icon size={24} />
+                  <header.icon size={32} strokeWidth={1.5} />
```
CarlosBM18, line 62: Should we use tokens for these numbers?

CarlosBM18 (approved):
Nice!

lib/components/Overlays/Dialog/index.tsx
dani-moreno: Umm, I guess we won't be using Lucide icons in the near future, but rather our own icons. Since we're yet to incorporate them into One, I've put together an example to give a taste of how an icon should shape up.

PR #317, opened 2024-07-19
Use inter font by default

josepjaume:
Uses the inter font and ships it with the project so it can readily be used.


dani-moreno (approved):
My eyes can rest! 🙏

PR #318, opened 2024-07-19
Line Chart & Insight

dani-moreno:
### Context
Made a new **Line chart** and **Line insight** (mirroring @josepjaume's approach with the area chart) just to get the hang of how they're put together.

This is more of a learning exercise for me, but the chart component is already up and running 🏃 

### Screenshot
image

### DIFF
--- a/lib/components/Charts/LineChart/index.tsx
+++ b/lib/components/Charts/LineChart/index.tsx
@@ -1,4 +1,9 @@
-import { ChartContainer, ChartTooltip, ChartTooltipContent } from "@/ui/chart"
+import {
+  ChartConfig,
+  ChartContainer,
+  ChartTooltip,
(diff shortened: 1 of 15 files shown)


lib/components/Charts/LineChart/index.tsx
```diff
@@ -0,0 +1,98 @@
...
+
+type AxisConfig = {
+  hide?: boolean
+  tickFormatter?: (value: string) => string
+}
+
+export type LineChartConfig<Keys extends string = string> = Record<
+  Keys,
+  ChartConfig[string]
+>
```
josepjaume, line 29: We can probably reuse this type across charts!

dani-moreno: Creating a common type, right? Let me try it!

PR #333, opened 2024-07-23
Export Textarea component

rodrigo-picanco:
## 🚪 Why?
Was fiddling around and saw we're not exporting the textarea component. 

## 🔑 What?
Exports the textarea component


josepjaume (approved):
Nice catch!

PR #335, opened 2024-07-24
feat: bar chart

elacallefactorial:
## 🚪 Why?

Factorial one doesn't have a Bar component in order to use it with other charts components

## 🔑 What?

This PR adds a Bar chart to draw Bar components and Lines in the insights components and data graph component; this Bar chart has the following interface:

- **dataConfig** Configures each data point with a label and colour
- **xAxis** 
  - **hide** Hides the X-axis
  - **formatter** Arguments that receive a function to customize the X-axis labels
- **lines** Draw a line over the Bar components
- **data** required data by the Bar component to draw the Bars and Lines
- **yAxis** 
  - **hide** Hides the X-axis
  - **formatter** Arguments that receive a function to customize the X-axis labels

## 🏡 Context

https://www.figma.com/design/9pLsRzdinid0ha0qRWta2D/Employee-Profile?node-id=378-19759&t=cFvdNpMxrdkyqDqn-0

## 📸 Visual proof

Screenshot 2024-07-24 at 11 38 39

### DIFF
--- a/lib/components/Charts/BarChart/index.stories.tsx
+++ b/lib/components/Charts/BarChart/index.stories.tsx
@@ -1,0 +1,50 @@
+import { Meta, StoryObj } from "@storybook/react"
+import { BarChart } from "."
+
+const dataConfig = {
+  desktop: {
+    label: "Desktop",
+  },
+  mobile: {
+    label: "Mobile",
+  },
+}
+
+const Component = BarChart<typeof dataConfig>
+
+const meta: Meta<typeof Component> = {
+  component: Component,
+  title: "Charts/BarChart",
+  args: {},
+} satisfies Meta<typeof Component>
+
+export default meta
+
+type Story = StoryObj<typeof meta>
+
+export const Default: Story = {
+  args: {
+    dataConfig: {
+      desktop: {
+        label: "Desktop",
+        color: "hsl(var(--chart-1))",
+      },
+      mobile: {
+        label: "Mobile",
+        color: "hsl(var(--chart-2))",
+      },
+    },
+    xAxis: {
+      hide: false,
+    },
+    lines: true,
+    data: [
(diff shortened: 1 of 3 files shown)


dani-moreno:
Hey, good job @elacallefactorial, looking good!

If you take a look at the [recharts Bar API](https://recharts.org/en-US/api/Bar), you'll see that there are a few props that may be worth adding. Especially `layout`, and probably `label`.

lib/components/Charts/BarChart/index.stories.tsx
```diff
@@ -0,0 +1,50 @@
...
+      },
+      mobile: {
+        label: "Mobile",
+        color: "hsl(var(--chart-2))",
+      },
+    },
+    xAxis: {
+      hide: false,
+    },
+    lines: true,
```
josepjaume, line 40: Is that something we really need? From the screenshot, it looks a bit redundant.

lib/components/Charts/utils/bar.ts
```diff
@@ -0,0 +1,5 @@
+import { ChartItem } from "./types"
+
+export function prepareData<Keys extends string>(data: ChartItem<Keys>[]) {
+  return data.map((item) => ({ x: item.label, ...item.values }))
+}
```
josepjaume, line 5: Could we perhaps use this function on the `LineChart` component as well so it's not duplicated?

josepjaume (approved):
Requested a couple changes but preemptively approving.

josepjaume:
Nice job! Could you maybe consider adding an equivalent `BarChartInsight` just like the rest? Should be a handful of lines. Also please remember to add it on their `exports.ts` file, as it's what we actually export (the rest stays private).

elacallefactorial:
> recharts Bar API

Hi @dani-moreno I've created other PR with this feature https://github.com/factorialco/factorial-one/pull/341

PR #336, opened 2024-07-24
Revamp icons

josepjaume:
Revamps icons again to use a similar approach to `lucide` which exports each icon individually as a component instead.


RamProg (commented):
I see all svgs are now React components, rather than simple .svg What is the benefit we expect to get from converting them like this? Usually when you have to work with a design you get an SVG and it's an extra task to convert it from pure SVG to a react component.

josepjaume:
Hey @RamProg the problem lies with bundling and shipping the artifacts. We need to be able to ship these icons in a way that they're usable from any consumers, and we can't trust that the consumer will know how to import an SVG and transform it to a react component behind the scenes (which is what always ends happening).

In this case, we do that work for you so you don't have to add a plugin like [vite-plugin-svgr](https://www.npmjs.com/package/vite-plugin-svgr).

Also, there's a few extra steps we are doing there:

* Cleaning up the svg with the svg optimizer (could be done when exporting but still)
* More importantly, we're replacing the colors with `currentColor` so we can change them via the current font color.

josepjaume:
That said - I just pushed a commit adding the source `svg` into the package, so you still have access to them via:

```
import AddAvatar from "@factorialco/factorial-one/assets/icons/AddAvatar.svg"
```

dani-moreno (approved):
Yay! This change may seem small at first but it's huge for our workflow!

PR #341, opened 2024-07-25
feat: bar chart horizontal layout

elacallefactorial:
## 🚪 Why?

The first version of Bar chat didn't have a setting to configure the bar layout

## 🔑 What?

This PR adds the `layout` property with two available options:

**vertical**: This property draws the Bar chart component in vertical direction  

**horizontal**: This property draws the Bar chart component in horizontal direction  

## 📸 Visual proof

Screenshot 2024-07-25 at 15 50 49

### DIFF
--- a/lib/components/Charts/BarChart/index.stories.tsx
+++ b/lib/components/Charts/BarChart/index.stories.tsx
@@ -1,0 +1,50 @@
+import { Meta, StoryObj } from "@storybook/react"
+import { BarChart } from "."
+
+const dataConfig = {
+  desktop: {
+    label: "Desktop",
+  },
+  mobile: {
+    label: "Mobile",
+  },
+}
+
+const Component = BarChart<typeof dataConfig>
+
+const meta: Meta<typeof Component> = {
+  component: Component,
+  title: "Charts/BarChart",
+  args: {},
+} satisfies Meta<typeof Component>
+
+export default meta
+
+type Story = StoryObj<typeof meta>
+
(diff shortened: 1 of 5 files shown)


josepjaume:
Cool! This took a surprising amount of fiddling, huh?

That said - I see there's an example on ShadCN's site with a horizontal chart that doesn't seem to need that much data mangling. I wonder if we should maybe have this as two separate charts instead - one that's horizontal, and another one that's vertical. I don't think we plan on allowing users to just change layouts, right? Even, so it could be done with different components. Here's how it looks:

image

You can find it on shad's page:

https://ui.shadcn.com/charts

elacallefactorial:
> Cool! This took a surprising amount of fiddling, huh?
> 
> That said - I see there's an example on ShadCN's site with a horizontal chart that doesn't seem to need that much data mangling. I wonder if we should maybe have this as two separate charts instead - one that's horizontal, and another one that's vertical. I don't think we plan on allowing users to just change layouts, right? Even, so it could be done with different components. Here's how it looks:
> 
> image
> You can find it on shad's page:
> 
> https://ui.shadcn.com/charts

Hi @josepjaume,

I wasn't sure how to tackle this layout issue. I think a single component should not duplicate the Bar chart code. However, handling two layouts in the same code increases its complexity.

Creating another component to render the Bar chart with a horizontal layout is easier than managing complex logic in a single component. If the second option is acceptable, I will create another Bar chart component.

PR #345, opened 2024-07-26
fix: remove Line component inside the Bar chart

elacallefactorial:
## 🚪 Why?

We don't need to paint lines in the Bar chart component

## 🔑 What?

This PR deletes the Line component inside the Bar chart component

## 🏡 Context

https://factorialteam.slack.com/archives/C07C80SHZL6/p1721981311282789?thread_ts=1721980570.989029&cid=C07C80SHZL6


sauldom102 (approved):
🧹

PR #349, opened 2024-07-26
feat: vertical bar chart 

elacallefactorial:
## 🚪 Why?

The first version of Bar chat didn't have a setting to configure the bar layout 

## 🔑 What?

In this PR, we have extracted the Bar component into a generic BaseBarChart component. This new component contains the common code shared between the two types of bars. 

Each specific component, Bar and VBar, now has its own code to render the bars in horizontal and vertical views, respectively.

## 📸 Visual proof

Screenshot 2024-07-26 at 17 39 59


lib/components/Charts/VBarChart/index.tsx
```diff
@@ -0,0 +1,74 @@
...
+      layout="vertical"
+      xAxisComponent={<XAxis {...xAxis} hide={config.xAxis?.hide} />}
+      yAxisComponent={<YAxis {...yAxis} hide={config.yAxis?.hide} />}
+      ref={ref}
+      {...config}
+    />
+  )
+}
+
+export const VBarChart = fixedForwardRef(_VBarChart)
```
josepjaume, line 74: This might be more of a personal preference, but I don't like using shortcuts for things like this. Having `VerticalBarChart` is totally acceptable IMHO.

lib/components/Charts/BaseBarChart/index.tsx
```diff
@@ -0,0 +1,84 @@
...
+      </ComposedChart>
+    </ChartContainer>
+  )
+}
+
+_BaseBarChart.defaultProps = {
+  layout: "horizontal",
+}
+
+export const BarChart = fixedForwardRef(_BaseBarChart)
```
josepjaume, line 84: I feel we should somehow clarify that this is private. Maybe if we create a `BarChart` folder that contains both and this within with a `_` prepended? Also it might not need its own folder, as it doesn't have stories. That's the convention I'm generally pushing for for private classes, but maybe it's not the best.

josepjaume (commented):
Thanks for the work! I'm, however, skeptical that extracting common things into `BaseBarChart` is actually helping that much. I think it makes the code a bit more difficult to read, and it couples both implementations a bit too much perhaps? I can imagine a future where the two kinds of graphs evolve in slightly different ways and this is maybe rigidifying the system a bit too much.

In the end, we're already relying on primitives provided by shadcn which simplify the `rechart` API surface, so this feels like an unnecessary step.

That said, I don't have a strong opinion. What matters the most is that the public API makes sense and this seems to have been achieved.

lib/components/Charts/BaseBarChart/index.tsx
elacallefactorial: @josepjaumeI was concerned about repeating the same code in two places. Having duplicated code could become problematic if updates are needed in the future, as changes would have to be made in multiple locations. Therefore, I initially considered a middle ground between duplicating the code and creating a single component to manage the layout.

However, since there are currently only two components and adding complexity to the logic isn't ideal, I've decided to keep the bar components without any abstraction.

lib/components/Insights/Charts/VBarChartInsight/index.stories.tsx
```diff
@@ -0,0 +1,26 @@
+import type { Meta, StoryObj } from "@storybook/react"
+
+import BarChartStory from "@/components/Charts/VerticalBarChart/index.stories"
+import { VerticalBarChartInsight } from "."
+import { containerStoryArgs } from "../storybook-utils"
+
+const meta = {
```
josepjaume, line 7: You're gonna kill me for this, but the folder name is still the old one (VBarChartInsight) 👀

josepjaume (approved):
I'll approve it regardless :)

PR #355, opened 2024-07-29
Area chart enhancements

hossamElfarFactorial:
## 🔑 What?
- Make the footer optional
- Add Legend, color gradient, and axis tick count to `AreaChart`

## 📸 Visual proof

SCR-20240729-khgq

### DIFF
--- a/lib/components/Charts/AreaChart/index.tsx
+++ b/lib/components/Charts/AreaChart/index.tsx
@@ -1,4 +1,10 @@
-import { ChartContainer, ChartTooltip, ChartTooltipContent } from "@/ui/chart"
+import {
+  ChartContainer,
+  ChartLegend,
+  ChartLegendContent,
+  ChartTooltip,
+  ChartTooltipContent,
+} from "@/ui/chart"
 import { ForwardedRef } from "react"
 import {
   Area,
@@ -8,7 +14,7 @@ import {
   YAxis,
 } from "recharts"
 import { autoColor } from "../utils/colors"
-import { cartesianGridProps, xAxisProps, yAxisProps } from "../utils/elements"
+import { cartesianGridProps } from "../utils/elements"
 import { fixedForwardRef } from "../utils/forwardRef"
 import { prepareData } from "../utils/muncher"
 import { ChartConfig, ChartPropsBase, InferChartKeys } from "../utils/types"
@@ -18,6 +24,7 @@ export type AreaChartProps<
   Keys extends string = InferChartKeys<DataConfig>,
 > = ChartPropsBase<DataConfig, Keys> & {
   lineType?: "natural" | "linear" | "step"
+  showLegend?: boolean
 }
 
 export const _AreaChart = <
@@ -30,6 +37,7 @@ export const _AreaChart = <
     xAxis,
     yAxis,
     lineType,
+    showLegend,
   }: AreaChartProps<DataConfig, Keys>,
   ref: ForwardedRef<HTMLDivElement>
 ) => {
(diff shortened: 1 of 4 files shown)


lib/components/Charts/AreaChart/index.tsx
```diff
@@ -30,6 +37,7 @@ export const _AreaChart = <
     xAxis,
     yAxis,
     lineType,
+    showLegend,
```
josepjaume, line 40: Wondering if this should be optional or not. In the end, this design system is all about having opinions and reducing the amount of choices.

lib/components/Charts/AreaChart/index.tsx
```diff
@@ -43,22 +51,62 @@ export const _AreaChart = <
...
+          />
+        )}
         <ChartTooltip
           cursor
-          content={<ChartTooltipContent indicator="dot" />}
+          content={<ChartTooltipContent indicator="line" />}
         />
+        <defs>
+          {areas.map((area, index) => (
+            <linearGradient id={`fill${area}`} x1="0" y1="0" x2="0" y2="1">
```
josepjaume, line 78: Nice job!

lib/components/Charts/utils/elements.tsx
```diff
@@ -9,6 +9,7 @@ export const xAxisProps = (
   tickLine: false,
   axisLine: false,
   tickMargin: 8,
+  tickCount: config?.tickCount,
```
josepjaume, line 12: What's the use case for this? Do we really want to make it configurable?

lib/components/Insights/InsightsContainer/index.tsx
```diff
@@ -21,7 +21,7 @@ export interface InsightsContainerProps {
     title: string
     description: string
   }
-  footer: {
+  footer?: {
```
josepjaume, line 24: I guess it's fine if footer is optional? In any case this should change according to what @dani-moreno and Desirée end up with.

lib/components/Insights/InsightsContainer/index.tsx
```diff
@@ -33,22 +33,24 @@ export const InsightsContainer = forwardRef<
 >(({ header, footer, children }, ref) => (
   <Card ref={ref}>
     <CardHeader>
-      <CardTitle>{header.title}</CardTitle>
       <CardDescription>{header.description}</CardDescription>
+      <CardTitle>{header.title}</CardTitle>
```
josepjaume, line 37: Why the change of order?

josepjaume (commented):
Added a few comments! Also: Could you maybe add some more storybook examples to these charts to showcase these new properties?

lib/components/Charts/AreaChart/index.tsx
hossamElfarFactorial: Hmmm I agree, I found some examples at figma that we can use legend and some not, so I made it optional. But since the design is not final, I will always show it for now

lib/components/Charts/utils/elements.tsx
hossamElfarFactorial: We had this use case on the employees' page, we needed to add a chart in a small place that with many y points, caused the chart to look off.

lib/components/Insights/InsightsContainer/index.tsx
hossamElfarFactorial: It's based on the designs at figma, but I will revert it since the design is not yet final.

lib/components/Charts/AreaChart/index.tsx
hossamElfarFactorial: [always show area chart legend](https://github.com/factorialco/factorial-one/pull/355/commits/dac7b65721f0e2486892a1f1dfe55624601f7013)

lib/components/Charts/utils/elements.tsx
josepjaume: Okay! Should we default it to false, perhaps?

josepjaume (approved):

PR #356, opened 2024-07-29
Insights pie chart component

hossamElfarFactorial:
## 🔑 What?
- Insights pie chart component

## 📸 Visual proof
SCR-20240729-lies
SCR-20240729-lizj


lib/components/Charts/PieChart/index.tsx
```diff
@@ -0,0 +1,49 @@
...
+  return (
+    <ChartContainer config={dataConfig} ref={ref}>
+      <PieChartPrimitive accessibilityLayer margin={{ left: 12, right: 12 }}>
+        <ChartTooltip cursor content={<ChartTooltipContent hideLabel />} />
+        <Pie
+          nameKey={"label"}
+          dataKey={"value"}
+          data={preparedData}
+          innerRadius={donutPieChart ? 60 : 0}
+        />
```
josepjaume, line 43: Interesting that `Pie` has way fewer configuration overhead than the rest of the chart types!

josepjaume (approved):
Neat! Let's merge this. That said, could you maybe consider adding this as a type of insight as well? I'm still not sure we'll follow that pattern but as you can see, we're providing some "ready-to-use" insights already.

lib/components/Charts/PieChart/index.tsx
```diff
@@ -0,0 +1,49 @@
...
+  color?: string
+}
+
+export type PieChartProps<
+  DataConfig extends ChartConfig = ChartConfig,
+  Keys extends string = InferChartKeys<DataConfig>,
+> = {
+  dataConfig: ChartConfig<Keys>
+  data: PieChartItem[]
+  donutPieChart?: boolean
```
josepjaume, line 20: Do we really want to support both? I'm in favor of starting with fewer options and deciding later.

hossamElfarFactorial: I think we will need it, we have some use cases for using donut pie chart at some places.

lib/components/Insights/Charts/PieChartInsight/index.stories.tsx
```diff
@@ -0,0 +1,60 @@
...
+        { label: "june", value: 214 },
+      ],
+    },
+  },
+} satisfies Meta<typeof PieChartInsight>
+
+export default meta
+type Story = StoryObj<typeof meta>
+
+export const Default: Story = {}
```
josepjaume, line 60: I'm thinking that we might want to add a story for the pie version and another one for the donut version?

PR #360, opened 2024-07-31
Feat/fct 16897/new tab layout

RamProg:
## 🚪 Why?

We are working on a new Profile with a whole new layout. We want to start this Layout in Factorial One and potentially reuse it in different screens.

## 🔑 What?

This Layout has a breadcrumb, a header, a tab navigation system and a content. The content will change when a different tab is selected. The Layout should be able to know how to handle this behaviour without including any content within itself, so it should be smart enough to navigate, but without needing to know what's inside.

Responsiveness is not required **yet**. This is my first PR on F1 so please feel free to point out any practices that could be improved. We want this to be a solid base for building future new screens in Factorial.

## 🏡 Context

https://www.figma.com/design/9pLsRzdinid0ha0qRWta2D/Employee-Profile?node-id=378-17717&m=dev
https://factorialmakers.atlassian.net/browse/FCT-16897

## 📸 Visual proof

image


lib/components/Blocks/Header/index.tsx
```diff
@@ -0,0 +1,31 @@
...
+  title: string
+  subtitle: string
+  src: string
+  alt: string
+}
+
+export const Header = forwardRef<HTMLDivElement, HeaderProps>(
+  ({ title, subtitle, src, alt }, ref) => {
+    return (
+      <FlexBox ref={ref} className="px-10 py-6">
```
josepjaume, line 15: I'm personally not sold yet on using `Stack`, `Split` and `FlexBox` internally. I think they are nice components as an escape hatch for the consumers, but maybe using tailwind classes here would work better? I can see how some components like `Stack` will generate friction when we need to make things responsive and react to container queries.

lib/components/Blocks/Header/index.tsx
```diff
@@ -0,0 +1,31 @@
...
+        <Avatar size="xlarge">
```
josepjaume, line 16: I think we might want to make the `Avatar` API more strict so it doesn't allow so much composability. I think these two things could be passed as props. The internals (shadcn implementation) could stay nicely decoupled, but we could expose another component that's more strict, perhaps?

That's what we do with `Button`, in case you need an example.

lib/components/Blocks/Header/index.tsx
```diff
@@ -0,0 +1,31 @@
...
+          <AvatarImage src={src} alt={alt} />
+          <AvatarFallback>{alt}</AvatarFallback>
+        </Avatar>
+        <FlexBox className="flex-col gap-4 pl-5">
+          <h1 className="pt-2 text-[1.69rem]/[1.6rem] font-medium text-foreground">
```
josepjaume, line 21: This font size is a bit weird, I would try to use one of the existing sizes even if it looks a bit off, and then we normalize everything at once. I think Tailwind took it too far by basically re-allowing any CSS property. How doesn't that go against their own principles?

lib/components/Navigation/Breadcrumb/index.tsx
```diff
@@ -0,0 +1,26 @@
...
+interface BreadcrumbsType {
+  icon: IconType
+  route: string
+  title: string
+}
+
+export const Breadcrumb = forwardRef<HTMLDivElement, BreadcrumbsType>(
+  ({ icon, route, title }, ref) => {
+    return (
+      <FlexBox
```
josepjaume, line 15: Same thing about `FlexBox` I said before 👀 . Actually, if you look at the filename, it starts with `_` which tries to convey "private" 😆

lib/components/Utilities/Icon/index.tsx
```diff
@@ -15,24 +15,31 @@ const iconVariants = cva("inline-block fill-current", {
       sm: "h-4 w-4",
       xs: "h-3 w-3",
     },
+    color: {
+      primary: "text-foreground",
+      secondary: "text-secondary-foreground",
+      tertiary: "text-primary-foreground",
+    },
```
josepjaume, line 22: Do we need these properties? I mean they don't hurt, but we default to using whatever color the container has. I guess if you don't set the property, it will just not add that class and still have this behavior?

src/playground/Pages/Tabs/index.stories.tsx
```diff
@@ -0,0 +1,67 @@
...
+  component: Tabs,
+  tags: ["autodocs"],
+} satisfies Meta<typeof Tabs>
+
+export default meta
+type Story = StoryObj<typeof meta>
+
+const Overview = () => (
+  <div className="grid grid-cols-[2fr_1fr] divide-x divide-y-0 divide-dashed divide-muted">
+    <AutoGrid tileSize="md" className="col-span-2 pl-10 pr-8 pt-6">
```
josepjaume, line 17: We just recently added a `Dashboard` component meant for this use case! It just uses `AutoGrid` but I think it's better for scalability.

lib/components/Blocks/Header/index.tsx
RamProg: Makes sense, I assumed they were there to be reused, but I don't see the upside either.

lib/components/Utilities/Icon/index.tsx
RamProg: I can get rid of these, but I have 2 icons in the same container with different colors, so that forces me to wrap them and get it a bit dirtier, I'll push it so you can see what I mean.

lib/components/Information/Avatar/index.tsx
```diff
@@ -1 +1,23 @@
...
+  Avatar as AvatarComponent,
+  AvatarFallback,
+  AvatarImage,
+} from "@/ui/avatar"
+import { forwardRef } from "react"
+
+interface AvatarType {
+  alt: string
+  src?: string
+  size?: "medium" | "small" | "xsmall" | "large" | "xlarge" | "xxlarge"
```
josepjaume, line 11: Should we follow tailwind's conventions? They usually use `md`, `sm`, `xs`, `lg`, `xl` and `2xl`. Not that important, I guess.

josepjaume, line 11: If you want to delegate to the sizes that `AvatarComponent` allows, you can always do:

```suggestion
  size?: ComponentProps<typeof AvatarComponent>["size"]
```

lib/components/Navigation/Breadcrumb/index.tsx
```diff
@@ -0,0 +1,29 @@
...
+import { ChevronRight } from "@/icons"
+import { forwardRef } from "react"
+
+interface BreadcrumbsType {
+  icon: IconType
+  route: string
+  title: string
+}
+
+export const Breadcrumb = forwardRef<HTMLDivElement, BreadcrumbsType>(
```
josepjaume, line 11: Have you considered checking out ShadCN's breadcrumb component? https://ui.shadcn.com/docs/components/breadcrumb

src/playground/Pages/Tabs/index.stories.tsx
```diff
@@ -0,0 +1,76 @@
...
+
+  const Component = Widgets[index % Widgets.length]
+  return <Component />
+}
+
+const Overview = () => (
+  <div className="grid grid-cols-[2fr_1fr] divide-x divide-y-0 divide-dashed divide-muted">
+    <div className="pl-10 pr-8 pt-6">
+      <Dashboard>
+        <>{Array.from({ length: 6 }).map((_, i) => renderWidget(i))}</>
```
josepjaume, line 39: Careful - you're setting a fragment as the first element. This should be directly the array so the masonry does its thing.

src/playground/Pages/Tabs/index.stories.tsx
```diff
@@ -0,0 +1,76 @@
+import { AreaChartWidget } from "@/components/Widgets/Charts/AreaChartWidget"
```
josepjaume, line 1: I can't comment on the file name, but should this be called `Overview` for consistency, perhaps?

src/playground/Pages/Tabs/index.tsx
```diff
@@ -0,0 +1,54 @@
+import { Header } from "@/components/Blocks/Header"
+import { Breadcrumb } from "@/components/Navigation/Breadcrumb"
+import {
+  Tabs as TabsComponent,
+  TabsContent,
+  TabsList,
+  TabsTrigger,
+} from "@/components/Navigation/Tabs"
```
josepjaume, line 8: I feel we'll enventually have to create different kinds of tabs or at least differentiate the "navigational" ones (only used in pages) from other kinds of tabs (forms, etc). But this is OK for now.

src/playground/Pages/Tabs/index.tsx
```diff
@@ -0,0 +1,54 @@
...
+
+export const Tabs = forwardRef<HTMLDivElement, TabsProps>(
+  ({ tabs, title, subtitle, src, alt, defaultTab = tabs[0].key }, ref) => (
+    <main ref={ref}>
+      <Breadcrumb icon={User} route={"Employees"} title={"Alba Horneros"} />
+      <Header title={title} subtitle={subtitle} src={src} alt={alt}></Header>
+      <TabsComponent defaultValue={defaultTab}>
+        <TabsList className="h-auto w-full justify-start rounded-none border-b border-l-0 border-r-0 border-t-0 border-solid border-b-muted bg-transparent px-10 py-3">
+          {tabs.map((tab: TabType) => (
+            <TabsTrigger
```
josepjaume, line 35: This is really interesting. Eventually, we'll want to use a React Router `<Link>` element here, but we don't want to get coupled to it as a library. Since this uses radix behind the scenes, we can use the `asChild` prop and just pass the `<Link />` component as children and they will merge.

This is how it works: https://www.radix-ui.com/primitives/docs/utilities/slot

josepjaume (commented):
Nearly there! Sorry for the torture haha.

src/playground/Pages/Tabs/index.stories.tsx
RamProg: This one is outdated

src/playground/Pages/Tabs/index.tsx
RamProg: Agreed

RamProg:
> Nearly there! Sorry for the torture haha.

Don't hesitate to comment, the whole point of this is to agree on solid "foundations" to create a new and better version of Factorial right? So let's work on that, as long as we have the capacity, the better we can build this, the happier we'll be in 2025 and so on.

src/playground/Pages/Tabs/index.tsx
josepjaume: What I mean is - we probably want to eventually find a way to pass down a component as a trigger?

RamProg: Agreed. I haven't experimented yet with connecting this component to Factorial, the goal here was simply creating the Layout UI, I think there's still some work left before jumping to that, but it's definitely a challenge I have in mind. Thanks for the doc.

lib/components/Information/Avatar/index.tsx
RamProg: done

lib/components/Blocks/Header/index.stories.tsx
```diff
@@ -0,0 +1,25 @@
...
+    subtitle: "Product Designer",
+    src: "https://github.com/dani-moreno.png",
+    alt: "DM",
+  },
+} satisfies Meta<typeof Header>
+
+export default meta
+type Story = StoryObj<typeof meta>
+
+export const Default: Story = {
```
josepjaume, line 18: I have the feeling that this might fit better into a `Pages` category, right? Actually, I don't really like "Blocks" and the only other element we have now is `Card` which should probably be ditched in favor of `Widget` (card shouldn't be publicly exposed).

lib/components/Information/Avatar/index.stories.tsx
```diff
@@ -10,13 +10,14 @@ const meta = {
   tags: ["autodocs"],
   args: {
     size: "medium",
+    src: "https://github.com/dani-moreno.png",
```
josepjaume, line 13: This is not for now, but I think we should create an `ImageSrcProvider` the same way we have a `LinkProvider`. Ideally, the image provider would be able to set the image dimensions as params and even leverage [srcset](https://developer.mozilla.org/en-US/docs/Learn/HTML/Multimedia_and_embedding/Responsive_images) for responsive images.

lib/components/Navigation/Breadcrumb/index.tsx
```diff
@@ -0,0 +1,49 @@
...
+interface RouteType {
+  title: string
+  url: string
+}
+
+export const Breadcrumb = forwardRef<HTMLDivElement, BreadcrumbsType>(
+  ({ icon, routes, title }, ref) => {
+    return (
+      <BreadcrumbComponent
+        className="flex min-h-6 flex-row border-x-0 border-t-0 border-dashed border-b-muted px-5 py-3"
```
josepjaume, line 27: I have the feeling that we should bake in all these classes inside ShadCN's component, right? What I usually try to do is to change the original component's API as little as possible, but change its aspect if needed.

This whole "trying to keep ShadCN alive" is quite messy though, so both options are good if the public API stays stable.

src/playground/Pages/Tabs/index.stories.tsx
```diff
@@ -0,0 +1,91 @@
...
+)
+
+const Profile = () => <div>Profile</div>
+const Personal = () => <div>Personal</div>
+const Agreement = () => <div>Agreement</div>
+const TimeOff = () => <div>Time off</div>
+const Competencies = () => <div>Competencies</div>
+const Activity = () => <div>Activity</div>
+
+export const Default: Story = {
```
josepjaume, line 61: What would you feel about calling this whole story `EmployeeProfile` to make it more explicit? `Tabs` looks too generic to me.

RamProg: I was aiming for generic, my reasoning was that the same layout could be reused in a different screen.

lib/components/Navigation/Breadcrumb/index.tsx
RamProg: Yeah, I don't think there's a right answer, end of day what we are exposing is the same, regardless of where we put the styles. I can't think of any major pros / cons of either option.

lib/components/Blocks/Header/index.stories.tsx
RamProg: done

josepjaume (approved):
Let's do this!

PR #372, opened 2024-08-02
Widget insights

dani-moreno:
## 🚪 Why?

As our first milestone working on Insights, we're aiming to swap out the current charts in the Employees list with One components. This PR sets the stage for us to do just that.

## 🔑 What?

First take on creating and customizing One components to look like our [Figma charts](https://www.figma.com/design/O8578qyVUfgzm6ZkVkBJwW/Charts-UI-Kit-2.0?node-id=1-9276&t=Yx1rZtksNxfRus2B-0).

- Style changes.
- Modified `Card` and `ChartContainer` to mirror what we made in Figma.
- Made an example in the Dashboard component that is akin to what we have in the Employees list.

## 📸 Visual proof

image


josepjaume:
Lookin lovely! 
image

lib/components/Widgets/Dashboard/index.stories.tsx
```diff
@@ -18,7 +18,8 @@ const renderWidget = (index: number) => {
     () => <AreaChartWidget {...AreaChartWidgetStoriesMeta.args} />,
     () => (
       <WidgetContainer
-        header={{ title: "A form widget", description: "Enter your data" }}
+        header={{ title: "A form widget" }}
+        description="Enter your data"
```
josepjaume, line 22: There's something about this API that makes me a bit uncomfortable. I understand how having `header` separately groups things logically, but it also somehow bleeds implementation details.

My question is: Do we really have a use case for `description`? Should we just remove it?

lib/components/Widgets/WidgetContainer/index.tsx
```diff
@@ -1,48 +1,39 @@
...
   HTMLDivElement,
   WidgetContainerProps & { children: ReactNode }
->(({ header, footer, children }, ref) => (
-  <Card ref={ref}>
+>(({ header, description, children }, ref) => (
+  <Card ref={ref} className="min-h-[200px]">
     <CardHeader>
-      <CardTitle>{header.title}</CardTitle>
-      <CardDescription>{header.description}</CardDescription>
+      <div className="flex grow flex-row gap-1.5">
```
josepjaume, line 29: I think this `flex` needs `items-center` to keep things aligned in the vertical axis.

lib/components/Widgets/WidgetContainer/index.tsx
```diff
@@ -1,48 +1,39 @@
...
+        <CardTitle>{header.title}</CardTitle>
+        {header.subtitle && <CardSubtitle>{header.subtitle}</CardSubtitle>}
+        {header.info && <CardInfo>{header.info}</CardInfo>}
+      </div>
+      {header.link && <CardLink href={header.link} />}
     </CardHeader>
+    <CardDescription>{description}</CardDescription>
```
josepjaume, line 36: As I said before - do we really need this?

lib/ui/card.tsx
```diff
@@ -2,14 +2,25 @@ import * as React from "react"
...
+
 const Card = React.forwardRef<
   HTMLDivElement,
   React.HTMLAttributes<HTMLDivElement>
 >(({ className, ...props }, ref) => (
   <div
     ref={ref}
     className={cn(
-      "rounded-2xl border border-solid border-border bg-card text-card-foreground shadow-sm",
+      "min-w-60 rounded-2xl border border-solid border-border bg-card text-card-foreground",
```
josepjaume, line 23: Not entirely sold we need a `min-w`, this should be a last resort I think. If you're doing this so things display better on the storybook example, we can easily create a decorator that wraps the card inside a container that has a min width and height.

lib/components/Widgets/WidgetContainer/index.tsx
```diff
@@ -1,48 +1,39 @@
...
+  <Card ref={ref} className="min-h-[200px]">
     <CardHeader>
-      <CardTitle>{header.title}</CardTitle>
-      <CardDescription>{header.description}</CardDescription>
+      <div className="flex grow flex-row gap-1.5">
+        <CardTitle>{header.title}</CardTitle>
+        {header.subtitle && <CardSubtitle>{header.subtitle}</CardSubtitle>}
+        {header.info && <CardInfo>{header.info}</CardInfo>}
+      </div>
+      {header.link && <CardLink href={header.link} />}
```
josepjaume, line 34: What if we tried making the whole card header clickable, not only the arrow?

lib/ui/card.tsx
```diff
@@ -33,25 +44,71 @@ CardHeader.displayName = "CardHeader"
...
+  React.HTMLAttributes<HTMLDivElement>
+>(({ className, ...props }, ref) => (
+  <div ref={ref} className={cn("-ml-1 h-6 w-6", className)}>
+    <TooltipProvider>
+      <Tooltip>
+        <TooltipTrigger
+          className="cursor-default text-muted-foreground"
+          title="Info"
+        >
+          <Icon icon={InfoCircleLine} size="md" />
```
josepjaume, line 75: `className="icon-info"` would maybe help?

lib/ui/card.tsx
```diff
@@ -33,25 +44,71 @@ CardHeader.displayName = "CardHeader"
...
+CardLink.displayName = "CardLink"
 
 const CardDescription = React.forwardRef<
   HTMLParagraphElement,
   React.HTMLAttributes<HTMLParagraphElement>
 >(({ className, ...props }, ref) => (
   <p
     ref={ref}
-    className={cn("text-sm text-muted-foreground", className)}
+    className={cn("px-4 text-sm text-muted-foreground", className)}
```
josepjaume, line 111: This `PX` looks a bit suspicious to me. It almost looks as if `description` should live inside of `CardHeader` but we're bringing it outside just to add padding back again.

I would consider putting everything inside of CardHeader or removing `description` altogether, no?

josepjaume (commented):
Added a few comments! Do you wanna pair on it?

lib/ui/card.tsx
dani-moreno: Done!

lib/components/Widgets/WidgetContainer/index.tsx
dani-moreno: Done!

lib/ui/card.tsx
dani-moreno: Done! Added decorators to stories :)

lib/components/Widgets/WidgetContainer/index.tsx
dani-moreno: As we talked, we're opting for the arrow for the time being, but it's subject to change.

lib/components/Blocks/Card/index.stories.tsx
```diff
@@ -39,8 +39,8 @@ export const WithTitle: Story = {
       <Card {...props}>
         <CardHeader>
           <CardTitle>Card title</CardTitle>
-          <CardDescription>Description of the title</CardDescription>
         </CardHeader>
+        <CardDescription>Description of the title</CardDescription>
```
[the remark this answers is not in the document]
dani-moreno: Done! Removed `CardDescription` as we don't really have a need for it at the moment.

lib/components/Widgets/Charts/ChartContainer.tsx
```diff
@@ -10,8 +10,18 @@ export const ChartContainer = forwardRef<
...
   }
->(({ chart, ...props }, ref) => (
+>(({ chart, summaries, ...props }, ref) => (
   <WidgetContainer ref={ref} {...props}>
+    {summaries && (
+      <div className="-mt-2 flex flex-row pb-3">
+        {summaries.map((summary, index) => (
+          <div key={index} className="grow">
+            <div className="text-sm text-muted-foreground">{summary.label}</div>
+            <div className="text-2xl font-medium">{summary.value}</div>
```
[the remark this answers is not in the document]
dani-moreno: Yay, done!
image

lib/ui/card.tsx
```diff
@@ -36,32 +47,73 @@ const CardTitle = React.forwardRef<
...
+>(({ className, ...props }, ref) => (
+  <div
+    ref={ref}
+    className={cn("-ml-1 flex h-6 w-6 items-center justify-center", className)}
+  >
+    <TooltipProvider>
+      <Tooltip>
+        <TooltipTrigger
+          className="h-5 w-5 cursor-help text-muted-foreground"
+          title="Info"
```
josepjaume, line 80: This isn't cool 👀 , and also it's in english. Can we use `aria-label` instead with the contents of the tooltip?

lib/ui/card.tsx
```diff
@@ -36,32 +47,73 @@ const CardTitle = React.forwardRef<
...
+        >
+          <Icon icon={InfoCircleLine} size="md" />
+        </TooltipTrigger>
+        <TooltipContent>
+          <p {...props} />
```
josepjaume, line 85: This is a bit weird. Shouldn't we just receive the contents as a string and put it as children of `<p>`?

lib/ui/card.tsx
```diff
@@ -36,32 +47,73 @@ const CardTitle = React.forwardRef<
...
+  HTMLAnchorElement,
+  React.ComponentPropsWithoutRef<"a">
+>(({ className, ...props }) => {
+  return (
+    <a
+      className={cn(
+        "h-6 w-6 text-muted-foreground transition-colors hover:text-foreground",
+        className
+      )}
+      aria-label="Link"
```
josepjaume, line 103: This is a magic string and it's not translated. We should use something that we receive as a prop. That way, we can even smoothly transition to using a link with text instead in case we wanted in the future.

josepjaume (changes requested):

lib/ui/card.tsx
josepjaume: A way to do this is to receive the link as a parameter like this:

```
link?: {
  title: string,
  url: string
}
```

This way you can check if `link` is there, and then, in case it exists, it's forced to have `title` and `url`.

lib/ui/card.tsx
dani-moreno: Hah, right! Fixed :)

josepjaume (approved):
Amazing work!

PR #373, opened 2024-08-02
Add link controller

josepjaume:
I've been wondering how can we support client-side links without getting coupled to any of the routing libraries. They usually rely on you explicitly using a `<Link>` component, but I didn't want us to have to do that, with the potential resulting inconsistencies.

This adds a `<LinkProvider>` component that is merged into the `<FactorialOne />` provider. This link provider allows us to define a `Component` for any link.

You can define a component like this:

```tsx
<LinkProvider
  component: (props, ref) => (
    <a
      ref={ref}
      {...props}
      onClick={(event, ...args) => {
        console.log(props.href)
        props?.onClick?.(event, ...args)
        event.preventDefault()
      }}
    />
  ),
>
  <App />
</LinkProvider>
```

As you can see, `component` receives the props from the link, and it allows you to define a new component. This means that we can easily use our `<Link />` component of choice coming from our routing library!

### DIFF
--- a/.storybook/preview.tsx
+++ b/.storybook/preview.tsx
@@ -13,7 +13,6 @@ import { FactorialOneProvider } from "../lib/lib/one-provider"
 import lightTheme, { darkTheme } from "./FactorialOne"
 import { DocsContainer } from "./DocsContainer"
 import { useDarkMode } from "storybook-dark-mode"
-import { fn } from "@storybook/test"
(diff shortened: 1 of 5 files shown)


RamProg (approved):

sauldom102 (approved):
Ohh quite powerful 😮 I guess we could add some custom behavior with this like opening certain types of links in a popover or something like that

PR #374, opened 2024-08-03
Add base font sizes

josepjaume:
This allows customizing base font sizes, which seem to be slightly different from the default Tailwind ones:

https://tailwindcss.com/docs/font-size

We don't have to merge this PR yet, but this is worth checking out.

### DIFF
--- a/lib/components/Actions/Button/index.stories.tsx
+++ b/lib/components/Actions/Button/index.stories.tsx
@@ -2,7 +2,7 @@ import type { Meta, StoryObj } from "@storybook/react"
 import { fn } from "@storybook/test"
 
 import Add from "@/icons/Add"
-import Bell from "@/icons/Bell"
+import Archive from "@/icons/Archive"
 import { Button } from "."
 
 const meta = {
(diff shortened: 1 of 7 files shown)


RamProg (approved):
This makes a lot of sense, but as I've said before, designers need to be 100% in sync with what are the font options and I imagine there will be more than 2. Also, it could happen that we want the same size, but a different line height for two fonts. We could consider moving from the default names to our own names so we have more control and they make more sense (ie text-title, text-cta, etc). Last thing, interesting that customizing sizes changed the order in Tailwind? I imagine that's unexpected.

josepjaume:
Yeah @RamProg you're right that we need for designers to agree on this, as this decision cascades down and could have unintended effects.

As for using our own names - I would try and stick to the tailwind defaults as much as possible for two reasons:

* ShadCN trusts them, so any future component we want to import will already have a good baseline without much customization
* They are good defaults! I trust the designers behind Tailwind UI and the whole reasoning behind sizing, names, etc. and I doubt we could reach something that's better

That said, we should probably wait to merge this until we discuss it further with the designers!

PR #377, opened 2024-08-05
Story/fct 16898/create details

RamProg:
## 🚪 Why?

This is part of the new Profile Layout that includes a Details bar on the right

## 🔑 What?

Create Details Component and integrate it in the Tab Layout storie

## 🏡 Context

https://factorialmakers.atlassian.net/browse/FCT-16898
https://www.figma.com/design/9pLsRzdinid0ha0qRWta2D/Employee-Profile?node-id=946-24240&t=IlDuhabx9R5Fm6DO-0

## 📸 Visual proof

image


lib/components/Information/Badge/index.tsx
```diff
@@ -1 +1,33 @@
...
+  imageSrc?: ComponentProps<typeof Avatar>["src"]
+  alt?: ComponentProps<typeof Avatar>["alt"]
+}
+
+export const Badge = forwardRef<HTMLDivElement, BadgeProps>(
+  ({ text, hasAvatar = false, imageSrc, alt, variant = "name" }) => {
+    return (
+      <BadgeComponent
+        variant={variant}
+        className={cn(hasAvatar ? "pl-[0.15rem] pr-2" : "px-2")}
```
josepjaume, line 19: I would set `flex gap-1 items-center` and prevent passing down a `className` to `Avatar` down below.

lib/components/Information/Badge/index.tsx
```diff
@@ -1 +1,33 @@
...
+      >
+        {hasAvatar && (
+          <Avatar
+            alt={alt || text[0]}
+            src={imageSrc}
+            size="xxsmall"
+            className="mr-1 bg-orange-500"
```
josepjaume, line 26: Why `bg-orange-500` in this specific instance? Can we solve this in a different way? Should this background color be on the avatar itself?

lib/ui/avatar.tsx
```diff
@@ -17,6 +18,7 @@ export const sizes = [
 const avatarVariants = cva("relative flex shrink-0 overflow-hidden", {
   variants: {
     size: {
+      xxsmall: "h-6 w-6 rounded-xl text-base text-white",
```
josepjaume, line 21: Maybe we should just shrink the `xsmall` size? We don't want that many variations around, do we?

src/playground/Weekdays/index.tsx
```diff
@@ -0,0 +1,41 @@
...
+  "Thursday",
+  "Friday",
+  "Saturday",
+  "Sunday",
+]
+
+export const Weekdays = forwardRef<HTMLDivElement, WeekdaysProps>(
+  ({ daysOfTheWeek = DAYS_OF_THE_WEEK, activatedDays = [] }, ref) => {
+    return (
+      <ToggleGroup
```
josepjaume, line 22: Nice job *but* I don't think these are actually supposed to be interactive 👀

RamProg: They are not interactive. activated days is an array of the active ones, the rest will be deactivated and you can't modify the state by clicking on them.

Toggle and ToggleGroup were the most similar shadCN components to what we needed. It was easier to create it, but I considered that by using the Toggle we could (if we wanted) also have the toggle feature in the future, so I made a call to use Toggle rather than creating a more basic, limited, but similar component.

lib/components/Information/Badge/index.tsx
RamProg: I wasn't sure about this one... When there's no specific avatar it will take the first letter of the name and create the avatar. I don't know if the background color is random or there's a specific color for each team for example. In any case if you send an image source it will override this and just show the image, this is more of a fallback mechanism.

lib/components/Information/Badge/index.tsx
RamProg: The issue here was that the Avatar is supossed to be very close to the left, almost touching the border: 
image

Because the image is rounded, then it looks fine, but if you put text so close to the border it won't look good. So for this badge with avatar I need this custom padding on the left.

I'm not really passing className to avatar, but to Badge, another would be to just make this padding part of the variant, do you want to request that change?

lib/ui/avatar.tsx
RamProg: Makes sense to me.

src/playground/Weekdays/index.tsx
josepjaume: I understand the reasoning. Something feels off about using "disabled" elements just for displaying purposes in this scenario, as I don't see how it could become interactive in the future. It feels like cutting corners, no?

lib/components/Information/Badge/index.tsx
josepjaume: I feel like this logic should go in the `AvatarFallback` component then, no?

lib/components/Information/Badge/index.tsx
```diff
@@ -1 +1,33 @@
...
+interface BadgeProps {
+  text: string
+  hasAvatar?: boolean
+  variant?: ComponentProps<typeof BadgeComponent>["variant"]
+  imageSrc?: ComponentProps<typeof Avatar>["src"]
+  alt?: ComponentProps<typeof Avatar>["alt"]
+}
+
+export const Badge = forwardRef<HTMLDivElement, BadgeProps>(
+  ({ text, hasAvatar = false, imageSrc, alt, variant = "name" }) => {
```
josepjaume, line 15: Can't we just use a property called `avatarSrc` and infer `hasAvatar` from there? Feels redundant.

lib/components/Information/Badge/index.tsx
josepjaume: Okay, I understand! I think it's ok then. The fact that we're keeping the component split between shadcn and our implementation is getting a bit in the middle. In the end, as with the rest, what matters is the public API.

lib/components/Information/Avatar/index.tsx
```diff
@@ -9,12 +9,13 @@ interface AvatarType {
   alt: string
   src?: string
   size?: ComponentProps<typeof AvatarComponent>["size"]
+  className?: ComponentProps<typeof AvatarComponent>["className"]
```
josepjaume, line 12: I would really like to try and prevent people being able to pass `className` to components, as this will inevitably end with hacks. Can you think of any other solution to this involving the wrapper?

src/playground/Weekdays/index.tsx
RamProg: I actually thought I was doing the opposite, creating this from scratch was much easier, it's just a container with a string, but having a more powerful component that we could use in the future made more sense. I imagine the workable days need to be set at some point right? Why not using the same one, but make it dynamic in that case? This component can work both ways.

I'm happy to get rid of it if my reasoning was wrong.

lib/components/Information/Badge/index.tsx
RamProg: I had the same thought, but even if there's no avatarSrc, we could still have a fallback avatar that we want to display. So, you can provide no avatar and not alt text, but still want to show an image.

lib/components/Information/Avatar/index.tsx
RamProg: Yeah, I was thinking that might be the case, I'll take a look.

src/playground/Weekdays/index.tsx
josepjaume: I understand. No biggie, let's go for this and we can always change our minds later. I would say though this will need a few stories to document it. Maybe in the `Forms` section?

src/playground/Weekdays/index.tsx
```diff
@@ -0,0 +1,41 @@
...
+
+const DAYS_OF_THE_WEEK = [
+  "Monday",
+  "Tuesday",
+  "Wednesday",
+  "Thursday",
+  "Friday",
+  "Saturday",
+  "Sunday",
+]
```
josepjaume, line 17: I feel this should come as a parameter as it needs to be translated and it's also company-specific (start of the week could be different from company to company)

RamProg: Agreed. That's why this is a fallback in case you don't send your own days of the week.

lib/ui/avatar.tsx
```diff
@@ -18,7 +18,7 @@ export const sizes = [
 const avatarVariants = cva("relative flex shrink-0 overflow-hidden", {
   variants: {
     size: {
-      xsmall: "h-8 w-8 rounded-xl text-xs",
+      xsmall: "h-6 w-6 rounded-xl text-base",
```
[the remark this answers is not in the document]
RamProg: Yeah, that's why before I had created xxsmall. The reason for the sizing is simply follow the figma sizes and spacing. I'll show you how each case looks like:

text-xs
image
image

text-sm
image
image


text-base
image
image

lib/components/Information/Badge/index.tsx
RamProg: I'll try it out

src/playground/Weekdays/index.tsx
RamProg: Good point, I'll add it

src/playground/Weekdays/index.tsx
josepjaume: Hahaha I understand. Still, this is quite dangerous. I think we'll eventually have to add a `<LocalizationProvider />` in which we define things like:

* Error validation messages
* Default labels for obvious things: Submit buttons, tooltip aria-labels, etc
* Time things: Default week day, months, days, formatting, etc
* Maybe even money, etc?

So I would say that we can keep it like this while it's on the playground, but create that provider before we promote this to a full-fledged component.

lib/ui/avatar.tsx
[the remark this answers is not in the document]
josepjaume: I think so! For consistency, no? Otherwise we're using a bigger size for a smaller variant? Feels weird.

josepjaume (approved):
Let's do this!

PR #385, opened 2024-08-06
Use rem for typography

josepjaume:
Swaps the usage of `px` with `rem` on typography sizes.

### DIFF
--- a/tailwind.config.ts
+++ b/tailwind.config.ts
@@ -25,14 +25,14 @@ export default {
       sans: ["Inter", "sans-serif"],
     },
     fontSize: {
-      xs: ["11px", "16px"],
-      sm: ["13px", "20px"],
-      base: ["15px", "24px"],
-      lg: ["17px", "24px"], // Not really used but we keep it for scale
-      xl: ["19px", "28px"],
-      "2xl": ["21px", "28px"], // Same
-      "3xl": ["24x", "32px"], // Same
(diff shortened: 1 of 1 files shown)


tailwind.config.ts
```diff
@@ -25,14 +25,14 @@ export default {
...
-      sm: ["13px", "20px"],
-      base: ["15px", "24px"],
-      lg: ["17px", "24px"], // Not really used but we keep it for scale
-      xl: ["19px", "28px"],
-      "2xl": ["21px", "28px"], // Same
-      "3xl": ["24x", "32px"], // Same
-      "4xl": ["27x", "32px"],
+      xs: ["0.6875rem", "1rem"],
+      sm: ["0.8125rem", "1.25rem"],
+      base: ["0.9375rem", "1.5rem"],
```
RamProg, line 30: If our base font is 15px instead of the default 16 in tailwind for example, should be set 1 rem to be 15 px instead of 16 ? Reference: https://tailwindcss.com/docs/font-size

josepjaume: The thing is -- all the rest of the measures like `p-1`, `p-2`, `h-4` are in terms of the base `rem`, so this would effectively modify all sizes and what's worth, make some not dividable by two.

I think we need to interpret the base `1em` as the "baseline measure" and then the typography is just "a bit smaller than the baseline". Otherwise things get weird fast.

PR #392, opened 2024-08-07
Widget skeleton loading

dani-moreno:
Adds a Skeleton version of the WidgetContainer, as a subcomponent. This approach lets you show a Skeleton version while the content is loaded, and replace it with the actual WidgetContainer once the content is ready:

```
const { data, loading } = useQuery(...)

if(loading && !data) return <Widget.Skeleton />

return (<AreaChartWidget ... />)
```

image


josepjaume:
As discussed offline, `loading` props in components will always compete with their required attributes. I think the best way to handle this would be to provide `Skeletons` that are drop-in replacement for components while they are loading.

```tsx
const { data, loading } = fetchData()
if(loading) return <Widget.Skeleton />
return (<Widget title="Whatever">Content</Widget>)
```

We could decide to optionally pass a title to the skeleton even without corrupting the props on `Widget` (as sometimes the title can be dynamic).

josepjaume:
This is powerful and gives more control to users in scenarios where we could have nested loading states:

```tsx
<Widget title="Something we already know">
  <List>Something we know</List>
  <EventList.Skeleton>This is loading</EventList.Skeleton>
</Widget>
```

lib/components/Widgets/WidgetContainer/index.tsx
```diff
@@ -16,23 +17,43 @@ export interface WidgetContainerProps {
...
-        <CardTitle>{header.title}</CardTitle>
-        {header.subtitle && <CardSubtitle>{header.subtitle}</CardSubtitle>}
-        {header.info && <CardInfo content={header.info} />}
-      </div>
-      {header.link && (
-        <CardLink href={header.link.url} title={header.link.title} />
+      {loading ? (
+        <div className="flex h-6 w-full items-center" aria-hidden={true}>
+          <Skeleton className="h-4 w-full max-w-28" />
+        </div>
```
RamProg, line 32: Is the div wrapping the Skeleton really necessary? Also, instead of sending a className to the Skeleton maybe we want to simply change the looks in the Skeleton itself? Ideally the Skeleton will already know how to present itself and it won't need another component to tell it how to look like.

RamProg (approved):
Looks great! Just a comment, but not a blocker.

josepjaume (approved):
Excellent work, I love it! ❤️

lib/components/Widgets/WidgetContainer/index.tsx
dani-moreno: @RamProg Good catch! The div is there to ensure the header maintains the same height as when there's a title. The skeleton bar is a bit less lower (drawing it at the same size as the text makes it looks a bit chunky, hah).

Regarding the Skeleton component, I'm quite taken with the freedom of passing classes. It's in line with shadcn's approach, after all. Maybe we could add some predefined styles to it.

PR #396, opened 2024-08-07
Use 14px as the default font size

josepjaume:
The current default font size for Factorial is `14px`, and adopting it seems like the best path forward to make the transition as smooth as possible.

We moved to using dimensions in pixels, which gives us the most flexibility - the ability of changing the default font size for accessibility purposes, but keeping the dimensions of things.


dani-moreno (approved):

nlopin:
I would like to spawn a new round of discussions since setting the size of the font in pixels breaks the default font size setting in browsers, harming the essential accessibility rule.

Have we considered other options? Is the initial body font hardcoded in all our products? 

Thinking about the design system in general, I'd consider using a relative unit together with `!important` to avoid collisions for some time (e.g. `font-size: 87.5% !important`), while tackling the whole issue of the initial font size setting on the product level

PR #404, opened 2024-08-09
Add `withSkeleton` helper

josepjaume:
Adds a helper that allows appending skeletons to components extremely easy!


dani-moreno (approved):
🔥

PR #422, opened 2024-08-14
CartesianGrid on charts

dani-moreno:
Adds the `CartesianGrid` to those charts that were missing it, and uses the props defined on `elements.tsx`.
Also, adds a dashed stroke to this `CartesianGrid`, as we have in Figma.


CarlosBM18 (approved):
Nice!

PR #424, opened 2024-08-14
Chart counter animation

dani-moreno:
Adds a simple and subtle animation to numbers in the `ChartContainer` :-)

https://github.com/user-attachments/assets/91e59417-3399-4614-b828-d9a5629a74e2


CarlosBM18 (approved):
💅🏼

PR #427, opened 2024-08-15
Add NumberInput

viniciusmeneses:
## 🚪 Why?

We recently introduced a `NumberField` component to Gamma to solve many problems we had with the numeric input components. Since we still don't have a numeric input in Factorial One and using an input with `type="number"` will not offer the best UX for our users, we need to port that component to Factorial One.

## 🔑 What?

Add the `NumberInput` component.

## 🏡 Context

- [Gamma NumberField](https://github.com/factorialco/factorial/pull/47179)

## 📸 Visual proof

https://github.com/user-attachments/assets/4d4bb57c-dae0-49d6-996a-1b7c57c5ccfe

### DIFF
--- a/lib/components/Forms/Fields/Input/index.tsx
+++ b/lib/components/Forms/Fields/Input/index.tsx
@@ -1,11 +1,13 @@
 export * from "@/ui/input"
 import { Input as ShadcnInput } from "@/ui/input"
-import { ComponentProps } from "react"
+import { ComponentProps, HTMLInputTypeAttribute } from "react"
 
-type InputProps = Pick<
+export type InputProps = Pick<
   ComponentProps<typeof ShadcnInput>,
-  "ref" | "type" | "disabled" | "size" | "onChange" | "value" | "placeholder"
->
+  "ref" | "disabled" | "size" | "onChange" | "value" | "placeholder"
+> & {
+  type?: Exclude<HTMLInputTypeAttribute, "number">
+}
 
 const Input: React.FC<InputProps> = ShadcnInput
 
--- a/lib/components/Forms/Fields/NumberInput/index.stories.tsx
+++ b/lib/components/Forms/Fields/NumberInput/index.stories.tsx
@@ -1,5 +1,6 @@
 import type { Meta, StoryObj } from "@storybook/react"
 
+import { useState } from "react"
 import { NumberInput } from "."
 
 const meta = {
@@ -34,6 +35,18 @@ type Story = StoryObj<typeof meta>
 
 export const Primary: Story = {
   args: {},
+}
+
+export const WithStep: Story = {
+  args: {
+    min: 1,
+    max: 5,
+    step: 1,
+  },
+  render: (props) => {
+    const [value, setValue] = useState<number | null>(1)
+    return <NumberInput {...props} value={value} onChange={setValue} />
+  },
 }
 
 export const Disabled: Story = {
--- a/lib/components/Forms/Fields/NumberInput/index.stories.tsx
+++ b/lib/components/Forms/Fields/NumberInput/index.stories.tsx
@@ -1,0 +1,45 @@
+import type { Meta, StoryObj } from "@storybook/react"
+
+import { NumberInput } from "."
+
+const meta = {
+  render: (props) => <NumberInput key={JSON.stringify(props)} {...props} />,
+  tags: ["autodocs"],
+  args: {
+    disabled: false,
+    placeholder: "Placeholder text here",
+    locale: "en-US",
+  },
+  argTypes: {
+    value: {
+      control: { type: "number" },
+    },
+    locale: {
+      options: ["en-US", "es-ES", "pt-BR"],
+      control: { type: "select" },
+    },
+    maxDecimals: {
+      control: { type: "number" },
+    },
+  },
+  parameters: {
+    jsx: {
+      filterProps: (_: unknown, propName: string) => propName !== "key",
+    },
+  },
+} satisfies Meta<typeof NumberInput>
+
+export default meta
+type Story = StoryObj<typeof meta>
+
+export const Primary: Story = {
+  args: {},
+}
+
+export const Disabled: Story = {
+  args: {
+    value: 32.5,
+    disabled: true,
+    locale: "es-ES",
+  },
+}
--- a/lib/components/Forms/Fields/NumberInput/index.stories.tsx
+++ b/lib/components/Forms/Fields/NumberInput/index.stories.tsx
@@ -44,7 +44,7 @@ export const WithStep: Story = {
     step: 1,
   },
   render: (props) => {
-    const [value, setValue] = useState<number | null>(1)
+    const [value, setValue] = useState<number | null>(props.value ?? 1)
     return <NumberInput {...props} value={value} onChange={setValue} />
   },
 }
--- a/lib/components/Forms/Fields/NumberInput/index.tsx
+++ b/lib/components/Forms/Fields/NumberInput/index.tsx
@@ -1,0 +1,77 @@
+import { forwardRef, useEffect, useState } from "react"
+import { Input, InputProps } from "../Input"
+
+const FORMAT = /^([0-9]+)$/
+// eslint-disable-next-line no-useless-escape
+const FORMAT_WITH_DECIMALS = /^([0-9]+)(?:[\.,]([0-9]+)?)?$/
+
+const parseValue = (value: string) => parseFloat(value.replace(",", "."))
+
+type NumberInputProps = Omit<InputProps, "value" | "type" | "onChange"> & {
+  locale: string
+  value?: number
+  maxDecimals?: number
+  onChange?: (value: number | null) => void
+}
+
+export const NumberInput = forwardRef<HTMLInputElement, NumberInputProps>(
+  ({ locale, value, maxDecimals, onChange, ...props }, ref) => {
+    const formatValue = (value: number, maxDecimals?: number) =>
+      new Intl.NumberFormat(locale, {
+        maximumFractionDigits: maxDecimals,
+        useGrouping: false,
+      }).format(value)
+
+    const [fieldValue, setFieldValue] = useState<string>(() =>
+      value != null ? formatValue(value, maxDecimals) : ""
+    )
+
+    const handleChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+      const value = event.target.value
+
+      if (!value) {
+        setFieldValue("")
+        return onChange?.(null)
+      }
+
+      const match = value.match(
+        maxDecimals !== 0 ? FORMAT_WITH_DECIMALS : FORMAT
+      )
+      if (!match) return
+
+      const [_, integers, decimals] = match
+      if (maxDecimals && decimals && decimals.length > maxDecimals) return
+
+      const valueWithNoLeadingZero = value.match(/^0\d$/)
+        ? value.slice(1)
+        : value
+
+      const valueAsNumber = parseFloat(`${integers}.${decimals ?? 0}`)
+
+      setFieldValue(valueWithNoLeadingZero)
+      onChange?.(valueAsNumber)
+    }
+
+    // We need the logic "parseValue(fieldValue) !== value" below to ensure we don't remove the decimal separator while the user is modifying the decimals.
+    // For example, if the input value is 18,3 and I want to change it to 18,5. I'll delete the last digit and the input will become "18,"
+    // At this point, we have fired the "onChange" callback with the value 18 because it's the most precise valid value ("18," isn't a valid one),
+    // then the useEffect below will be triggered, since the "value" changed from "18,3" to "18", and it'll not modify the "fieldValue" to keep the decimal separator
+    // unless if the value really changed.
+    useEffect(() => {
+      if (value === null) return setFieldValue("")
+      if (value != null && parseValue(fieldValue) !== value)
+        setFieldValue(formatValue(value, maxDecimals))
+      // eslint-disable-next-line react-hooks/exhaustive-deps
+    }, [fieldValue, value])
+
+    return (
+      <Input
+        type="text"
+        ref={ref}
+        value={fieldValue}
+        onChange={handleChange}
+        {...props}
+      />
+    )
+  }
+)
--- a/lib/components/Forms/Fields/NumberInput/index.tsx
+++ b/lib/components/Forms/Fields/NumberInput/index.tsx
@@ -3,12 +3,7 @@ import { ChevronDown, ChevronUp } from "@/icons"
 import { Input } from "@/ui/input"
 import { forwardRef, useEffect, useState } from "react"
 import { InputProps } from "../Input"
-
-const FORMAT = /^([0-9]+)$/
-// eslint-disable-next-line no-useless-escape
-const FORMAT_WITH_DECIMALS = /^([0-9]+)(?:[\.,]([0-9]+)?)?$/
-
-const parseValue = (value: string) => parseFloat(value.replace(",", "."))
+import { extractNumber } from "./extractNumber"
 
 const formatValue = (value: number, locale: string, maxDecimals?: number) =>
   new Intl.NumberFormat(locale, {
@@ -33,27 +28,13 @@ export const NumberInput = forwardRef<HTMLInputElement, NumberInputProps>(
     )
 
     const handleChange = (value: string) => {
-      if (!value) {
-        setFieldValue("")
-        return onChange?.(null)
-      }
+      const extractedData = extractNumber(value, { maxDecimals })
+      if (!extractedData) return
 
-      const match = value.match(
-        maxDecimals !== 0 ? FORMAT_WITH_DECIMALS : FORMAT
-      )
-      if (!match) return
+      const { formattedValue, value: parsedValue } = extractedData
 
-      const [_, integers, decimals] = match
-      if (maxDecimals && decimals && decimals.length > maxDecimals) return
-
-      const valueWithNoLeadingZero = value.match(/^0\d$/)
-        ? value.slice(1)
-        : value
-
-      const valueAsNumber = parseFloat(`${integers}.${decimals ?? 0}`)
-
-      setFieldValue(valueWithNoLeadingZero)
-      onChange?.(valueAsNumber)
+      setFieldValue(formattedValue)
+      onChange?.(parsedValue)
     }
 
     const handleStep = (type: "increase" | "decrease") => () => {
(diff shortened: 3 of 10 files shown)


lib/components/Forms/Fields/NumberInput/index.tsx
```diff
@@ -0,0 +1,77 @@
...
+import { Input, InputProps } from "../Input"
+
+const FORMAT = /^([0-9]+)$/
+// eslint-disable-next-line no-useless-escape
+const FORMAT_WITH_DECIMALS = /^([0-9]+)(?:[\.,]([0-9]+)?)?$/
+
+const parseValue = (value: string) => parseFloat(value.replace(",", "."))
+
+type NumberInputProps = Omit<InputProps, "value" | "type" | "onChange"> & {
+  locale: string
```
viniciusmeneses, line 11: The locale is a required prop, I'm wondering if there's a better approach...

viniciusmeneses:
@josepjaume porting the numeric input to Factorial One as you suggested [here](https://github.com/factorialco/factorial/pull/47179#pullrequestreview-2205152491)

lib/components/Forms/Fields/Input/index.tsx
```diff
@@ -1,11 +1,13 @@
...
+import { ComponentProps, HTMLInputTypeAttribute } from "react"
 
-type InputProps = Pick<
+export type InputProps = Pick<
   ComponentProps<typeof ShadcnInput>,
-  "ref" | "type" | "disabled" | "size" | "onChange" | "value" | "placeholder"
->
+  "ref" | "disabled" | "size" | "onChange" | "value" | "placeholder"
+> & {
+  type?: Exclude<HTMLInputTypeAttribute, "number">
```
nlopin, line 9: Why should we ban a `number` type? I understand they intersect with the `NumberInput`, but they are not the same.

`NumberInput` does not cover the case when you have a counter with steps (classic `type="number"` use case)

I think we should keep the number type available. 

@josepjaume wdyt?

lib/components/Forms/Fields/NumberInput/index.tsx
nlopin: we can have locale undefined. according to the docs it will use the locale of the runtime (i.e. browser) will be used:
> The runtime's default locale is used when undefined is passed or when none of the specified locale identifiers is supported.

viniciusmeneses: Yes, but making the locale prop optional can lead us to inconsistencies when the browser is in a different language than the user-defined language in Factorial, people will eventually forget to pass the prop if it's not required...

Rogermax: I agree with vinicius, I wouldn't make it optional. If this component was in factorial we could do it because we can access to the employee locale, but here...

lib/components/Forms/Fields/NumberInput/index.tsx
```diff
@@ -0,0 +1,77 @@
...
+    // For example, if the input value is 18,3 and I want to change it to 18,5. I'll delete the last digit and the input will become "18,"
+    // At this point, we have fired the "onChange" callback with the value 18 because it's the most precise valid value ("18," isn't a valid one),
+    // then the useEffect below will be triggered, since the "value" changed from "18,3" to "18", and it'll not modify the "fieldValue" to keep the decimal separator
+    // unless if the value really changed.
+    useEffect(() => {
+      if (value === null) return setFieldValue("")
+      if (value != null && parseValue(fieldValue) !== value)
+        setFieldValue(formatValue(value, maxDecimals))
+      // eslint-disable-next-line react-hooks/exhaustive-deps
+    }, [fieldValue, value])
```
Rogermax, line 102: Can we change this for a const? I think is more performance, because useEffect requieres extra renders

```suggestion
const fieldValue = !value || parseValue(value) !== value ? formatValue(value) : value
```

lib/components/Forms/Fields/NumberInput/index.tsx
```diff
@@ -0,0 +1,77 @@
...
+  ({ locale, value, maxDecimals, onChange, ...props }, ref) => {
+    const formatValue = (value: number, maxDecimals?: number) =>
+      new Intl.NumberFormat(locale, {
+        maximumFractionDigits: maxDecimals,
+        useGrouping: false,
+      }).format(value)
+
+    const [fieldValue, setFieldValue] = useState<string>(() =>
+      value != null ? formatValue(value, maxDecimals) : ""
+    )
```
Rogermax, line 33: this can be removed with the other suggestion

lib/components/Forms/Fields/NumberInput/index.tsx
```diff
@@ -0,0 +1,77 @@
...
+  onChange?: (value: number | null) => void
+}
+
+export const NumberInput = forwardRef<HTMLInputElement, NumberInputProps>(
+  ({ locale, value, maxDecimals, onChange, ...props }, ref) => {
+    const formatValue = (value: number, maxDecimals?: number) =>
+      new Intl.NumberFormat(locale, {
+        maximumFractionDigits: maxDecimals,
+        useGrouping: false,
+      }).format(value)
```
Rogermax, line 29: This function seems a utility, I will move it out (adding the locale as a param)

lib/components/Forms/Fields/NumberInput/index.tsx
```diff
@@ -0,0 +1,118 @@
...
+
+    const [fieldValue, setFieldValue] = useState<string>(() =>
+      value != null ? formatValue(value, maxDecimals) : ""
+    )
+
+    const handleChange = (value: string) => {
+      if (!value) {
+        setFieldValue("")
```
Rogermax, line 37: With the other changes this can be removed

lib/components/Forms/Fields/NumberInput/index.tsx
```diff
@@ -0,0 +1,118 @@
...
+      const [_, integers, decimals] = match
+      if (maxDecimals && decimals && decimals.length > maxDecimals) return
+
+      const valueWithNoLeadingZero = value.match(/^0\d$/)
+        ? value.slice(1)
+        : value
+
+      const valueAsNumber = parseFloat(`${integers}.${decimals ?? 0}`)
+
+      setFieldValue(valueWithNoLeadingZero)
```
Rogermax, line 55: With the other changes this can be removed

Rogermax (commented):
Vinicius... I made my suggestions by mind.. maybe I'm wrong and you can't do it this way because you lose the decimals separator.. but my intuition tell me that the useEffect can be removed by a const

lib/components/Forms/Fields/Input/index.tsx
viniciusmeneses: > Why should we ban a number type? I understand they intersect with the NumberInput, but they are not the same.

If we allow both, we'll have different UX across the app as pointed out [here](https://github.com/factorialco/factorial/pull/47179), so the idea is to have a single way to render a numeric input to avoid those issues.

> NumberInput does not cover the case when you have a counter with steps (classic type="number" use case)

The gamma version has the steps and I just added it for this one as well. I opened the PR some weeks ago without it because at that time I didn't manage to add the icons for some reason that I don't remember 😅 

https://github.com/user-attachments/assets/d2998f77-16a1-4ec4-9b39-cca421459196

lib/components/Forms/Fields/NumberInput/index.tsx
viniciusmeneses: We must have both a state (`fieldValue`) and the `useEffect` inside this component to handle [this scenario that we already discussed in Gamma's PR](https://github.com/factorialco/factorial/pull/47179#discussion_r1695697766). If we don't have another state, it makes impossible to keep the decimal separator while the user is typing.

Rogermax: But if the user type a , and this has no sence (because in the local defined should be a .) makes sense to replace it while typing no?

Rogermax: Yeah I remember... The problem is the input is a number, then you loose that information (the decimal point)... ok ok

viniciusmeneses: > But if the user type a , and this has no sence (because in the local defined should be a .) makes sense to replace it while typing no?

It depends if we want to force to use the same separator defined in the locale or not. Right now we don't restrict that, so we allow both separators independently of the user locale. I've implemented it like this because that's the same behavior of the native browser's `input type="number"`.

Rogermax: Then if I'm in Spanish can I type 1.223,43?

viniciusmeneses: Not really because we don't support thousand separators for now. to keep the same behavior of the native input type number

Rogermax (approved):
Maybe we shouldn't replicate current type=number input standard behaviour, and make one better ;)

But for me its ok for now

viniciusmeneses:
@Rogermax Totally, I’m just replicating somethings to maintain what users are used to, but we can change/improve later for sure

lib/components/Forms/Fields/Input/index.tsx
josepjaume: Interesting! I'm also struggling to decide to which extent we want to mask the underlying `input` props in our components or not. I'm leaning a bit more towards @viniciusmeneses's approach. passing the `type` to `input` feels like it belongs to an underlying layer.

Should we explicitly provide `PasswordInput`, `TextInput` and things like these instead (following the `NumberInput` example)? They convey more meaning and one could see their UX evolving in slightly different ways (passwords having a built-in complexity validation, numbers having arrows...).

lib/components/Forms/Fields/NumberInput/index.stories.tsx
```diff
@@ -0,0 +1,58 @@
...
+    disabled: false,
+    placeholder: "Placeholder text here",
+    locale: "en-US",
+  },
+  argTypes: {
+    value: {
+      control: { type: "number" },
+    },
+    locale: {
+      options: ["en-US", "es-ES", "pt-BR"],
```
josepjaume, line 19: Wondewring if we can type the allowed locales with a TypeScript `enum`?

lib/components/Forms/Fields/NumberInput/index.tsx
```diff
@@ -0,0 +1,118 @@
...
+      if (value === null) return setFieldValue("")
+      if (value != null && parseValue(fieldValue) !== value)
+        setFieldValue(formatValue(value, locale, maxDecimals))
+      // eslint-disable-next-line react-hooks/exhaustive-deps
+    }, [fieldValue, value])
+
+    return (
+      <div className="group relative">
+        <Input
+          type="text"
```
rubenmoya, line 107: Using `type="text"` would be problematic in mobile phones since it won't use the numeric keyboard, like the `number` input does. We can fix it by specifying an `inputmode` ([more info](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/inputmode))

rubenmoya (commented):
This component looks quite complex to not have tests. We should check that different separators, locales, and numbers are formatted as we expect. And we should also test that the increment/decrement works as expected.

lib/components/Forms/Fields/NumberInput/index.stories.tsx
viniciusmeneses: TS doesn't offer an `enum` or `union` with the locales supported by `Intl`, so I think the only way would be to define a union [like this](https://gist.github.com/g4rcez/e072eb9e1203679645985a9c89722117#file-locale-ts) (or with just the main locales used in Factorial). 

However, something to have in mind is that `factorial-i18n` doesn't type the locale as well, so when rendering this component inside Factorial we'd need to assert the type: `<NumberInput locale={i18n.locale as Locales} />`

WDYT?

josepjaume: Okay, I understand. If a locale list was somehow standardized, that would probably be the way to go - but since it's not the case, better to leave it as a string and provide some options just on storybook, exactly like you did.

The other way around looks safer at a first glance, but it's equally dangerous for the reasons you mention (bound to do an assertion at one point or another!).

josepjaume:
As @rubenmoya says, this might benefit from test. Good news is - we have `vitest` in Factorial One, which is extremely faster and also has a browser extension that makes everything super easy!

There's a couple of examples already, maybe we could explore that?

lib/components/Forms/Fields/Input/index.tsx
nlopin: I'd rather rendered `NumberInput` when an engineer types `<Input type="number">`, but both approaches work as long as they are consistently applied.

It would be nice to somehow notify the user about the alternative

lib/components/Forms/Fields/NumberInput/index.tsx
```diff
@@ -0,0 +1,118 @@
+import { Icon } from "@/components/Utilities/Icon"
+import { ChevronDown, ChevronUp } from "@/icons"
+import { Input } from "@/ui/input"
+import { forwardRef, useEffect, useState } from "react"
+import { InputProps } from "../Input"
+
+const FORMAT = /^([0-9]+)$/
+// eslint-disable-next-line no-useless-escape
+const FORMAT_WITH_DECIMALS = /^([0-9]+)(?:[\.,]([0-9]+)?)?$/
```
lochiego, line 9: This numberinput doesn't support negative numbers. I have a [PR ](https://github.com/factorialco/factorial/pull/49953)to introduce negative support to NumberField in gamma, if it's accepted perhaps you could port that logic to this component.

lib/components/Forms/Fields/Input/index.tsx
josepjaume: The problem with this @nlopin is that I can't see how TypeScript would apply the right types. The signature of the props of `<NumberInput />` will potentially be different from `Input` in the mid term (that's why we're creating a separate component) so that would get in the middle.

lib/components/Forms/Fields/NumberInput/index.tsx
```diff
@@ -0,0 +1,118 @@
...
+    // For example, if the input value is 18,3 and I want to change it to 18,5. I'll delete the last digit and the input will become "18,"
+    // At this point, we have fired the "onChange" callback with the value 18 because it's the most precise valid value ("18," isn't a valid one),
+    // then the useEffect below will be triggered, since the "value" changed from "18,3" to "18", and it'll not modify the "fieldValue" to keep the decimal separator
+    // unless if the value really changed.
+    useEffect(() => {
+      if (value === null) return setFieldValue("")
+      if (value != null && parseValue(fieldValue) !== value)
+        setFieldValue(formatValue(value, locale, maxDecimals))
+      // eslint-disable-next-line react-hooks/exhaustive-deps
+    }, [fieldValue, value])
```
josepjaume, line 102: I wonder if this could be done in `handleChange` to avoid having to observe both fields and trigger an extra re-render?

lib/components/Forms/Fields/Input/index.tsx
lochiego: > ...passing the type to input feels like it belongs to an underlying layer.

> Should we explicitly provide PasswordInput, TextInput and things like these instead (following the NumberInput example)?

💯 agree with this approach. I prefer separate components that provide a more constrained set of handles that impact what we are trying to do specifically in that context. It lets our interfaces read cleaner (values, handlers, etc.) and the implementation can be more cleanly maintained than in a component that handles each case or internally delegates to another component (which gets us back to this).

> The signature of the props of <NumberInput /> will potentially be different from Input in the mid term

*Technically* we could support this using Union Types, but not my favorite. In the following excerpt TS will correctly infer that `maxDecimals` is not supported when type is `password` and the type of the incoming value for `onChange`:

Screenshot 2024-09-12 at 10 55 45

lib/components/Forms/Fields/NumberInput/index.tsx
lochiego: Merged the PR mentioned, that should help simplify the useEffect need as well.

lib/components/Forms/Fields/Input/index.tsx
josepjaume: > Technically we could support this using Union Types


Yeah, union types will always be a PITA (autocompletes are awkward) and we'd end up having a pretty complicated bunch of conditionals to distinguish components. It feels to me that going the other way around is going to be more effective: separate entrypoints, and sharing logic as we see fit in their implementations (as opposed to sharing logic by default and then having to artifficially separate the code paths).

lochiego: 💯

lib/components/Forms/Fields/NumberInput/index.tsx
viniciusmeneses: Thanks! I ported that logic to here 🙂

lib/components/Forms/Fields/NumberInput/index.tsx
viniciusmeneses: Yes, I removed this by porting the same logic of [this PR](https://github.com/factorialco/factorial/pull/49953) to here.

tsconfig.json
```diff
@@ -24,7 +24,8 @@
     /* Shadcn */
     "baseUrl": ".",
     "types": [
-      "vite/client"
+      "vite/client",
+      "@testing-library/jest-dom"
```
viniciusmeneses, line 28: Adding this to be able to use RTL assertions

lib/components/Forms/Fields/NumberInput/index.tsx
lochiego: There's an important consideration here, and that's that in the linked PR I overlooked that NumberField is an uncontrolled component. I had to add this hook back in to support a case where the value is changed externally (as in the *new mileage expense* form) in #50139, but then update the comparison logic in #50173.

viniciusmeneses: Thanks for letting me know, I completely forgot about this 😅

lib/components/Forms/Fields/NumberInput/index.tsx
nlopin: Thanks for this learning, folks!

nlopin (approved):

lochiego:
I don't know if you want to integrate this here, but I have [another PR](https://github.com/factorialco/factorial/pull/50247) to the NumberField in gamma to honor the `min` and `max` props when user types.

nlopin:
Let's make it a separate PR if it is not a huge issue for you.

PR #439, opened 2024-08-23
FCT-18022: create tasks insight

RamProg:
## 🚪 Why?

Part of the new Profile work, we need a widget to show Tasks

## 🔑 What?

Creates a new component which is an Insight for Tasks 

## 🏡 Context

https://www.figma.com/design/9pLsRzdinid0ha0qRWta2D/Employee-Profile?node-id=604-7408&t=MkGHhq2Zza655sXj-0

## 📸 Visual proof

image

### DIFF
--- a/src/playground/Insights/ui/separator.tsx
+++ b/src/playground/Insights/ui/separator.tsx
@@ -1,0 +1,13 @@
+import React from "react"
+
+export const Separator: React.FC = () => {
+  return (
+    <div
+      className="my-4 h-[1px] w-full"
+      style={{
+        backgroundImage:
+          "repeating-linear-gradient(to right, hsl(var(--secondary-intermediate)) 0, hsl(var(--secondary-intermediate)) 3px, transparent 3px, transparent 7px)",
+      }}
+    />
+  )
+}
--- a/src/playground/Insights/Tasks/index.stories.tsx
+++ b/src/playground/Insights/Tasks/index.stories.tsx
@@ -1,0 +1,36 @@
+import type { Meta, StoryObj } from "@storybook/react"
+
+import { TasksInsight } from "."
+
+const meta: Meta = {
+  component: TasksInsight,
+  parameters: {
+    tags: ["autodocs"],
+  },
+  args: {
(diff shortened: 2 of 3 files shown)


sauldom102 (approved):
LGTM! ✅

src/playground/Insights/ui/separator.tsx
```diff
@@ -0,0 +1,13 @@
+import React from "react"
+
+export const Separator: React.FC = () => {
+  return (
+    <div
+      className="my-4 h-[1px] w-full"
+      style={{
+        backgroundImage:
+          "repeating-linear-gradient(to right, hsl(var(--secondary-intermediate)) 0, hsl(var(--secondary-intermediate)) 3px, transparent 3px, transparent 7px)",
```
dani-moreno, line 9: Can't this be done with a `border-dashed`?

RamProg: The think is that border dashed looks like - - - - - and the design is more like _ _ _ _ _ so I had to do this workaround.

dani-moreno: Not sure I’m following, do you mean the line is thicker in the design? If that's the case, I'm pretty sure we can find a happy medium and keep the separator simpler (even though I do like the workaround!).

RamProg: Sorry, I thought I did a great job explaining that, but evidently I didn't. The default dashed line in css is very different to the one in Figma
image
image

dani-moreno: Oh, gotcha! You did a fantastic job explaining it, just a classic Monday for me 😅. Still wondering if we're overcomplicating things just for a separator, but let's roll with it!

dani-moreno (approved):

src/playground/Insights/ui/separator.tsx
RamProg: This is super easy to rollback and won't create any dependencies, so we can move forward with it and change it afterwards if we decide it.

PR #440, opened 2024-08-25
feat: add numeric summaries widget

hossamElfarFactorial:
## 🚪 Why?
We have some usecase where we need to display numeric values without charts.
## 🔑 What?
This PR adds a `SummariesWidget` widget that can take `summaries` prop and displays  numeric values same as the `summaries` we are displaying for the other widgets with charts

## 📸 Visual proof

image

### DIFF
--- a/lib/components/Widgets/WidgetContainer/index.tsx
+++ b/lib/components/Widgets/WidgetContainer/index.tsx
@@ -34,17 +34,17 @@ const Container = forwardRef<
         <div className="flex flex-1 flex-col truncate">
           <div className="flex flex-row justify-between">
             <div className="flex min-h-6 grow flex-row items-center gap-1.5 truncate">
-              {header?.title && <CardTitle>{header?.title}</CardTitle>}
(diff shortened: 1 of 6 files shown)


lib/components/Widgets/WidgetContainer/index.tsx
```diff
@@ -29,21 +29,25 @@ const Container = forwardRef<
...
+              {header?.info && <CardInfo content={header?.info} />}
+            </div>
+            {header?.link && (
+              <CardLink href={header.link.url} title={header.link.title} />
+            )}
           </div>
-          {header.link && (
-            <CardLink href={header.link.url} title={header.link.title} />
-          )}
+          {header?.comment && <CardComment>{header.comment}</CardComment>}
```
RamProg, line 47: The truthy value of header is being checked in 32 already right? Looks like the code might be checking the same value several times.

RamProg (approved):
Minor comment

PR #445, opened 2024-08-26
Feat/fct 18131/create insight timeoff

RamProg:
## 🚪 Why?

Part of the new Profile work, we are creating insights

## 🔑 What?

This PR creates a Timeoff widget

## 🏡 Context

https://www.figma.com/design/9pLsRzdinid0ha0qRWta2D/Employee-Profile?node-id=604-7408&t=dFJO9AADh1bYqZcy-0

## 📸 Visual proof

image

### DIFF
--- a/src/playground/Insights/Tasks/index.tsx
+++ b/src/playground/Insights/Tasks/index.tsx
@@ -1,7 +1,7 @@
 import { Button } from "@/components/Actions/Button"
+import { WidgetContainer } from "@/components/Widgets/WidgetContainer"
 import { Circle } from "@/icons"
 import AlertCircle from "@/icons/AlertCircle"
-import { Card, CardContent, CardHeader, CardLink, CardTitle } from "@/ui/card"
 import { forwardRef } from "react"
 import { Indicator } from "../ui/indicator"
 import { Separator } from "../ui/separator"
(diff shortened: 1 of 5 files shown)


src/playground/Insights/ui/event.tsx
```diff
@@ -0,0 +1,49 @@
+import { cn } from "@/lib/utils"
+import { forwardRef } from "react"
+
+interface EventProps {
+  title: string
+  length: number
+  lengthUnit: string
+  from: string
+  until: string
```
sauldom102, line 9: Don't you think we should leave this more open to `title` & `subtitle` props only? And then in Factorial we just form these texts based on the data that we have

sauldom102 (commented):
These are shifts in the mobile app, they have some icons too. Don't we have those in web?

Screenshot 2024-08-27 at 09 42 42

src/playground/Insights/Tasks/index.tsx
```diff
@@ -67,49 +67,48 @@ export const TasksInsight = forwardRef<HTMLDivElement, TasksInsightProps>(
...
+        <WidgetContainer
+          header={{
+            title,
+            link: { title: linkTitle, url: linkUrl },
+          }}
+        >
+          <div className="grid grid-cols-3">
+            {taskCategories.map(({ label, count, icon, color }) => (
+              <Indicator
+                label={label}
```
sauldom102, line 79: We're missing a `key` here

src/playground/Insights/Tasks/index.tsx
```diff
@@ -67,49 +67,48 @@ export const TasksInsight = forwardRef<HTMLDivElement, TasksInsightProps>(
...
+                count={count}
+                icon={icon}
+                color={color}
+              />
+            ))}
+          </div>
+          <Separator />
+          <div className="flex flex-col gap-3">
+            {!tasks?.length ? (
+              <p>{emptyStateText ?? "No tasks"}</p>
```
sauldom102, line 89: I think this text should be mandatory as a prop and not assign any default here

src/playground/Insights/Timeoff/index.tsx
```diff
@@ -0,0 +1,105 @@
...
+                    until={item.until}
+                    color={item.color}
+                    isPending={true}
+                  />
+                ))
+            )}
+          </div>
+          <Separator />
+          <p className="mb-4 font-medium text-muted-foreground">
+            Current and upcoming
```
sauldom102, line 75: This text should come localized in a prop right? 🤔

src/playground/Insights/Tasks/index.tsx
RamProg: Makes sense

src/playground/Insights/Tasks/index.tsx
RamProg: Good catch

RamProg:
> These are shifts in the mobile app, they have some icons too. Don't we have those in web?
> 
> Screenshot 2024-08-27 at 09 42 42

I'm not sure, we would need to ask @lily-factorial 
Looks like we can always ask them later, right?

sauldom102:
> > These are shifts in the mobile app, they have some icons too. Don't we have those in web?
> > Screenshot 2024-08-27 at 09 42 42
> 
> I'm not sure, we would need to ask @lily-factorial Looks like we can always add them later, right?

Sure! It just took my attention cause I recently did this component in mobile

PR #448, opened 2024-08-27
Adds Category Bar Chart

dani-moreno:
Adds a new type of chart to our Charts category: **Category Bar**, as we have in [Figma](https://www.figma.com/design/O8578qyVUfgzm6ZkVkBJwW/%E2%8C%9B%EF%B8%8F-Insight-UI-Kit-2.0?node-id=500-43508&t=ZmEsXqhlWOXmyzRJ-4).

https://github.com/user-attachments/assets/baa2c67d-8dd4-4c99-831a-8176e7b3df09

### DIFF
--- a/lib/components/Charts/CategoryBarChart/index.tsx
+++ b/lib/components/Charts/CategoryBarChart/index.tsx
@@ -40,10 +40,14 @@ export function CategoryBar({ data, legend = true }: CategoryBarProps) {
                   className="h-full cursor-default overflow-hidden rounded-[2px]"
                   style={{ width: `${percentage}%` }}
                   title={category.name}
+                  asChild
                 >
                   <div
                     className="relative h-full w-full"
                     style={{ backgroundColor: color }}
+                    role="img"
+                    title={category.name}
+                    tabIndex={0}
                   >
                     {category.muted && (
                       <div
@@ -51,8 +55,8 @@ export function CategoryBar({ data, legend = true }: CategoryBarProps) {
                         style={{
                           backgroundImage: `repeating-linear-gradient(
                           135deg,
-                          rgba(255,255,255,0.1),
-                          rgba(255,255,255,0.1) 2px,
+                          rgba(255,255,255,0.2),
+                          rgba(255,255,255,0.2) 2px,
                           transparent 2px,
                           transparent 6px
                         )`,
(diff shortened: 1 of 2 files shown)


lib/components/Charts/CategoryBarChart/index.tsx
```diff
@@ -0,0 +1,113 @@
...
+              const percentage = (value / total) * 100
+              return percentage % 1 === 0
+                ? percentage.toFixed(0)
+                : percentage.toFixed(1)
+            }
+
+            return (
+              <Tooltip key={category.name}>
+                <TooltipTrigger
+                  className="h-full cursor-default overflow-hidden rounded-[2px]"
```
RamProg, line 40: I think we are avoiding the custom values like rounded-[2px], but I understand it's the only way

lib/components/Charts/CategoryBarChart/index.tsx
```diff
@@ -0,0 +1,113 @@
...
+                    tabIndex={0}
+                  >
+                    {category.muted && (
+                      <div
+                        className="absolute inset-0"
+                        style={{
+                          backgroundImage: `repeating-linear-gradient(
+                          135deg,
+                          rgba(255,255,255,0.2),
+                          rgba(255,255,255,0.2) 2px,
```
RamProg, line 59: can we use an existing color so it's compatible with dark mode?

lib/components/Charts/CategoryBarChart/index.tsx
```diff
@@ -0,0 +1,113 @@
...
+                          transparent 6px
+                        )`,
+                        }}
+                      />
+                    )}
+                  </div>
+                </TooltipTrigger>
+                <TooltipContent className="flex items-center gap-1 text-sm">
+                  <div
+                    className="h-2.5 w-2.5 shrink-0 translate-y-[1px] rounded-full"
```
RamProg, line 70: same here about custom values

lib/components/Charts/CategoryBarChart/index.tsx
```diff
@@ -0,0 +1,113 @@
...
+            const color = category.color || autoColor(index)
+
+            return (
+              <div
+                key={category.name}
+                className="flex items-center gap-1"
+                role="listitem"
+              >
+                <div
+                  className="h-2 w-2 shrink-0 translate-y-[1px] rounded-full"
```
RamProg, line 100: same here about custom values

RamProg (approved):
Nice! I left a few comments to take a look

lib/components/Charts/CategoryBarChart/index.tsx
dani-moreno: Good catch, there's even a Tailwind class that does the same thing, `translate-y-px`, so no need for a custom value! Changing it!

lib/components/Charts/CategoryBarChart/index.tsx
dani-moreno: Umm, this is being displayed on top of the bar, and with the opacity it looks good IMO no matter the theme.
image

PR #449, opened 2024-08-27
adjust area chart styling and aspect

hossamElfarFactorial:
## 🚪 Why?
- I needed to add a new aspect to the area chart.
- The current line type caused the area chart lines to goes above the max value or below the min value if the values between the ticks fluctuates between the min and max. 

## 🔑 What?
- Add a new aspect `small` and use by default for the area chart widget
- Add `recharts` line types and use `monotoneX` as a default line type
- Fix the summary value when the value is decimal
- Always show the x axis ticks by setting `interval` to 0. This is required as using the default value `preserveEnd ` could potential drop some ticks from the x axis.
- Show the area chart legend only if we have more than 1 area, and hide it otherwise
## 📸 Visual proof
image

### DIFF
--- a/lib/components/Charts/AreaChart/index.tsx
+++ b/lib/components/Charts/AreaChart/index.tsx
@@ -21,22 +21,7 @@ import { fixedForwardRef } from "../utils/forwardRef"
 import { prepareData } from "../utils/muncher"
 import { ChartConfig, ChartPropsBase, InferChartKeys } from "../utils/types"
 
-type allowedLineTypes =
-  | "basis"
-  | "basisClosed"
-  | "basisOpen"
-  | "bumpX"
-  | "bumpY"
-  | "bump"
-  | "linear"
-  | "linearClosed"
-  | "natural"
-  | "monotoneX"
-  | "monotoneY"
-  | "monotone"
-  | "step"
-  | "stepBefore"
-  | "stepAfter"
+type allowedLineTypes = "natural" | "linear" | "step" | "monotoneX"
 
 export type AreaChartProps<
   DataConfig extends ChartConfig = ChartConfig,
--- a/lib/components/Charts/AreaChart/index.tsx
+++ b/lib/components/Charts/AreaChart/index.tsx
@@ -1,4 +1,11 @@
-import { ChartContainer, ChartTooltip, ChartTooltipContent } from "@/ui/chart"
+import { cn } from "@/lib/utils"
+import {
+  ChartContainer,
+  ChartLegend,
+  ChartLegendContent,
+  ChartTooltip,
+  ChartTooltipContent,
+} from "@/ui/chart"
 import { nanoid } from "nanoid"
 import { ForwardedRef } from "react"
 import {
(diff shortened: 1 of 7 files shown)


lib/components/Charts/AreaChart/index.tsx
```diff
@@ -25,7 +18,22 @@ export type AreaChartProps<
   DataConfig extends ChartConfig = ChartConfig,
   Keys extends string = InferChartKeys<DataConfig>,
 > = ChartPropsBase<DataConfig, Keys> & {
-  lineType?: "natural" | "linear" | "step"
+  lineType?:
```
elacallefactorial, line 21: You might create a type in order no to have a long piece of code with the allowed line types

`type lineType =  "basis"  | "basisClosed" | "basisOpen" | "bumpX" | "bumpY" | "bump"....`

```
  DataConfig extends ChartConfig = ChartConfig,
  Keys extends string = InferChartKeys<DataConfig>,
> = ChartPropsBase<DataConfig, Keys> & {
  lineType?: lineType
```

lib/components/Widgets/Charts/ChartContainer.tsx
```diff
@@ -20,18 +19,18 @@ const Container = forwardRef<
             <div className="mb-0.5 text-sm text-muted-foreground">
               {summary.label}
             </div>
-            <div className="flex flex-row items-end gap-0.5">
-              <ChartCounter value={summary.value} />
```
dani-moreno, line 24: Why was this needed? if it is because of decimals, we can probably adjust the ChartCounter component.
Not sure about the unit being a prefix instead of a suffix. Maybe we need both cases?

lib/components/Charts/AreaChart/index.tsx
```diff
@@ -21,11 +21,28 @@ import { fixedForwardRef } from "../utils/forwardRef"
 import { prepareData } from "../utils/muncher"
 import { ChartConfig, ChartPropsBase, InferChartKeys } from "../utils/types"
 
+type allowedLineTypes =
```
dani-moreno, line 24: We probably don’t need all of them, right? Not entirely sure what the monotoneX one does, but it seems to get the job done. Maybe we can stick with that and simplify the options?
(This would be a great topic to discuss with Desi).

lib/components/Charts/AreaChart/index.tsx
```diff
@@ -118,11 +136,13 @@ export const _AreaChart = <
             strokeWidth={1.5}
           />
         ))}
-        <ChartLegend
-          className={cn("flex justify-start", fullWidth && "ml-10")}
-          iconType="star"
-          content={<ChartLegendContent />}
-        />
+        {Object.keys(dataConfig).length > 1 && (
```
dani-moreno, line 139: Yay, perfect! 🙌

dani-moreno (changes requested):

lib/components/Charts/AreaChart/index.tsx
hossamElfarFactorial: [remove extra area chart line charts](https://github.com/factorialco/factorial-one/pull/449/commits/33c7ed9153eb2c955673b76044dcf202484c3d95)

lib/components/Widgets/Charts/ChartContainer.tsx
hossamElfarFactorial: [fix chart counter decimal points and add post and prefix units to cha…](https://github.com/factorialco/factorial-one/pull/449/commits/fd1c0f890c4f58327ef8c7c17ad55932c69d3f26)

dani-moreno (approved):
Nice, looks good @hossamElfarFactorial !

The ChartCounter animation looks a bit weird with decimals now, but that's something we can take a look in another PR!

Rogermax:
I have a PR that makes charts reponsive (without aspect-ratio), can you check this PR: https://github.com/factorialco/factorial-one/pull/454 ?

PR #453, opened 2024-09-02
Dashed config

Rogermax:
## 🔑 What?

Add dashed config to ChartConfig

https://github.com/user-attachments/assets/799e4660-8447-44c7-8283-3640001ba350


dani-moreno:
Looks good, thanks @Rogermax!

Told Desi (Insights designer) about your PR yesterday, and she already had the dashed line in mind. But we weren't sure about the dashed bar chart, do you have a use case for it?

Rogermax:
> Looks good, thanks @Rogermax!
> 
> Told Desi (Insights designer) about your PR yesterday, and she already had the dashed line in mind. But we weren't sure about the dashed bar chart, do you have a use case for it?

No, but since it's a property common to all the charts, I wanted to create a complex example of how it could be implemented for the other cases.

dani-moreno:
> > Looks good, thanks @Rogermax!
> > Told Desi (Insights designer) about your PR yesterday, and she already had the dashed line in mind. But we weren't sure about the dashed bar chart, do you have a use case for it?
> 
> No, but since it's a property common to all the charts, I wanted to create a complex example of how it could be implemented for the other cases.

Ah yes, got it. Not sure if it is possible to only have it available on the line chart, as it looks a bit weird in other charts (I'm a bit worried about opening a _caja de pandora_ here).

Rogermax:
> > > Looks good, thanks @Rogermax!
> > > Told Desi (Insights designer) about your PR yesterday, and she already had the dashed line in mind. But we weren't sure about the dashed bar chart, do you have a use case for it?
> > 
> > 
> > No, but since it's a property common to all the charts, I wanted to create a complex example of how it could be implemented for the other cases.
> 
> Ah yes, got it. Not sure if it is possible to only have it available on the line chart, as it looks a bit weird in other charts (I'm a bit worried about opening a _caja de pandora_ here).

...the problem is that I can't found any better place to add this config. Sure we can make a type only for Line charts that use an extended type for that config

I don't know :)

Rogermax:
@dani-moreno talking with Desirée I removed from BarChart the dashed type and also simplified the types a bit

dani-moreno (approved):
Amazing @Rogermax! Thanks for the huge effort!

PR #454, opened 2024-09-02
Fix to allow charts be responsive

Rogermax:
## 🚪 Why?

To have responsive charts. Now we have by default "aspect: wide" (aspect-ratio: 16/9) this makes imposible that a line chart expand only in one direction. So for example a line chart that I want to be just 200px height and allow to expand full width is not posible, but the library recharts allow it: https://recharts.org/en-US/examples/ComposedResponsiveContainer
But we are wrapping the "ResponsiveContainer" with an aspect ratio loosing this responsive feature.

## 🔑 What?

* Remove the default aspect: wide
* If aspect is not given, then add css aspect-auto and h-full to let "Responsive" component to act correctly
* Modify stories to fix a nice height
* modify widgets to fix a height

https://github.com/user-attachments/assets/4d54a42e-8b2f-4a1a-bc5a-c2b3d9e67669

### DIFF
--- a/lib/ui/chart.tsx
+++ b/lib/ui/chart.tsx
@@ -70,7 +70,7 @@ const ChartContainerComponent = (
     id,
     className,
     children,
-    aspect = "wide",
+    aspect,
     config,
     ...props
   }: ChartContainerComponentProps,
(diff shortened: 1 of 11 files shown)


Rogermax:
@dani-moreno could you check this

lib/ui/chart.tsx
```diff
@@ -70,7 +70,7 @@ const ChartContainerComponent = (
     id,
     className,
     children,
-    aspect = "wide",
+    aspect,
```
Rogermax, line 73: this is like the most important change!

Rogermax:
@josepjaume @sauldom102 any thoughts? this PR does not break things just allow to not force people tie to an aspect ratio, allowing horizontal charts to expands as its container (really responsive) I need this to make posible this figma: 

Screenshot 2024-09-06 at 11 31 47

This long charts are not any of the current aspect-ratio

hossamElfarFactorial (approved):
The change LGTM

PR #455, opened 2024-09-02
Widget structure update

dani-moreno:
Updated the `WidgetContainer` with two common items:

- Action button on footer.
- Summaries (translated from `ChartContainer` to `WidgetContainer`).

image


lib/components/Widgets/Charts/chart-counter.tsx
dani-moreno, line 1: Got rid of this component. It was a simple animation, but it looked a bit off in some cases (like with decimals, for example). Keeping the summaries simple for now, will make another one better!

Rogermax (approved):

lib/components/Widgets/Charts/ChartContainer.tsx
```diff
@@ -1,14 +1,8 @@
...
-import ChartCounter from "./chart-counter"
 
 export type ChartContainerPropsBase = WidgetContainerProps & {
-  summaries?: Array<{
-    label: string
-    value: number
-    prefixUnit?: string
-    postfixUnit?: string
-  }>
+  summaries?: WidgetContainerProps["summaries"]
```
josepjaume, line 5: This shouldn't be needed as the `ChartContainerProps` now are just `WidgetContainerProps`. Have you tried just doing `ChartContainerPropsBase = WidgetContainerProps`?

lib/components/Widgets/WidgetContainer/index.stories.tsx
```diff
@@ -43,3 +44,13 @@ export const WithLink: Story = {
     },
   },
 }
+
+export const WithAction: Story = {
+  args: {
+    ...meta.args,
+    action: {
+      label: "Show more",
+      onClick: fn(),
```
josepjaume, line 53: Awesome!

lib/components/Widgets/WidgetContainer/index.tsx
```diff
@@ -22,12 +24,19 @@ export interface WidgetContainerProps {
     info?: string
     link?: { title: string; url: string }
   }
+  action?: ButtonProps
+  summaries?: Array<{
+    label: string
+    value: number
+    prefixUnit?: string
+    postfixUnit?: string
```
josepjaume, line 32: I feel this will eventually become part of a shared type to deal with units, but that's for the future

josepjaume (approved):
Good work!!

lib/components/Widgets/Charts/ChartContainer.tsx
dani-moreno: Ah, right! Done 🫶

PR #456, opened 2024-09-02
Remove `muted` prop from CategoryBar

dani-moreno:
I initially added a `muted` prop to the CategoryBar chart component since I noticed it was being used in several places within Factorial. This prop displayed that bar with a lines pattern, setting it apart from the others.

image

Even though it's in use, it doesn't really hold any meaning and lacks a genuine use case. After some analysis, Desi and I decided to get rid of it.

image

_The scientists are me basically._


sauldom102 (approved):
LGTM

PR #459, opened 2024-09-03
Add Progress Bar component

dani-moreno:
Adds a `Progress Bar` component, which is essential for the Employee Profile and Insights.

image

This `Progress Bar` serves as a wrapper for the `Progress` shadcn component, adding a bit more functionality, similar to what we do with our Button component, for example.

The component allows the user to define a label displayed to its right, and the color is customizable (or it defaults to one of our chart colors if not specified), exactly as we have in [Figma](https://www.figma.com/design/O8578qyVUfgzm6ZkVkBJwW/%E2%8C%9B%EF%B8%8F-Charts-UI-Kit-2.0?node-id=604-42741&t=c8rb5Q5q1GFByc3a-4).

### DIFF
--- a/lib/components/Charts/ProgressChart/index.tsx
+++ b/lib/components/Charts/ProgressChart/index.tsx
@@ -25,8 +25,8 @@ export function ProgressBar({
           value={percentage}
           className="w-full"
           aria-valuemin={0}
-          aria-valuemax={100}
-          aria-valuenow={percentage}
+          aria-valuemax={max}
+          aria-valuenow={value}
           aria-label={`${percentage.toFixed(1)}%`}
         />
       </div>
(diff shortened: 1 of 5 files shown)


lib/components/Charts/ProgressChart/index.tsx
```diff
@@ -0,0 +1,29 @@
...
+      <div className="flex-grow">
+        <Progress
+          color={barColor}
+          value={value}
+          className="w-full"
+          aria-label={`${value}%`}
+        />
+      </div>
+      {label && (
+        <div className="flex-shrink-0 text-sm font-medium" aria-live="polite">
```
nlopin, line 23: What if we move `aria-live` to the wrapper container? The `value` is also a subject of change and the user should be aware about its updates even if there is no label

nlopin (approved):
Hey! I commented a couple of small things, it looks great

lib/components/Charts/ProgressChart/index.tsx
dani-moreno: Thanks, great catch! And welcome, happy to have you on board! 🫶

PR #460, opened 2024-09-03
Update `WidgetContainer` content as array

dani-moreno:
This PR updates the WidgetContainer to support an array as children, displaying a separator between each one to achieve a result like this:

image

This means users can do this while using the WidgetContainer, providing a simpler approach for them:

image


RamProg (approved):
Great work!

josepjaume (approved):
This is super cool! Incredible work Dani, you're becoming a TypeScript/React master in no time!

PR #467, opened 2024-09-05
Add Alert prop to WidgetContainer component

sauldom102:
## 🔑 What?

Add Alert prop to WidgetContainer component.

## 🚪 Why?

By looking at Figma we can start to realize that there's a pattern with alerts in the widgets being part of the component.

## 👁️ Visual proof

Screenshot 2024-09-05 at 14 25 33


RamProg (approved):

nlopin:
Where do we use it? Could you share examples?

PR #468, opened 2024-09-05
List component for Projects and Competencies widget

dani-moreno:
## 🔑 What?

Adds a new `List` component to the Widgets folder.

This component has two variants: `List` (default) and `List.Progress`. The first is useful for the Projects widget, while the latter is designed for the Competencies widget.

There are also some other changes to make the widgets looks good:
- Added a separator between summaries and the content of a WidgetContainer when there are summaries.
- Ajusted the label of the ProgressBar to use [tabular numbers](https://arc.net/l/quote/ayptyjfp).

Also, I'm adding here:
- Two examples of this List component is used in the Projects and Competencies widget.
- Moved the Timesheet example to the "Examples" folder inside Widgets.

## 👀 A vé?
image
image


src/playground/Widgets/List/index.tsx
```diff
@@ -0,0 +1,63 @@
...
+}: {
+  item: ListItem
+  variant?: "default" | "progress"
+}) => (
+  <div className="mb-2 flex w-full items-center last:mb-0">
+    <span className="flex-grow truncate text-muted-foreground">
+      {item.name}
+    </span>
+    {variant === "default" ? (
+      <span className="ml-4 flex-shrink-0 text-right text-sm font-medium">
```
sauldom102, line 35: Do we need this `ml-4` here?

sauldom102: With `flex-grow` in the span on the left maybe we won't need `text-right` either

src/playground/Widgets/List/index.tsx
```diff
@@ -0,0 +1,63 @@
...
+  value: number
+}
+
+type ListItem = DefaultListItem | ProgressListItem
+
+interface ListProps {
+  items: ListItem[]
+  title?: string
+  variant?: "default" | "progress"
+}
```
sauldom102, line 21: We could improve this type to better restrict it, so that we don't allow anyone to pass both progress items and text items

```suggestion
type ListProps = {
  title?: string
} & ({
  items: DefaultListItem[]
} | {
  items: ProgressListItem[]
})
```

We won't need the `variant` either

src/playground/Widgets/List/index.tsx
```diff
@@ -0,0 +1,63 @@
...
+  variant = "default",
+}: {
+  item: ListItem
+  variant?: "default" | "progress"
+}) => (
+  <div className="mb-2 flex w-full items-center last:mb-0">
+    <span className="flex-grow truncate text-muted-foreground">
+      {item.name}
+    </span>
+    {variant === "default" ? (
```
sauldom102, line 34: According to the suggestion above, you won't need to check the variant here but `typeof item.value` instead

src/playground/Widgets/List/index.tsx
```diff
@@ -0,0 +1,63 @@
...
+    )}
+  </div>
+)
+
+const List = ({ items, title, variant = "default" }: ListProps) => (
+  <div className="w-full">
+    {title && <h2 className="mb-2 text-base font-medium">{title}</h2>}
+    <div className="w-full">
+      {items.map((item, index) => (
+        <ListItem key={index} item={item} variant={variant} />
```
sauldom102, line 51: We should avoid using `index` as key, but in case you need to use because we don't have any real unique attribute in our item, we should at least cast it to string like `${index}`

sauldom102 (commented):
We should probably come up with some other name for this `List` component, it sounds too generic for me personally

src/playground/Widgets/List/index.tsx
dani-moreno: Right! Made a few changes here.

src/playground/Widgets/List/index.tsx
```diff
@@ -0,0 +1,63 @@
...
+  <div className="mb-2 flex w-full items-center last:mb-0">
+    <span className="flex-grow truncate text-muted-foreground">
+      {item.name}
+    </span>
+    {variant === "default" ? (
+      <span className="ml-4 flex-shrink-0 text-right text-sm font-medium">
+        {item.value}
+      </span>
+    ) : (
+      <div className="ml-4 w-24 shrink-0">
```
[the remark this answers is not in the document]
dani-moreno: Yep, thanks! I added a `gap-4` on the container div which was the intended result.

src/playground/Widgets/List/index.tsx
dani-moreno: I've changed the key to the name, but now that I rethink it, it may not be an unique value, right? 🤔

src/playground/Widgets/List/index.tsx
dani-moreno: Cool!

dani-moreno:
> What happens if we pass a really long item title? 👀

There's a `truncate` in the title, so the value/progressbar keeps visible.

src/playground/Widgets/List/index.tsx
sauldom102: Well, the name should be unique actually, so it could be good too!

sauldom102 (approved):
Great work so far!

nlopin:
> > What happens if we pass a really long item title? 👀
> 
> There's a `truncate` in the title, so the value/progressbar keeps visible.

I noticed truncate is loved here and I suffer with my love to big font sizes 😢 

CleanShot 2024-09-06 at 16 34 38@2x

src/playground/Widgets/List/index.tsx
```diff
@@ -0,0 +1,50 @@
...
+    ) : (
+      <span className="flex-shrink-0 text-sm font-medium">{item.value}</span>
+    )}
+  </div>
+)
+
+function List({ items, title }: ListProps) {
+  return (
+    <div className="w-full">
+      {title && <h2 className="mb-2 text-base font-medium">{title}</h2>}
```
nlopin, line 40: it would be nicer to have the wrapper above managing the spaces between the smaller components. 

I suggest we try replacing `mb-2` here and in the `ListItem` component with flexbox + gap-2 on the main wrapper on line 39

src/playground/Widgets/List/index.tsx
```diff
@@ -0,0 +1,50 @@
...
+  title?: string
+  items: ListItem[]
+}
+
+const ListItem = ({ item }: { item: ListItem }) => (
+  <div className="mb-2 flex w-full items-center gap-4 last:mb-0">
+    <span className="flex-grow truncate text-muted-foreground">
+      {item.name}
+    </span>
+    {typeof item.value === "number" ? (
```
nlopin, line 27: If we automatically decide to show a progress bar, we assume that a number passed is in percents. This is not clear from the API and it could be surprising for the user. 

Shouldn't we have a `variant` or `type` field that will tell us how to represent the passed number?

dani-moreno:
> I noticed truncate is loved here and I suffer with my love to big font sizes 😢

Hah, you're right, `truncate` is the lazy solution, to be honest. Not happy with how I've ended building this, I think it'll work better with a table structure, I think a table structure would work better, allowing for longer names with multiple lines without breaking anything.

I’m switching this back to draft and will make some adjustments, thanks for all the feedback! :)

PR #470, opened 2024-09-06
Introduce typography scale

nlopin:
CleanShot 2024-09-06 at 16 14 06@2x

## 🚪 Why?

We are setting up foundations of the design system: colors, spaces, typography, etc.

## 🔑 What?

- sets a font scale in the Tailwind theme; We now have only 5 allowed font sizes for external users
- adds a basic documentation page
- removes the 15px font-size set to the root of the page
- adjusts font-sizes of existing components

Factorial app uses fixed font size scale set in pixels. The new scale, in contrary, uses relative units to provide better accessibility. This approach will not create any style mismatches when used together with the old design system.

## 🏡 Context

- [🎨 Typography in Figma](https://www.figma.com/design/b0A4ttd9C576bp2roJh0hX/Foundations?node-id=1-2&t=RaffcNrezm0ObtqA-1)

### DIFF
--- a/docs/typography.mdx
+++ b/docs/typography.mdx
@@ -1,0 +1,33 @@
+import { Meta, Unstyled } from "@storybook/blocks"
+
+<Meta title="essentials/Typography" />
+
+# Typography
+
+The typography comprises five predefined font sizes. The base font size of the
+app is 14px when used with the default browser settings.
+
+<Unstyled>
+  <div className="grid grid-cols-[auto_minmax(0,1fr)] items-baseline gap-4">
+    <p className="text-right text-sm text-gray-600">
+      <pre className="inline">text-2xl</pre>, 26px
+    </p>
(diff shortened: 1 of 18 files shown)


docs/typography.mdx
```diff
@@ -0,0 +1,33 @@
+import { Meta, Unstyled } from "@storybook/blocks"
+
+<Meta title="essentials/Typography" />
+
+# Typography
+
+The typography comprises five predefined font sizes. The base font size of the
+app is 14px when used with the default browser settings.
```
josepjaume, line 8: Do we agree that `14px` is the baseline? I remember discussing 14-15-16px with Mario. I'm fine with 14px, but just saying.

nlopin: CleanShot 2024-09-09 at 11 59 58

josepjaume (approved):
Nice job! Let's do this!

PR #472, opened 2024-09-09
Adds an overridable link component

josepjaume:
This adds a link component that can be overriden via a provider (please check out the tests) so the consumer can customize which routing library to use -- which are usually quite coupled to a react `<Link />` component they provide.


josepjaume:
Done! Tests were failing because of accessibility & contrast, which is nice to see 💅

PR #473, opened 2024-09-09
Foundations colors and semantic tokens

dani-moreno:
> [!IMPORTANT]  
> This PR looks huge, but most of the changes are just replacements for colors! Interesting changes are on [styles.css](https://github.com/factorialco/factorial-one/pull/473/files#diff-380b7b38760dd442e897eb0164c58f6a17da966ccaca6318017a468c163979b1), and [tailwind.config.ts](https://github.com/factorialco/factorial-one/pull/473/files#diff-655dc9e3d0aa561e3fa164bf48bd89cb0f5da65e0a567f8ebbf9dd791a0e7f40).

## 🚪 Why?

A color system **enables us to speak a common language** and ensures consistent designs across products. Instead of constantly questioning which colors to use for, say, displaying a positive message to the user, we have a set of tokens that make their function immediately recognizable.

Instead of using `grass-50` for a positive alert, we use `positive`, which serves as an alias for that color. `positive` can be an alias of a different color depending on the theme, giving us the flexibility to create themes in the near future.

To do so, we've built this system, which has three layers:

image

### 1st layer: Primitives
Tokens named after how the look. These are the current color palette we have available, replacing the Tailwind ones. Examples of these primitives would be `orange-50`, `grass-50`, or `radical-50`.

### 2nd layer: Semantic values
Tokens named after what they mean. Built on top of the primitives. Examples of these semantic values would be `accent-50`, `warning-50` or `positive-50`.

### 3rd layer: Tokens
Tokens named after what they should be used for. These are the tokens we'll use in our interface. Built on top of the semantic values. Examples of these tokens would be `foreground-accent`, `background-neutral`, or `icon-info`.

## 🔑 What?

- [Set our colors](https://github.com/factorialco/factorial-one/pull/473/files#diff-655dc9e3d0aa561e3fa164bf48bd89cb0f5da65e0a567f8ebbf9dd791a0e7f40) (replacing Tailwind's), as we have in Figma.
- Set semantic tokens on [styles.css](https://github.com/factorialco/factorial-one/pull/473/files#diff-380b7b38760dd442e897eb0164c58f6a17da966ccaca6318017a468c163979b1).
- Extended our semantic tokens in Tailwind to be used as classes.

----

> **Why not use shadcn tokens?**
> That was my initial impression, but after a thorough review, the default tokens from shadcn aren't what we need. They lack semantic meaning and some of them were component-specific (card, popover), which we are avoiding.

> **Why replace Tailwind colors?**
> We have our own colors palette, that match our brand identity and style guide.

> **Some components don't look like they do in Figma**
> That's intentional! I just swapped out the colors so everything aligns with this new color system, trying to use colors that look similar to the ones currently being used, but I haven't made changes to the components yet (just the Button). We'll tackle those gradually. Let me know if I should change anything now that I might have missed!

## 🏡 Context

- [🎨 Figma](https://www.figma.com/design/b0A4ttd9C576bp2roJh0hX/Foundations?node-id=0-1&node-type=CANVAS&t=uHWkq3wob1cxWmdt-0)


dani-moreno:
> I see we are deleting classes like `fill-foreground-primary` and replacing them with `fill-neutral-100`. The former looks way more semantic than the latter.
> 
> As an engineer, I'd love to have classes I can apply to my UI situation. From that point of view `foreground-primary` or `background-secondary` is preferable

Valid point! I've had the same thought, but since we're in the early stages of design, we've aimed for a system that offers more flexibility and granularity, as we'll likely be updating things frequently. Also, I'm trying to avoid ambiguity, since terms like "primary" and "secondary" can mean different things in different contexts.

I agree that the learning curve, especially for developers, is steeper. It's aligned with Figma, and the nomenclature resembles Tailwind, but I'm not sure if that's enough.

nlopin:
> > I see we are deleting classes like `fill-foreground-primary` and replacing them with `fill-neutral-100`. The former looks way more semantic than the latter.
> > As an engineer, I'd love to have classes I can apply to my UI situation. From that point of view `foreground-primary` or `background-secondary` is preferable
> 
> Valid point! I've had the same thought, but since we're in the early stages of design, we've aimed for a system that offers more flexibility and granularity, as we'll likely be updating things frequently. Also, I'm trying to avoid ambiguity, since terms like "primary" and "secondary" can mean different things in different contexts.
> 
> I agree that the learning curve, especially for developers, is steeper. It's aligned with Figma, and the nomenclature resembles Tailwind, but I'm not sure if that's enough.

We should provide both. We should facilitate right decisions and exposing a level of semantic tokens like `foreground-`, `background-`, `border-` makes it easy to keep the UI consistent and contrast. 

The set of tokens you propose is one level below on that scale and they are crucial for making adjustments where semantic tokens don't fit. It must be available too. At the same time, they give too much flexibility. In the example with text color, the developer has to pick one out of ten shades of gray and they can easily mix a tone or two eventually making the whole app look messy. 

I understand we don't have the full picture which tokens do we need, but we can start small by adding what we already know. In case something changes based on the context, like primary colors, we can still name them primary, but replace the value of that token in the context it applies. Since we use css variables, it should be a trivial operation.

Happy to continue the discussion on a call or pair program on the issue! We are moving towards a great solution ❤️

dani-moreno:
> We should provide both. We should facilitate right decisions and exposing a level of semantic tokens like `foreground-`, `background-`, `border-` makes it easy to keep the UI consistent and contrast.
> 
> The set of tokens you propose is one level below on that scale and they are crucial for making adjustments where semantic tokens don't fit. It must be available too. At the same time, they give too much flexibility. In the example with text color, the developer has to pick one out of ten shades of gray and they can easily mix a tone or two eventually making the whole app look messy.
> 
> I understand we don't have the full picture which tokens do we need, but we can start small by adding what we already know. In case something changes based on the context, like primary colors, we can still name them primary, but replace the value of that token in the context it applies. Since we use css variables, it should be a trivial operation.
> 
> Happy to continue the discussion on a call or pair program on the issue! We are moving towards a great solution ❤️
> 
> P.S. I found some good naming examples in [Reshaped](https://reshaped.so/docs/tokens/color) and [IBM Carbon](https://carbondesignsystem.com/elements/color/tokens/) design systems

Ah, I completely agree with you, I was too worried about making the system too complex haha, but without that layer of actual semantic tokens (what I have here is probably just semantic values), it's going to be pretty difficult to use. Moving the conversation to Slack, thanks!

I think we can use some of the ideas that I had for Gamma, adapting them to F1.

nlopin (approved):
@dani-moreno do we have default colors exported by the tailwind or we overwrite them completely?

dani-moreno:
> @dani-moreno do we have default colors exported by the tailwind or we overwrite them completely?

They've been removed completely. Since we have our own color palette, I don't think we need them anymore.

PR #474, opened 2024-09-09
chore: remove fill when dashed

RamProg:
## 🚪 Why?

In Figma when the chart is dashed there is no filling in the area

## 🔑 What?

Add a condition to make fill invisible when the line is dashed

## 🏡 Context

https://www.figma.com/design/9pLsRzdinid0ha0qRWta2D/Employee-Profile?node-id=604-7408&node-type=FRAME&t=A0gGeFk4RpgbnesX-0

image


Rogermax (approved):
Ohh.. nice improve thx

PR #480, opened 2024-09-10
Updates spacing

nlopin:
## 🚪 Why?

We review the design system fundamentals

## 🔑 What?

Splits the absolute spacing scale in two:
- absolute scale in pixels, used in paddings, margins and other similar cases;
- relative scale defined in `rem` to define block sizes

This change aligns with our shift to relative font sizes in #470, ensuring that the UI scales according to the user's font preferences. While Tailwind defaults to `rem` units for all properties, we’ve chosen to retain pixels for paddings and margins to prevent these elements from consuming too much useful space. Despite mixing both scales, they output the same spacing on a 16px default font size.

Since tailwind uses one scale by default, I had to do a lot of copy-pasting from the [default tailwind config](https://github.com/tailwindlabs/tailwindcss/blob/main/stubs/config.full.js) just to apply another helper.

## 🏡 Context
- [🎨 Figma](https://www.figma.com/design/b0A4ttd9C576bp2roJh0hX/Foundations?node-id=307-4118&node-type=frame&m=dev)
- [MinHeight px vs rem exploration](https://codepen.io/Lopinopulos/pen/MWMxWYM)

CleanShot 2024-09-10 at 16 15 03@2x


tailwind/spacing.ts
```diff
@@ -0,0 +1,275 @@
...
+    60: "240px",
+    64: "256px",
+    72: "288px",
+    80: "320px",
+    96: "384px",
+  },
+  flexBasis: ({ theme }) => ({
+    auto: "auto",
+    ...theme("relativeSpacing"),
+    "1/2": "50%",
```
josepjaume, line 90: Can't we import this from tailwind's defaults instead of redefining it? The fewer custom code we have, the better.

tailwind/spacing.ts
```diff
@@ -0,0 +1,275 @@
...
+    "9/12": "75%",
+    "10/12": "83.333333%",
+    "11/12": "91.666667%",
+    full: "100%",
+  }),
+  height: ({ theme }) => ({
+    auto: "auto",
+    ...theme("relativeSpacing"),
+    "1/2": "50%",
+    "1/3": "33.333333%",
```
josepjaume, line 122: Same!

josepjaume (approved):
Nice job! I'm just worrying that we're being too specific with things we're actually not customizing, like the fractions (`1/2`, `1/6`...) which we could maybe delegate to the default implementation?

tailwind/spacing.ts
nlopin: I couldn't find a way to do so. They have the exact same code, but with `theme("spacing")` helper. Do you have an idea how we can make it?

nlopin:
> Nice job! I'm just worrying that we're being too specific with things we're actually not customizing, like the fractions (`1/2`, `1/6`...) which we could maybe delegate to the default implementation?

Here is a video with details how I tried to approach the issue https://drive.google.com/file/d/1l76duWujHire8s_foB3wig3cKcQF3j5D/view?usp=sharing

tailwind/spacing.ts
josepjaume: Wouldn't just splatting `...theme('spacing')` just before make sure we keep the same keys and just override the next ones?

```tsc
...theme("spacing"),
...theme("relativeSpacing"),
```

nlopin:
@josepjaume take a look

tailwind.config.ts
```diff
@@ -62,44 +64,19 @@ export default {
...
+      ...getRelativeSpacingOverride([
+        "flexBasis",
+        "height",
+        "maxHeight",
+        "maxWidth",
+        "minHeight",
+        "minWidth",
+        "textIndent",
+        "width",
+      ]),
```
josepjaume, line 79: This is an incredibly minor nitpick and I'm absolutely not blocking the PR for this, but I'm worried that this is not actually typed? These keys might change in a future Tailwind update (highly unlikely) and we wouldn't notice.

What's the advantage of this over:

```ts
{
  flexBasis: relativeSpacing,
  height: relativeSpacing,
  maxHeight: relativeSpacing
  // ...
}
```

I mean, other than the obnoxious repetition 😆

josepjaume (approved):

tailwind.config.ts
nlopin: it is typed, it only allows keys from the Tailwind theme object. At the same time, I agree, let's make it simple

PR #481, opened 2024-09-10
chore: add blur feature to area chart

RamProg:
## 🚪 Why?

The salary widget has a button to toggle the blur on salary data

## 🔑 What?

We add some props to conditionally allow the eye button that will hide or show the salary

## 🏡 Context

https://www.figma.com/design/9pLsRzdinid0ha0qRWta2D/Employee-Profile?node-id=378-17717&node-type=canvas&t=umPkgaSFRFR6L52Y-0

image


nlopin:
is it fine we still show it in the tooltip on hover?
CleanShot 2024-09-10 at 16 17 29@2x

RamProg:
> is it fine we still show it in the tooltip on hover? 

For now this is fine, we will run it by the designer Lily later to check details.

nlopin:
I can also copy the contend behind the blur. `user-select: none;` will fix the issue

nlopin:
What is the plan on other graphs? do you plan to add blur there too?

RamProg:
> I can also copy the content behind the blur. `user-select: none;` will fix the issue

That's not a problem because you can simply press the button and see the content. The goal here is to hide it from someone looking over your shoulder. It's not password protected or anything like that.

RamProg:
> What is the plan on other graphs? do you plan to add blur there too?

Not at the moment because we don't have any widgets designed that require that.

nlopin:
> > I can also copy the content behind the blur. `user-select: none;` will fix the issue
> 
> That's not a problem because you can simply press the button and see the content. The goal here is to hide it from someone looking over your shoulder. It's not password protected or anything like that.

That is not the problem because you solve specific task, but it is an issue on the design system. Another user will be less aware of this specifics. It would be good to have it behave as sanity check suggests

PR #482, opened 2024-09-10
chore: implement an indicator horizontal list

RamProg:
## 🚪 Why?

We are creating the Tasks Widget, so we need a list of horizontal indicators

## 🔑 What?

I extract and refactor this piece of code so we can use it to create the Tasks widget from F0

## 🏡 Context

https://www.figma.com/design/9pLsRzdinid0ha0qRWta2D/Employee-Profile?node-id=604-7408&node-type=frame&t=umPkgaSFRFR6L52Y-0

image

image


lib/components/Utilities/IndicatorsList/index.tsx
```diff
@@ -0,0 +1,26 @@
...
+import { forwardRef } from "react"
+import { IconType } from "../Icon"
+
+export interface IndicatorsListProps {
+  items: {
+    label: string
+    count: number
+    icon?: IconType
+    color: string
+  }[]
```
sauldom102, line 11: We could better get these attributes from the `Indicator` component props:

```suggestion
  items: IndicatorProps[]
```

sauldom102 (changes requested):
Are we placing widget "composables" inside "Utilities"? We should likely organize it into a common folder

RamProg:
> Are we placing widget "composables" inside "Utilities"? We should likely organize it into a common folder

I addressed all of your comments

sauldom102 (approved):
Nice! ✅

PR #492, opened 2024-09-12
fix: remove empty nodes separator

RamProg:
## 🚪 Why?

In some cases we have conditional parts of a widget that should not show a separator

## 🔑 What?

We check that the node actually has any content

## 🏡 Context

https://www.figma.com/design/9pLsRzdinid0ha0qRWta2D/Employee-Profile?node-id=604-7408&node-type=frame&t=vOiPf49zpJBIE1r9-0

image

### DIFF
--- a/lib/components/Widgets/WidgetContainer/index.stories.tsx
+++ b/lib/components/Widgets/WidgetContainer/index.stories.tsx
@@ -2,7 +2,7 @@ import type { Meta, StoryObj } from "@storybook/react"
 
 import { Placeholder } from "@/lib/storybook-utils"
 import { fn } from "@storybook/test"
-import { ComponentProps, Fragment } from "react"
+import { ComponentProps } from "react"
 import { WidgetContainer } from "."
 
 const meta = {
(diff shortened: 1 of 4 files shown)


lib/components/Widgets/WidgetContainer/index.stories.tsx
```diff
@@ -77,3 +77,29 @@ export const MultipleContent: Story = {
...
+          key={index}
+          className="rounded-lg bg-muted p-3 text-center text-foreground"
+        >
+          Content {index + 1}
+        </div>
+      )),
+      undefined,
+      null,
+      false,
+      <Fragment></Fragment>,
```
nlopin, line 101: Does `<></>` work? they are the same

nlopin (approved):

lib/components/Widgets/WidgetContainer/index.stories.tsx
```diff
@@ -77,3 +77,29 @@ export const MultipleContent: Story = {
     ],
   },
 }
+
+export const MultipleContentWihNulls: Story = {
```
josepjaume, line 81: This should probably be just a spec, I don't think it should be part of the documentation, right? It's extremely easy, check out any `.spec.ts` examples.

lib/components/Widgets/WidgetContainer/index.tsx
```diff
@@ -44,6 +44,17 @@ const Container = forwardRef<
...
+  const isRealNode = (node: React.ReactNode): boolean => {
+    return (
+      !!node &&
+      !(
+        React.isValidElement(node) &&
+        node.type === React.Fragment &&
+        React.Children.count(node.props.children) === 0
+      )
+    )
+  }
```
josepjaume, line 56: Nice trick!

josepjaume (approved):

lib/components/Widgets/WidgetContainer/index.stories.tsx
RamProg: done

lib/components/Widgets/WidgetContainer/index.stories.tsx
RamProg: I removed this and created a test

PR #499, opened 2024-09-13
Add addon designs

josepjaume:
Adds the design addons to link storybook with Figma.


sauldom102 (approved):
Supeeer!! 🔥

.github/workflows/chromatic.yml
```diff
@@ -24,7 +24,7 @@ jobs:
         with:
           node-version: "21.x"
           cache: npm
-      - run: npm ci
+      - run: npm i
```
nlopin, line 27: why? is it a temporary thing?

nlopin:
is there a way to add a link to the docs page? We don't have a lot of text on the `Button` page, however there is text on the typography page. It would be great to have an way to have a link there

josepjaume:
@nlopin Yeah... I think we can modify the default component to render pages and read from `parameters`. In case there's a `design` entry, we can just render a link :)

PR #500, opened 2024-09-13
Define supported font weights

nlopin:
## 🚪 Why?

We define foundations

## 🔑 What?

CleanShot 2024-09-13 at 17 24 23@2x

Sets three supported font weights, overwriting the full list provided by the Tailwind

## 🏡 Context

- [🎨 Figma](https://www.figma.com/design/b0A4ttd9C576bp2roJh0hX/Foundations?node-id=307-4037&node-type=frame&m=dev)


josepjaume (approved):
Awesome!

PR #503, opened 2024-09-13
fix: add comment to progress section

RamProg:
## 🚪 Why?

PogressSection was missing an option to add a text at the end like in Figma 

## 🔑 What?

Conditionally add text at bottom

## 🏡 Context

https://www.figma.com/design/9pLsRzdinid0ha0qRWta2D/Employee-Profile?node-id=604-7408&node-type=frame&t=lKMH3Bu0qtM1BT24-0

image


src/playground/Widgets/ProgressSection/index.tsx
```diff
@@ -11,6 +11,7 @@ interface ProgressSectionProps extends Omit<CategoryBarProps, "data"> {
   unit?: string
   primaryLabel?: string
   secondaryLabel?: string
+  bottomComment?: string
```
dani-moreno, line 14: What if we name it something else? The position might change, so it's better not to imply that in the name. I'm not thrilled with just `comment` as a name, but it could work if we don't come up with a better one.

src/playground/Widgets/ProgressSection/index.tsx
```diff
@@ -49,6 +51,7 @@ export function ProgressSection({
         </div>
       </div>
       <CategoryBar data={data} {...categoryBarProps} />
+      <p className="text-grey-50">{bottomComment}</p>
```
dani-moreno, line 54: I merged the color tokens a few hours ago (😬), so that color probably won't apply correctly. It would be great if you could merge main and check it out for yourself, but the token would be `text-f1-foreground-secondary`.

RamProg: oh ok, I noticed new colors when I did this, but that one I think didn't exist yet, I'll check it out now

nlopin (approved):
so cool to see the color tokens!

RamProg:
I found the ProgressBarSection, so this PR might not be necessary, I'll close it once confirmed.

PR #509, opened 2024-09-16
Define button icon colors

dani-moreno:
## 🚪 Why?

Button icons have their own color system. When a label is displayed, the icon should have a lighter color to give prominence to the label. When there's no label, the icon takes center stage, displaying a stronger color.

image

_Notice how the icon has a lighter color than the label_

## 🔑 What?

- Add icon color variants, with and without a label.
- Add needed tokens.

## 🏡 Context

- [🎨 Figma](https://www.figma.com/design/pZzg1KTe9lpKTSGPUZa8OJ/Web-Components?node-id=14-976)

### DIFF
--- a/lib/ui/button.tsx
+++ b/lib/ui/button.tsx
@@ -13,7 +13,7 @@ export const variants = [
 ] as const
 
 const buttonVariants = cva(
-  "focus-visible:ring-offset focus-visible:ring-ring inline-flex h-10 items-center justify-center gap-1 whitespace-nowrap rounded-xl border-none text-base font-medium transition-colors focus-visible:outline-none focus-visible:ring-1 focus-visible:ring-offset-1 disabled:pointer-events-none disabled:opacity-50",
+  "focus-visible:ring-offset focus-visible:ring-ring group inline-flex h-10 items-center justify-center gap-1 whitespace-nowrap rounded-xl border-none text-base font-medium transition-colors focus-visible:outline-none focus-visible:ring-1 focus-visible:ring-offset-1 disabled:pointer-events-none disabled:opacity-50",
   {
     variants: {
       variant: {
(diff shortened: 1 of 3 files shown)


lib/ui/button.tsx
```diff
@@ -13,7 +13,7 @@ export const variants = [
 ] as const
 
 const buttonVariants = cva(
-  "focus-visible:ring-offset focus-visible:ring-ring inline-flex h-10 items-center justify-center gap-1 whitespace-nowrap rounded-xl border-none text-base font-medium transition-colors focus-visible:outline-none focus-visible:ring-1 focus-visible:ring-offset-1 disabled:pointer-events-none disabled:opacity-50",
+  "focus-visible:ring-offset focus-visible:ring-ring group inline-flex h-10 items-center justify-center gap-1 whitespace-nowrap rounded-xl border-none text-base font-medium transition-colors focus-visible:outline-none focus-visible:ring-1 focus-visible:ring-offset-1 disabled:pointer-events-none disabled:opacity-50",
```
josepjaume, line 16: Oh wow, this group thing is very weird but makes sense. Didn't know about it!

lib/ui/button.tsx
```diff
@@ -42,6 +42,39 @@ const buttonVariants = cva(
...
+const iconVariants = cva("transition-colors", {
+  variants: {
+    variant: {
+      default: "text-f1-icon-inverse-secondary",
+      outline: "text-f1-icon",
+      neutral: "text-f1-icon",
+      critical:
+        "text-f1-icon-critical-bold group-hover:text-f1-icon-inverse-secondary",
+      ghost: "text-f1-icon",
+      promote: "text-f1-icon",
```
josepjaume, line 54: It's a shame that we're losing dark theme colors, but I guess we're going to work on it as a separate epic months (if not years) from now.

tailwind.config.ts
```diff
@@ -139,7 +139,10 @@ export default {
           foreground: {
             DEFAULT: "hsl(var(--neutral-100))",
             secondary: "hsl(var(--neutral-50))",
-            inverse: "hsl(var(--neutral-0))",
+            inverse: {
+              DEFAULT: "hsl(var(--neutral-0))",
+              secondary: "hsl(var(--neutral-0) / 0.8)",
+            },
```
josepjaume, line 145: It worries me that we keep adding "secondary" whenever we please. Is it too naive to try to solve this with opacity so we don't need to add extra colors, perhaps?

josepjaume (approved):
Good job! A couple of minor comments. Not that important.

lib/ui/button.tsx
dani-moreno: Ah, no! We're not, It's just that the dark colors are currently the same as the light ones. However, the theme system is in place and ready for when we define them at a design level.

nlopin:
Have we considered an option to add a css `filter` prop to darken the icon automatically?

nlopin (approved):

PR #511, opened 2024-09-16
chore: rename info component; integrate icons

RamProg:
## 🚪 Why?

Icons in Factorial rn are part of the old design system. We can't pass icons as props as that would create a dependency on F1 with the old design system.

## 🔑 What?

Now the component handles the icon internally. Also, the name was changed.

## 🏡 Context

https://www.figma.com/design/9pLsRzdinid0ha0qRWta2D/Employee-Profile?node-id=378-17717&node-type=canvas&t=sx6jR27IhjrT19F9-0


sauldom102 (approved):
LGTM!

lib/components/Widgets/Info/index.tsx
```diff
@@ -0,0 +1,29 @@
+import { Icon } from "@/components/Utilities/Icon"
+import BriefcaseIcon from "@/icons/Briefcase"
+import OfficeIcon from "@/icons/Office"
+import { forwardRef } from "react"
+
+const iconsMap = {
```
nlopin, line 6: Do we want to control what can be passed? Why can't we allow passing any icon available?

PR #513, opened 2024-09-16
Create TasksList component

sauldom102:
## 🔑 What?

Create TasksList component.

## 🚪 Why?

We need this sub-component for our Tasks widget component.

Screenshot 2024-09-16 at 13 16 59

## 🏡 Context

- [💼 Jira](https://factorialmakers.atlassian.net/browse/FCT-18772)


lib/icons/index.ts
```diff
@@ -1,17 +1,17 @@
 export { default as AcademicCap } from './AcademicCap'
 export { default as Add } from './Add'
 export { default as Ai } from './Ai'
-export { default as AlertCircleLine } from './AlertCircleLine'
+export { default as Alert } from './Alert'
 export { default as AlertCircle } from './AlertCircle'
+export { default as AlertCircleLine } from './AlertCircleLine'
 export { default as AlertSign } from './AlertSign'
-export { default as Alert } from './Alert'
 export { default as AlignTextCenter } from './AlignTextCenter'
```
sauldom102, line 8: Just sorted this file alphabetically in VSCode

src/playground/Widgets/Tasks/index.tsx
```diff
@@ -83,21 +84,12 @@ export const TasksInsight = forwardRef<HTMLDivElement, TasksInsightProps>(
...
-              </div>
-            ))
-          )}
-        </div>
+        {(inProgressTasks.length || pendingTasks.length) && (
+          <TasksList
+            inProgressTasks={inProgressTasks}
+            pendingTasks={pendingTasks}
+          />
+        )}
```
sauldom102, line 92: I think we don't need an empty state text here, it's self-explanatory seeing the counters already, and we're saving some vertical space as well.

Screenshot 2024-09-16 at 13 18 31

nlopin:
How does the long task title look like?

sauldom102:
> How does a long task title look like?

I just made a story with how it would look like for long titles ✨ I think it's fine adding some ellipsis at the end, we shouldn't have tasks with long names, as an alternative we could make it 2 lines maximum

Screenshot 2024-09-16 at 13 47 09

sauldom102:
> > How does a long task title look like?
> 
> I just made a story with how it would look like for long titles ✨ I think it's fine adding some ellipsis at the end, we shouldn't have tasks with long names, as an alternative we could make it 2-3 lines maximum
> 
> Screenshot 2024-09-16 at 13 47 09

This could be the alternative:

Screenshot 2024-09-16 at 13 53 23

src/playground/Widgets/TasksList/index.tsx
```diff
@@ -10,12 +10,12 @@ type TaskItemProps = {
...
 function TaskItem({ title, status }: TaskItemProps) {
   return (
-    <div className="flex flex-row items-center gap-2">
+    <div className="flex flex-row items-start gap-2">
       {(status === "due" || status === "no-due") && (
         <Circle width={24} opacity={status === "no-due" ? 0.4 : 1} />
       )}
       {status === "in-progress" && <InProgressTask />}
-      <p className="flex-1 truncate font-medium">{title}</p>
+      <p className="mt-[2px] line-clamp-2 flex-1 font-medium">{title}</p>
```
nlopin, line 18: you can do `mt-0.5` instead of `mt-[2px]` if i'm not wrong

RamProg: We shouldn't use custom values at all, so if mt-0.5 doesn't work just use whatever is available and closer to the value.

src/playground/Widgets/TasksList/index.tsx
```diff
@@ -0,0 +1,69 @@
...
+type TaskItemProps = {
+  title: string
+  status: TaskStatus
+}
+
+function TaskItem({ title, status }: TaskItemProps) {
+  return (
+    <div className="flex flex-row items-start gap-2">
+      {(status === "due" || status === "no-due") && (
+        <Circle width={24} opacity={status === "no-due" ? 0.4 : 1} />
```
RamProg, line 15: Why the opacity? Are you using it to change the color? We have shades of grey you can use.

src/playground/Widgets/TasksList/index.tsx
```diff
@@ -0,0 +1,69 @@
...
+  maxTasksToShow?: number
+}
+
+export function TasksList({
+  inProgressTasks,
+  dueTasks,
+  noDueTasks,
+  maxTasksToShow = 5,
+}: Props) {
+  const tasksToRender = useMemo(
```
RamProg, line 36: I don't think the design systems needs to have this logic to handle memory, it should focus in UI and be as "dumb" as possible.

src/playground/Widgets/TasksList/index.tsx
```diff
@@ -0,0 +1,69 @@
...
+}
+
+interface Props {
+  inProgressTasks: string[]
+  noDueTasks: string[]
+  dueTasks: string[]
+  maxTasksToShow?: number
+}
+
+export function TasksList({
```
RamProg, line 30: I think we can move this outside of the playground and straight to the main widgets folder. The playground was more for early development, but we are not really supposed to use it at all from Factorial,

src/playground/Widgets/Tasks/index.tsx
```diff
@@ -83,21 +86,13 @@ export const TasksInsight = forwardRef<HTMLDivElement, TasksInsightProps>(
...
-                key={`${task} ${i}`}
-                className="flex flex-row items-center gap-3"
-              >
-                <div className="border-f1-border h-5 min-w-5 rounded-md border border-solid" />
-                <p className="truncate font-medium">{task}</p>
-              </div>
-            ))
-          )}
-        </div>
+        {(inProgressTasks.length || dueTasks.length || noDueTasks.length) && (
```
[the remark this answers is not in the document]
sauldom102: Yeah I know, I just wanted to leave it as close as our real widget as possible

sauldom102:
> What is the next step for this component? do we plan to allow clicking on the icons? I was confused when trying to use the component. The icon begs to click on it, but nothing happens

I don't think it was thought to make them interactive like that. We also need to consider that a task can be in multiple states, not only done/undone.

nlopin:
> I don't think it was thought to make them interactive like that. We also need to consider that a task can be in multiple states, not only done/undone.

Do you think we can add an `onClick` handler for each item at least? I'd track how many users will click the items in the Amplitude to decide if we need to add more interactivity

RamProg (approved):
One step closer!

sauldom102:
> > I don't think it was thought to make them interactive like that. We also need to consider that a task can be in multiple states, not only done/undone.
> 
> Do you think we can add an `onClick` handler for each item at least? I'd track how many users will click the items in the Amplitude to decide if we need to add more interactivity

Sure! Good idea!

src/playground/Widgets/Tasks/index.tsx
```diff
@@ -7,16 +8,17 @@ import { forwardRef } from "react"
 
 export interface TasksInsightData {
   title: string
-  tasks: string[]
+  inProgressTasks: string[]
+  dueTasks: string[]
+  noDueTasks: string[]
   overdueLabel: string
-  overdueTasks: number
+  overdueTasksCount: number
```
nlopin, line 15: do we need a count? we have the list of the tasks above.

We can to `overdueTasks.length` and etc

PR #516, opened 2024-09-16
chore: update color

RamProg:
## 🚪 Why?

New token and colors broke the separator

## 🔑 What?

We update with new color name

## 🏡 Context

https://www.figma.com/design/9pLsRzdinid0ha0qRWta2D/Employee-Profile?node-id=604-7408&node-type=frame&t=6DlERfINeg4OUFjB-0

image

### DIFF
--- a/lib/ui/separator.tsx
+++ b/lib/ui/separator.tsx
@@ -11,7 +11,7 @@ export const Separator = forwardRef<
       className="my-4 h-[1px] w-full"
       style={{
         backgroundImage:
-          "repeating-linear-gradient(to right, hsl(var(--f1-border)) 0, hsl(var(--f1-border)) 3px, transparent 3px, transparent 7px)",
+          "repeating-linear-gradient(to right, hsl(var(--neutral-30)) 0, hsl(var(--neutral-30)) 3px, transparent 3px, transparent 7px)",
       }}
       {...props}
     />


nlopin:
is there a screenshot "before"?

RamProg:
> is there a screenshot "before"?

before new tokens or before the PR?

before new tokens it looked like figma https://www.figma.com/design/9pLsRzdinid0ha0qRWta2D/Employee-Profile?node-id=604-7408&node-type=frame&t=6DlERfINeg4OUFjB-0

before PR it was completely white, so it was invisible

dani-moreno:
Can't we find a way to use a token instead of a bare color? Tokens aren't exported as CSS variables, which is probably why it's not being applied. But we could try to display a token.

nlopin:
> Can't we find a way to use a token instead of a bare color? Tokens aren't exported as CSS variables, which is probably why it's not being applied. But we could try to display a token.

I think the issue here is that it is a gradient. Have you tried using [gradient utilities](https://tailwindcss.com/docs/background-image#linear-gradients) @RamProg ?

RamProg:
> > Can't we find a way to use a token instead of a bare color? Tokens aren't exported as CSS variables, which is probably why it's not being applied. But we could try to display a token.
> 
> I think the issue here is that it is a gradient. Have you tried using [gradient utilities](https://tailwindcss.com/docs/background-image#linear-gradients) @RamProg ?

Yeah, I agree with both and already tried it. Unfortunately, I can't achieve the same exact result with tailwind so I need to use plain CSS which is why I'm using the CSS tokens and not the tailwind tokens.

I think if we ever find another way around this, it's an easy fix as it's a very specific and small component, but this PR is only updating the colour so we can see the separator.

Issue #498, opened 2024-09-13
Charts: y-axis automatic width calculation


nlopin:
## What 

Y-axis has a [fixed width](https://github.com/factorialco/factorial-one/commit/79053987be81012d4aac0a4ea41c2f18a9391021) of 32px, which fits 3-4 characters.

If the label is bigger, it is truncated. For smaller labels you can see a gap. 

![CleanShot 2024-09-13 at 08 34 50@2x](https://github.com/user-attachments/assets/92f76712-4c57-4334-a50a-29f7bea96f09)

It would be nice to automatically calculate the width of the Y-axis component.

## Context

We use `recharts` library to render charts. This library converts components into an SVG with axises, areas, etc. The library uses fixed axis width that needs manual adjustments. 

The library has a [similar issue](https://github.com/recharts/recharts/issues/2027) to resolve. Not sure if we should tackle it directly in the library or just have an in-place calculation of the width that fits our case only.

Rogermax:
Ohh I see.. seems difficult and will take long that this fix is available for us. I will go ahead with our one solution for now

PR #521, opened 2024-09-17
add tickFormatter to ChartTooltip and width to yAxis

Rogermax:
## 🚪 Why?

To improve chart customization 

## 🔑 What?

* Add tickTransformation to tooltip match values (normally from yAxis)
* Remove fixed 32 wifth on yAxis and let user config by prop or by default is auto calculated from labels

https://github.com/user-attachments/assets/6de690a5-a4e5-4204-9904-ba9c8697c89a

https://github.com/user-attachments/assets/ec7b938c-c4bb-4785-8f6b-b62a395bb54a

### DIFF
--- a/lib/components/Charts/AreaChart/index.stories.tsx
+++ b/lib/components/Charts/AreaChart/index.stories.tsx
@@ -21,6 +21,8 @@ const meta: Meta<typeof AreaChart<typeof dataConfig>> = {
     },
     yAxis: {
       hide: true,
+      tickFormatter: (value: string) =>
+        `${Number.isNaN(parseFloat(value)) ? value : (parseFloat(value) / 100).toFixed(2) + "€"}`,
     },
     data: [
       { label: "January", values: { mobile: 1200, desktop: 500 } },
--- a/lib/components/Charts/AreaChart/index.stories.tsx
+++ b/lib/components/Charts/AreaChart/index.stories.tsx
@@ -22,7 +22,7 @@ const meta: Meta<typeof AreaChart<typeof dataConfig>> = {
     yAxis: {
       hide: true,
       tickFormatter: (value: string) =>
-        `${Number.isNaN(parseFloat(value)) ? value : (parseFloat(value) / 100).toFixed(2) + "€"}`,
+        `${Number.isNaN(parseFloat(value)) ? value : (parseFloat(value) / 100).toFixed(2) + ",00 Euros"}`,
     },
     data: [
       { label: "January", values: { mobile: 1200, desktop: 500 } },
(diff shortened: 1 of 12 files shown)


nlopin:
@Rogermax we have a task to make yAxis adjustable automatically (#498). Do you think we can work add it in this PR?

My current understanding is that we can roughly calculate the width of the axis based on the longest label.

nlopin:
@RamProg if we merge this, you will have to adjust the code on the profile page

Rogermax:
> @RamProg if we merge this, you will have to adjust the code on the profile page

The default width is 40 (recharts default) will break many things using 40 instead of 32? 8px (blank space) 
But we can handle this if you're using a widget (just update yAxis widget to have width: 32 as a param) 
Or are we using directly the component (not a widget?)

Rogermax:
> @Rogermax we have a task to make yAxis adjustable automatically (#498). Do you think we can work add it in this PR?
> 
> My current understanding is that we can roughly calculate the width of the axis based on the longest label.

Yeah let me try, sounds not too much difficult

nlopin:
I've seen your comment in the issue. What do you think about exposing the `width` but have the default value set to 32 so we don't break the code that already relies on this width?

lib/components/Charts/AreaChart/index.stories.tsx
```diff
@@ -21,6 +21,8 @@ const meta: Meta<typeof AreaChart<typeof dataConfig>> = {
     },
     yAxis: {
       hide: true,
+      tickFormatter: (value: string) =>
+        `${Number.isNaN(parseFloat(value)) ? value : (parseFloat(value) / 100).toFixed(2) + "€"}`,
```
RamProg, line 25: I'm not sure we should fix this behaviour, for salaries for example, we don't need decimal positions.

RamProg:
> I've seen your comment in the issue. What do you think about exposing the `width` but have the default value set to 32 so we don't break the code that already relies on this width?

It's just one usecase and it will take me a min to add a value there, so don't make a decision based on that. We should however find a way to verify that we are not breaking something in F0 when we change F1 because as this change effect is small, there could be much bigger ones.

lib/components/Charts/AreaChart/index.stories.tsx
Rogermax: This is only for storybook as an example and by default is hided this yAxis, so you will not see that at first.

Rogermax:
> @Rogermax we have a task to make yAxis adjustable automatically (#498). Do you think we can work add it in this PR?
> 
> My current understanding is that we can roughly calculate the width of the axis based on the longest label.

Done, check second video in description

lib/components/Charts/utils/elements.tsx
```diff
@@ -2,6 +2,19 @@ import { ComponentProps } from "react"
 import { XAxis, YAxis } from "recharts"
 import { AxisConfig } from "./types"
 
+export function measureTextWidth(
+  text: string,
+  font = "12px Inter, sans-serif"
+): number {
+  const canvas = document.createElement("canvas")
```
nlopin, line 9: clever approach

nlopin (approved):
That's awesome! Smart approach

PR #522, opened 2024-09-17
Add shadows

nlopin:
## 🔑 What?

Adds shadows from the [Figma](https://www.figma.com/design/b0A4ttd9C576bp2roJh0hX/Foundations?node-id=307-4459&node-type=frame&m=dev). The naming is also taken from there.

CleanShot 2024-09-17 at 15 35 20@2x


tailwind.config.ts
```diff
@@ -123,6 +123,14 @@ export default {
         },
       ],
     },
+    boxShadow: {
+      soft: "0 2px 20px 0 hsl(var(--neutral-100) / 0.04)",
+      DEFAULT: "0 4px 20px 0 hsl(var(--neutral-100) / 0.08)",
+      intense: "0 8px 30px 0 hsl(var(--neutral-100) / 0.12)",
+      strong: "0 12px 56px 0 hsl(var(--neutral-100) / 0.16)",
+      inner: "inset 0 2px 4px 0 hsl(var(--neutral-100) / 0.04)",
+      none: "none",
```
josepjaume, line 132: We're diverging quite a lot from `shadcn` and `tailwind`, so importing components will be more difficult. I guess that's ok? I still don't see why we don't just stick to the standard names (`xs`, `sm`, `md`...). They are actually quite informative. Looking at this list, I'm unsure about the difference between `strong` or `intense` for example.

@dani-moreno what do you think?

nlopin: happy to get back to standard names as long as we update sync it with figma

dani-moreno: Yes, I completely agree. It's easier to adapt Figma to code rather than the other way around. Plus, Tailwind has a good naming convention, in my opinion.

I've updated the naming in Figma.

tailwind.config.ts
```diff
@@ -123,6 +123,14 @@ export default {
         },
       ],
     },
+    boxShadow: {
+      soft: "0 2px 20px 0 hsl(var(--neutral-100) / 0.04)",
+      DEFAULT: "0 4px 20px 0 hsl(var(--neutral-100) / 0.08)",
+      intense: "0 8px 30px 0 hsl(var(--neutral-100) / 0.12)",
+      strong: "0 12px 56px 0 hsl(var(--neutral-100) / 0.16)",
+      inner: "inset 0 2px 4px 0 hsl(var(--neutral-100) / 0.04)",
```
dani-moreno, line 131: Checking designs, looks like we're not using inner shadows. Maybe we could just remove it for now until we find an use case.

tailwind.config.ts
josepjaume: Awesome! Can we do the same for the rest of customizations?

dani-moreno: Which ones?

josepjaume: You know what? None 🤣. My bad, I thought there were others but I see this was the only exception. Good!

tailwind.config.ts
josepjaume: Let's consider this then

tailwind.config.ts
nlopin: fixed. The names are updated and the inset is removed

dani-moreno (approved):

PR #523, opened 2024-09-17
Adds an `experimental` folder

josepjaume:
## 🚪 Why?

We need a way to explicitly state that some components are in an experimentation phase before promoting them, but at the same time we want to be able to use them in the application to prove they work.

## 🔑 What?

This introduces an "experimental" category to allow that. In some way, it should supersede the `playground`.

I would actually keep the playground but:

1. Remove its export, so it can't be used from the outside. Any such use should go through "experimental"
2. Probably not publish it on the production storybook but keep it in our local dev environments. This way we don't pollute the global search but we can use as a way to stitch together interface, test stuff, etc.

These changes would go in a second PR so we can orderly migrate things /cc @RamProg 

## 📸 Pics or didn't happen

image


eliseo-juan (approved):
Makes sense for me

nlopin (approved):

PR #525, opened 2024-09-17
chore: move details to main folder; update colors

RamProg:
EDITED
## 🚪 Why?

Need to expose composables to create the Details sidebar in the Profile.
 
## 🔑 What?

We expose two composables and create an example of the Details sidebar

## 🏡 Context

https://www.figma.com/design/9pLsRzdinid0ha0qRWta2D/Employee-Profile?node-id=946-24240&node-type=frame&t=hcwxRDOh0W1mgYnO-0

image


lib/components/Widgets/Details/index.tsx
```diff
@@ -0,0 +1,73 @@
...
+  details?: DetailsItemType[]
+  manager?: { title: string; name: string; avatar: string }
+  teams?: { title: string; list: string[] }
+  workableDays?: {
+    title: string
+    activatedDays?: ComponentProps<typeof Weekdays>["activatedDays"]
+  }
+}
+
+export const Details = forwardRef<HTMLDivElement, DetailsType>(
```
eliseo-juan, line 17: I see this suuuuuper tied to our business model, so I don't think that Factorial One is the place for this. I see a room for each detail type, but in a generic way. Convince me otherwise

eliseo-juan (changes requested):
I feel that conceptually it doesn't make sense to move this to main, I feel it makes sense to move it to Factorial, keeping the basic components in factorial one? Could be, but the detail as such I don't see that it makes sense.

lib/components/Widgets/Details/index.tsx
RamProg: I think you are right. I'll use this code to create this component from Factorial then and this PR to move and expose the DetailsItem component.

nlopin (approved):

tailwind.config.ts
```diff
@@ -267,9 +267,10 @@ export default {
       pattern: /^border-f1-/,
       variants: ["hover", "focus", "active"],
     },
-    {
-      pattern: /^icon-f1-/,
-      variants: ["hover", "focus", "active"],
-    },
+    // Had to comment this out as it was generating a warning that blocked CI because it's not being used, please reimplement when it's used
```
nlopin, line 270: merge `main`, I deleted this piece #532 today

eliseo-juan (approved):

PR #535, opened 2024-09-18
Add Tabs navigation

dani-moreno:
## 🚪 Why?

Tabs are part of the navigation components for Factorial One. This is one of the components we need for this first stage.

## 🔑 What?

- Add `tab-navigation` to ui folder (base component for this one).
- Add `Tabs` to experimental.

https://github.com/user-attachments/assets/e9d91840-64e6-4e17-8b39-c18c70d32243

<sup>_Tabs motion example_</sup>

Tabs has a `type` prop that changes the style of the tab bar, as [there's an option to have two tabs bars in some sections](https://www.figma.com/design/uImdC0GKxH59DkjzBqRJEF/Layouts-%26-Blocks?node-id=582-211787&t=qquPxamzKxKsQkZy-4).

image

<sup>_Tabs with secondary type_</sup>

## 🏡 Context

- [🎨 Figma](https://www.figma.com/design/pZzg1KTe9lpKTSGPUZa8OJ/Web-Components?node-id=175-4984)

## 🎨 Tokens

- This PR introduces a new token, `border-secondary`, used in non-interactive elements and separators.

---

> [!NOTE]
> PD: There is already a Tabs component in our codebase, that is based on the [Radix Tab component](https://www.radix-ui.com/primitives/docs/components/tabs), which is meant for tab panels. However, our Tabs function more like a [Navigation Menu](https://www.radix-ui.com/primitives/docs/components/navigation-menu), and that's what I based this component on. This way, it adheres to the [navigation role](https://www.w3.org/TR/wai-aria-1.2/#navigation) requirements.
We can remove the [Tabs](https://factorialco.github.io/factorial-one/?path=/docs/components-navigation-tabs--docs&args=variant:neutral) one, and adapt the Employee Profile to use this one.


lib/experimental/Navigation/Tabs/index.tsx
```diff
@@ -0,0 +1,36 @@
...
+  onTabChange,
+  type = "primary",
+}: TabsProps) {
+  return (
+    <TabNavigation type={type}>
+      {tabs.map((tab) => (
+        <TabNavigationLink
+          key={tab.label}
+          active={activeTab === tab.label}
+          onClick={() => onTabChange(tab.label)}
```
nlopin, line 27: We should provide the whole tab data, there is a chance the url is important or something

lib/ui/tab-navigation.tsx
```diff
@@ -0,0 +1,144 @@
...
+        : content,
+  })
+}
+
+interface TabNavigationProps
+  extends Omit<
+    React.ComponentPropsWithoutRef<typeof NavigationMenuPrimitives.Root>,
+    "orientation" | "defaultValue" | "dir"
+  > {
+  type?: "primary" | "secondary"
```
nlopin, line 28: what do you think if we have a `secondary` boolean prop and that's it. The user will do

`<Tabs ... />` for primary and `<Tabs secondary .../>` for secondary

lib/ui/tab-navigation.tsx
```diff
@@ -0,0 +1,144 @@
...
+    <NavigationMenuPrimitives.List
+      className={cn(
+        // base
+        "flex items-center justify-start gap-1 overflow-x-auto whitespace-nowrap border border-solid px-6 py-3 [scrollbar-width:none] [&::-webkit-scrollbar]:hidden",
+        // background
+        type === "primary"
+          ? "bg-f1-background-transparent"
+          : "bg-f1-background-secondary/25",
+        // border
+        "border-b border-transparent",
```
nlopin, line 45: do we need this line? Does `border-b` do something?
The transparent color is overwritten on the next line

lib/ui/tab-navigation.tsx
```diff
@@ -0,0 +1,144 @@
...
+      ...props
+    },
+    forwardedRef
+  ) => (
+    <NavigationMenuPrimitives.Item className="flex" aria-disabled={disabled}>
+      <NavigationMenuPrimitives.Link
+        data-active={active ? "true" : undefined}
+        aria-disabled={disabled}
+        className={cn(
+          "group relative flex shrink-0 select-none items-center justify-center rounded-xl no-underline focus-visible:outline-none focus-visible:ring-1 focus-visible:ring-f1-ring focus-visible:ring-offset-1",
```
nlopin, line 89: Apply the patch if you merge after #565 
```suggestion
          "group relative flex shrink-0 select-none items-center justify-center rounded-md no-underline focus-visible:outline-none focus-visible:ring-1 focus-visible:ring-f1-ring focus-visible:ring-offset-1",
```

lib/ui/tab-navigation.tsx
```diff
@@ -0,0 +1,144 @@
...
+        ref={forwardedRef}
+        onSelect={() => {}}
+        asChild={asChild}
+        {...props}
+      >
+        {getSubtree({ asChild, children }, (children) => (
+          <span
+            className={cn(
+              // base
+              "flex items-center justify-center whitespace-nowrap rounded-xl px-3 py-1.5 font-medium transition-all",
```
nlopin, line 101:
```suggestion
              "flex items-center justify-center whitespace-nowrap rounded-md px-3 py-1.5 font-medium transition-all",
```

lib/ui/tab-navigation.tsx
```diff
@@ -0,0 +1,144 @@
...
+              // background
+              type === "primary"
```
nlopin, line 103: let's maybe gather all the classes for primary and secondary to an object and inject it all together?

```ts
const classes = {
  primary: "bg-f1-background-transparent group-hover:bg-f1-background-secondary group-hover:text-f1-foreground",
secondary: "bg-f1-background/60 group-hover:border-f1-border"
}
```

Then we can get it by the type variable

lib/experimental/Navigation/Tabs/index.stories.tsx
```diff
@@ -0,0 +1,58 @@
+import { action } from "@storybook/addon-actions"
+import type { Meta, StoryObj } from "@storybook/react"
+import { useState } from "react"
+import { Tabs } from "."
+
+const meta: Meta<typeof Tabs> = {
+  component: Tabs,
+  tags: ["autodocs"],
+  argTypes: {
+    type: {
```
nlopin, line 10: looks like this control does not work, I don't see it in the storybook of the PR

lib/ui/tab-navigation.tsx
dani-moreno: Cool! That's a lot better. We probably won't have more types and that's super easy to read. Thanks 😎

lib/experimental/Navigation/Tabs/index.tsx
dani-moreno: right, done!

lib/experimental/Navigation/Tabs/index.stories.tsx
dani-moreno: I changed this now that I switched secondary to a boolean. Can you check again?

lib/ui/tab-navigation.tsx
```diff
@@ -0,0 +1,137 @@
...
+  return React.cloneElement(firstChild, {
+    children:
+      typeof content === "function"
+        ? content(firstChild.props.children)
+        : content,
+  })
+}
+
+const tabNavigationVariants = cva(
+  "flex items-center justify-start gap-1 overflow-x-auto whitespace-nowrap border-b border-b-f1-border-secondary px-6 py-3 [scrollbar-width:none] [&::-webkit-scrollbar]:hidden",
```
dani-moreno, line 25: @nlopin not sure what's going on here, if I set the border this way, it is not displayed (that's why I was doing this weird thing with border-transparent etc earlier) 😅

lib/ui/tab-navigation.tsx
dani-moreno: Done with `cva`, not sure if it looks cleaner now though that secondary is a boolean 😅

lib/ui/tab-navigation.tsx
nlopin: are you sure `border-b` does something?

dani-moreno: It does, but i was missing the solid type for the border. There's no way to add it only to the border bottom with Tailwind without adding an extended class, so I fixed it [like we do in the Breadcrumbs component](https://github.com/factorialco/factorial-one/blob/main/lib/components/Navigation/Breadcrumb/index.tsx), setting every other border to 0.

nlopin (approved):

PR #539, opened 2024-09-19
Adds explicit exports from the library

nlopin:
Describe what is possible to import from the package:
- all stable components from `@factorialco/factorial-one`
- experimental components from `@factorialco/factorial-one/experimental`
- styles 
- fonts

Everything else will throw an error on import


josepjaume (approved):
Love it! Good work. I wonder if there's some components in the playground that we should move to experimental too before this, as they are used in the app.

PR #552, opened 2024-09-23
chore: create profile layout

RamProg:
## 🚪 Why?

We are moving the Layout from Factorial to F1. This is a first version of the component, it will probably become more complex in a near future.

## 🔑 What?

We create the Profile Layout in the Experimental folder so we can use it, but it's not a final version.
 
## 🏡 Context

https://www.figma.com/design/9pLsRzdinid0ha0qRWta2D/Employee-Profile?node-id=946-24240&node-type=frame&t=y9TUKveVVCZxXEYe-0

image


lib/experimental/Layouts/ProfileLayout/index.tsx
```diff
@@ -0,0 +1,16 @@
+interface ProfileLayoutProps {
+  mainContent: React.ReactNode
+  sideContent: React.ReactNode
+}
+
+export const ProfileLayout: React.FC<ProfileLayoutProps> = ({
```
eliseo-juan, line 6: I don't feel like ProfileLayout is a good name for this Layout, I every team do the same we would have infinite number of layouts, and in that case, I don't understand why we need factorial-one. In addition, as a public library, a user could not understand what is Profile? I don't know what would be a good name, but something to do with there being a main panel and a side panel.

eliseo-juan (changes requested):
I would also think about create Separator component, let's talk with Mario about this

lib/experimental/Layouts/ProfileLayout/index.tsx
nlopin: `TwoPaneLayout` or `MainWithSidebarLayout`?

nlopin:
> I would also think about create Separator component, let's talk with Mario about this

100%, a very useful base component

lib/experimental/Layouts/ProfileLayout/index.tsx
RamProg: We had this conversation with @josepjaume and we agreed on ProfileLayout. I hesitated about the name at first, but I ended up accepting it. I don't have a strong opinion though, this is the experimental folder, so we can always rename when we have a general Layout strategy decided. Also, this Layout is not definitive, it's just what we need now, in the future it could include the tab navigation and header.

The arguments were something like this:
1) A Profile is not a Factorial component, it's like a user. It's a generic concept that different platforms understand and use similarly.
2) The name helps identify and find it easily.

But happy to agree on MainWithSidebarLayout !

RamProg:
> I would also think about create Separator component, let's talk with Mario about this

I didn't create it as a separate component as it's literally only a line with a colour, too simple to be a component. But if we decide to change the separator in the future it would be nice to have it centralised, so I'm happy to change it now.

lib/experimental/Layouts/ProfileLayout/index.tsx
nlopin: do you think it can wait for tomorrow and I merge it after a brief talk with @josepjaume? If the component is called `ProfileLayout` it suggests where it should be used, which is confusing.

nlopin:
> > I would also think about create Separator component, let's talk with Mario about this
> 
> I didn't create it as a separate component as it's literally only a line with a colour, too simple to be a component. But if we decide to change the separator in the future it would be nice to have it centralised, so I'm happy to change it now.

We have separators everywhere, by extracting the component, we avoid writing them over and over. Also external users probably prefer using a ready component in their UI

lib/experimental/Layouts/ProfileLayout/index.tsx
RamProg: It can wait for tomorrow morning, yeah, bus as it is a Layout it does block me so please let's try to move it forward early tomorrow. Thank you.

eliseo-juan: I can understand that we should not block the development of Profile, but honestly, the reasoning behind this decision for me does not make sense at all:
- We are using this layout for Profile but the same will be used by other teams for their own products in the revamp, so they are gonna be "our first customers"
- The name helps identify the layout when you are coping Profile or when you are finding Profile, but other teams would try to find "Expenses", and then they will use Profile.

I think we can make a little effort now, because this is a mandatory refactor for the future. Naming matters so much for me in this case.

josepjaume: Hey hey! So I did lean towards calling it the "Profile" layout precisely because in my mind, it should be restricted to profile-like things. Use cases are pages that have secondary details on the right -- profiles are on example, but it could also happen in other "resource" views like projects (I've seen a design with that).

TwoPaneLayout or MainWithSidebarLayout sounds too "presentational" to me. We should, whenever possible, try to find names that convey semantic meaning.

lib/experimental/Layouts/ProfileLayout/index.tsx
```diff
@@ -0,0 +1,20 @@
...
+  mainContent: React.ReactNode
+  sideContent: React.ReactNode
+}
+
+export const ProfileLayout = forwardRef<HTMLDivElement, ProfileLayoutProps>(
+  ({ mainContent, sideContent }, ref) => {
+    return (
+      <div
+        ref={ref}
+        className="grid min-h-svh grid-cols-1 py-6 sm:grid-cols-[7fr_3fr] sm:divide-x sm:divide-y-0 sm:divide-solid sm:divide-f1-foreground-disabled"
```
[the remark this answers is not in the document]
josepjaume: I think grid makes sense precisely because of that "reordering" property! It could of course be solved reversing the order of flex as well but grid feels more natural.

lib/experimental/Layouts/ProfileLayout/index.tsx
RamProg: I think we all agree on philosophy about what we are doing and how it needs to be done, but in this case we seem to be overengineering a name. If we can't all agree quickly, then we should merge this with any of the great options that we mentioned and revisit this when we have more Layouts and we decide on naming conventions for them.

nlopin: After long discussion with @josepjaume we stopped at `InfoPaneLayout` to highlight that there is an extra pane for information purposes together with the main content

RamProg:
All comments have been addressed, please take another look

nlopin (approved):

eliseo-juan (approved):

nlopin (approved):

PR #553, opened 2024-09-23
Add support to create salary widget empty state

sauldom102:
## 🔑 What?

Add support to create salary widget empty state.

## 👁️ Visual proof

Screenshot 2024-09-23 at 14 11 50

## 🏡 Context

- [💼 Jira](https://factorialmakers.atlassian.net/browse/FCT-18836)


lib/icons/AreaGraph.tsx
nlopin, line 1: is it automatically generated or manually injected? we should add `.svg` files to the `assets/icons` folder also

src/playground/Widgets/EmptyState/index.stories.tsx
```diff
@@ -10,16 +10,30 @@ const meta: Meta = {
...
-    ],
-    variant: "performance",
-    link: "/performance",
+    buttonLabel: "Try it for free",
+    buttonVariant: "promote",
+    icon: "area-graph",
   },
+  decorators: [
+    (Story) => (
+      <div className="w-[386px]">
```
nlopin, line 19: why do we need a fix width here? shouldn't the widget container provide sizing?

src/playground/Widgets/EmptyState/index.tsx
```diff
@@ -1,55 +1,48 @@
...
   content: string
-  link: string
-  variant: Variants
-  buttons: {
-    label: string
-    variant: ComponentProps<typeof ShadcnButton>["variant"]
-  }[]
+  icon: Icon
+  buttonLabel?: string
+  buttonVariant?: ComponentProps<typeof ShadcnButton>["variant"]
```
nlopin, line 14: do we need to have such a granular setup? are there examples where we use different button variants?

lib/icons/AreaGraph.tsx
sauldom102: At the moment they are manually injected. I remember that we had in gamma a way to sync the icons from figma to our frontend, we should likely start doing it in F1 too!

src/playground/Widgets/EmptyState/index.stories.tsx
sauldom102: The `WidgetContainer` goes full width. When the widget is placed in our layout with multiple columns it is auto-stretched to fit in it, but we're just showing this widget in the playground, that's why I provided this fixed width in the wrapper, so that it's closer to our design

src/playground/Widgets/Performance/index.tsx
```diff
@@ -23,12 +23,9 @@ export const PerformanceInsight = forwardRef<
...
         content={`See how ${name}'s performance evolved over time`}
-        link={link}
-        variant="performance"
-        buttons={[
-          { label: "Try it for free", variant: "neutral" },
-          { label: "Book a demo", variant: "outline" },
-        ]}
+        icon="area-graph"
+        buttonLabel="Try it for free"
+        buttonVariant="promote"
```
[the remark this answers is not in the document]
sauldom102: Haven't seen more than two buttons anymore in figma

src/playground/Widgets/EmptyState/index.tsx
sauldom102: Screenshot 2024-09-24 at 08 43 27

sauldom102: We could also provide a `promote?: boolean` prop to use the `promote` variant for the button and, if not provided, it will fall into the default variant

lib/icons/AreaGraph.tsx
nlopin: Could you also inject the `svg` files? 100% we need an automatic solution

src/playground/Widgets/EmptyState/index.tsx
nlopin: I like the idea! It will suggest that the whole widget is in the "promotion" state

RamProg:
I think some of the component used here changed names recently, try resolving conflicts please, some stuff might get broken.

src/playground/Widgets/EmptyState/index.tsx
```diff
@@ -1,55 +1,47 @@
...
-  },
-})
+const Icons: Record<Icon, ExoticComponent<{ className: string }>> = {
+  "area-graph": AreaGraph,
+  cash: Cash,
+}
 
 export const EmptyState = forwardRef<HTMLDivElement, EmptyStateType>(
-  ({ title, link, variant, content, buttons }, ref) => {
+  ({ title, content, icon, buttonLabel, promote = false }, ref) => {
```
RamProg, line 22: what about an action for the button?

sauldom102: Added!

RamProg (approved):

PR #556, opened 2024-09-23
chore: implement Details Item List

RamProg:
## 🚪 Why?

We are missing some composables to build the Details.

## 🔑 What?

We create the DetailsItemsList as clean way to build the Details and include a Title to the list of details.

## 🏡 Context

https://www.figma.com/design/9pLsRzdinid0ha0qRWta2D/Employee-Profile?node-id=946-24240&node-type=frame&t=wV2GNEoAhdWXziPx-0


nlopin:
`Avatar` should be a separate component to have a leaner API for the `Badge` component. It will be also discoverable for the devs

RamProg:
> `Avatar` should be a separate component to have a leaner API for the `Badge` component. It will be also discoverable for the devs. I see it is not in the scope of this PR though

Avatar is a separate component, but the badge has it inside for consistency in size and positioning.

nlopin (approved):

PR #569, opened 2024-09-24
Button size variants

dani-moreno:
## 🚪 Why?

There are several [size variants](https://www.figma.com/design/pZzg1KTe9lpKTSGPUZa8OJ/Web-Components?node-id=41-1256&t=DAwXjJrhRku80x3Z-4) for the Button in Figma that we were missing in code. This PR adds them.

## 🔑 What?

- `size` prop in Button.
- Add a new icon size, `sm`, that is 16px (pretty common in our designs and we were missing it).

### DIFF
--- a/lib/ui/button.tsx
+++ b/lib/ui/button.tsx
@@ -12,7 +12,7 @@ export const variants = [
   "promote",
 ] as const
 
-export const sizes = ["small", "medium", "large"] as const
+export const sizes = ["sm", "md", "lg"] as const
 
 const buttonVariants = cva(
   "group inline-flex items-center justify-center gap-1 whitespace-nowrap rounded-md border-none text-base font-medium transition-colors focus-visible:outline-none focus-visible:ring-1 focus-visible:ring-f1-ring focus-visible:ring-offset-1 disabled:pointer-events-none disabled:opacity-50",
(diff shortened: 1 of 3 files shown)


lib/ui/button.tsx
```diff
@@ -30,19 +32,24 @@ const buttonVariants = cva(
...
     defaultVariants: {
       variant: "default",
+      size: "medium",
       round: false,
     },
   }
 )
 
-const iconVariants = cva("transition-colors", {
+const iconVariants = cva("-ml-0.5 transition-colors", {
```
dani-moreno, line 52: This is a visual adjustment for the icon. When the icon is present, the left padding of the button should be a bit lower. Not sure if this is the best approach though.

josepjaume: This is weird, no? Maybe it's because of the box model? Let's merge this and I can check out what's happening :)

josepjaume (approved):
Nice job!

lib/ui/button.tsx
dani-moreno: It's weird indeed, haha. The basic idea is: the button has a left padding of 8px (for example), but when the icon is present, it should be 6px. This is a hacky way of getting that result but is, well, hacky 😅

josepjaume: Aaaaaah I understand. I think we can conditionally change this on the button itself when we do have an `icon` property instead of doing it on the icon. What do you think?

dani-moreno: The component that knows if there’s an icon is our Button, but the styling is here in the shadcn button, so I wasn’t exactly sure how to handle that. 😬

josepjaume: Aaaaah dammit. Will give it a look.

PR #573, opened 2024-09-25
Center link

josepjaume:
The link's icon was not centered. This fixes it.


lib/components/Actions/Link/index.stories.tsx
```diff
@@ -10,12 +11,23 @@ const meta = {
   tags: ["autodocs"],
   args: {
     children: "This is a link",
-    href: "https://www.youtube.com/watch?v=dQw4w9WgXcQ",
```
nlopin, line 13: I hope it was Rick Astley

eliseo-juan:
It's clear that we need links in F1, but I cannot find a 1:1 match with [Figma](https://www.figma.com/design/pZzg1KTe9lpKTSGPUZa8OJ/Web-Components?node-id=1-3&t=QcnbrRE6g0Noz3VC-1).

Maybe we should move it to experimental, or ask Dani or Mario to add it to the Figma, what do you think?

lib/components/Actions/Link/index.stories.tsx
josepjaume: Duh 😆

nlopin:
Could you also add the target switch to the args table?

nlopin (approved):

PR #574, opened 2024-09-25
Fix button icon variants location

josepjaume:
This PR started as an attempt to remove the negative margin on button, but after some consideration, I think it's OK. Still, I moved the icon variants to our Button, as they fit more naturally there (we were just exporting them and re-importing without actually using them on shadcn).

### DIFF
--- a/lib/components/Actions/Button/index.tsx
+++ b/lib/components/Actions/Button/index.tsx
@@ -1,9 +1,6 @@
 import { Icon, IconType } from "@/components/Utilities/Icon"
-import {
-  Button as ShadcnButton,
-  iconOnlyVariants,
-  iconVariants,
-} from "@/ui/button"
+import { Button as ShadcnButton } from "@/ui/button"
+import { cva } from "class-variance-authority"
 import { ComponentProps, forwardRef, useState } from "react"
 
 export type ButtonProps = Pick<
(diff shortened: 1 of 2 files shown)


josepjaume:
Dammit, no, bad approach. I think I understand how it works now 😆 .

josepjaume:
This PR started as an attempt to remove the negative margin on button, but after some consideration, I think it's OK. Still, I moved the `icon` variants to our `Button`, as they fit more naturally there (we were just exporting them and re-importing without actually using them on shadcn).

lib/components/Actions/Button/index.tsx
```diff
@@ -19,6 +16,39 @@ export type ButtonProps = Pick<
   hideLabel?: boolean
 }
 
+const iconVariants = cva("-ml-0.5 transition-colors", {
```
nlopin, line 19: > Fix button negative margin

`cva("-ml-0.5` 🤔

josepjaume: Let me quote myself 😆 

> This PR started as an attempt to remove the negative margin on button, but after some consideration, I think it's OK. Still, I moved the icon variants to our Button, as they fit more naturally there (we were just exporting them and re-importing without actually using them on shadcn).

nlopin (approved):

PR #577, opened 2024-09-25
Adds an assets folder only meant to be used by storybook

josepjaume:
## 🚪 Why?

One of our components was showing different images at each Chromatic run, which made the CI flaky.

## 🔑 What?

The image was using a URL that would return a random element. This adds a `storybook-assets` folder that we can use in this situations.


eliseo-juan (approved):
Healthy envy of Isabella

PR #578, opened 2024-09-25
chore: update composables

RamProg:
## 🚪 Why?

Adjusting widgets to updated design 

## 🔑 What?

Adjustments to Status Widget composables to fit design

## 🏡 Context

https://www.figma.com/design/9pLsRzdinid0ha0qRWta2D/Employee-Profile?node-id=604-7408&node-type=frame&t=Du1ir1wtJRXblJI7-0


lib/experimental/Widgets/Content/EventsList/event.tsx
```diff
@@ -41,9 +41,9 @@ export const Event = forwardRef<HTMLDivElement, EventProps>(
           }
         />
         <div className="flex flex-col gap-0.5">
-          <div className="flex flex-row gap-1">
+          <div className="flex flex-row gap-1 font-medium">
             <p>{title}</p>
-            <p className="text-f1-foreground-secondary">{subtitle}</p>
```
sauldom102, line 46: Why do they have the same color now? 🤔 The subtitle should be lighter as on figma

lib/experimental/Widgets/Content/EventsList/index.tsx
```diff
@@ -6,32 +7,34 @@ export interface EventsListProps {
...
+      <div ref={ref}>
+        {title && (
+          <p className="mb-3 text-sm font-semibold uppercase text-f1-foreground-secondary">
+            {title}
+          </p>
+        )}
+        <div className="flex flex-col gap-2">
+          {events.slice(0, limit).map((item) => (
+            <Event
+              key={item.title}
```
sauldom102, line 26: We should probably use a combination of multiple properties for the key, they can have the same title:

```suggestion
              key={`${item.title} ${item.subtitle} ${item.description}`}
```

lib/experimental/Widgets/Content/EventsList/event.tsx
RamProg: I think I know where the confussion might come from. The subtitle is the bit between parenthesys. As you can see in the image, it's the same colour.

image

RamProg: The lighter bit would be this one 
image

nlopin (approved):

lib/experimental/Widgets/Content/IconText/index.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+const iconsMap = {
+  office: OfficeIcon,
+  briefcase: BriefcaseIcon,
+  home: HomeIcon,
+}
+
+interface IconTextProps {
+  icon: keyof typeof iconsMap
+  texts: string[]
```
sauldom102, line 15: I'm not sure about this prop change, why not joining the text in our frontend and passing it as a `text`  string prop as we were doing before?

lib/experimental/Widgets/Content/IconTextsList/index.tsx
```diff
@@ -0,0 +1,20 @@
...
+  ({ list }, ref) => {
+    return (
+      <div ref={ref} className="mt-2 flex flex-col gap-2">
+        {list.map((item) => (
+          <IconText key={item.texts[0]} icon={item.icon} texts={item.texts} />
+        ))}
+      </div>
+    )
+  }
+)
```
sauldom102, line 18: How common is it to have a list of icon text components together? Is it worth having a dedicated component for it?

RamProg: The only usecase I've seen is a list (status widget)

lib/experimental/Widgets/Content/IconText/index.tsx
RamProg: Because now there's a small dot between the values, I need to insert it between them.

lib/experimental/Widgets/Content/EventsList/event.tsx
sauldom102: Oooh true

lib/experimental/Widgets/Content/EventsList/index.tsx
RamProg: This make sense

PR #579, opened 2024-09-25
Add Shortcut component

dani-moreno:
## 🚪 Why?

[Shortcut](https://www.figma.com/design/pZzg1KTe9lpKTSGPUZa8OJ/Web-Components?node-id=172-2793) is one of our web components. We need it for the sidebar.

## 🔑 What?

- Add `shortcut` to experimental
- `shortcut` accepts an array of keys and displays them as keystrokes.
- With some special keys, it displays an icon. [^1]

## 👀 A vé?
image

[^1]: We decided to display an icon in some cases to be able to better control how a character is displayed. We'll have [some custom icons for some keys](https://www.figma.com/design/pZzg1KTe9lpKTSGPUZa8OJ/Web-Components?node-id=224-13281).


lib/experimental/Information/Shortcut/index.tsx
```diff
@@ -0,0 +1,35 @@
...
+  keys: string[]
+}
+
+const iconMap: Record<string, IconProps["icon"]> = {
+  cmd: Cmd,
+}
+
+function Shortcut({ keys }: ShortcutProps) {
+  return (
+    <div className="inline-flex items-center space-x-0.5">
```
josepjaume, line 15: You probably want to use `gap` instead of `space-x` as it's a flex container.

lib/experimental/Information/Shortcut/index.stories.tsx
```diff
@@ -0,0 +1,29 @@
...
+export default meta
+type Story = StoryObj<typeof meta>
+
+export const Default: Story = {}
+
+export const MultipleKeys: Story = {
+  args: {
+    keys: ["↑", "↑", "↓", "↓", "←", "→", "←", "→", "B", "A"],
+  },
+}
```
josepjaume, line 29: A couple more examples wouldn't hurt. What if there's no horizontal space and you have multiple keystroke elements? How does it wrap?

I think we can get used to adding examples in context.

josepjaume (approved):
Nice job! A couple non-blocking comments.

josepjaume:
Hahaha KONAMI CODE!

PR #580, opened 2024-09-25
Kill chart flakiness

josepjaume:
## 🚪 Why?

Chart stories are flaky and report failures on the CI.

## 🔑 What?

This changes the Chart Widget's decorators so they use a fixed width instead of a `min-width` which I believe will lead us towards the path of success 💪


sauldom102 (approved):
Great!!

dani-moreno (approved):
Ahh that makes sense!

PR #595, opened 2024-09-26
chore: add gap and stories

RamProg:
## 🚪 Why?

Alerts have no padding

## 🔑 What?

add gap and stories

## 🏡 Context

https://www.figma.com/design/9pLsRzdinid0ha0qRWta2D/Employee-Profile?node-id=604-7408&node-type=frame&t=yyyyi1rRyI2asVNY-0

image


sauldom102 (approved):

eliseo-juan (commented):
I would say that is aligned on the top but Its not aligned at the bottom

nlopin:
Try to use baseline alignment, it should do the trick

Also the dot in between the title and the text looks cut

RamProg:
yeah sorry this is the updated screen

image

PR #596, opened 2024-09-26
Beautify Storybook 💅

dani-moreno:
_🧙✨ whooosh, the **mage of beautiful docs** has arrived in this Storybook to make it look good_

image

_Light theme_

image

_Dark theme_

---

PD: Had to do some weird things to make things looks correctly in the dark theme, as using filters to invert colors. As we don't have dark colors defined, they don't look properly with a dark theme. We can change that once we have dark colors.


josepjaume (approved):
OMG DANI THIS IS AWESOME 🌈

PR #598, opened 2024-09-26
Add experimental warning

josepjaume:
## 🚪 Why?

People have started using experimental components and we'd like to prevent that until they become stable.

## 🔑 What?

Adds a banner on top of experimental components.


sauldom102 (approved):
🚀

PR #600, opened 2024-09-26
Compile in docs mode in production

josepjaume:
## 🚪 Why?

The amount of information in Storybook can be overwhelming.

## 🔑 What?

Storybook has something called the "docs" mode, in which the resulting storybook project only includes the documentation (either manual or autogenerated) and not every story as a separate page.

I believe this will help us keep the project decluttered, clearer and what's important - more useful.


nlopin (approved):

dani-moreno:
image

These links don't work as the URL has changed. I'm pretty sure there are better ways to link things inside Storybook, but we could change the URL meanwhile.

dani-moreno:
Other than that, it looks **way better** this way! It was quite overwhelming before.

PR #601, opened 2024-09-26
Beautify docs tables 💅

dani-moreno:
Another bunch of classes to make tables look better and make them easier to scan and read.

| Before  | After |
| ------------- | ------------- |
| image  | image |

I also took the opportunity to turn the Typography page into a table, making it more readable:

image


nlopin (approved):
love it 😍

PR #602, opened 2024-09-26
Multiline events support for `EventsList`

nlopin:
## 🔑 What?

- allow up to 3 lines of event title (before it was breaking the layout)
- added an example to the storybook

CleanShot 2024-09-26 at 18 30 39@2x


eliseo-juan (approved):
Beautiful!

PR #614, opened 2024-09-27
Add some shadow to our widgets

sauldom102:
## 🔑 What?

Add some shadow to our widgets.

## 👁️ Visual proof

Screenshot 2024-09-27 at 17 17 25

## 🏡 Context

- [💼 Jira](https://factorialmakers.atlassian.net/browse/FCT-19213)

### DIFF
--- a/lib/experimental/Widgets/Layout/Dashboard/index.tsx
+++ b/lib/experimental/Widgets/Layout/Dashboard/index.tsx
@@ -54,7 +54,7 @@ const DashboardComponent = forwardRef<HTMLDivElement, DashboardProps>(
                       style={{
                         width: `${Math.floor((1 / columns) * 10000) / 100 - 0.05}%`,
                       }}
-                      className="pb-4 pr-4"
+                      className="pb-4 pr-4 *:shadow-md"
                     >
                       {child}
                     </div>
--- a/lib/experimental/Widgets/Layout/Dashboard/index.tsx
+++ b/lib/experimental/Widgets/Layout/Dashboard/index.tsx
@@ -39,7 +39,7 @@ const DashboardComponent = forwardRef<HTMLDivElement, DashboardProps>(
     }, [setColumns])
 
     return (
-      <div ref={ref} className="overflow-hidden">
+      <div ref={ref}>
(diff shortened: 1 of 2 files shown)


lib/experimental/Widgets/Layout/Dashboard/index.tsx
```diff
@@ -39,7 +39,7 @@ const DashboardComponent = forwardRef<HTMLDivElement, DashboardProps>(
     }, [setColumns])
 
     return (
-      <div ref={ref} className="overflow-hidden">
```
sauldom102, line 42: This is needed to prevent shadows to be cropped on the sides

nlopin (approved):

josepjaume:
This looks cool but unfortunately it breaks the WidgetStrip a bit 😢 . 

image

The whole "Widgets can be scrolled horizontally" looks like isn't going anywhere (haven't seen it on the latest designs) so maybe we can ditch it in favor of simply a responsive behaviour.

lib/experimental/Widgets/Layout/Dashboard/index.tsx
josepjaume: Unfortunately this also means that we are dropping the scroll bars 👀

nlopin: can we add some padding to accommodate the shadows?

josepjaume: But then we would need to add negative margin to the container. I think the best solution, as spoken with @sauldom102 is to make the parent (container) dictate the shadows via `[&>*]:shadow-md`. The dashboard benefits from shadows, but a strip doesn't.

If you think about it, it makes sense that this is the responsability of the parent, the same way gap or overall layout is.

josepjaume: Sorry, I'm mixing things:

1. Yes we need to remove `overflow-hidden` from dashboard, 100%
2. We can't have widgets have shadows, we need to let the container decide.

PR #615, opened 2024-09-28
Force display name

josepjaume:
## 🚪 Why?

We want our documentation to be as clean as possible, and we do have a lot of unnamed components at the moment.

## 🔑 What?

This forces `displayName` on components via an eslint rule. Since it adds the react rules hook, it also fixed some other stuff.

It basically does two things:

* Adds `displayName` to some components
* Alternatively, if using `forwardRef`, you can also use a named function like `function ComponentName` instead of an anonymous one (`() =>`) and the transpiler will use it as `displayName`.


nlopin:
the area chart changes look unrelated to display names or am I wrong?

sauldom102 (approved):
Perfect!

PR #623, opened 2024-09-30
`Weekdays`: update colors and typography

nlopin:
## 🔑 What?

- sets new colors 
- uses full day names for ARIA. Screen readers will read the full day name instead of one letter abbreviation

| Before | After |
|--------|--------|
| CleanShot 2024-09-30 at 12 54 25@2x | CleanShot 2024-09-30 at 12 48 15@2x |


sauldom102 (approved):
🚀

PR #624, opened 2024-09-30
Create Tag component

RamProg:
## 🚪 Why?

We need to update the looks of the team and employee badges.

## 🔑 What?

Create the Tag component to use in the Details sidebar.

Keep in mind I intentionally don't change the avatar looks for now.

Colors will be updated on a future PR.

## 🏡 Context

image
image

image

image


lib/experimental/Layouts/Utils/DetailsItemsList/index.stories.tsx
```diff
@@ -17,8 +17,8 @@ const weekdays = (
   />
 )
 const teams = (
-  <BadgesList
-    badges={[
+  <TagsList
+    tags={[
       { text: "Management", avatar: { alt: "M" } },
       { text: "Engineering", avatar: { alt: "E" } },
       { text: "Managers", avatar: { alt: "M" } },
```
sauldom102, line 24: Why not just taking the first letter of the `text` prop? I see `avatar->alt` kind of redundant

RamProg: I think I remember cases where they took the first two letters? If that doesn't happen they yeah, I agree

sauldom102 (commented):
In case we have an employee full name we should probably show two initials in the avatar, so in my case it would be "SD"

lib/experimental/Layouts/Utils/DetailsItemsList/index.stories.tsx
nlopin: maybe take the first one if `alt` is not there to give a meaningful default

lib/experimental/Layouts/Utils/Tag/index.tsx
```diff
@@ -0,0 +1,25 @@
...
+}
+
+export const Tag = forwardRef<HTMLDivElement, TagProps>(
+  ({ text, avatar }, ref) => {
+    return (
+      <div
+        ref={ref}
+        className="flex flex-row items-center justify-start rounded-md border border-solid border-f1-border py-1 pl-1 pr-2 font-medium text-f1-foreground"
+      >
+        <span className="mr-1.5">
```
nlopin, line 16: the parent div is flex, why not using `gap-1.5`?

nlopin (approved):

sauldom102 (commented):
Shall we add an optional `onClick` prop to make them clickable in case we want to add some navigation with it later on? 👀

sauldom102 (approved):

PR #625, opened 2024-09-30
Task/add dynamic colors

RamProg:
## 🚪 Why?

We want the background color of the tags to be dynamic but stable

## 🔑 What?

We take the same function and the same key names of the colors used in Factorial to Factorial One so we can keep stability between them, but there will be a difference as F1 has a different color system. But they will keep a certain stability.

## 🏡 Context

image


nlopin:
Don't you want to target the #624 branch to show only the color change?

lib/components/Information/Avatar/index.tsx
```diff
@@ -1,24 +1,33 @@
...
   AvatarFallback,
   AvatarImage,
 } from "@/ui/avatar"
 import { ComponentProps, forwardRef } from "react"
-
 interface AvatarType {
   alt: string
   src?: string
   size?: ComponentProps<typeof AvatarComponent>["size"]
+  color?: string
```
josepjaume, line 12: We should type this. We don't want people adding arbitrary colors here!

lib/components/Information/Avatar/index.tsx
```diff
@@ -1,24 +1,33 @@
...
-  )
-})
+export const Avatar = forwardRef<HTMLDivElement, AvatarType>(
+  ({ src, alt, size, color }, ref) => {
+    return (
+      <AvatarComponent size={size} ref={ref}>
+        <AvatarImage src={src} alt={alt} />
+        <AvatarFallback
+          className={cn(
+            color && `bg-${color}`,
```
josepjaume, line 22: This approach is usually quite problematic with tailwind. Tailwind parses the component body statically and determines which classes to keep in the compiled artifact. Since this is changing dynamically, chances are that classes are only available by accident (because other components use them).

The usual approach to do this is to map each color to a class, and this way you also have types for free:

```ts
class colorClassNames = {
  'whatever': 'bg-whatever',
  'foo': 'bg-foo'
}

type AvailableColors = typeof keyof colors
```

lib/experimental/Layouts/Utils/helper.ts
```diff
@@ -0,0 +1,29 @@
+const avatarColors: Record<string, string> = {
```
josepjaume, line 1: We're widening the types here, and this is not a good idea, as it will allow any color. Instead, use the approach mentioned before:

```tsx
type AvailableColors = keyof typeof avatarColors
```

josepjaume (changes requested):

lib/experimental/Layouts/Utils/helper.ts
josepjaume: You could maybe do something like:

```suggestion
const avatarColors: Record<AvailableColors, string> = {
```

RamProg: That's a circular dependency

RamProg: Remember that AvailableColors = keyof typeof avatarColors

RamProg: Yeah, that creates an error somewhere else, but let me see if I can find another solution to that

RamProg: Keep in mind I'm trying to keep the same function we have on the Frontend to keep some level of consistency. Just updated a new version of the types, check it out please. Thank you.

josepjaume: I understand, but this whole project is about rethinking the foundations. It might make sense to keep a similar signature, but let's not get carried away or we can easily replicate some of the mistakes as well :)

josepjaume (approved):
Nice work!

PR #629, opened 2024-09-30
Prevent legend from overflowing the chart

sauldom102:
## 🔑 What?

Prevent legend from overflowing the chart.

## 🚪 Why?

Moving the legend like that was causing it to overflow the widget container.


nlopin:
we do want to shift it. What is the issue with the overflow?

PR #631, opened 2024-09-30
Revert "Chromatic takes care of deployment now"

nlopin:
Reverts factorialco/factorial-one#604

Reason: we cannot point custom domain to our chromatic builds on the Standard plan:
CleanShot 2024-09-30 at 18 08 28@2x

### DIFF
--- a/.github/workflows/deploy.yaml
+++ b/.github/workflows/deploy.yaml
@@ -1,0 +1,29 @@
+name: Build and Publish Storybook to GitHub Pages
+
+on:
+  push:
+    branches:
+      - "main"
+
+permissions:
+  contents: read
+  pages: write
+  id-token: write
+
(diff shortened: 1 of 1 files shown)


nlopin:
We can also follow the [Gamma steps and dump everything to an S3 bucket](https://github.com/factorialco/factorial/blob/main/.github/workflows/release-gamma.yml), but docs in the branch is more standard solution.

josepjaume:
Thanks! The issue with this is that Storybook links don't really work well when deployed in a sub-folder. We should [use a custom domain](https://docs.github.com/en/pages/configuring-a-custom-domain-for-your-github-pages-site) so it's deployed on the root (`/`).

josepjaume (approved):

PR #635, opened 2024-10-01
Reorganize components (stable/experimental)

josepjaume:
## 🚪 Why?

We have too many components that are mistakenly categorized as "stable" instead of experimental. Some of them were even leftovers that shouldn't be there in the first place.

## 🔑 What?

This removes some components, moves others to "experimental" and as a cherry on top, disables exporting components from "playground"


dani-moreno:
Cool 😎 I'd say even `Avatar` and `Alert` should be in experimental for now, as they don't reflect what we have in Figma now.

josepjaume:
Let's move these two as well!

PR #636, opened 2024-10-01
chore: update shadow; add mobile shadow

RamProg:
## 🚪 Why?

Shadows don't match figma

## 🔑 What?

Correct the shadow and add shadow on mobile (1 column)

## 🏡 Context

https://www.figma.com/design/9pLsRzdinid0ha0qRWta2D/Employee-Profile?node-id=378-17717&node-type=canvas&t=6ZVacn3T1RX514Kp-0

### DIFF
--- a/lib/experimental/Widgets/Layout/Dashboard/index.tsx
+++ b/lib/experimental/Widgets/Layout/Dashboard/index.tsx
@@ -42,13 +42,7 @@ const _Dashboard = forwardRef<HTMLDivElement, DashboardProps>(
       <div ref={ref} className="text-f1-foreground">
         <div ref={containerRef}>
           {columns === 1 ? (
-            <div className="flex flex-col gap-4">
-              {children?.map((child, i) => (
-                <span className="*:shadow" key={i}>
-                  {child}
-                </span>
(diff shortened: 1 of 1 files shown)


lib/experimental/Widgets/Layout/Dashboard/index.tsx
```diff
@@ -42,7 +42,13 @@ const _Dashboard = forwardRef<HTMLDivElement, DashboardProps>(
       <div ref={ref} className="text-f1-foreground">
         <div ref={containerRef}>
           {columns === 1 ? (
-            <div className="flex flex-col gap-4">{children}</div>
+            <div className="flex flex-col gap-4">
+              {children?.map((child, i) => (
+                <span className="*:shadow" key={i}>
```
nlopin, line 47: why not adding the shadow selector to the class above, assuming that the only children allowed are widgets? 

`span` is not a good tag for the purpose because it is an inline element. It will be an invalid html if they will contain block elements, like _divs_, inside. If the above is not possible, let's at least use `div` instead of `span` here

nlopin (approved):

PR #637, opened 2024-10-01
Do not show tooltip when we are blurring the chart

sauldom102:
## 🔑 What?

Do not show tooltip when we are blurring the chart.

## 🚪 Why?

We shouldn't show any tooltip when hovering the chart in case it's blurred.

## 👁️ Visual proof

https://github.com/user-attachments/assets/4a5e00b8-f7d0-445b-9308-fc3de43a8a2d

### DIFF
--- a/lib/components/Charts/AreaChart/index.tsx
+++ b/lib/components/Charts/AreaChart/index.tsx
@@ -182,15 +182,17 @@ export const _AreaChart = <K extends LineChartConfig>(
             className={cn(yAxis?.isBlur && "blur-sm")}
           />
         )}
-        <ChartTooltip
-          cursor
-          content={
-            <ChartTooltipContent
-              indicator="dot"
-              yAxisFormatter={yAxis?.tickFormatter}
-            />
-          }
-        />
+        {!yAxis?.isBlur && (
+          <ChartTooltip
+            cursor
+            content={
+              <ChartTooltipContent
+                indicator="dot"
+                yAxisFormatter={yAxis?.tickFormatter}
+              />
+            }
+          />
+        )}
         {areas.map((area, index) => (
           <Area
             isAnimationActive={false}


RamProg (approved):

lib/components/Charts/AreaChart/index.tsx
```diff
@@ -182,15 +182,17 @@ export const _AreaChart = <K extends LineChartConfig>(
...
+          <ChartTooltip
+            cursor
+            content={
+              <ChartTooltipContent
+                indicator="dot"
+                yAxisFormatter={yAxis?.tickFormatter}
+              />
+            }
+          />
+        )}
```
josepjaume, line 195: I think we're pushing too much logic on the charts. I would totally rethink this "blur" thing and try to find a higher-order solution (maybe even a higher-order component / wrapper). It can easily pollute the whole codebase.

nlopin: let's then start rethinking it until it is too late

josepjaume: You mean "before it's too late" because "until it's too late" it's what we want to avoid 😆

nlopin: [Before would have been better, but before is over](https://youtube.com/watch?v=WqH4Bpe7Mjk&t=109)

RamProg: If we wanna blur the YAxis, then the chart needs to know if it should be blurred, right? How else will be blur it? Adding an overlay doesn't make a lot of sense and the alternative would be passing down the className prop which would give too much control of the UI. I understand the concern, I just don't see a better option. It makes sense to me that the chart knows how to blur itself.

josepjaume: Creating a <PrivacyProvider /> and then letting the card act upon its state would help us avoid that `blur` property in the contract and also encourage a global solution. What do you think?

PR #638, opened 2024-10-01
Add Focus Ring class

dani-moreno:
## 🚪 Why?

We use a [focus ring/indicator](https://web.dev/articles/style-focus) in a few components, adding several Tailwind classes to create it.

If we want to change it, we'd have to go through each component one by one to update those classes. It’s also easy to misalign and establish a different focus ring, leading to inconsistency. That’s why I built this utility class to apply the styles consistently.

## 🔑 What?

- Add a `focusRing` utility class. This class accepts extra classes, in case it is needed.
- Apply it to several components already using a focus ring.

### DIFF
--- a/lib/lib/utils.ts
+++ b/lib/lib/utils.ts
@@ -4,3 +4,10 @@ import { twMerge } from "tailwind-merge"
 export function cn(...inputs: ClassValue[]) {
   return twMerge(clsx(inputs))
 }
+
+export function focusRing(extraClasses?: string) {
+  return cn(
+    "focus-visible:outline-none focus-visible:ring-1 focus-visible:ring-f1-ring focus-visible:ring-offset-1",
+    extraClasses
+  )
+}
--- a/lib/experimental/Navigation/Sidebar/Menu/index.tsx
+++ b/lib/experimental/Navigation/Sidebar/Menu/index.tsx
@@ -1,6 +1,6 @@
 import { Counter } from "@/experimental/Information/Counter"
 import * as Icons from "@/icons"
-import { cn } from "@/lib/utils"
+import { cn, focusRing } from "@/lib/utils"
 import {
   Collapsible,
   CollapsibleContent,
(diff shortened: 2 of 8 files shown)


lib/lib/utils.ts
```diff
@@ -4,3 +4,10 @@ import { twMerge } from "tailwind-merge"
 export function cn(...inputs: ClassValue[]) {
   return twMerge(clsx(inputs))
 }
+
+export function focusRing(extraClasses?: string) {
+  return cn(
+    "focus-visible:outline-none focus-visible:ring-1 focus-visible:ring-f1-ring focus-visible:ring-offset-1",
+    extraClasses
+  )
+}
```
josepjaume, line 13: I wonder why you use a function here instead of exporting a simple string:

```suggestion
export const focusRing = `focus-visible:outline-none focus-visible:ring-1 focus-visible:ring-f1-ring focus-visible:ring-offset-1`
```

dani-moreno: Some elements need more classes to build the focus ring as it should look, that's why I built it as a function that accepts extra classes. This way everything related to the focus ring stays in the same place.

I put this together as a starting point, but I'm very open to change it if there are better ways!

dani-moreno: Yes, but they kind of get lost in some long lists of classes. This way, everything stays on the same place, like in the [Menu component](https://github.com/factorialco/factorial-one/pull/638/files#diff-28e3f62644251b16bc2bebe038208020fa6f7a14f1af65b81704a64231331ebc), for example:

```
className={cn(
    "flex w-full cursor-pointer items-center justify-between border-t border-dashed border-transparent border-t-f1-border px-1.5 pb-2 pt-4 text-sm font-semibold uppercase tracking-wide text-f1-foreground-secondary",
    focusRing("focus-visible:rounded-md")
 )}
```

Just trying to make it a bit more readable.

josepjaume: Okay, I buy it!

josepjaume: It's nice how we're basically inventing CSS all over again, but on top of Tailwind 😆

PR #648, opened 2024-10-04
Add ModuleAvatar

dani-moreno:
## 🚪 Why?

ModuleAvatar is one of our web components, mainly used in the header alongside breadcrumbs. We need it to recreate our navigation components design.

## 🔑 What?

- Add ModuleAvatar.

image

## 🏡 Context

- [🎨 Figma](_figma_link_)


lib/experimental/Information/ModuleAvatar/index.tsx
```diff
@@ -0,0 +1,103 @@
...
+    React.ComponentType<React.SVGProps<SVGSVGElement>>
+  >
+)
+
+export type IconType = keyof typeof iconComponents
+
+const moduleAvatarVariants = cva("relative flex items-center justify-center", {
+  variants: {
+    size: {
+      sm: "h-5 w-5",
```
nlopin, line 50: we need to set `shrink-0` to the base style, otherwise it will be squished in narrow flex containers

lib/experimental/Information/ModuleAvatar/index.tsx
```diff
@@ -0,0 +1,103 @@
...
+      md: "h-6 w-6",
+      lg: "h-8 w-8",
+    },
+  },
+  defaultVariants: {
+    size: "md",
+  },
+})
+
+const iconSizeVariants = cva("relative z-10 text-f1-foreground-inverse", {
```
nlopin, line 60: why do we need a z-index?

dani-moreno:
> The naming is confusing. Is it the name from Figma?

Yep it is, but I also think it can be confusing. Maybe we can come up with a better name.

lib/experimental/Information/ModuleAvatar/index.tsx
dani-moreno: Ooops, not really!

nlopin (approved):

PR #660, opened 2024-10-08
chore: create progess bar duo; adapt two columns list

RamProg:
## 🚪 Why?

We need to display a two column list with a progress bar for the competencies widget

## 🔑 What?

We create the progress bar duo component and adapt the two colum list together with a story.

## 🏡 Context

https://www.figma.com/design/9pLsRzdinid0ha0qRWta2D/Employee-Profile?node-id=604-7408&node-type=frame&t=9ILoPFa3hL2KCP3V-0

image

### DIFF
--- a/lib/experimental/Widgets/Content/ProgressBarDuo/index.tsx
+++ b/lib/experimental/Widgets/Content/ProgressBarDuo/index.tsx
@@ -1,0 +1,32 @@
+import { autoColor } from "@/components/Charts/utils/colors"
+import { cn } from "@/lib/utils"
+import { forwardRef } from "react"
+
+interface DualProgressBarProps {
+  value: number
+  max?: number
+}
+
+export const ProgressBarDuo = forwardRef<HTMLDivElement, DualProgressBarProps>(
+  function ProgressBarDuo({ value, max = 100 }: DualProgressBarProps, ref) {
+    const percentage = Math.min(100, Math.max(0, (value / max) * 100))
+
+    return (
+      <div ref={ref} className="flex h-2 w-full gap-0.5 overflow-hidden">
+        <div
+          className={cn(
+            "rounded transition-all duration-300 ease-in-out",
+            `bg-[${autoColor(0)}]`
+          )}
+          style={{ width: `${percentage}%` }}
+        />
+        <div
+          className="rounded bg-f1-foreground-disabled transition-all duration-300 ease-in-out"
+          style={{ width: `${100 - percentage}%` }}
+        />
+      </div>
+    )
+  }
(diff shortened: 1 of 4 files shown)


lib/experimental/Widgets/Content/TwoColumnsList/index.tsx
```diff
@@ -1,28 +1,34 @@
...
-  function TwoColumnsList({ list }, ref) {
+  function TwoColumnsList({ title, list }, ref) {
     return (
-      <div ref={ref} className="grid grid-cols-[1fr_auto] gap-2">
-        {list.map((item) => (
-          <Item key={item.title} title={item.title} info={item.info} />
-        ))}
+      <div ref={ref} className="flex flex-col gap-2">
+        {title && <div className="font-medium">{title}</div>}
+        <div className="flex flex-col gap-2">
```
nlopin, line 27: this div looks redundant, the parent one is exactly the same. If you delete it, does it change anything?

lib/experimental/Widgets/Content/TwoColumnsList/index.tsx
```diff
@@ -1,28 +1,34 @@
...
   list: TwoColumnsItemType[]
 }
 
 const Item = ({ title, info }: TwoColumnsItemType) => (
-  <>
-    <div className="line-clamp-2 text-f1-foreground-secondary">{title}</div>
-    <div className="font-medium">{info}</div>
-  </>
+  <div className="flex items-center justify-between">
+    <div className="flex text-f1-foreground-secondary">{title}</div>
```
nlopin, line 15: Consider switching to a paragraph to stay semantic, or maybe it should even be an item list.

I also dropped the `flex` since there are no block children inside
```suggestion
    <p className="text-f1-foreground-secondary">{title}</p>
```

lib/experimental/Widgets/Content/ProgressBarDuo/index.tsx
nlopin, line 1: is there a chance we get more segments to color, not only two? We can use the same approach [`CategoryBarSection`](https://one.factorial.dev/?path=/docs/experimental-widgets-content-categorybarsection--documentation) does with its `data` prop to keep the api the same

RamProg: I'm not aware of any usecases where there are more than 2 segments

RamProg: For sure not for widgets

nlopin (approved):

PR #663, opened 2024-10-08
Improve `Link` in navigation

josepjaume:
This PR does two things:

* Creates a `NavigationItem` type meant to be used by all navigation elements. It makes sure we normalize how we express navigation items across the platform.
* Adds missing exports

### DIFF
--- a/lib/lib/linkHandler.tsx
+++ b/lib/lib/linkHandler.tsx
@@ -39,7 +39,9 @@ export const useLinkContext = () => {
   }
 }
 
-export type LinkProps = AnchorHTMLAttributes<HTMLAnchorElement>
+export type LinkProps = AnchorHTMLAttributes<HTMLAnchorElement> & {
+  exactMatch?: boolean
+}
 
 export const useNavigation = () => {
   const { currentPath } = useLinkContext()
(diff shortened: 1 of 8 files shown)


lib/lib/linkHandler.tsx
```diff
@@ -65,7 +67,7 @@ export const Link = forwardRef<HTMLAnchorElement, LinkProps>(
     const { isActive } = useNavigation()
 
     const overridenProps = {
-      "data-is-active": isActive(props.href) || undefined,
+      "data-is-active": isActive(props.href, { exact: props.exactMatch }),
```
nlopin, line 70: We should set `exactMatch` default to `false` for the sake of reducing the the optional prop pollution. If we don't stop it here, it will spread everywhere

nlopin (approved):

lib/lib/linkHandler.tsx
josepjaume: Not sure I follow! `exactMatch` defaults to `false` in the implementation of `isActive`:

https://github.com/factorialco/factorial-one/pull/663/files#diff-1386e11f651c9b9dd4ccd2245ad6444068c5d8176915645c3554db1d2b6ffb46R51

PR #665, opened 2024-10-08
Add barbones ApplicationFrame component

josepjaume:
Adds a barebones `ApplicationFrame` component for @dani-moreno to improve.


dani-moreno (approved):
Let's goooo! 🚀

PR #666, opened 2024-10-08
Add Breadcrumbs & Header

dani-moreno:
## 🚪 Why?

We're working on the main navigation components for teams to adapt. These two are key elements of that.

## 🔑 What?

- Add `Breadcrumbs`.
- Add `Header`.

---

`Breadcrumbs` is a basic component that accepts a data structure of links and displays them.

image

---

`Header` is the page header component in Factorial's layout. It uses Breadcrumbs and includes a few other necessary elements.

**Important:** This is the basic component, but it needs more work, which I'll handle in a different PR. Next tasks are:
- Make it responsive by collapsing items into a dropdown.
- The right menu buttons are placeholders for now; they also need to collapse when there are too many.

image

## 🏡 Context

- [🎨 Figma](https://www.figma.com/design/pZzg1KTe9lpKTSGPUZa8OJ/Web-Components?node-id=38-709&t=3qBwKfPMtcG0JzcY-4)


lib/experimental/Navigation/Header/Breadcrumbs/index.tsx
```diff
@@ -0,0 +1,75 @@
...
+} from "@/experimental/Information/ModuleAvatar"
+
+import { ChevronRight } from "@/icons"
+import { Link } from "@/lib/linkHandler"
+import { cn, focusRing } from "@/lib/utils"
+import { NavigationItem } from "../../utils"
+
+export type BreadcrumbItemType = NavigationItem & {
+  icon?: IconType
+}
```
josepjaume, line 22: Cool!

lib/experimental/Navigation/Header/Header/index.tsx
```diff
@@ -0,0 +1,86 @@
...
+  tree = [],
+  actions = [],
+  menu,
+}: HeaderProps) {
+  const breadcrumbsTree: BreadcrumbItemType[] = [
+    { label: module.name, href: module.href, icon: module.icon },
+    ...tree,
+  ]
+
+  const isFirstLevel = !tree.length
```
josepjaume, line 41: Careful with this - in JavaScript, `0` is `falsey`. Which means that `!0 == true`. I don't think it matters in this case, but it reads weird. Maybe `tree.length > 0` would be better?

dani-moreno: Ah JavaScript is weird, haha. Changed the check.

lib/experimental/Navigation/Header/Breadcrumbs/index.tsx
```diff
@@ -0,0 +1,75 @@
...
+      )}
+    </ShadBreadcrumbItem>
+  )
+}
+
+interface BreadcrumbsProps {
+  tree: BreadcrumbItemType[]
+}
+
+export default function Breadcrumbs({ tree }: BreadcrumbsProps) {
```
josepjaume, line 61: Please remember to add this on `exports.tsx`.

lib/experimental/Navigation/Header/Breadcrumbs/index.tsx
```diff
@@ -0,0 +1,75 @@
...
+        </>
+      ) : (
+        <BreadcrumbPage>{item.label}</BreadcrumbPage>
+      )}
+    </ShadBreadcrumbItem>
+  )
+}
+
+interface BreadcrumbsProps {
+  tree: BreadcrumbItemType[]
```
josepjaume, line 58: This isn't a `tree` no? It's just an array 👁️ . The name is a bit misleading, perhaps?

lib/experimental/Navigation/Header/Header/index.tsx
```diff
@@ -0,0 +1,86 @@
...
+  type IconType as ModuleAvatarIconType,
+} from "@/experimental/Information/ModuleAvatar"
+
+type HeaderProps = {
+  module: {
+    name: string
+    href: string
+    icon: ModuleAvatarIconType
+  }
+  tree?: BreadcrumbItemType[]
```
josepjaume, line 18: What about just using `breadcrumbs`?

lib/experimental/Navigation/Header/Header/index.tsx
```diff
@@ -0,0 +1,86 @@
...
+  actions?: {
+    label: string
+    icon: IconType
+    onClick: () => void
+  }[]
+  menu: {
+    show: boolean
+    onClick: () => void
+  }
```
josepjaume, line 27: Maybe we should be more prescriptive in terms of what actions do we allow here? I thought we wanted to encourage "settings" and "help" or somethin like that. Do we want to go specific, or give more freedom to devs? I'm afraid of abuse.

lib/experimental/Navigation/Header/Header/index.tsx
```diff
@@ -0,0 +1,86 @@
...
+}
+
+export default function Header({
+  module,
+  tree = [],
+  actions = [],
+  menu,
+}: HeaderProps) {
+  const breadcrumbsTree: BreadcrumbItemType[] = [
```
josepjaume, line 36: Not a tree 👀 😆

lib/experimental/Navigation/Header/Header/index.tsx
```diff
@@ -0,0 +1,86 @@
...
+          {actions.map((action, index) => (
+            <Button
+              hideLabel
+              round
+              key={index}
+              variant="outline"
+              onClick={action.onClick}
+              label={action.label}
+              icon={action.icon}
+            />
```
josepjaume, line 80: Should actions be links? Or should they accept one or the other? Or maybe it's OK that they are buttons, as they shouldn't navigate away?

lib/experimental/Navigation/Header/Breadcrumbs/index.tsx
dani-moreno: Hah, you're completely right, I think I got carried away as I made the menu just before this one.

lib/experimental/Navigation/Header/Header/index.tsx
dani-moreno: I reviewed this with Mario this morning, and it's quite open. Actions aren't limited to just a few, they can be anything to support every case we have at Factorial.

josepjaume (approved):
Nice, let's go!

PR #668, opened 2024-10-09
Update Select options & styling

dani-moreno:
## 🚪 Why?

`Select` is one of our main web components, and we need to use it for our navigation components (for the Company Selector). That's why I updated the styling and some options needed.

## 🔑 What?

- Update Select.
- Option to add an `icon` and `description` to each item.
- Option to have custom triggers (needed for the Company Selector).
- Option to disable it.

image

## 🏡 Context

- [🎨 Figma](https://www.figma.com/design/pZzg1KTe9lpKTSGPUZa8OJ/Web-Components?node-id=252-11020&t=3qBwKfPMtcG0JzcY-4)

### DIFF
--- a/lib/experimental/Forms/Fields/Select/index.tsx
+++ b/lib/experimental/Forms/Fields/Select/index.tsx
@@ -32,7 +32,7 @@ const SelectItem = ({ item }: { item: SelectItemProps<string> }) => {
   return (
     <SelectItemPrimitive value={item.value}>
       <div className="flex items-start gap-1.5">
-        {Icon && <Icon className="h-5 w-5 text-f1-icon" />}
+        {Icon && <Icon className="h-5 w-5 shrink-0 text-f1-icon" />}
         <div className="flex flex-col">
           <span className="font-medium">{item.label}</span>
           {item.description && (
@@ -50,7 +50,7 @@ const SelectValue = ({ item }: { item: SelectItemProps<string> }) => {
   const Icon = item.icon && Icons[item.icon]
   return (
     <div className="flex items-center gap-1.5">
-      {Icon && <Icon className="h-5 w-5 text-f1-icon" />}
+      {Icon && <Icon className="h-5 w-5 shrink-0 text-f1-icon" />}
       {item.label}
     </div>
   )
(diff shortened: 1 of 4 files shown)


nlopin:
Have you tried putting the checkmark in the middle of the section? How does it look?

lib/experimental/Forms/Fields/Select/index.tsx
```diff
@@ -1,31 +1,100 @@
...
+      </div>
+    </SelectItemPrimitive>
+  )
 }
 
+const SelectValue = ({ item }: { item: SelectItemProps<string> }) => {
+  const Icon = item.icon && Icons[item.icon]
+  return (
+    <div className="flex items-center gap-1.5">
+      {Icon && <Icon className="h-5 w-5 text-f1-icon" />}
```
nlopin, line 53: should we put `shrink-0` here or we already have in on the Icon component level?

lib/ui/select.tsx
```diff
@@ -118,14 +108,15 @@ const SelectItem = React.forwardRef<
...
     className={cn(
-      "relative flex w-full cursor-default select-none items-center rounded-xs py-2 pl-2 pr-8 text-sm outline-none focus:bg-f1-background-secondary focus:text-f1-foreground data-[disabled]:pointer-events-none data-[disabled]:opacity-50",
+      "relative flex w-full cursor-default select-none items-center rounded p-2 outline-none transition-colors focus:bg-f1-background-secondary focus:text-f1-foreground data-[disabled]:pointer-events-none data-[disabled]:opacity-50",
+      "data-[state=checked]:bg-f1-background-selected-bold/5 hover:data-[state=checked]:bg-f1-background-selected-bold/10",
       className
     )}
     {...props}
   >
-    <span className="absolute right-2 flex h-3.5 w-3.5 items-center justify-center">
+    <span className="absolute right-2 top-2 flex items-center justify-center">
```
nlopin, line 117: does `span` work well here? if one of the inner children is not an inline element, it won't be semantically correct to use `span` and we should switch to `div` instead

nlopin (approved):

lib/experimental/Forms/Fields/Select/index.tsx
dani-moreno: Ah, good catch! The Icon component does have `shrink-0`, but as I'm using the icon directly it is not applied, so adding it.

Now that I think about it, I should be using the Icon component instead...

lib/ui/select.tsx
dani-moreno: Right! That's what shadcn provides by default, but you're completely right, it's good that we review it.

PR #670, opened 2024-10-09
Add SidebarIcon

dani-moreno:
## 🔑 What?

Adds `SidebarIcon`, a small part of the sidebar that lets the user collapse or expand it. The component displays a dynamic icon depending on the sidebar's state (expanded or not) with a subtle, cool animation 😎 ✨  

---

https://github.com/user-attachments/assets/2656640b-5406-4987-98f3-e8e22d2347dd

Collapse

---

https://github.com/user-attachments/assets/1286c6f5-f909-4dde-aa64-71dd4b8cc77f

Expand

---

## 🏡 Context

- [🎨 Figma](https://www.figma.com/design/pZzg1KTe9lpKTSGPUZa8OJ/Web-Components?node-id=182-6313&t=avsQvT5FQ3iTx8rS-4)

### DIFF
--- a/lib/experimental/Navigation/Sidebar/Icon/index.stories.tsx
+++ b/lib/experimental/Navigation/Sidebar/Icon/index.stories.tsx
@@ -1,0 +1,17 @@
+import type { Meta, StoryObj } from "@storybook/react"
+import { SidebarIcon } from "."
+
+const meta: Meta<typeof SidebarIcon> = {
+  component: SidebarIcon,
+  tags: ["autodocs"],
+}
+
+export default meta
+
+type Story = StoryObj<typeof SidebarIcon>
+
+export const Default: Story = {
+  args: {
+    isExpanded: true,
+  },
+}
--- a/lib/experimental/Navigation/Sidebar/Icon/index.tsx
+++ b/lib/experimental/Navigation/Sidebar/Icon/index.tsx
@@ -67,6 +67,7 @@ export function SidebarIcon({ isExpanded, onClick }: SidebarIconProps) {
       round
       onClick={onClick}
       className="group"
+      title="Toggle Sidebar"
     >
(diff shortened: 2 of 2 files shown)


lib/experimental/Navigation/Sidebar/Icon/index.stories.tsx
nlopin, line 1: do we need a story for it? with all the love to the amazing work done, it is too small to be a separate thing. I don't see how it can be useful outside of the sidebar navigation

lib/experimental/Navigation/Sidebar/Icon/index.tsx
```diff
@@ -0,0 +1,75 @@
...
+
+export function SidebarIcon({ isExpanded, onClick }: SidebarIconProps) {
+  return (
+    <Button
+      variant="outline"
+      size="md"
+      round
+      onClick={onClick}
+      className="group"
+      title="Toggle Sidebar"
```
nlopin, line 70: we will need translations for that 😞

lib/experimental/Navigation/Sidebar/Icon/index.stories.tsx
dani-moreno: Ah, sure! I guess we don't need stories for all these little 'pieces'. I just use them to work, but I can remove them to reduce the noise.

Let me keep them until I finish these components, and then I can prepare a PR to remove the unnecessary ones. They kind of help me work for now.

lib/experimental/Navigation/Sidebar/Icon/index.tsx
dani-moreno: What if this is a prop? So the translation comes from Factorial, not the design system?

nlopin (approved):

PR #672, opened 2024-10-10
Animate menu icon

dani-moreno:
Transition the color of the icon between states, as well as the menu item does.
Also, uses the correct token for the active icon.

### DIFF
--- a/lib/experimental/Navigation/Sidebar/Menu/index.tsx
+++ b/lib/experimental/Navigation/Sidebar/Menu/index.tsx
@@ -43,8 +43,8 @@ const MenuItemContent = ({
       <div className="flex items-center gap-1.5 font-medium text-f1-foreground">
         <IconComponent
           className={cn(
-            "h-5 w-5",
-            active ? "text-f1-foreground" : "text-f1-icon"
+            "h-5 w-5 transition-colors",
+            active ? "text-f1-icon-bold" : "text-f1-icon"
           )}
         />
         <span>{item.label}</span>


lib/experimental/Navigation/Sidebar/Menu/index.tsx
```diff
@@ -43,8 +43,8 @@ const MenuItemContent = ({
       <div className="flex items-center gap-1.5 font-medium text-f1-foreground">
         <IconComponent
           className={cn(
-            "h-5 w-5",
-            active ? "text-f1-foreground" : "text-f1-icon"
+            "h-5 w-5 transition-colors",
+            active ? "text-f1-icon-bold" : "text-f1-icon"
```
josepjaume, line 47: Is it like really fast? Not sure I'm seeing a transition there 👀 .

Also, random thought: Should we eventually add some guidelines on animations? Like, which things we believe are "good" animations and which things aren't?

josepjaume (approved):

lib/experimental/Navigation/Sidebar/Menu/index.tsx
dani-moreno: Ah, there seem to be some override with the Navigation Item component and the animation does not happen? It works with a regular element.

dani-moreno: Btw, about animation guidelines, I will! It's a bit soon yet but we can have something like [this](https://polaris.shopify.com/design/motion) in the near near future.

josepjaume: That would be amazing!

PR #673, opened 2024-10-10
Use `<link />` on Dropdown

dani-moreno:
Adapts the Dropdown component to use the `Link` component for links.

### DIFF
--- a/lib/experimental/Navigation/Dropdown/index.tsx
+++ b/lib/experimental/Navigation/Dropdown/index.tsx
@@ -98,7 +98,7 @@ export function Dropdown({ items, children }: DropdownProps) {
               key={index}
               item={{
                 ...item,
-                onClick: item.onClick ? () => item.onClick?.() : undefined,
+                onClick: item.onClick,
               }}
             />
           ))}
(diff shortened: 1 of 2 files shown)


lib/experimental/Navigation/Dropdown/index.tsx
```diff
@@ -79,9 +98,7 @@ export function Dropdown({ items, children }: DropdownProps) {
               key={index}
               item={{
                 ...item,
-                onClick: () => {
-                  item.onClick()
-                },
+                onClick: item.onClick ? () => item.onClick?.() : undefined,
```
RamProg, line 101: This looks redundant, you are checking twice for item.onClick

RamProg:
```suggestion
                onClick: () => item.onClick?.()
```

dani-moreno: Thanks! Can't it be just `onClick: item.onClick`, as it is not defined, it's going to be `undefined` anyway? Or am I missing something?

RamProg: Yep, that should work too

RamProg: What wouldn't work is the original because if onClick is undefined it will break

RamProg (approved):

PR #674, opened 2024-10-10
Add scrollbar to experimental

Rogermax:
## 🚪 Why?

To expose ScrollBar in experimental

## 🔑 What?

Add the export in experimental file


nlopin:
I double checked the code, the shadcn documentation and discussed the export with the team. We decided not to expose the scroll bar since it is an internal component for the `ScrollArea` component. 

`ScrollArea` gives a similar functionality, exposing another component creates a point where a developer needs to take a decision what approach to use. We prefer to avoid this by providing one clear solution for scrolling

PR #676, opened 2024-10-10
Add blend to scroll area if not at bottom

Rogermax:
## 🚪 Why?

To add a max height to widgets

## 🔑 What?

Add ScrollArea to Widget with a max height to avoid widgets grow too much. Making the scroll with ScrollArea and improving ScrollArea component to show a blend in case you don't reach the bottom

https://github.com/user-attachments/assets/52d295ae-9c62-4281-aa33-e9baad6e137f

### DIFF
--- a/lib/experimental/Widgets/Widget/index.stories.tsx
+++ b/lib/experimental/Widgets/Widget/index.stories.tsx
@@ -78,6 +78,22 @@ export const MultipleContent: Story = {
   },
 }
 
+export const LongContentScrollArea: Story = {
+  args: {
+    ...meta.args,
+    children: [
+      Array.from({ length: 6 }).map((_, index) => (
+        <div
+          key={index}
+          className="rounded-sm bg-f1-background-secondary p-3 text-center text-f1-foreground"
+        >
+          Content {index + 1}
+        </div>
+      )),
+    ],
+  },
+}
+
 export const Full: Story = {
   args: {
     ...meta.args,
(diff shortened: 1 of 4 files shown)


eliseo-juan (commented):
Honestly, I'm not sure widgets should have scrollable content. By the nature of widgets, they are already going to appear in places with their own scroll, the truth is that I don't know if we want to scroll inside a widget inside a scrolling screen.

Rogermax:
> Honestly, I'm not sure widgets should have scrollable content. By the nature of widgets, they are already going to appear in places with their own scroll, the truth is that I don't know if we want to scroll inside a widget inside a scrolling screen.

We want to fix that dashboard (now feels horrible): 


https://github.com/user-attachments/assets/be711400-7ffa-453d-86bb-fdc49b89a62e


The idea is to let dashboard configurable to allow has the minWidth widgets bigger
And we want to avoid inifinite long widgets

eliseo-juan (changes requested):
After talking with the team we all agreed that adding scroll to widgets could be a big rabbit hole, user has to discover if one have it or not, and also makes the scroll something impossible to navigate, you can experience it in the current dashboard. 

The aim of the widget is to provide an insight or a subset of information that could be expanded in another view, but not to group information like in that dashboard. René is gonna talk with Sam to find a solution for that.

Feel free to open this kind of discussions in #factorial-one

Rogermax:
> After talking with the team we all agreed that adding scroll to widgets could be a big rabbit hole, user has to discover if one have it or not, and also makes the scroll something impossible to navigate, you can experience it in the current dashboard.
> 
> The aim of the widget is to provide an insight or a subset of information that could be expanded in another view, but not to group information like in that dashboard. René is gonna talk with Sam to find a solution for that.
> 
> Feel free to open this kind of discussions in #factorial-one

Can we, at least, keep the improvement "fade-out effect" in the ScrollArea component where you can scroll?

PR #677, opened 2024-10-10
Responsive breadcrumbs

dani-moreno:
## 🚪 Why?

In some cases, the header breadcrumbs will have a long list of items (e.g., Documents, where you can go as deep as you want in folders). That's why we need a way to truncate them.

We're going the "Notion" route, so when there are too many items to display:
- The first item is always shown, so it's easy for the user to know which section they're in.
- The last items are displayed until there's no more space.
- The remaining items are collapsed into a dropdown.

https://github.com/user-attachments/assets/367f30b0-d0bf-4cc4-b658-f68cb7ddb0d0

## 🔑 What?

Items that has to be collapsed are calculated at load and when the screen resizes. This is briefly what's happening:

- Get the breadcrumb container width.
- Define a fixed width for each breadcrumb item.
- From right to left, subtract an item width until there's not enough space.
- The final count represents how many breadcrumbs can be shown.
- items beyond this count will be collapsed into the dropdown.

The function prioritizes showing the last (rightmost) breadcrumbs, while ensuring there's space for the dropdown if needed.

----

_Note: This is the 'safest' approach and a good starting point, but calculating the actual width of each item instead of using a fixed width will lead to a more accurate calculation. The current method leaves a lot of blank space on the right, as it assumes every item has the maximum possible width._

## 🏡 Context

- [🎨 Figma](https://www.figma.com/design/pZzg1KTe9lpKTSGPUZa8OJ/Web-Components?node-id=14-564)

### DIFF
--- a/lib/experimental/Navigation/Header/Breadcrumbs/index.tsx
+++ b/lib/experimental/Navigation/Header/Breadcrumbs/index.tsx
@@ -130,9 +130,9 @@ export default function Breadcrumbs({ breadcrumbs }: BreadcrumbsProps) {
         {collapsedItems.length > 0 && (
           <>
             <Dropdown items={collapsedItems as DropdownItemWithoutIcon[]}>
-              <div className="rounded-sm px-1.5 py-0.5 font-medium text-f1-foreground no-underline transition-colors hover:bg-f1-background-secondary">
+              <li className="rounded-sm px-1.5 py-0.5 font-medium text-f1-foreground no-underline transition-colors hover:bg-f1-background-secondary">
                 ...
-              </div>
+              </li>
             </Dropdown>
             <BreadcrumbSeparator>
               <ChevronRight className="h-4 w-4 text-f1-icon-secondary" />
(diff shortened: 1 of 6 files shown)


lib/experimental/Navigation/Header/Breadcrumbs/index.tsx
```diff
@@ -59,14 +68,82 @@ interface BreadcrumbsProps {
...
+
+    const updateVisibleBreadcrumb = () => {
+      if (!containerRef.current || breadcrumbs.length <= 2) {
+        setVisibleCount(breadcrumbs.length)
+        return
+      }
+
+      const containerWidth = containerRef.current.offsetWidth
+      const itemWidth = 150
+      const dropdownWidth = 50
```
sauldom102, line 86: Do we use these "magic" numbers in another place? Maybe we could have them declared outside of the `Breadcrumbs` component as a global const variable

sauldom102 (approved):
Great work so far! 💪🏼

PR #679, opened 2024-10-10
TabBar: allow onClick handler

nlopin:
## 🚪 Why?

We want to track tab clicks. For example, we have tracking when user goes to the notification tab on the Inbox screen


RamProg (approved):
LGTM

PR #680, opened 2024-10-10
Add Sidebar User Menu

dani-moreno:
## 🔑 What?

Add the component for the user menu, that will stay at the bottom of the sidebar.

image

## 🏡 Context

- [🎨 Figma](https://www.figma.com/design/uImdC0GKxH59DkjzBqRJEF/Layouts-%26-Blocks?node-id=572-138330&t=6a75e6tnBiCCYazo-4)

### DIFF
--- a/lib/experimental/Navigation/Sidebar/User/index.stories.tsx
+++ b/lib/experimental/Navigation/Sidebar/User/index.stories.tsx
@@ -1,0 +1,23 @@
+import type { Meta, StoryObj } from "@storybook/react"
+import { User } from "."
+
+const meta = {
+  component: User,
+  tags: ["autodocs"],
+} satisfies Meta<typeof User>
+
+export default meta
+type Story = StoryObj<typeof meta>
+
+export const Default: Story = {
+  args: {
+    name: "Dani Moreno",
+    avatarUrl: "https://github.com/dani-moreno.png",
+    avatarAlt: "DM",
+    options: [
+      { label: "Preferences", href: "/preferences" },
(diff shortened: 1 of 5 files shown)


lib/experimental/Navigation/Sidebar/User/index.stories.tsx
```diff
@@ -0,0 +1,23 @@
...
+} satisfies Meta<typeof User>
+
+export default meta
+type Story = StoryObj<typeof meta>
+
+export const Default: Story = {
+  args: {
+    name: "Dani Moreno",
+    avatarUrl: "https://github.com/dani-moreno.png",
+    avatarAlt: "DM",
```
sauldom102, line 16: I'm not sure if we should be passing this `alt` ourselves, it seems a bit redundant, we could calculate it inside the `Avatar` component based on the full name, so we would only need to pass the full name

dani-moreno: Yep, I agree. The `Avatar` doesn't have a name prop right now, but we could change it to require a name instead of the `alt`. We could then use that name as a title or something similar.

I can do that, but I'd prefer to handle it in a different PR. This way, I can update how we're using the Avatar everywhere without mixing it with changes to this menu. Wdyt?

sauldom102: Yeah sure! It could be (and probably should be) on another PR

sauldom102 (approved):

PR #681, opened 2024-10-10
Fix Select check icon

dani-moreno:
The check icon for the selected option could overlap the label if the label is the longer one.

Screenshot 2024-10-10 at 18 57 23

---

This PR fixes that by using a grid to hold the icon instead of an absolute position:

Screenshot 2024-10-10 at 19 05 08

### DIFF
--- a/lib/ui/select.tsx
+++ b/lib/ui/select.tsx
@@ -108,14 +108,14 @@ const SelectItem = React.forwardRef<
   <SelectPrimitive.Item
     ref={ref}
     className={cn(
-      "relative grid w-full cursor-default select-none grid-cols-[1fr_20px] items-center gap-x-1.5 rounded p-2 outline-none transition-colors focus:bg-f1-background-secondary focus:text-f1-foreground data-[disabled]:pointer-events-none data-[disabled]:opacity-50",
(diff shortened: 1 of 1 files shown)


lib/ui/select.tsx
```diff
@@ -108,19 +108,16 @@ const SelectItem = React.forwardRef<
...
-    <div className="absolute right-2 top-2 flex items-center justify-center">
-      <SelectPrimitive.ItemIndicator>
-        <CheckCircle className="h-5 w-5 text-f1-icon-selected" />
-      </SelectPrimitive.ItemIndicator>
-    </div>
-
     <SelectPrimitive.ItemText>{children}</SelectPrimitive.ItemText>
+    <SelectPrimitive.ItemIndicator className="flex">
+      <CheckCircle className="size-5 text-f1-icon-selected" />
+    </SelectPrimitive.ItemIndicator>
```
sauldom102, line 120: I think using a `grid` here is a bit too much, we could use a simpler option like flex

dani-moreno: To achieve the same layout with flex, you'd need to add classes to the child elements, right? I used grid to avoid that, but if it's overkill I can change it, no probs! 😀

josepjaume: I think both are fine and performance-wise it's not that one is more heavy than the other (trust me, I've asked ChatGPT, I'm an engineer)

nlopin (approved):

PR #682, opened 2024-10-11
Make dashboard widget size configurable

Rogermax:
## 🚪 Why?

To avoid using grid of two columns

## 🔑 What?

Add a configurable param to Dasboard layout

https://github.com/user-attachments/assets/f6e95eb5-806c-4b57-97b3-6a2bfc97de7b

### DIFF
--- a/lib/experimental/Widgets/Layout/Dashboard/index.tsx
+++ b/lib/experimental/Widgets/Layout/Dashboard/index.tsx
@@ -5,6 +5,7 @@ import {
   forwardRef,
   ReactNode,
   useEffect,
+  useMemo,
   useRef,
   useState,
 } from "react"
@@ -12,11 +13,23 @@ import { Masonry } from "react-masonry"
 import { Widget } from "../../Widget"
 
 type DashboardProps = {
+  widgetWidth?: "sm" | "md" | "lg"
   children?: ReactNode[]
 }
 
 const _Dashboard = forwardRef<HTMLDivElement, DashboardProps>(
-  function Dashboard({ children }, ref) {
+  function Dashboard({ children, widgetWidth = "sm" }, ref) {
+    const maxWidgetWidth = useMemo(() => {
+      switch (widgetWidth) {
+        case "sm":
+          return 340
+        case "md":
+          return 480
+        case "lg":
+          return 640
+      }
+    }, [widgetWidth])
+
     const [columns, setColumns] = useState<number | undefined>()
 
     const arrayChildren = Children.toArray(children)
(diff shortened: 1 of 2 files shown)


nlopin:
Hi! Could you narrow the scope of this PR? It is hard to understand which part of it is related to the widget size

Rogermax:
> Hi! Could you narrow the scope of this PR? It is hard to understand which part of it is related to the widget size

Upss.. srry I bring a lot of changes, let me clean it

josepjaume:
I can live with this. @dani-moreno what do you think? Do we need to sync this up with Figma?

Rogermax:
@nlopin @dani-moreno can we merge this as designers has agreed with that?

josepjaume (approved):

lib/experimental/Widgets/Layout/Dashboard/index.tsx
```diff
@@ -5,18 +5,31 @@ import {
...
+    const maxWidgetWidth = useMemo(() => {
+      switch (widgetWidth) {
+        case "sm":
+          return 340
+        case "md":
+          return 480
+        case "lg":
+          return 640
+      }
+    }, [widgetWidth])
```
josepjaume, line 31: Sorry, was just reviewing this and I approved without thinking. Why do we need a `useMemo` for this? It's not an expensive computation.

Rogermax: Because is used in the useEffect below, without useMemo, the useEffect will execute in every render

lib/experimental/Widgets/Layout/Dashboard/index.tsx
```diff
@@ -5,18 +5,31 @@ import {
...
-  function Dashboard({ children }, ref) {
+  function Dashboard({ children, widgetWidth = "sm" }, ref) {
+    const maxWidgetWidth = useMemo(() => {
+      switch (widgetWidth) {
+        case "sm":
+          return 340
+        case "md":
+          return 480
+        case "lg":
+          return 640
```
josepjaume, line 29: Another way to implement this so it's exhaustively checked is to define something like:

```tsx
const WidgetWidth = "sm" | "md" | "lg"

const widths: Record<WidgetWidth, number> = {
  "sm": 340,
  "md": 480,
  "lg": 640
}
```

lib/experimental/Widgets/Layout/Dashboard/index.tsx
josepjaume: That's not the case! The hook will be checked by value, and an integer is a stable value.

Rogermax: mmm... true!

josepjaume: True is also a stable value 😆

Rogermax: let me do a quick PR to make these changes

PR #683, opened 2024-10-11
Create PrivacyModeProvider and components to handle it

sauldom102:
## 🔑 What?

Create PrivacyModeProvider and components to handle it.

## 🚪 Why?

We need a cleaner way to define this privacy mode without having to pass props all over our components.

## 🏡 Context

- [💼 Jira](https://factorialmakers.atlassian.net/browse/FCT-19755?atlOrigin=eyJpIjoiYTJhYThmMjllMjdjNGNjOTkyNmUxMWJlMzBhMDRiYjUiLCJwIjoiaiJ9)

### DIFF
--- a/lib/experimental/Utilities/PrivateBox/index.stories.tsx
+++ b/lib/experimental/Utilities/PrivateBox/index.stories.tsx
@@ -1,0 +1,43 @@
+// Replace your-framework with the name of your framework
+import { Button } from "@/../lib/components/Actions/Button"
+import { PrivacyModeProvider, usePrivacyMode } from "@/lib/privacyMode"
+import type { Meta, StoryObj } from "@storybook/react"
+import { PrivateBox } from "."
+
+const meta: Meta = {
+  decorators: [
+    (Story) => (
+      <PrivacyModeProvider>
+        <Story />
+      </PrivacyModeProvider>
+    ),
+  ],
+}
+
+type Story = StoryObj
+
+export const PrivateBoxExample: Story = {
+  render: () => {
+    const { toggle, enabled } = usePrivacyMode()
+
+    return (
+      <div className="flex flex-col gap-4">
+        <p>
+          My salary is{" "}
+          <PrivateBox>
+            <span>60,000 €/year</span>
+          </PrivateBox>
+        </p>
+        <div className="self-start">
+          <Button
+            label={`${enabled ? "Disable " : "Enable "} Privacy Mode`}
+            onClick={toggle}
+            variant="neutral"
+          />
+        </div>
+      </div>
+    )
+  },
+}
+
+export default meta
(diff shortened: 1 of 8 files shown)


lib/experimental/Widgets/Widget/index.tsx
```diff
@@ -60,91 +61,93 @@ const Container = forwardRef<
...
-            <div className="flex flex-row items-center justify-between gap-2">
-              <div className="flex min-h-6 grow flex-row items-center gap-1.5 truncate">
-                {header.title && <CardTitle>{header.title}</CardTitle>}
-                {header.subtitle && (
-                  <>
-                    <InlineDot />
-                    <CardSubtitle>{header.subtitle}</CardSubtitle>
-                  </>
+    <PrivacyModeProvider>
+      <Card className="flex gap-4" ref={ref}>
```
sauldom102, line 65: Notice how we've added this `PrivacyModelProvider` to our widget, so that we can always have this option to control it at widget level without having to use the provider at all

josepjaume: I think we should just put this provider alongside the rest of the providers in the `FactorialOne` provider umbrella. This way when you set privacy = off in one of the widgets, it will happen on the rest of the app accordingly :).

I think this might be a bit weird at the moment since we have the "eye" button inside the widget, but this should be moved eventually to the layout itself /cc @dani-moreno .

josepjaume: BTW this provider might benefit from a prop that sets an initial state like `initialPrivacy`, so we can then decide if we store it on `localStorage`, user settings, etc (because the app can inject that value when instantiating `<FactorialOne />`.

lib/experimental/Utilities/PrivateBox/index.stories.tsx
```diff
@@ -0,0 +1,43 @@
+// Replace your-framework with the name of your framework
+import { Button } from "@/../lib/components/Actions/Button"
+import { PrivacyModeProvider, usePrivacyMode } from "@/lib/privacyMode"
+import type { Meta, StoryObj } from "@storybook/react"
+import { PrivateBox } from "."
+
+const meta: Meta = {
+  decorators: [
+    (Story) => (
+      <PrivacyModeProvider>
```
josepjaume, line 10: We are already adding the `FactorialOneProvider` in all stories, so this is redundant now :)

lib/lib/privacyMode.tsx
```diff
@@ -0,0 +1,45 @@
...
+
+  return (
+    <PrivacyModeContext.Provider value={{ enable, disable, toggle, enabled }}>
+      {children}
+    </PrivacyModeContext.Provider>
+  )
+}
+
+export const usePrivacyMode = () => {
+  return useContext(PrivacyModeContext)
```
josepjaume, line 44: The problem with this is that the return value of this can be `undefined`. A trick people generally use with this is to check for existance and raise otherwise - this way you get proper typing:

```suggestion
  let context = useContext(PrivacyModeContext)
  if(context == undefined) throw "usePrivacyMode requires wrapping the component in a PrivacyModeProvider`
  return context
```

josepjaume (changes requested):

lib/lib/privacyMode.tsx
sauldom102: Oh I see, makes sense

PR #685, opened 2024-10-13
Add Tooltip

dani-moreno:
## 🚪 Why?

Tooltip is one of our web components. We use it in our navigation components.

## 🔑 What?

- Add `Tooltip`, which accepts a label, description, and a shortcut.
- The `children` are used to wrap the content that will display the tooltip.

image

- Also add [a new variant](https://www.figma.com/design/pZzg1KTe9lpKTSGPUZa8OJ/Web-Components?node-id=172-2793) for `Shortcut` (and the concept of variants in it), that is used in `Tooltip`.

image

## 🏡 Context

- [🎨 Figma](https://www.figma.com/design/pZzg1KTe9lpKTSGPUZa8OJ/Web-Components?node-id=121-2188)


lib/experimental/Overlays/Tooltip/index.tsx
```diff
@@ -0,0 +1,40 @@
...
+  TooltipProvider,
+  TooltipTrigger,
+} from "@/ui/tooltip"
+import React from "react"
+
+interface TooltipProps {
+  label: string
+  description?: string
+  children: React.ReactNode
+  shortcut?: string[]
```
josepjaume, line 15: This feels a bit out of place, but at the same time I understand why we have it. 

Additionally, I'm worried that we need to put a bit more work into `shortcuts`, as they might be platform-dependant (windows vs mac) and might need better types.

We can leave this be as it's on `experimental` but I think we need to think deeper about it.

josepjaume: Also, we should probably type this better. At the minimum least, I would delegate to whatever the `Shortcut` component accepts via `shortcut?: ComponentProps<typeof Shortcut>["keys"]`

tailwind.config.ts
```diff
@@ -193,7 +193,7 @@ export default {
               hover: "hsl(var(--neutral-20))",
             },
             tertiary: "hsl(var(--neutral-5))",
-            bold: "hsl(var(--neutral-100))",
+            bold: "hsl(var(--neutral-90))",
```
nlopin, line 196: isn't it used elsewhere?

dani-moreno:
> how do we position the Tooltip? is it handled by some library?

It is always positioned on top, unless there's not enough space. It's basically a wrapper of the [Radix Tooltip](https://www.radix-ui.com/primitives/docs/components/tooltip).

PR #689, opened 2024-10-14
chore: create copyable label

RamProg:
## 🚪 Why?

UI updates in Profile

## 🔑 What?

Create new component that changes bg and changes cursor when hovered and copies the text when clicked

## 🏡 Context

https://www.figma.com/design/uImdC0GKxH59DkjzBqRJEF/Layouts-%26-Blocks?node-id=1469-190164&node-type=frame&t=QF6wrXCHdvPmANq9-0

https://github.com/user-attachments/assets/cc0cd3fd-9f57-4327-8776-357170cb4aa6

### DIFF
--- a/lib/experimental/Widgets/Content/CopyLabel/index.tsx
+++ b/lib/experimental/Widgets/Content/CopyLabel/index.tsx
@@ -31,7 +31,7 @@ export const CopyLabel = forwardRef<HTMLDivElement, CopyLabelType>(
       <div
         ref={ref}
         className={cn(
-          "relative flex h-8 list-none items-center justify-between rounded-md p-1.5 font-medium text-f1-foreground-secondary transition-colors duration-300 hover:cursor-pointer",
+          "relative flex list-none items-center justify-between rounded-md bg-f1-background-secondary p-1.5 font-medium text-f1-foreground-secondary transition-colors duration-300 hover:cursor-pointer",
           copied
             ? "hover:bg-f1-background-positive"
             : "hover:bg-f1-background-secondary"
--- a/lib/experimental/Widgets/Content/CopyLabel/index.tsx
+++ b/lib/experimental/Widgets/Content/CopyLabel/index.tsx
@@ -31,7 +31,7 @@ export const CopyLabel = forwardRef<HTMLDivElement, CopyLabelType>(
       <div
         ref={ref}
         className={cn(
-          "relative flex list-none items-center justify-between rounded-md p-1.5 font-medium text-f1-foreground-secondary transition-colors duration-300 hover:cursor-pointer",
+          "group relative flex h-8 items-center justify-between rounded-md p-1.5 font-medium text-f1-foreground-secondary transition-colors duration-300 hover:cursor-pointer",
           copied
             ? "hover:bg-f1-background-positive"
             : "hover:bg-f1-background-secondary"
@@ -40,21 +40,23 @@ export const CopyLabel = forwardRef<HTMLDivElement, CopyLabelType>(
         role="button"
         aria-label={`Copy ${text} to clipboard`}
       >
-        <p className="truncate pr-6 text-f1-foreground">{text}</p>
+        <p className="flex-1 truncate pr-8 text-f1-foreground" title={text}>
+          {text}
+        </p>
         <Icon
           icon={LayersFront}
           size="md"
           className={cn(
-            "absolute right-1.5 transition-all duration-300",
-            copied ? "scale-75 opacity-0" : "scale-100 opacity-100"
+            "absolute right-1.5 opacity-0 transition-all duration-300",
+            !copied && "group-hover:opacity-100"
           )}
         />
         <Icon
           icon={Check}
           size="md"
           className={cn(
-            "absolute right-1.5 transition-all duration-300",
-            copied ? "scale-100 opacity-100" : "scale-75 opacity-0"
+            "absolute right-1.5 opacity-0 transition-all duration-300",
+            copied && "group-hover:opacity-100"
           )}
         />
       </div>
--- a/lib/experimental/Widgets/Content/CopyLabel/index.tsx
+++ b/lib/experimental/Widgets/Content/CopyLabel/index.tsx
@@ -1,0 +1,31 @@
+import { Icon } from "@/components/Utilities/Icon"
+import { LayersFront } from "@/icons"
+import { forwardRef } from "react"
+interface CopyLabelType {
+  text: string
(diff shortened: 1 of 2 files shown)


lib/experimental/Widgets/Content/CopyLabel/index.tsx
nlopin, line 1: could it be a feature of the current label component? (do we have one?)

nlopin: It could be a `CopyContainer` that adds a copy icon to all the children inside. I'm not quite sure about the reusability of this component in other modules. Do you see it used elsewhere?

RamProg: Before creating this for the Details Sidebar, we were using just a regular <p>

I haven't seen this used anywhere else, but I believe it's brand new, Mario showed it last week. I do think it's something that could be reused, seems like a basic enough thing to have other usecases.

nlopin (changes requested):
We are drastically lacking the feedback when the button is clicked. Please, align with Mario about the approach there. At the moment it is not clear whether the text was copied to the buffer

lib/experimental/Widgets/Content/CopyLabel/index.tsx
```diff
@@ -0,0 +1,33 @@
+import { Icon } from "@/components/Utilities/Icon"
+import { LayersFront } from "@/icons"
+import { forwardRef } from "react"
+interface CopyLabelType {
+  text: string
```
josepjaume, line 5: We could accept arbitrary elements as `children` but I'm afraid of abuse. I think we can limit this to a `label` property or `text` as you have and maybe later open it to other use cases.

lib/experimental/Widgets/Content/CopyLabel/index.tsx
```diff
@@ -0,0 +1,33 @@
...
+}
+
+export const CopyLabel = forwardRef<HTMLDivElement, CopyLabelType>(
+  ({ text }, ref) => {
+    const copyHandler = async () => {
+      try {
+        await navigator.clipboard.writeText(text)
```
josepjaume, line 12: At the minimum least, this deserves a [Toast](https://ui.shadcn.com/docs/components/toast) or a [Sonner](https://ui.shadcn.com/docs/components/sonner). We don't have that element yet, but we do need some kind of feedback. I would consider doing both at the same time.

But then... If we have "toasts" we need to think of other use cases like this one.

lib/experimental/Widgets/Content/CopyLabel/index.tsx
```diff
@@ -0,0 +1,33 @@
...
+      } catch (error) {
+        void error
+      }
+    }
+
+    return (
+      <div
+        ref={ref}
+        className="flex h-8 list-none items-center justify-between rounded-md p-1.5 font-medium text-f1-foreground-secondary hover:cursor-pointer hover:bg-f1-background-secondary"
```
josepjaume, line 21: I wouldn't set a fixed `height` here. We shuould be able to reach this otherwise. If the icon is taking up too much space, then let's remove `p-1.5` and apply a smaller one?

josepjaume (changes requested):
I would consider polishing this a bit more. We're doing something a bit more profound than a simple component - we're adding "clipboard" functionality to the platform and we need a visual language for this.

lib/experimental/Widgets/Content/CopyLabel/index.tsx
nlopin: we can also show a hint or change the icon to "copied!". we need a design decision here

RamProg:
I asked the designer about it, will finish this when we have a decision.

RamProg:
@josepjaume @nlopin Added feedback, please take a look

nlopin:
@dani-moreno we need your animation touch to make it super polished

lib/experimental/Widgets/Content/CopyLabel/index.tsx
```diff
@@ -0,0 +1,67 @@
...
+  text: string
+}
+
+export const CopyLabel = forwardRef<HTMLDivElement, CopyLabelType>(
+  ({ text }, ref) => {
+    const [copied, setCopied] = useState(false)
+
+    useEffect(() => {
+      if (copied) {
+        const timer = setTimeout(() => setCopied(false), 1500)
```
nlopin, line 15: I'd extracted the delay to a constant, like `const COPIED_SHOWN_MS = 1500`

Have you tried 1s or even 0.75s? 1.5s looks a bit long for my taste. If user missed the notification, one more click is cheap to make

nlopin (approved):

lib/experimental/Widgets/Content/CopyLabel/index.tsx
RamProg: A matter of taste I guess, I tried 1000 first, but I liked it more at 1500. I'll extract the delay.

josepjaume (approved):
Great job! Thanks for taking in all the feedback ❤️

lib/experimental/Widgets/Content/CopyLabel/index.tsx
```diff
@@ -0,0 +1,69 @@
...
+        ref={ref}
+        className={cn(
+          "group relative flex h-8 items-center justify-between rounded-md p-1.5 font-medium text-f1-foreground-secondary transition-colors duration-300 hover:cursor-pointer",
+          copied
+            ? "hover:bg-f1-background-positive"
+            : "hover:bg-f1-background-secondary"
+        )}
+        onClick={copyHandler}
+        role="button"
+        aria-label={`Copy ${text} to clipboard`}
```
miriam-mr90, line 43: Hey @RamProg !! Just curious 👀 Shouldn't the `aria-label` text come from translations?

nlopin: great catch Miriam! We are exploring a unified approach to translations in the design system. Until it is done, we use English to keep things moving

nlopin:
After a sync with Mario, I found out that the copy interaction must be a part of the `DataList.Item` component, together with navigate to another page. This component will be done in a separate PR.

`CopyLabel` is discarded

eliseo-juan:
> After a sync with Mario, I found out that the copy interaction must be a part of the `DataList.Item` component, together with navigate to another page. This component will be done in a separate PR.
> 
> `CopyLabel` is discarded

sounds fair to me

PR #690, opened 2024-10-14
Add Page component

dani-moreno:
## 🔑 What?

Add `Page` component, which serves as a container for the right side of the layout, handling the content and navigation outside of the sidebar.

It's a simple component for now, accepting a `header` prop (where Header and Tabs will sit) and then the rest of the content.

image

With these changes, the layout looks as in the design:

image

## 🏡 Context

- [🎨 Figma](https://www.figma.com/design/uImdC0GKxH59DkjzBqRJEF/Layouts-%26-Blocks?node-id=4-4687&t=RXKvTrwCEmrMbidR-4)


josepjaume (changes requested):
Great work! Please check out my only comment.

nlopin:
how does it behave on smaller screens?

lib/experimental/Navigation/Page/index.tsx
```diff
@@ -0,0 +1,15 @@
+interface PageProps {
+  children?: React.ReactNode
+  header?: React.ReactNode
+}
+
+export function Page({ children, header }: PageProps) {
+  return (
+    <div className="flex w-full flex-col overflow-hidden rounded-xl bg-f1-background shadow">
+      {header && <div className="flex flex-col">{header}</div>}
+      <div className="flex flex-1 flex-col gap-4 overflow-y-auto px-6 py-5">
```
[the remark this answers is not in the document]
dani-moreno: Aw, yes, the layouts are not defined yet but we're pretty sure we won't always need that padding.

Love the idea of `PageLayouts`! I made a couple of them just as examples that we can iterate later, what do you think?

lib/experimental/Navigation/ApplicationFrame/index.tsx
```diff
@@ -10,7 +10,7 @@ export const ApplicationFrame: React.FC<ApplicationFrameProps> = ({
   return (
     <div className="flex h-full flex-row gap-3">
       <div className="w-64 pl-3">{sidebar}</div>
-      <div className="flex flex-1">{children}</div>
+      <div className="flex flex-1 py-1 pr-1">{children}</div>
```
josepjaume, line 13: Weird that we need that padding, no?

lib/experimental/Navigation/Page/Layouts/index.tsx
```diff
@@ -0,0 +1,21 @@
...
+
+export function Layout({ layout, children }: LayoutProps) {
+  switch (layout) {
+    case "fullWidth":
+      return <FullWidthLayout>{children}</FullWidthLayout>
+    case "standard":
+    default:
+      return <StandardLayout>{children}</StandardLayout>
+  }
+}
```
josepjaume, line 21: At this point, why not just allowing people to use `<FullWidthLayout>` directly under the `Page` children? I understand this way you force *something* to be there, but at the same time you limit props. I can imagine a near future where we add an `InboxLayout` where we want to pass a prop for `list` and another one for `detail`. This API design would make it quite difficult.

josepjaume (commented):
Great job - added a couple of comments. Nearly there!

lib/experimental/Navigation/ApplicationFrame/index.tsx
dani-moreno: The content has this small padding, but the sidebar doesn't. It shouldn't depend on a breakpoint anyway, I can do it in another PR.

lib/experimental/Navigation/Page/index.stories.tsx
```diff
@@ -0,0 +1,75 @@
...
+    ),
+    layout: "fullWidth",
+    children: (
+      <>
+        {Array(25)
+          .fill(0)
+          .map((_, index) => (
+            <Placeholder key={index} className="min-h-24" />
+          ))}
+      </>
```
josepjaume, line 72: I think you can just ignore the `<>` parts here and return the array

lib/experimental/Navigation/Page/index.stories.tsx
```diff
@@ -0,0 +1,75 @@
...
+    header: (
+      <>
+        <Header
+          {...HeaderStories.FirstLevel.args}
+          menu={{ show: false, onClick: () => {} }}
+        />
+        <Tabs {...TabsStories.Primary.args} />
+      </>
+    ),
+    layout: "fullWidth",
```
josepjaume, line 64: This is no longer used right?

josepjaume (approved):
Nice job!

lib/experimental/Navigation/Page/index.stories.tsx
dani-moreno: Oops! You're right, that entire story is no longer used!

PR #691, opened 2024-10-14
Add Navigation in Header

dani-moreno:
## 🚪 Why?

The `Header` component can contain navigation between items. For example, when navigating employees, tasks, and so on.

## 🔑 What?

- Add `Navigation` component inside Header.
- Uses `Navigation` inside the `Header` component.

image

## 🏡 Context

- [🎨 Figma](https://www.figma.com/design/pZzg1KTe9lpKTSGPUZa8OJ/Web-Components?node-id=38-709&t=4HWatFmVqrVyRgCP-4)

### DIFF
--- a/lib/experimental/Navigation/Header/Header/index.stories.tsx
+++ b/lib/experimental/Navigation/Header/Header/index.stories.tsx
@@ -1,5 +1,6 @@
 import EllipsisHorizontal from "@/icons/EllipsisHorizontal"
 import Settings from "@/icons/Settings"
+import { action } from "@storybook/addon-actions"
 
 import type { Meta, StoryObj } from "@storybook/react"
 import Header from "."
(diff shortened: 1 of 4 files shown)


josepjaume (commented):
Great work (it looks gorgeous!) but honestly, I would think twice before adding this on the platform. I'm ok with a dropdown, but any implementation of this will be super buggy, as we don't know the context from which this was called, so we'll have different ways of filtering things and the user will suffer.

@somebody32 this is what we were talking about the other day.

somebody32:
I think it is a valid case only if you can see the detail and the list on the same page (ie master-detail pattern). In that case, it is obvious and follows the most important UX pattern of our lives: understand what is going to happen before you click.

But if that is thought as a control for a "hidden" or assumed list, then it will break that rule as soon as you can't see and understand the entity you're navigating

josepjaume:
Strongly agree with what @somebody32 says here!

nlopin:
Hey! What's the next step here? are we parking this idea for the time being therefore closing the PR?

nlopin:
Discussed and closed. There is no up and down relations between resources at the moment. We will proceed with a dropdown with a simple filter as an alternative solution

PR #692, opened 2024-10-14
Fix buggy `Menu` animation

dani-moreno:
The open/close animation of a category in the `Menu` component is a bit bugged. This PR solves the issue, separating the style of the list, from the div that has the motion applied. Somehow mixing them in the same div does weird things if you're animating the height.

### Before

_Look at the small "jump" the close animation does near the end._

https://github.com/user-attachments/assets/9205ad65-a94c-421f-a734-796e4fe2ff74

---

### After

https://github.com/user-attachments/assets/c75ebdfe-98b7-4170-b710-042029df245a


josepjaume (approved):
Awesome! hahaha

PR #698, opened 2024-10-15
Add a working example of the layout

dani-moreno:
This PR adds a working example of the layout, with the show/hide sidebar functionality and the rest of the layout adapting to it.

https://github.com/user-attachments/assets/67d12af1-6c46-4c59-ba5e-1145bbd0e3e5

---

**Note:** The layout isn't entirely finished as it lacks a few things and has a few bugs, but it's enough to try out. I'll keep working on it when I'm back! 👀🪛

**Things to be improved:**
- I'm using the `useMediaQuery` hook in some places, but I'd like to investigate if I can avoid it in some places (probably yes!).
- A few details yet to be done.

### DIFF
--- a/lib/experimental/Navigation/ApplicationFrame/FrameProvider.tsx
+++ b/lib/experimental/Navigation/ApplicationFrame/FrameProvider.tsx
@@ -1,0 +1,87 @@
+import React, {
+  createContext,
+  PointerEvent,
+  useCallback,
+  useContext,
+  useState,
+} from "react"
+import { useMediaQuery } from "usehooks-ts"
+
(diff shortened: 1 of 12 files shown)


eliseo-juan:
Is it me, or the bottom padding for each category when collapsed is to large?
image

josepjaume:
Amazing work! Since you'll be off for a week or two, I'll add a few commits with some improvement I think could be made before merging :)

josepjaume:
@nlopin you're right. Fortunately enough Framer Motion does this by itself and I believe the `tailwind-animations` module does as well. Let me fix this for the other cases.

josepjaume (approved):

PR #706, opened 2024-10-18
improve code on dashboard component

Rogermax:
## 🚪 Why?

To implement some @josepjaume suggestions from other PR

## 🔑 What?

Refactor some code, remove useMemo (not necessary because is a primitive type and will not re-execute useEffect in every render)


josepjaume (approved):
Thanks so much! These things matter :)

PR #720, opened 2024-10-22
C

nlopin:
## 🚪 Why?

[why]: # (✍️ _What needs is this Pull Request solving or how is it improving the product?_)

## 🔑 What?

[what]: # (✍️ _What changes is this Pull Request introducing in the code?_) 

## 🏡 Context

- [💼 Jira](_jira_link_)
- [📖 Notion](_notion_link_)
- [🎨 Figma](_figma_link_)
- [💻 Other PR](_github_link_)


nlopin:
One thing I want to deliver before merge is:

- [ ] extract presentational view of an item to a component

lib/experimental/Lists/DataList/ItemContainer.tsx
```diff
@@ -0,0 +1,90 @@
...
+      return (
+        <a
+          {...rest}
+          className={cn(
+            rest.className,
+            "text-inherit group flex items-center gap-1.5 rounded p-1.5 text-f1-foreground no-underline hover:bg-f1-background-hover focus-visible:outline focus-visible:outline-2 focus-visible:outline-f1-border-selected-bold"
+          )}
+        >
+          {children}
+        </a>
```
josepjaume, line 85: This should actually use our `Link` component from `lib/linkHandler` if we want it to work with React Router. Shouldn't be a big issue as its API is 1:1 with `a`.

lib/experimental/Lists/DataList/ItemWithCopyButton.tsx
```diff
@@ -0,0 +1,78 @@
...
+
+  return (
+    <ItemContainer
+      as="button"
+      text={text}
+      leftIcon={IconComponent}
+      onClick={copyHandler}
+      className={
+        copied
+          ? "transition-colors duration-300 hover:bg-f1-background-positive focus-visible:bg-f1-background-positive"
```
josepjaume, line 44: Nice

lib/experimental/Lists/DataList/LinkItem.tsx
```diff
@@ -0,0 +1,40 @@
...
+      leftIcon={icon}
+      href={href}
+      onClick={(e) => {
+        if (e.ctrlKey || e.metaKey || e.button === 1) {
+          // Let the browser open the link in a new tab, no custom navigation needed
+          return
+        }
+        e.preventDefault()
+        onClick(e)
+      }}
```
josepjaume, line 27: This shouldn't be necessary when using `Link` from `linkHandler` as the react router's `Link` component already handles this logic.

josepjaume (changes requested):
Good job! Left you a few comments related to link handling.

lib/components/Utilities/Icon/index.tsx
```diff
@@ -7,7 +7,7 @@ import {
   SVGProps,
 } from "react"
 
-const iconVariants = cva("inline-block shrink-0", {
+const iconVariants = cva("inline-block shrink-0 text-f1-icon", {
```
rubenmoya, line 10: Sorry to jump in, since I don't have much context, but, wouldn't it be better to let the icon "inherit" the text color from the container where it is used by default?

That will make it easier for it to adapt to different states like hover/focus without having to pass down any more classes to this component

lib/experimental/Lists/DataList/index.tsx
```diff
@@ -0,0 +1,58 @@
...
+
+const _DataList = forwardRef<HTMLUListElement, DataListProps>(
+  ({ children, label }, ref) => {
+    return (
+      <ul
+        className="flex min-w-32 max-w-72 flex-col gap-0.5 px-1.5 py-1"
+        ref={ref}
+      >
+        {label && (
+          <p className="px-1.5 text-f1-foreground-secondary">{label}</p>
```
rubenmoya, line 27: A `p` element is not allowed inside a `ul` element [more info in MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/ul)

image

Probably we'd need to move the `p` outside and add an `aria-labelledby` to indicate what is is labeling.

lib/experimental/Lists/DataList/ItemWithCopyButton.tsx
```diff
@@ -0,0 +1,78 @@
...
+import { ItemProps } from "@/experimental/Lists/DataList"
+import { ItemContainer } from "@/experimental/Lists/DataList/ItemContainer"
+import { CheckCircle, LayersFront } from "@/icons"
+import { cn } from "@/lib/utils"
+
+const COPIED_SHOWN_MS = 750
+
+type CopyButtonProps = ItemProps
+
+export const ItemWithCopyButton = ({
```
rubenmoya, line 13: Just playing devils advocate, could there be a case where we'd want a different item type to also be able to copy its content? 🤔

Make the "copy" action an item itself could be limiting or confusing in the future.

Also, will always the shown text be the one we want to copy to the clipboard?

lib/experimental/Lists/DataList/ItemWithCopyButton.tsx
```diff
@@ -0,0 +1,78 @@
...
+      as="button"
+      text={text}
+      leftIcon={IconComponent}
+      onClick={copyHandler}
+      className={
+        copied
+          ? "transition-colors duration-300 hover:bg-f1-background-positive focus-visible:bg-f1-background-positive"
+          : undefined
+      }
+      aria-label={copied ? "Copied!" : `Copy ${text}`}
```
rubenmoya, line 47: What's the benefit of doing this? The `Copied!` text won't reach the user. Also, it is hardcoded in english 🤔

lib/components/Utilities/Icon/index.tsx
nlopin: It does not prevent you from overriding the default color and adapt the icon to your needs.

The issue with inheriting the text color is that in all figma files we have, the icon color is slightly lighter than the color of the main text. Since this is a subtle difference, it is easy to overlook and miss it. I noticed the color difference in this component, but there are several components ignoring the difference. This default class is a way to tackle the problem. 

What would be your suggestion? Also @josepjaume, maybe you have an opinion here too

lib/experimental/Lists/DataList/index.tsx
nlopin: great catch 👍  on it

lib/experimental/Lists/DataList/ItemWithCopyButton.tsx
nlopin: I initially went the generic way, growing the complexity until I found out we don't have such requirement yet. It prefer to expand the functionality when needed, rather than delivering a flexible solution and support it

lib/components/Utilities/Icon/index.tsx
josepjaume: Yeah, discussed this with @dani-moreno and had the exact same concerns as you @rubenmoya. An option here would be to use `filter: darken(75%)` or comething like that? In tailwind, that'd be `brightness-75`.

nlopin: ok, will adjust the class

lib/experimental/Lists/DataList/LinkItem.tsx
nlopin: do I get it right that I don't even need the `onClick` handler?

josepjaume (approved):
Lookin nice! 🚀

lib/components/Utilities/Icon/index.tsx
josepjaume: Dammit

PR #729, opened 2024-10-23
Make avatar images cover instead of stretching

josepjaume:
Avatars should always look nice, and this means using `object-fit: cover` instead of stretching them :)


eliseo-juan (approved):
Indeed!

PR #731, opened 2024-10-23
Rename page header

josepjaume:
Renames page header from a very generic `Header` component to `PageHeader`.


josepjaume:
Excuse the merge, but this is quite straightforward and also fixes a wrong export (we were exporting default before) and it's blocking me.

PR #739, opened 2024-10-25
Create AlertTag, StatusTag and BalanceTag components

sauldom102:
## 🔑 What?

Create AlertTag, StatusTag and BalanceTag components.

Screenshot 2024-10-25 at 10 33 37


lib/experimental/Information/AlertTag/index.tsx
```diff
@@ -0,0 +1,55 @@
...
+import { Icon } from "@/factorial-one"
+import { AlertCircle, InfoCircle, Warning } from "@/icons"
+import { cn } from "@/lib/utils"
+import { forwardRef } from "react"
+
+type Variant = "info" | "warning" | "critical"
+
+interface Props {
+  text: string
+  variant: Variant
```
nlopin, line 11: What about having more semantic name, like `level` or `importance`?

lib/experimental/Information/BalanceTag/index.stories.tsx
nlopin, line 1: hm, I thought we will provide the calculation logic ourselves here with an API like `<BalanceDelta previous={100} current={200} />` and it will do all the decision about positive and negative. We can use this implementation as an internal, building the logic on top so devs can always go lower level

nlopin:
what about multiline support? what do we allow? I think we can also create a type forbidding empty strings and remove the `throw` behavior

sauldom102:
> what about multiline support? what do we allow? I think we can also create a type forbidding empty strings and remove the `throw` behavior

I don't think we should support multiline on this component. What do you mean about "a type forbidding empty strings"? 🤔

lib/experimental/Information/BalanceTag/index.stories.tsx
nlopin: I think this is partially the point, to provide a unified money format. We will have to get the locale for proper formatting though

nlopin:
> What do you mean about "a type forbidding empty strings"?

something like this https://gist.github.com/wesbos/20babb2f78acf596f4d24322d0e4fbe0

lib/experimental/Information/BalanceTag/index.stories.tsx
sauldom102: Could we do this in a next PR exporting another higher level component called `BalanceDeltaTag`?

sauldom102:
> > What do you mean about "a type forbidding empty strings"?
> 
> something like this https://gist.github.com/wesbos/20babb2f78acf596f4d24322d0e4fbe0
> 
> not sure is it possible, but worth spending half an hour

We can't do this in typescript as this text could be very variable, it needs to be done on runtime, I just created a hook utility to do that!

lib/experimental/Information/BalanceTag/index.stories.tsx
nlopin: could you create an issue for this plz?

lib/experimental/Information/StatusTag/index.tsx
```diff
@@ -0,0 +1,48 @@
...
+          {
+            neutral: "bg-f1-background-secondary text-f1-foreground-secondary",
+            info: "bg-f1-background-info text-f1-foreground-info",
+            positive: "bg-f1-background-positive text-f1-foreground-positive",
+            warning: "bg-f1-background-warning text-f1-foreground-warning",
+            critical: "bg-f1-background-critical text-f1-foreground-critical",
+          }[variant]
+        )}
+      >
+        <div
```
nlopin, line 30: can we use our internal `Icon` component here?

lib/experimental/Information/StatusTag/index.tsx
```diff
@@ -0,0 +1,48 @@
...
+
+export const StatusTag = forwardRef<HTMLDivElement, Props>(
+  ({ text, variant }, ref) => {
+    useTextFormatEnforcer(text, { disallowEmpty: true })
+
+    return (
+      <div
+        ref={ref}
+        className={cn(
+          "flex flex-row items-center justify-start gap-1.5 rounded-md px-2 py-0.5 text-base font-medium",
```
nlopin, line 20: let's add `line-clamp-1`?

nlopin (approved):

lib/experimental/Information/BalanceTag/index.stories.tsx
sauldom102: Sure!

lib/experimental/Information/StatusTag/index.tsx
sauldom102: Yep!

lib/experimental/Information/StatusTag/index.tsx
sauldom102: Not really, it's just a dot, I don't see the point of using a SVG here 🤔

PR #749, opened 2024-10-28
Create dumb WidgetSection component

sauldom102:
## 🔑 What?

Create dumb WidgetSection component.

## 🚪 Why?

I just noticed a pattern of every content separated in our widgets, some of them have some title, but they shouldn't actually be tied to the component itself that is there but it should be more related to the widget itself. We have this title accompanying various components that we have here, and we didn't have a common component to spread this change of removing the uppercase in all of them.

Screenshot 2024-10-28 at 12 35 34

## 👁️ Visual proof

Screenshot 2024-10-28 at 12 38 07

### DIFF
--- a/lib/experimental/Widgets/WidgetSection/index.tsx
+++ b/lib/experimental/Widgets/WidgetSection/index.tsx
@@ -11,9 +11,9 @@ export const WidgetSection = forwardRef<HTMLDivElement, Props>(
 
     return (
       <div ref={ref} className="flex flex-col gap-2">
-        <h4 className="text-base font-medium text-f1-foreground-secondary">
+        <p className="text-base font-medium text-f1-foreground-secondary">
           {title}
-        </h4>
+        </p>
         {children}
       </div>
     )
--- a/lib/experimental/Widgets/WidgetSection/index.tsx
+++ b/lib/experimental/Widgets/WidgetSection/index.tsx
@@ -11,9 +11,9 @@ export const WidgetSection = forwardRef<HTMLDivElement, Props>(
 
     return (
       <div ref={ref} className="flex flex-col gap-2">
-        <p className="text-base font-medium text-f1-foreground-secondary">
+        <h4 className="text-base font-medium text-f1-foreground-secondary">
           {title}
-        </p>
+        </h4>
         {children}
       </div>
     )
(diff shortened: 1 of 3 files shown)


lib/experimental/Widgets/WidgetSection/index.tsx
```diff
@@ -0,0 +1,23 @@
...
+  title?: string
+}
+
+export const WidgetSection = forwardRef<HTMLDivElement, Props>(
+  ({ title, children }, ref) => {
+    useTextFormatEnforcer(title, { disallowEmpty: true })
+
+    return (
+      <div ref={ref} className="flex flex-col gap-2">
+        <p className="text-base font-medium text-f1-foreground-secondary">
```
nlopin, line 14: shouldn't it be `h3` or `h4`?

nlopin:
Don't we have the gaps already set by the widget itself?

sauldom102:
> Don't we have the gaps already set by the widget itself?

Only those gaps that separate each section by a dashed line, but not these

lib/experimental/Widgets/WidgetSection/index.tsx
sauldom102: Yep, it's better semantically, done!

sauldom102: Actually we cannot add a `h4` and some `className` at the same time, it's detected as an accesibility violation

nlopin:
I see that the dashed divider is not used that often. I think we should have the gap you defined by default and the divider as an optional thing via the group. WDYT? 

I recall we discussed this with @josepjaume but haven't written down any decision

sauldom102:
> I see that the dashed divider is not used that often. I think we should have the gap you defined by default and the divider as an optional thing via the group. WDYT?
> 
> I recall we discussed this with @josepjaume but haven't written down any decision

I think the divider is defined in a good way currently, it's declarative and it will show up depending on how we pass the children, if we add two children it will show in the middle with the spacing that it already has, how should we express it?

sauldom102:
> > I think the divider is defined in a good way currently, it's declarative and it will show up depending on how we pass the children, if we add two children it will show in the middle with the spacing that it already has, how should we express it?
> 
> my concern is about which pattern is more common - gaps between chunks of widget or dividers. In my perception the gaps are used more often, so they should be the default and easy way. Dividers are less common, therefore it require more code.
> 
> Currently it is other way around. It is not a blocker for this PR, but something to review in the follow up

Sure, we should review it

nlopin (approved):

PR #751, opened 2024-10-28
Create rest of tag components

sauldom102:
## 🔑 What?

Create rest of tag components.


sauldom102 (commented):
I'm going to polish the Clock-in widget and rest of stuff using these new components and deprecate the other components that we have using the other "Tag" component that was done previously

dani-moreno (approved):

PR #752, opened 2024-10-28
Add skip to content button to the application frame

nlopin:
## 🚪 Why?

To show some love to screen reader and keyboard users

## 🔑 What?

Add a new `a` component at the top of the sidebar with maximal z-index. It gets the focus first. When clicked it moves the focus to the main content area

## 👁️ Demo

https://github.com/user-attachments/assets/456db68b-3afc-4491-b265-fd602e53954b


lib/experimental/Navigation/ApplicationFrame/index.tsx
```diff
@@ -19,57 +19,74 @@ export function ApplicationFrame({ children, sidebar }: ApplicationFrameProps) {
...
 
+const SkipToContentButton = ({ contentId }: { contentId?: string }) => {
+  return (
+    <a
+      href={`#${contentId}`}
+      className={focusRing(
+        "absolute z-50 -translate-y-full translate-x-4 bg-f1-background-secondary-hover p-2 text-lg no-underline transition-transform duration-200 focus-visible:translate-y-0"
+      )}
+    >
+      Skip to content
```
josepjaume, line 30: What about translations? Should we start thinking about it?

lib/experimental/Navigation/ApplicationFrame/index.tsx
```diff
@@ -19,57 +19,74 @@ export function ApplicationFrame({ children, sidebar }: ApplicationFrameProps) {
...
-        </AnimatePresence>
-        <div
-          className={cn(
-            { "transition-all": !shouldReduceMotion },
-            sidebarState === "locked" ? "w-64 pl-3" : "w-0"
-          )}
-        >
-          {sidebar}
+    <>
+      <SkipToContentButton contentId="content" />
```
josepjaume, line 41: The fact that you need to know the `contentId` here and we're shifting the responsibility to the user makes me think this belongs to the `ApplicationFrame` and not the sidebar.

josepjaume:
Nice job! Looks neat! Should we ask for some proper design?

lib/experimental/Navigation/ApplicationFrame/index.tsx
nlopin: yes, that will be the next step and a separate PR

nlopin:
> Nice job! Looks neat! Should we ask for some proper design?

updated according to the design

lib/experimental/Navigation/ApplicationFrame/index.tsx
```diff
@@ -19,57 +19,74 @@ export function ApplicationFrame({ children, sidebar }: ApplicationFrameProps) {
...
+      <MotionConfig
+        reducedMotion={shouldReduceMotion ? "always" : "never"}
+        transition={{
+          ease: [0.25, 0.1, 0.25, 1],
+          duration: shouldReduceMotion ? 0 : 0.2,
```
somebody32, line 46: curios: why passing shouldReduceMotion twice in that case? should global prop kill all the params from the transition?

lib/experimental/Navigation/ApplicationFrame/index.tsx
```diff
@@ -19,57 +19,74 @@ export function ApplicationFrame({ children, sidebar }: ApplicationFrameProps) {
...
+          </div>
+          <motion.main
+            id="content"
+            className={cn(
+              "flex max-w-full flex-1 overflow-x-hidden xs:py-1 xs:pr-1",
+              sidebarState === "locked" ? "pl-0" : "xs:pl-1"
+            )}
+            layout
+            transition={{
+              duration: shouldReduceMotion ? 0 : 0.3,
```
somebody32, line 80: I thought you can tell framer motion one time in general and not to pass that time always, am I wrong? https://www.framer.com/motion/guide-accessibility/#reducedmotion

lib/experimental/Navigation/ApplicationFrame/index.tsx
josepjaume: Okay!

lib/experimental/Navigation/ApplicationFrame/index.tsx
josepjaume: Sir, you are a gentleman and a scholar.

lib/experimental/Navigation/ApplicationFrame/index.tsx
josepjaume: Unfortunately, the moment you explicitly state a transition, the default behavior of framer motion doesn't work. Sadly it only does it for defaults, which admittedly pollutes the code.

josepjaume (approved):

josepjaume:
Let's do this, no?

PR #753, opened 2024-10-28
AvatarList

Rogermax:
## 🔑 What?

New component for making widget list

## 🏡 Context

Screenshot 2024-10-28 at 19 39 50


nlopin:
@dani-moreno we need avatar component to unify all the options

nlopin:
@Rogermax should we put the image at the top left corner? I see it is kinda centered on the screenshot, but I'm not sure it was the intention or not

Rogermax:
> @Rogermax should we put the image at the top left corner? I see it is kinda centered on the screenshot, but I'm not sure it was the intention or not

Better in the middle no? when you have a list of items you want the image on the center no?
Screenshot 2024-10-30 at 12 51 11

lib/experimental/Information/Tags/BalanceTag/index.tsx
```diff
@@ -26,7 +26,7 @@ export const BalanceTag = forwardRef<HTMLDivElement, Props>(
       <BaseTag
         ref={ref}
         className={cn(
-          "pl-1",
+          "line-clamp-1 flex flex-row items-center justify-between gap-0.5 rounded-md py-0.5 pl-1 pr-2 text-base font-medium",
```
sauldom102, line 29: Why are these changes needed?

lib/experimental/Widgets/Content/TasksList/TaskItem/index.tsx
```diff
@@ -38,18 +42,22 @@ const Wrapper: React.FC<WrapperProps> = ({ onClick, className, children }) => {
   )
 }
 
+export function getWidgetListItemClassname(clickable = false) {
+  return cn(
+    "flex flex-row items-center gap-2 rounded-[8px] border border-solid border-transparent px-2 py-[6px]",
+    clickable
+      ? "cursor-pointer hover:bg-f1-background-tertiary focus:border-f1-background-selected-bold focus:outline-none"
+      : undefined
```
sauldom102, line 50: I recently changed some styles here for the same component for the widget that we have in the employee profile https://github.com/factorialco/factorial-one/pull/767 I see that you changed the gap, are we sure about this?

lib/experimental/Information/Tags/BalanceTag/index.tsx
Rogermax: I think is a bad conflict resolution, let me put it as before

lib/experimental/Widgets/Content/TasksList/TaskItem/index.tsx
sauldom102: `py-[6px]` could better be expressed with `py-1.5`

lib/experimental/Information/Tags/BalanceTag/index.tsx
sauldom102: Perfect, thanks!

josepjaume (changes requested):
Thanks! Here are some comments :D

josepjaume:
BTW I would try to find a way to make this handle big chunks of content better. At a first glance, I would either make things wrap, or set a container query breakpoint.

We do use that already, check it out: https://github.com/tailwindlabs/tailwindcss-container-queries

Rogermax:
> BTW I would try to find a way to make this handle big chunks of content better. At a first glance, I would either make things wrap, or set a container query breakpoint.
> 
> We do use that already, check it out: https://github.com/tailwindlabs/tailwindcss-container-queries

Totally agree... 
I think we are running to much without thinking (me the first) 
There are some basic foundamentals not clear (at least from my point of view):
* Widgets height dependes on what? content? or max-height?
* What happens with overflow items? I know we don't want to put scroll.. but then we need a solution to show the full list
  * Adding a button for more elements its ok.. but if we click what? a modal? sidepanel?
* This common behaviour, of listing things and having a more elements button, should be a high order component... but components like TaskList, AvatarList, etc.. have a prop with explicit props (items, tasks).. this don't allow to render childrens (of certain class, like WidgetItems)

I have the hurry to have something to show in our dashboard for the alignment... but maybe we should stop and think a bit

PR #759, opened 2024-10-29
Add StatusTag in the PageHeader for the first level

jesusgonzalezrivera:
## 🔑 What?
Add the `StatusTag` component for the first level `PageHeader`

## 🚪 Why?
In ShiftManagement we have a status related to the shifts in the view. In the approved designs from some weeks ago, we included it in the top bar when we saw the first level view
image
image

## 🏡 Context
- [🎨 Figma](https://www.figma.com/design/MbxR3tIzVFCwvXZ9gmpU0w/%F0%9F%9F%A3-Factorial-One-%C2%B7-Nav?node-id=120-18140&t=v805BtucFRgGnUrq-0)

### DIFF
--- a/lib/experimental/Information/Tags/StatusTag/index.tsx
+++ b/lib/experimental/Information/Tags/StatusTag/index.tsx
@@ -3,7 +3,7 @@ import { cn } from "@/lib/utils"
 import { forwardRef } from "react"
 import { BaseTag } from "../BaseTag"
 
-export type Variant = "neutral" | "info" | "positive" | "warning" | "critical"
+type Variant = "neutral" | "info" | "positive" | "warning" | "critical"
 
 export type StatusVariant = Variant
 
--- a/lib/experimental/Information/Tags/StatusTag/index.tsx
+++ b/lib/experimental/Information/Tags/StatusTag/index.tsx
@@ -3,7 +3,7 @@ import { cn } from "@/lib/utils"
 import { forwardRef } from "react"
 import { BaseTag } from "../BaseTag"
 
-type Variant = "neutral" | "info" | "positive" | "warning" | "critical"
+export type Variant = "neutral" | "info" | "positive" | "warning" | "critical"
 
 export type StatusVariant = Variant
 
--- a/lib/experimental/Navigation/Header/PageHeader/index.tsx
+++ b/lib/experimental/Navigation/Header/PageHeader/index.tsx
@@ -1,5 +1,7 @@
 import { Button } from "@/components/Actions/Button"
 import { IconType } from "@/components/Utilities/Icon"
+import type { Variant } from "@/experimental/Information/Tags/StatusTag"
+import { StatusTag } from "@/experimental/Information/Tags/StatusTag"
 import { useSidebar } from "@/experimental/Navigation/ApplicationFrame/FrameProvider"
 import AlignTextJustify from "@/icons/app/AlignTextJustify"
 import { cn } from "@/lib/utils"
@@ -14,6 +16,10 @@ type HeaderProps = {
     href: string
     icon: IconType
   }
+  statusTag?: {
+    text: string
+    variant: Variant
+  }
   breadcrumbs?: BreadcrumbItemType[]
   actions?: {
     label: string
@@ -24,6 +30,7 @@ type HeaderProps = {
 
 export function PageHeader({
   module,
+  statusTag = undefined,
   breadcrumbs = [],
   actions = [],
 }: HeaderProps) {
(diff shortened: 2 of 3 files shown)


nlopin:
I see the first image contains the tag glued to the module title. The storybook places it near actions. What is the agreed placement?

jesusgonzalezrivera:
> I see the first image contains the tag glued to the module title. The storybook places it near actions. What is the agreed placement?

@nlopin Near to the actions as far as I know. There was a discussion about it some weeks ago in [this thread](https://factorialteam.slack.com/archives/C06QT46115K/p1729858677481679)

lib/experimental/Navigation/Header/PageHeader/index.tsx
```diff
@@ -74,21 +83,31 @@ export function PageHeader({
...
-        </div>
-      )}
+      <div className="flex items-center">
+        {!hasNavigation && hasStatus && (
+          <div className="pe-3">
+            <StatusTag text={statusTag.text} variant={statusTag.variant} />
+          </div>
+        )}
+        {hasStatus && hasActions && (
+          <div className="right-0 h-4 w-px bg-f1-border-secondary"></div>
```
nlopin, line 93: can you take a look if we have an existing divider component?

nlopin (approved):

lib/experimental/Navigation/Header/PageHeader/index.tsx
jesusgonzalezrivera: It existed but @josepjaume told me we have removed it to make use of borders

josepjaume: The problem with the old divider is that it was used once and in the wrong place (that should have been a border), so I removed without knowing you were using it in this PR :)

I honestly think this is fine, as a divider is not much else than a bunch of tokens. I would think about it later if we find more use cases.

lib/experimental/Navigation/Header/PageHeader/index.tsx
```diff
@@ -74,21 +83,31 @@ export function PageHeader({
...
+                hideLabel
+                round
+                key={index}
+                variant="outline"
+                onClick={action.onClick}
+                label={action.label}
+                icon={action.icon}
+              />
+            ))}
+          </div>
```
josepjaume, line 108: We don't have to do it now, but I can't help but think this could be expressed with something like:

```
{
  [
    hasStatus ? <div className ...> : undefined,
    hasActions ? actions.map(() => ...) : undefined,
  ].filter(Boolean).join(<div className={} ... />)
}
```

lib/experimental/Information/Tags/StatusTag/index.tsx
```diff
@@ -3,7 +3,7 @@ import { cn } from "@/lib/utils"
 import { forwardRef } from "react"
 import { BaseTag } from "../BaseTag"
 
-type Variant = "neutral" | "info" | "positive" | "warning" | "critical"
+export type Variant = "neutral" | "info" | "positive" | "warning" | "critical"
```
josepjaume, line 6: We were already exporting this as `StatusVariant` down below, which is a bit more explicit as well naming-wise :)

josepjaume (approved):
Sorry that it took so much time for me to review this! I think it's fine, but feel free to check my comments anyways.

Thanks for the work @jesusgonzalezrivera !

lib/experimental/Navigation/Header/PageHeader/index.tsx
nlopin: > I honestly think this is fine, as a divider is not much else than a bunch of tokens. I would think about it later if we find more use cases.

A bunch of tokens everyone does from scratch and slightly differently. Let's proceed here anyways

nlopin (approved):

lib/experimental/Navigation/Header/PageHeader/index.tsx
josepjaume: I understand, yes. The thing is that it looks like this specific divider might be just scoped to a "Toolbar" no? We probably need to define that component?

nlopin:
@jesusgonzalezrivera could you review the comments and address or resolve them? we are ready to merge

jesusgonzalezrivera:
Thanks a lot @nlopin 🙇‍♂️ I didn't receive any notification 🫠 I have fixed the one about the `Variant` but I would prefer to leave the other proposed change for another moment, so I can continue with the migration on Monday 🙏

jesusgonzalezrivera:
Anyway, I am still waiting for the manual review. Should I ping someone else? @nlopin @josepjaume 

image

PR #760, opened 2024-10-29
Update icon sizes

sauldom102:
## 🔑 What?

Update icon sizes.

## 🚪 Why?

The sizes that we have for our icons now are not the ones declared in F1 in Figma.

## 🏡 Context

- [🧵 Slack](https://factorialteam.slack.com/archives/C07LQSJ5UQY/p1730205706523929)

### DIFF
--- a/lib/components/Utilities/Icon/index.tsx
+++ b/lib/components/Utilities/Icon/index.tsx
@@ -10,16 +10,13 @@ import {
 const iconVariants = cva("inline-block shrink-0", {
   variants: {
     size: {
-      xl: "h-12 w-12",
-      lg: "h-8 w-8",
-      md: "h-5 w-5",
-      sm: "h-4 w-4",
-      xs: "h-3 w-3",
+      lg: "w-[24px]",
+      md: "w-[20px]",
+      sm: "w-[16px]",
     },
   },
-
   defaultVariants: {
-    size: "xl",
+    size: "md",
   },
 })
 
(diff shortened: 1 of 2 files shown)


lib/components/Utilities/Icon/index.tsx
```diff
@@ -10,16 +10,13 @@ import {
 const iconVariants = cva("inline-block shrink-0", {
   variants: {
     size: {
-      xl: "h-12 w-12",
-      lg: "h-8 w-8",
-      md: "h-5 w-5",
-      sm: "h-4 w-4",
-      xs: "h-3 w-3",
+      lg: "w-[24px]",
```
dani-moreno, line 13: What if you use [size](https://tailwindcss.com/docs/size)? That will set width and height, so no need for the `aspect-square`?

lib/components/Utilities/Icon/index.tsx
```diff
@@ -10,16 +10,13 @@ import {
...
+      md: "w-[20px]",
```
dani-moreno, line 14: If you look at [spacing.ts](https://github.com/factorialco/factorial-one/blob/main/tailwind/spacing.ts), you can see how 24px would be 6, 20px is 5 and 16px is 4.

I think you may be getting other values with the hover preview of Tailwind IntelliSense (happens to me too), but those values are not the real ones.

lib/components/Utilities/Icon/index.tsx
sauldom102: Ooh didn't know that one, great!

lib/components/Utilities/Icon/index.tsx
sauldom102: Ohh really? I'm gonna double-check, thank you!

lib/components/Utilities/Icon/index.tsx
sauldom102: Let's better use `w-6` with `aspect-square`, because `size` doesn't seem to be using `rem`

nlopin (approved):

PR #763, opened 2024-10-29
Clean-up of some widget content components that we don't need

sauldom102:
## 🔑 What?

Clean-up of some widget content components that we don't need (I'll remove the old `Tag` one when updating the sidebar with the `DataList`, because we're using it there).


nlopin (approved):
see -200 loc, approve without checking 😂

PR #764, opened 2024-10-29
Remove Badge

dani-moreno:
## 🚪 Why?

`Badge` is getting deprecated. It was never created in Figma and can be replaced with Tags.

## 🔑 What?

- Remove `Badge`.
- Replace its usage with tags.


lib/components/exports.ts
```diff
@@ -1,4 +1,3 @@
 export * from "./Actions/exports"
 export * from "./Charts/exports"
-export * from "./Information/exports"
```
sauldom102, line 3: Oh we need that, I don't think we should remove it

dani-moreno: If I remove the `Badge` component, there's nothing inside of the folder, right? Should I left it even if it's empty?

nlopin (approved):

src/playground/Widgets/Examples/details.stories.tsx
```diff
@@ -114,10 +108,10 @@ export const Details: Story = {
               <div className="flex flex-row flex-wrap gap-2">
                 {teams.list.map((team, index) => {
                   return !team ? null : (
-                    <Badge
+                    <TeamTag
                       key={team + index}
-                      text={team}
-                      avatar={{ alt: team[0] }}
+                      teamName={team}
+                      teamImageUrl={team[0]}
```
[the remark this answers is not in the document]
nlopin: is it related to this PR?

Speaking of `TeamTag` API, we indeed should derive the letter from the name

PR #765, opened 2024-10-29
Set lighter color for widget separator

sauldom102:
## 🔑 What?

Set lighter color for widget separator.

### DIFF
--- a/lib/ui/separator.tsx
+++ b/lib/ui/separator.tsx
@@ -14,7 +14,7 @@ export const Separator = forwardRef<HTMLDivElement, SeparatorProps>(
         className={cn("-mx-4 h-[1px]", !bare ? "my-4" : undefined)}
         style={{
           backgroundImage:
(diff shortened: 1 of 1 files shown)


dani-moreno:
Have you checked with Mario? The color used in Figma is `neutral-30`. If we want this change, we should change it in both places.

sauldom102:
> Have you checked with Mario? The color used in Figma is `neutral-30`. If we want this change, we should change it in both places.

It's now 20. I think he changed it along with the widget border colors.

Screenshot 2024-10-30 at 10 11 29

dani-moreno:
Hey! Checked with Mario, and [this](https://www.figma.com/design/pZzg1KTe9lpKTSGPUZa8OJ/Web-Components?node-id=73-4896) is how it should work (sorry for the back and forth! it wasn't updated in the employee profile it seems):

- The dashed line should use neutral-30.
- The default line should use neutral-10.

nlopin (approved):

PR #766, opened 2024-10-29
Fix progress bar duo color

sauldom102:
## 🔑 What?

Fix progress bar duo color.

### DIFF
--- a/lib/experimental/Widgets/Content/ProgressBarDuo/index.tsx
+++ b/lib/experimental/Widgets/Content/ProgressBarDuo/index.tsx
@@ -1,5 +1,4 @@
 import { autoColor } from "@/components/Charts/utils/colors"
-import { cn } from "@/lib/utils"
 import { forwardRef } from "react"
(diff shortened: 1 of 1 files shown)


lib/experimental/Widgets/Content/ProgressBarDuo/index.tsx
```diff
@@ -14,11 +13,8 @@ export const ProgressBarDuo = forwardRef<HTMLDivElement, DualProgressBarProps>(
     return (
       <div ref={ref} className="flex h-2 w-full gap-0.5 overflow-hidden">
         <div
-          className={cn(
-            "rounded transition-all duration-300 ease-in-out",
-            `bg-[${autoColor(0)}]`
```
sauldom102, line 19: We should avoid passing interpolated classnames like this as tailwind works at build time and it won't recognise it when we're running it

eliseo-juan (approved):

PR #767, opened 2024-10-29
Fix look of tasks widget

sauldom102:
## 🔑 What?

Fix look of tasks widget.

This is just an example, we need to update some of these things in our main repo as well as we're not using this component directly there. Along the way I realized about the changes we need to do. I also updated the `Indicator` component so that the label and the icon is below and `TasksList` component with some sizing for the icons and spacing.

## 👁️ Visual proof

Before:

Screenshot 2024-10-29 at 20 49 08

After:

Screenshot 2024-10-29 at 20 48 19

### DIFF
--- a/lib/experimental/Widgets/Content/TasksList/TaskItem/index.tsx
+++ b/lib/experimental/Widgets/Content/TasksList/TaskItem/index.tsx
@@ -1,7 +1,8 @@
-import { Badge } from "@/components/Information/Badge"
 import { Icon } from "@/components/Utilities/Icon"
 import { Counter } from "@/experimental/Information/Counter"
-import { DottedCircle, InProgressTask } from "@/icons"
+import { AlertTag } from "@/experimental/Information/Tags/AlertTag"
+import { RawTag } from "@/experimental/Information/Tags/RawTag"
+import { Calendar, DottedCircle, InProgressTask } from "@/icons"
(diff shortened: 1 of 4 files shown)


lib/experimental/Widgets/Content/TasksList/TaskItem/index.tsx
```diff
@@ -56,16 +57,16 @@ export function TaskItem({
...
               ? "hsl(var(--neutral-40))"
               : "hsl(var(--neutral-50))"
           }
         />
       )}
-      {!hideIcon && status === "in-progress" && <InProgressTask />}
+      {!hideIcon && status === "in-progress" && <Icon icon={InProgressTask} />}
       <p className="mt-0.5 line-clamp-2 flex-1 font-medium">{task.text}</p>
       {!!task.badge && (
         <Badge
```
eliseo-juan, line 72: I think you are ready to use your brand-new Tag component!

eliseo-juan (approved):
Good job! I cannot wait to merge everything!

lib/experimental/Widgets/Content/TasksList/TaskItem/index.tsx
sauldom102: Ohh true!

PR #774, opened 2024-10-30
Avatar update

dani-moreno:
## 🚪 Why?

Our current `Avatar` is outdated in code, as we have three different Avatar components in Figma, depending on the 'resource' it showcases.

These three avatars are:

- [UserAvatar](https://www.figma.com/design/pZzg1KTe9lpKTSGPUZa8OJ/Web-Components?node-id=5-17&t=k3kLGgxILv0Z3Pty-4), for users.
- [TeamAvatar](https://www.figma.com/design/pZzg1KTe9lpKTSGPUZa8OJ/Web-Components?node-id=1038-13279&t=k3kLGgxILv0Z3Pty-4), for teams.
- [CompanyAvatar](https://www.figma.com/design/pZzg1KTe9lpKTSGPUZa8OJ/Web-Components?node-id=240-10208&t=k3kLGgxILv0Z3Pty-4), for companies, used in the CompanySelector.

| UserAvatar | TeamAvatar | CompanyAvatar |
|-|-|-|
| image | image | image |

## 🔑 What?

Following the same approach as the tags, I've created three different components for these three avatars.

- They all share the same base (`BaseAvatar`), which handles almost everything:
  - Handles generating the initials shown in the fallback.
  - Handles generating a random color, depending on the string passed.

---

#### UserAvatar

Accepts a `firstName` and `lastName`, and shows a random color.

#### TeamAvatar

Accepts a `name`, and shows a random color.

#### CompanyAvatar

Accepts a `name`, and always shows viridian as its color, as in the design.

## 🏡 Context

- [🎨 Figma](https://www.figma.com/design/pZzg1KTe9lpKTSGPUZa8OJ/Web-Components?node-id=1-3)


lib/experimental/Layouts/Utils/DetailsItemsList/index.stories.tsx
dani-moreno, line 1: This one was super problematic. Are you deprecating it in favor of the new item component, @nlopin? If I'm wrong, I can try to adapt it.

lib/experimental/Information/Avatars/BaseAvatar/index.tsx
```diff
@@ -0,0 +1,67 @@
...
+  size?: ShadAvatarProps["size"]
+  color?: ShadAvatarProps["color"] | "random"
+}
+
+function getInitials(
+  name: string | string[],
+  size?: ShadAvatarProps["size"]
+): string {
+  const nameArray = Array.isArray(name) ? name : [name]
+  const hasSmallSize = size === "xsmall" || size === "small"
```
nlopin, line 24: I propose to reduce some redundancy in the name, because _small_ is always about size.

```suggestion
  const isSmall = size === "xsmall" || size === "small"
```

lib/experimental/Information/Avatars/BaseAvatar/index.tsx
```diff
@@ -0,0 +1,67 @@
...
+
+type Props = {
+  type: ShadAvatarProps["type"]
+  name: string | string[]
+  src?: string
+  size?: ShadAvatarProps["size"]
+  color?: ShadAvatarProps["color"] | "random"
+}
+
+function getInitials(
```
nlopin, line 19: this thing worth some unit tests, we have 4 different inputs and 3 different outputs

lib/experimental/Information/Avatars/BaseAvatar/index.tsx
```diff
@@ -0,0 +1,67 @@
...
+  ({ src, name, size, type, color }, ref) => {
+    const initials = getInitials(name, size)
+    const avatarColor =
+      color === "random"
+        ? getAvatarColor(Array.isArray(name) ? name.join("") : name)
+        : color
+
+    return (
+      <AvatarComponent size={size} type={type} color={avatarColor} ref={ref}>
+        <AvatarImage src={src} alt={initials} />
```
nlopin, line 60: Initials is a very poor alternative text (see the comment above)

lib/experimental/Information/Avatars/BaseAvatar/index.tsx
```diff
@@ -0,0 +1,67 @@
...
+  color as AvatarColors,
+  Avatar as AvatarComponent,
+  AvatarFallback,
+  AvatarImage,
+} from "@/ui/avatar"
+import { ComponentProps, forwardRef } from "react"
+
+type ShadAvatarProps = ComponentProps<typeof AvatarComponent>
+
+type Props = {
```
nlopin, line 11: can we move props closer to the component?

lib/experimental/Information/Avatars/BaseAvatar/index.tsx
```diff
@@ -0,0 +1,67 @@
...
+  return AvatarColors[index]
+}
+
+export const BaseAvatar = forwardRef<HTMLDivElement, Props>(
+  ({ src, name, size, type, color }, ref) => {
+    const initials = getInitials(name, size)
+    const avatarColor =
+      color === "random"
+        ? getAvatarColor(Array.isArray(name) ? name.join("") : name)
+        : color
```
nlopin, line 56: if `color` is not set, do we render white background? shouldn't `random` be the default value?

lib/experimental/Information/Avatars/BaseAvatar/index.tsx
josepjaume: Maybe it should even belong to a separate `utils.ts` file?

lib/experimental/Information/Avatars/BaseAvatar/index.tsx
```diff
@@ -0,0 +1,67 @@
...
+}
+
+function getInitials(
+  name: string | string[],
+  size?: ShadAvatarProps["size"]
+): string {
+  const nameArray = Array.isArray(name) ? name : [name]
+  const hasSmallSize = size === "xsmall" || size === "small"
+
+  if (hasSmallSize) return nameArray[0][0].toUpperCase()
```
josepjaume, line 26: Surprised that this isn't making typescript complain, as `nameArray[0]` can be undefined. Or is it?

lib/experimental/Information/Avatars/BaseAvatar/index.tsx
```diff
@@ -0,0 +1,67 @@
...
+  }
+
+  const index =
+    ((hash % AvatarColors.length) + AvatarColors.length) % AvatarColors.length
+
+  return AvatarColors[index]
+}
+
+export const BaseAvatar = forwardRef<HTMLDivElement, Props>(
+  ({ src, name, size, type, color }, ref) => {
```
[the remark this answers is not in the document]
josepjaume: image

lib/experimental/Layouts/Utils/Tag/index.tsx
```diff
@@ -1,24 +1,21 @@
...
-import { getColorFromText } from "../helper"
 
 interface TagProps {
   text: string
-  avatar?: ComponentProps<typeof Avatar>
+  avatar?: {
+    firstName?: string
+    lastName?: string
+    src?: string
+    size?: ComponentProps<typeof UserAvatar>["size"]
```
josepjaume, line 11: Do we want everything to be optional, or just the `avatar` prop? My guess is something needs to be required here.

lib/experimental/Navigation/Sidebar/User/index.stories.tsx
```diff
@@ -11,9 +11,9 @@ type Story = StoryObj<typeof meta>
 
 export const Default: Story = {
   args: {
-    name: "Dani Moreno",
+    firstName: "Dani",
+    lastName: "Moreno",
```
josepjaume, line 15: You're not the center of the universe, Dani

dani-moreno: Becoming John Doe, one step at a time

lib/experimental/Information/Avatars/BaseAvatar/index.tsx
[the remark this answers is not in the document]
dani-moreno: Hey, I added a way to define `aria-label` or `aria-labelledby` when using any of the avatars, but I'm not sure if that's the best way to do it. Pretty new to this stuff, heh. Can you check? 🙏

lib/experimental/Information/Avatars/BaseAvatar/index.tsx
dani-moreno: What’s the best move here? Should we use the content of ARIA tags as alt? If we’re covering the accessibility description at the root level with those tags, maybe this image is just decorative? 🤔

lib/experimental/Information/Avatars/BaseAvatar/index.tsx
dani-moreno: Ah yes, you’re totally right, thanks!

lib/experimental/Information/Avatars/BaseAvatar/index.tsx
dani-moreno: Moved everything to a `utils.ts` file. What should I test here? No experience with unit tests tbh 😓

lib/experimental/Layouts/Utils/Tag/index.tsx
dani-moreno: To be honest, I just adapted it a bit to work, but this Tag is getting deprecated soon, replacing it with #751

lib/experimental/Information/Avatars/BaseAvatar/index.tsx
nlopin: added tests

lib/experimental/Information/Avatars/BaseAvatar/index.tsx
nlopin: somehow it thinks it will get a string. I added both the test and the fix for this case

lib/experimental/Information/Avatars/BaseAvatar/index.tsx
nlopin: I think you have it handled by allowing to pass an aria-label

nlopin (approved):
Looks great 👍

PR #776, opened 2024-10-30
Document and fix layouts

josepjaume:
## 🚪 Why?

Layouts are a core part of our application and yet, we haven't put much thought into it.

## 🔑 What?

This adds some proper examples to layouts, and standardizes, well... the StandardLayout. While migrating to Factorial One, I've found out that most layouts either use a full-width content approach (tables?), a narrow content approach (for readability) or fullscreen (org chart).

We're in the process to adding more layouts, but this is a necessary step for backwards compatibility.


lib/experimental/Layouts/InfoPaneLayout/index.tsx
```diff
@@ -11,7 +11,7 @@ export const InfoPaneLayout = forwardRef<HTMLDivElement, InfoPaneLayoutProps>(
     return (
       <div
         ref={ref}
-        className="flex h-full flex-col-reverse gap-4 text-f1-foreground sm:flex-row sm:gap-0"
+        className="flex h-full flex-col-reverse gap-4 overflow-auto text-f1-foreground sm:flex-row sm:gap-0"
       >
         <div className="sm:my-6 sm:ms-6 sm:basis-3/4">{mainContent}</div>
```
sauldom102, line 16: What about passing the `mainContent` as the `children` directly? 👀

sauldom102 (approved):

lib/experimental/Layouts/InfoPaneLayout/index.tsx
josepjaume: Absolutely, that should have been done like this.

josepjaume: I'll do it!

nlopin:
Are they planned to be used only inside the `ApplicationFrame`?

josepjaume:
Yes @nlopin - I would actually call them Page Layouts. Should i change the name?

lib/experimental/Layouts/FullscreenLayout/index.tsx
```diff
@@ -0,0 +1,9 @@
+import React from "react"
+
+export interface StandardLayoutProps {
+  children: React.ReactNode
+}
+
+export function FullscreenLayout({ children }: StandardLayoutProps) {
+  return <div className="relative flex flex-1 [&>div]:flex-1">{children}</div>
```
nlopin, line 8: What is your thought on using `flex` as a default container across layouts? 

I thought we can have it HTML default value to keep things familiar

lib/experimental/Layouts/StandardLayout/index.tsx
```diff
@@ -1,13 +1,27 @@
...
-export interface StandardLayoutProps {
-  children: React.ReactNode
+export interface StandardLayoutProps
+  extends VariantProps<typeof layoutVariants> {
+  children?: React.ReactNode
 }
 
-export function StandardLayout({ children }: StandardLayoutProps) {
+const layoutVariants = cva(
+  "relative flex w-full flex-1 flex-col gap-4 place-self-center overflow-y-auto px-6 py-5",
```
nlopin, line 11: why do we set `relative`?

nlopin:
I think every layout must have [`isolate` class](https://tailwindcss.com/docs/isolation) to reduce the possibility of z-index collision with the rest of the frame.

This means all children of the layout will be _below_ the sidebar. It can be an issue itself if we don't use portals for our modals and popups, however I believe we do use them

nlopin:
> Yes @nlopin - I would actually call them Page Layouts. Should i change the name?

no, all good. I tried to understand why do we rely so hard on flex positioning

lib/experimental/Layouts/StandardLayout/index.tsx
josepjaume: To kind of create a "box" nobody can escape, even when using `absolute` positioning. Isn't this correct?

lib/experimental/Layouts/FullscreenLayout/index.tsx
josepjaume: In this case, what I'm trying to do is for the content to stretch. Think the `Org chart` layout, for example. Maybe it's too much and we should just let people create their own layouts in this scenario?

josepjaume:
@nlopin I will do it! Actually, the `isolate` property might be better set at the `page` level to guarantee every layout will always have this.

BTW another interesting thing we need to standardize are loading states. I guess we should create a `PageLayoutBaseProps` interface that every layout extends from.

nlopin (approved):

PR #794, opened 2024-11-05
WidgetList component

Rogermax:
## 🚪 Why?

Make foundations for a better widget layout

## 🔑 What?

New component WidgetList

## 🏡 Context

https://github.com/user-attachments/assets/6e642cdb-df3d-448e-872c-45c6c77aaa62

### DIFF
--- a/lib/experimental/Widgets/Content/TasksList/index.stories.tsx
+++ b/lib/experimental/Widgets/Content/TasksList/index.stories.tsx
@@ -29,6 +29,8 @@ export const Default: Story = {
         { id: 3, text: "Write changelog" },
         { id: 4, text: "Product review" },
         { id: 5, text: "Final conclusions" },
+        { id: 6, text: "Final conclusions 2" },
+        { id: 7, text: "Final conclusions 3" },
       ],
     },
   },
--- a/lib/experimental/Widgets/WidgetList/WidgetListItem/ImageWidget.tsx
+++ b/lib/experimental/Widgets/WidgetList/WidgetListItem/ImageWidget.tsx
@@ -1,0 +1,8 @@
+import React from "react"
+import { ImageWidgetProps } from "../WidgetListItemTypes"
+
+export const ImageWidget: React.FC<ImageWidgetProps> = ({ imageUrl }) => (
+  <div className="border-gray-200 border-b p-4">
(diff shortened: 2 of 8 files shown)


nlopin:
I have doubts about the approach, but because there is a PR and a lot of work already done, it is hard to share it. What do you think if next time we discuss the approach before writing the code?

I see the value in providing the default shell for widget, especially that many of them use this concept of "+n more". At the same time, the current approach relies on setting a maximum height as a way to control the content which does not give us enough sense of the amount of data user will see. I think the height of a widget is depends on the content, meaning the content dictates the height, not the other way around.

In addition, the manual calculation of the height is a performance bottleneck, since [`offsetHeight` triggers reflow](https://gist.github.com/paulirish/5d52fb081b3570c81e3a) (more about this on [google dev blog](https://web.dev/articles/avoid-large-complex-layouts-and-layout-thrashing))

We already have a bunch of list component and `WidgetList` could be a wrapper of them but giving an obvious API. The developer can provide the amount of items they want to show and optionally the amount of items to show in the "+n items" text together with the url where to redirect the user. If the number is not provided, we calculate it from the list of items provided.

Rogermax:
> I have doubts about the approach, but because there is a PR and a lot of work already done, it is hard to share it. What do you think if next time we discuss the approach before writing the code?
> 
> I see the value in providing the default shell for widget, especially that many of them use this concept of "+n more". At the same time, the current approach relies on setting a maximum height as a way to control the content which does not give us enough sense of the amount of data user will see. I think the height of a widget is depends on the content, meaning the content dictates the height, not the other way around.
> 
> In addition, the manual calculation of the height is a performance bottleneck, since [`offsetHeight` triggers reflow](https://gist.github.com/paulirish/5d52fb081b3570c81e3a) (more about this on [google dev blog](https://web.dev/articles/avoid-large-complex-layouts-and-layout-thrashing))
> 
> We already have a bunch of list component and `WidgetList` could be a wrapper of them but giving an obvious API. The developer can provide the amount of items they want to show and optionally the amount of items to show in the "+n items" text together with the url where to redirect the user. If the number is not provided, we calculate it from the list of items provided.

I was trying to make a better framework to avoid having to handle in every component the same issues with height, check this comment -> https://github.com/factorialco/factorial-one/pull/753#issuecomment-2447866494

PR #803, opened 2024-11-06
Use Indicator in IndicatorsList component

sauldom102:
## 🔑 What?

Use Indicator in IndicatorsList component.

## 🚪 Why?

We were reinventing the wheel in `IndicatorsList` component by not using the `Indicator` that we already had.


sauldom102 (commented):
Actually, we're not directly exporting `Indicator` component from F1, so we were observing some widgets using `IndicatorsList` not getting updated

nlopin (approved):

PR #804, opened 2024-11-06
Add Avatars in Dropdown and Select

dani-moreno:
## 🚪 Why?

Dropdown and Select should have the option to [render an avatar](https://www.figma.com/design/pZzg1KTe9lpKTSGPUZa8OJ/Web-Components?node-id=72-1968) in each item.

## 🔑 What?

- Add the option to display avatars in `Dropdown` and `Select`.
- Add a `utils` file in Avatars that handles the render an avatar depending on the type.
- Displays the logo in the `CompanySelector`.

image

## 🏡 Context

- [🎨 Figma](https://www.figma.com/design/pZzg1KTe9lpKTSGPUZa8OJ/Web-Components?node-id=72-1968)


lib/experimental/Information/Avatars/utils.tsx
```diff
@@ -0,0 +1,53 @@
...
+
+export type AvatarVariant =
+  | ({ type: "person" } & PersonAvatarProps)
+  | ({ type: "team" } & TeamAvatarProps)
+  | ({ type: "company" } & CompanyAvatarProps)
+
+export function renderAvatar(
+  avatar: AvatarVariant,
+  size: (typeof sizes)[number] = "xsmall"
+) {
```
josepjaume, line 19: There's the risk that we add a new type of avatar (`avatar.type`) and we forget to deal with it.

Here's a trick:

```suggestion
export function renderAvatar(
  avatar: AvatarVariant,
  size: (typeof sizes)[number] = "xsmall"
): ReactNode {
```

By doing this, you're telling the TypeScript compiler that this component needs to *always* return something. Go and try to remove `case "team"` for example: it will complain.

Otherwise, without manually specifying, the return type of the function will be `ReactNode | undefined` which is not what we want!

lib/experimental/Navigation/Dropdown/index.stories.tsx
```diff
@@ -107,3 +107,40 @@ export const WithLinks: Story = {
...
+      },
+      {
+        label: "Saúl Domínguez",
+        avatar: {
+          type: "person",
+          firstName: "Saúl",
+          lastName: "Domínguez",
+          src: "https://github.com/sauldom102.png",
+          "aria-label": "Saúl Domínguez avatar",
+        },
```
josepjaume, line 142: No Dani this time?

josepjaume (approved):
Awesome!

lib/experimental/Navigation/Dropdown/index.stories.tsx
dani-moreno: It's even better, a list of cool folks 😎

PR #806, opened 2024-11-07
Simplify TasksList component and add stories for TaskItem

sauldom102:


lib/experimental/Widgets/Content/TasksList/index.tsx
```diff
@@ -2,8 +2,7 @@ import { Task, TaskItem, TaskStatus } from "./TaskItem"
 
 interface TasksList {
   inProgress?: (Task | string)[]
-  noDue?: (Task | string)[]
-  due?: (Task | string)[]
```
sauldom102, line 6: This should be only `todo`

lib/experimental/Widgets/Content/TasksList/TaskItem/index.tsx
```diff
@@ -45,7 +45,7 @@ export function TaskItem({
   hideIcon = false,
 }: TaskItemProps) {
   const className = cn(
-    "flex flex-row items-center gap-1 rounded-sm border border-solid border-transparent px-2 py-1.5",
+    "flex flex-row items-center gap-1 rounded-sm border border-solid border-transparent px-2 py-1.5 text-f1-foreground",
```
sauldom102, line 48: Needed to add this class because otherwise items show up with a link color

Screenshot 2024-11-07 at 17 28 48

dani-moreno:
I noticed we usually vertically align the icon at the top in Figma, but here it is centered.

image


Just asked Mario to make sure, it should be at the top.

sauldom102:
> I noticed we usually vertically align the icon at the top in Figma, but here it is centered.
> 
> image
> Just asked Mario to make sure, it should be at the top.

Yeah, makes more sense

sauldom102:
> > I noticed we usually vertically align the icon at the top in Figma, but here it is centered.
> > image
> > Just asked Mario to make sure, it should be at the top.
> 
> Yeah, makes more sense

I'll better fix it in this subsequent PR @dani-moreno https://github.com/factorialco/factorial-one/pull/811

dani-moreno (approved):

PR #809, opened 2024-11-07
Fix line in Tabs component

dani-moreno:
| Before | After |
|--------|-------|
| image | image |

The line of the selected element was misaligned with the border by 1px, and it kept me up at night.

I had to mimic the bottom border with an absolute element since the line of the selected element couldn't be drawn on top of the container's border.


sauldom102 (commented):
Your attention to the detail is incredible! ❤️

sauldom102 (approved):

PR #812, opened 2024-11-08
Alpha base colors

dani-moreno:
## Description

Changes our base grey colors to use opacity, as we do in Figma. This purely visual change lets us make transparent elements that blend with the background, [especially important to achieve the Home design](https://www.figma.com/design/060q78tnGIGaRc1sxecTvV/Home-Web?node-id=944-46421&t=KPFwn0mNOGFtNBe8-4).

Also, did a pair session with Mario to review a few things after the change:
- Made some color adjustments to a few sidebar items.
- Fixed some color names that didn't match between Figma and the code.
- Created some new tokens and adjusted the `Shortcut` component, where they are used.
- Mobile sidebar has a new design, closer to how the floating web sidebar looks.
image


sauldom102 (approved):
Great!

nlopin (approved):

PR #813, opened 2024-11-09
Sidebar better implementation

dani-moreno:
## Description

Made a few changes to the sidebar to get a better implementation:

- Removed `AnimatePresence`, resulting in better performance by avoiding unnecessary unmounting/remounting
- Added proper visibility and pointer-events handling to fix accessibility (hidden items can't be clicked or focused)
- Achieved the same smooth animations with simplified logic

----

Also, I took the opportunity to standardize all icons to use our `Icon` component to leverage built-in stroke-width, sizing, and other features.


sauldom102 (approved):
Great!

PR #825, opened 2024-11-13
Create HomeLayout component

sauldom102:
## Screenshots

Large size:

Screenshot 2024-11-13 at 12 28 04

Medium size:

Screenshot 2024-11-13 at 12 28 21

Small size:

Screenshot 2024-11-13 at 12 28 35


styles.css
```diff
@@ -122,3 +122,21 @@
 body {
   @apply font-sans text-base text-f1-foreground antialiased;
 }
+
+@media (max-width: 992px) {
+  .widget-strip > div {
+    @apply min-w-96;
+  }
+}
```
josepjaume, line 130: You should be able to achieve the same behavior by using this class: `lg:[&>div]:min-w-96`

styles.css
```diff
@@ -122,3 +122,21 @@
...
+
+@media (max-width: 992px) {
+  .widget-strip > div {
+    min-width: calc(50vw - 48px);
+  }
+}
```
josepjaume, line 136: I think the same but with:

```
lg:[&>div]:min-w-[calc(50vw-48px)]
```

Or, if using a container query:

```
@home/lg:[&>div]:min-w-[calc(50vw-48px)]
```

styles.css
```diff
@@ -122,3 +122,21 @@
...
+
+@media (max-width: 743px) {
+  .widget-strip > div {
+    min-width: calc(100vw - 40px - 42px);
+  }
+}
```
josepjaume, line 142: Aaand:

```
@home/md:[&>div]:min-w-[calc(100vw-40px-42px)]
```

Weird huh?

styles.css
josepjaume: In any case, I would use Container Queries. We do have support for it and are using it already!

https://github.com/tailwindlabs/tailwindcss-container-queries

The benfit is that you would care about the size of the dashboard itself, not the whole page. So if we change the sidebar's width for example, things would behave as expected.

lib/experimental/Widgets/Layout/HomeLayout/index.tsx
```diff
@@ -0,0 +1,43 @@
...
+  children?: ReactNode
+}
+
+export const HomeLayout = forwardRef<HTMLDivElement, Props>(function Dashboard(
+  { widgets, children },
+  ref
+) {
+  const isSmallerScreen = useMediaQuery("(max-width: 992px)", {
+    initializeWithValue: true,
+  })
```
josepjaume, line 16: What if instead of checking the media query we check for the width of the element? That's very easy to achieve with `useResizeObserver` which comes bundled with the `usehooks-ts` library we already have:

https://usehooks-ts.com/react-hook/use-resize-observer

josepjaume (changes requested):
Very good job! I've added a few comments, hope you don't mind!

styles.css
sauldom102: Makes total sense! I was wrongly assuming that this layout is always gonna take the same space without considering the sidebar, I'm gonna use container queries!

lib/experimental/Widgets/Layout/HomeLayout/index.tsx
sauldom102: Done! ✅

dani-moreno (approved):

PR #828, opened 2024-11-13
Fixes the index tab behavior

josepjaume:
## Description

The following PR fixes some wrong highlighting of tags when one of the tabs is an index one. Since index tabs are usually `/` while other tabs are `/some-other-path`, we need to find the right tab by checking if the current path is active without the index first, and then checking if it's active with the index.

Otherwise, we would incorrectly match the index tab as active, resulting in two tabs being active at the same time.

Not applicable.

### Figma Link

Not applicable.

---

## Type of Change


josepjaume:
I ninja-merged this as I needed the hotfix to work!

PR #833, opened 2024-11-14
Add support to add multiline text to details item

sauldom102:
## Description

In Finance they have a valid use case of this component for showing an address that has more that one line.

## Screenshots

Screenshot 2024-11-14 at 13 23 11


dani-moreno:
I wonder if the icon should be aligned on top or at the center 🤔 maybe it's worth asking Mario.

sauldom102:
> I wonder if the icon should be aligned on top or at the center 🤔 maybe it's worth asking Mario.

Yeah I'll ask him

PR #834, opened 2024-11-14
Create OverviewLayout component

sauldom102:
## Description

We could be more prescriptive with this layout as it's a common one that we'll use for all overview pages, not allowing to place anything else on the sidepanel apart from the `DetailsItemsList` component.

The idea with this component is to deprecate the `InfoPaneLayout` and stop exporting it from FactorialOne, this will close the door to some unexpected customization.

I wish we could have closed this even more, so that the `children` is not any html, but we have the Widgets abstraction in our main frontend codebase and we can't entirely do that.

Proposed here: https://factorialteam.slack.com/archives/C07LQSJ5UQY/p1731587350661809

## Screenshots

Screenshot 2024-11-14 at 14 02 47


CarlosBM18 (approved):
LGTM!

lib/experimental/PageLayouts/OverviewLayout/index.stories.tsx
```diff
@@ -0,0 +1,44 @@
...
+const meta = {
+  component: OverviewLayout,
+  tags: ["autodocs"],
+  decorators: [PageDecorator],
+  args: {
+    children: <Placeholder className="h-[450px]">Main</Placeholder>,
+    sidepanel: {
+      title: DETAILS_ITEMS_ARGS.title,
+      items: DETAILS_ITEMS_ARGS.details,
+    },
```
josepjaume, line 24: I'm worried that this won't scale well in the future, perhaps? But I guess it's easier to make an API less strict than more strict, so I'm not opposing the idea.

lib/experimental/PageLayouts/OverviewLayout/index.stories.tsx
```diff
@@ -0,0 +1,44 @@
...
+  },
+  parameters: {
+    a11y: {
+      config: {
+        rules: [{ id: "svg-img-alt", enabled: false }],
+      },
```
josepjaume, line 30: Do we need this?

lib/experimental/PageLayouts/OverviewLayout/index.tsx
```diff
@@ -0,0 +1,27 @@
...
+
+export interface OverviewLayoutProps {
+  children: ReactNode
+  sidepanel: {
+    title: string
+    items: DetailsItemType[]
+  }
+}
+
+export const OverviewLayout = forwardRef<HTMLDivElement, OverviewLayoutProps>(
```
josepjaume, line 14: Way better name than `InfoPane`. Great job!

josepjaume (approved):
Great job! Please keep the design team on the loop. I think the naming makes total sense.

lib/experimental/PageLayouts/OverviewLayout/index.stories.tsx
sauldom102: Probably not, I just copied over some other story and forgot to remove that!

lib/experimental/PageLayouts/OverviewLayout/index.stories.tsx
sauldom102: Yep, maybe we end up having more cases, but I also thought it's easier to do that change in the future if we don't have this component too open and working in many different ways in our frontend

PR #842, opened 2024-11-15
Fix tab matching logic

josepjaume:
## Description

The logic to find the active tag when using `index` was overly complicated. This simplifies things by quite a lot and makes the more understandable.

None

### Figma Link

None

---

## Type of Change


sauldom102 (approved):
Great!

PR #843, opened 2024-11-15
Add Home Reaction component

dani-moreno:
## Description

In [the new Home](https://www.figma.com/design/060q78tnGIGaRc1sxecTvV/Home-Web?node-id=987-21551&t=anWU04NUVRrWfwop-4), we have emoji reactions in each post of communities (like Slack).

This PR adds a few components, that form the reaction part of a post.

## `reaction`

First is the `reaction` component, which is the emoji button that users can click to add a reaction with that same emoji. It's designed to be very engaging and encourage people to react.

https://github.com/user-attachments/assets/92cbc7c7-5ccf-4ddf-9da5-82c756a660a4

It adds and uses `canvas-confetti` to shoot particles when the user clicks on the button, and `number-flow` for the numbers spinning when clicking.

---

## `picker`

The `picker` component lets the user react quick with a few selected emojis. It's basically a popover with some emojis.

https://github.com/user-attachments/assets/5fd1337a-c99c-4862-b809-5f90e1630104

> [!NOTE]
> This component needs some more work, as we also should let the user pick any emoji they want, with a complete emoji picker. I'll tackle that in another PR, didn't want this one to become huuuge.

---

## `reactions`

Merges the other two components. This will be the one used in posts. It has a `onInteraction` prop to handle adding/removing emojis.

---

## `twemoji` helper

I've also added a few helper functions to draw emojis from `twemoji`. As you can see in our designs, we don't use native emojis because, on devices like Mac or iPhone, they have this kinda 3D look that doesn't fit our flat design.

---

## Type of Change

---

- [ ] Component added to the

- [x] **Responsiveness**

  - [x] Verified the component works across various screen sizes

- [x] **Accessibility**
  - [x] Accessibility standards meet at least AA level requirements
  - [x] All interactive elements are keyboard-navigable and have focus states
  - [x] Proper ARIA attributes are used where needed


sauldom102 (approved):
In love with the confetti animation, it's being so nice discovering this library https://www.kirilv.com/canvas-confetti/

lib/lib/emojis.tsx
```diff
@@ -0,0 +1,53 @@
+import { cva, type VariantProps } from "class-variance-authority"
+import { parse } from "twemoji-parser"
+
+interface ParseObject {
+  url: string
+  indices: [number, number]
+  text: string
+}
+
```
sauldom102, line 9: Small comment here, don't we have two levels of `lib` here? Maybe it's better to have something like `lib/emojis/index.tsx` instead of `lib/lib/emojis.tsx`

package.json
```diff
@@ -136,5 +140,10 @@
   },
   "optionalDependencies": {
     "@rollup/rollup-linux-x64-gnu": "^4.24"
+  },
+  "dependencies": {
+    "@number-flow/react": "^0.3.0",
+    "canvas-confetti": "^1.9.3",
+    "twemoji-parser": "^14.0.0"
```
sauldom102, line 147: Probably we should better have these as `peerDependencies`

lib/lib/emojis.tsx
dani-moreno: Umm, it seems weird, but we have almost everything inside the first lib folder (components, experimental, ui...), so it doesn't seem right to have emojis at the same level, does it?

PR #851, opened 2024-11-18
Fix avatar background with image

dani-moreno:
## Description

In avatars, when using an image with a transparent background, you can see the randomly picked color background behind, [which blends in weird and unintended ways](https://factorialteam.slack.com/archives/C071PC4TL1L/p1731945764284419). This PR changes that, forcing a white background when there's an image specified, so we make it look good.

(Also supporting dark mode)

## Screenshots

https://github.com/user-attachments/assets/a78f1ab7-e50d-43dc-a9fa-76d1e6f77cc7

---

## Type of Change


eliseo-juan (approved):

nlopin:
is it a fix on the video?

PR #856, opened 2024-11-19
Create WidgetHighlightButton component

sauldom102:
## Description

Create WidgetHighlightButton component.

## Screenshots

Screenshot 2024-11-19 at 10 30 17

### Figma Link

https://www.figma.com/design/pZzg1KTe9lpKTSGPUZa8OJ/Web-Components?node-id=2559-11116&t=JRkXvpmtZLqIXB1M-4

### DIFF
--- a/lib/experimental/Widgets/Content/Highlights/WidgetHighlightButton/index.tsx
+++ b/lib/experimental/Widgets/Content/Highlights/WidgetHighlightButton/index.tsx
@@ -1,0 +1,63 @@
+import { Icon, IconType } from "@/components/Utilities/Icon"
+import { cn } from "@/lib/utils"
+
+type Props = {
+  label: string
+  icon: IconType
+  iconClassName?: string
+  count: number
+  onClick?: () => void
+}
+
+export type WidgetHighlightButtonProps = Props
+
+type WrapperProps = {
+  onClick?: (ev: React.MouseEvent<HTMLAnchorElement>) => void
+  children: React.ReactNode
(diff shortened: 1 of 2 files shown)


dani-moreno (approved):
Don't wanna block this but, maybe this component can be more generic, and not belong to a widget?

lib/experimental/Widgets/Content/Highlights/WidgetHighlightButton/index.tsx
```diff
@@ -0,0 +1,63 @@
...
+  onClick?: (ev: React.MouseEvent<HTMLAnchorElement>) => void
+  children: React.ReactNode
+}
+
+const Wrapper: React.FC<WrapperProps> = ({ onClick, children }) => {
+  const className =
+    "block rounded-lg border border-solid border-transparent p-[1px]"
+
+  return onClick ? (
+    <a
```
rubenmoya, line 24: What's the purpose of having an `a` tag without an `href`? It won't be read as a link by assistive technologies, and it won't be focusable by default, so we need to add a `tabIndex`, and we just use it for an `onClick`, why not a `button` directly? 🤔

sauldom102:
> Don't wanna block this but, maybe this component can be more generic, and not belong to a widget?

Mario was actually unsure how much we're gonna reuse this and if it had to be a component or not, he has named it `WidgetHighlightButton`, maybe this evolves later on on something apart from widgets 🤔

PR #860, opened 2024-11-19
Add DaytimePage

dani-moreno:
## Description

`DaytimePage` is a variant of the `Page` component that will be used on the Home. It's pretty similar but features a gradient background that changes depending on the time of day: morning, afternoon, or evening.

image

_Morning_

image

_Afternoon_

image

_Evening_

### Figma Link

[Link to Figma Design](https://www.figma.com/design/060q78tnGIGaRc1sxecTvV/Home-Web?node-id=1234-75084&t=anWU04NUVRrWfwop-4)

### DIFF
--- a/lib/experimental/Navigation/Page/index.stories.tsx
+++ b/lib/experimental/Navigation/Page/index.stories.tsx
@@ -1,6 +1,6 @@
 import { Placeholder } from "@/lib/storybook-utils/placeholder"
 import type { Meta, StoryObj } from "@storybook/react"
-import { Page } from "."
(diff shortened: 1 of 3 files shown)


nlopin:
isn't it a one-off thing?

dani-moreno:
> isn't it a one-off thing?

What do you mean? Maybe it should be its own component?

nlopin:
> What do you mean? Maybe it should be its own component?

maybe it should be in the factorial repo, if we use it only once

nlopin (approved):

PR #861, opened 2024-11-19
Explicit exports from the package

nlopin:
## Description

List all allowed exports. It will give us more control over what we expose as well as forbids import of package internals

For users the change will forces them to use `@factorial/factorial-one/experimental` instead of `@factorial/factorial-one/dist/experimental` (no **dist** folder anymore)

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,6 +5,56 @@
   "typings": "dist/factorial-one.d.ts",
   "type": "module",
   "packageManager": "npm@10.8.2",
+  "exports": {
+    ".": {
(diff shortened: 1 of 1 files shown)


josepjaume (approved):
Awesome! This renders the whole `exports.ts` file useless - I guess we could just use `index.ts` from now on?

nlopin:
I think so. I'm not fully sure exports work as expected in the repo. It works magnificently locally, but I have doubts it is that easy

PR #862, opened 2024-11-19
Add Celebration

dani-moreno:
## Description

[Celebration](https://www.figma.com/design/pZzg1KTe9lpKTSGPUZa8OJ/Web-Components?node-id=2341-7709&t=Fgn2XzaLydjTLTcW-4) is one of the Communities components we have in the new Home. It's a card with some info about what the celebration is about, whom it involves, etc.

The component may seem simple at first but it has a few things going on, that's why I separated a few parts of it _(let me know if this is the best way to do it!)_:

- `/components/` include subcomponents of it. For now, it includes the avatar, which involves a clip-path to show a button on the bottom right corner. Eventually, it can include the working button that will open an emoji picker.
- `/hooks/` includes the particles falling when the user hover over the card. It's using canvas-confetti for it.
- `types` stores a few types needed for the component.

> [!WARNING]  
> The emoji reaction is missing its function as we need to work on the emoji picker first. Will do that in another PR!

## Screenshots

https://github.com/user-attachments/assets/e9a455e3-de60-422a-a8e0-7d35b7a1b9ae

_Normal look, and hover_

image

_With no avatar image, the background gets the same background color as the avatar_

image

_Skeleton state_

### Figma Link

[Link to Figma Design](https://www.figma.com/design/pZzg1KTe9lpKTSGPUZa8OJ/Web-Components?node-id=2341-7709&t=Fgn2XzaLydjTLTcW-4)


lib/experimental/Information/Communities/Celebration/components/avatar.tsx
```diff
@@ -0,0 +1,67 @@
...
+          <div className="absolute -bottom-[3px] -right-0.5">
+            <Button
+              label="React"
+              hideLabel
+              round
+              variant="neutral"
+              size="sm"
+              icon={Reaction}
+            />
+          </div>
```
sauldom102, line 62: We should have an `onClickReact` callback as a prop for this component, in case it's not specified we shouldn't show this button

sauldom102: Sounds good! We just need to consider that we'll have some celebrations that we won't be able to react to, so we should have a way to hide this button

lib/experimental/Information/Communities/Celebration/components/avatar.tsx
```diff
@@ -0,0 +1,67 @@
...
+              getAvatarColor(
+                [firstName, lastName].join("")
+              ) as keyof typeof BACKGROUND_COLORS
+            ]
+          : ""
+      )}
+    >
+      {src && (
+        <div
+          className="absolute inset-0 bg-cover bg-center bg-no-repeat opacity-10"
```
sauldom102, line 34: Shall we increase this opacity to make it more vivid? 👀

lib/experimental/Information/Communities/Celebration/components/avatar.tsx
dani-moreno: Ahh, didn't have that into account, thanks @sauldom102!

sauldom102 (approved):

PR #868, opened 2024-11-20
Add Carousel

dani-moreno:
## Description

In the new Home, we are using carousels at some places: first level of widgets, and celebrations. This PR adds a flexible, generic component to display a carousel, with several options.

- Optional navigation arrows
- Optional navigation dots
- Autoplay functionality (meant for celebrations)
- Responsive item sizing through Tailwind classes

## Screenshots

https://github.com/user-attachments/assets/4018bb17-f5f4-4374-b96e-08ae201964a7

---

Also added a story showcasing a "Sneak peek" styling, useful for the widgets, as we want to show a bit of the next widget.

https://github.com/user-attachments/assets/1573472e-fbf8-4835-b572-0c9e2b0efa0b

### Figma Link

[Link to Figma Design](https://www.figma.com/design/060q78tnGIGaRc1sxecTvV/Home-Web?node-id=847-27898&t=JLK4magebnLen8Od-4)

---

## Type of Change


sauldom102 (changes requested):
Maybe we could be more subtle with this hovering, I know I'm being very picky hahaha

https://github.com/user-attachments/assets/c8de0a61-6eb0-4ef0-98dc-ac0a47440b51

lib/experimental/Navigation/Carousel/index.stories.tsx
```diff
@@ -0,0 +1,70 @@
...
+export const SneakPeek: Story = {
+  decorators: [
+    (Story) => (
+      <div className="w-full py-6">
+        <Story />
+      </div>
+    ),
+  ],
+  args: {
+    itemClassName: "w-11/12 basis-auto",
```
sauldom102, line 66: This seems a bit hacky, we should allow to set a couple things via props:
- Whether or not we should show a little bit of the next item from next slide, this could be a boolean prop.
- Estimated number of columns for each endpoint, maybe something like:
```
{
    0: 3,
    750: 6,
    1024: 8
}
```
So that it's 3 items when size is <750px, 6 when size is 1024px, ...

Another option would be to have a simple `numberOfColumns` prop accepting a number and we have all that logic of different number of items per breakpoint outside of this component.

- `allowToGrowInItemWithSize={300}` meaning that, when specified, our slide can grow in items dynamically when a new item with 300px fits there

lib/experimental/Navigation/Carousel/index.tsx
```diff
@@ -0,0 +1,85 @@
...
+  itemClassName?: string
+}
+
+export const Carousel = ({
+  children,
+  showArrows = true,
+  showDots = true,
+  autoplay = false,
+  delay = 3000,
+  itemClassName = "sm:basis-1/2 md:basis-1/3 lg:basis-1/4",
```
sauldom102, line 29: I believe we shouldn't have a default size for the items, we should always be explicit when using this component with the amount of columns/items per slide we want

dani-moreno: Left at 1 item by default, and other breakpoints can be defined later when using the component.

sauldom102 (approved):
Very good work man!

PR #869, opened 2024-11-20
Do not allow more than one line in indicator item

sauldom102:
## Description

Do not allow more than one line in indicator item.

## Screenshots

Before:

Screenshot 2024-11-20 at 14 25 13

After:

Screenshot 2024-11-20 at 14 24 30


dani-moreno (approved):
Not a huuuge fan of truncating it, but let's go with it for now until we have a better approach!

PR #872, opened 2024-11-20
Add PostEvent

dani-moreno:
## Description

`PostEvent` is one of the pieces of a `Post` in Communities. It's displayed when the post is an event, showcasing relevant information, like the date, title, and so on.

The component accepts an optional image, so we can show it if the user uploaded any. You will be greeted with a random cat every time you visit the story in Storybook, how cool is that?!

## Screenshots

image

_Default look, with an image_

image

_No image_

image

_Skeleton loading state_

image

_It also looks pretty dope in dark mode, yay!_

image

_It also truncates text when needed_

### Figma Link

[Link to Figma Design](https://www.figma.com/design/pZzg1KTe9lpKTSGPUZa8OJ/Web-Components?node-id=2378-107879&t=enwdXpZ3onkWtwD8-4)

---

## Type of Change


lib/experimental/Information/Communities/Post/PostEvent/index.tsx
```diff
@@ -0,0 +1,71 @@
...
+
+type PostEventProps = {
+  title: string
+  image?: string
+  place: string
+  date: {
+    hour: string
+    day: number
+    month: string
+  }
```
sauldom102, line 13: I think we should agree on a format to pass these dates to our components in a simpler way, maybe like "YYYY-MM-DD" which is how we have it in backend, and then we just parse it inside and extract whatever we want

dani-moreno: Good point, I have no idea how we retrieve this info, so this is gold for me, haha. Maybe this is a topic to discuss on Slack. Do you mind if I open a thread there?

lib/experimental/Information/Communities/Post/PostEvent/index.tsx
```diff
@@ -0,0 +1,71 @@
...
+          <span className="flex min-w-0 flex-row gap-1 text-f1-foreground-secondary">
+            <span>{date.hour}</span>
+            <span>·</span>
+            <span className="truncate">{place}</span>
+          </span>
+        </div>
+        <div className="shrink-0">
+          <DateAvatar day={date.day} month={date.month} />
+        </div>
+      </div>
```
sauldom102, line 46: We should probably reuse this from a `CompanyEvent` component that we have declared outside here, what do you think?

https://www.figma.com/design/pZzg1KTe9lpKTSGPUZa8OJ/Web-Components?node-id=1687-16697&t=2XcYlDJHJR7XbBZ4-4

lib/experimental/Information/Communities/Post/PostEvent/index.tsx
```diff
@@ -0,0 +1,71 @@
...
+  image,
+  place,
+  date,
+}: PostEventProps) => {
+  return (
+    <div className="flex w-full flex-col gap-1 rounded-xl border border-solid border-f1-border-secondary bg-f1-background-inverse-secondary p-1 shadow">
+      {image && (
+        <img
+          src={image}
+          alt={title}
```
rubenmoya, line 27: Since this is just a presentational image we can omit the "alt". Using `alt=""` will make the image be `role="presentation"`, and it will be ignored by assistive technologies. Since we already have the title in text there is no need to repeat it here

lib/experimental/Information/Communities/Post/PostEvent/index.tsx
nlopin: Let's also consider getting a `Date` object instead and get all the dates from it

lib/experimental/Information/Communities/Post/PostEvent/index.tsx
```diff
@@ -0,0 +1,71 @@
...
+export const BasePostEvent = ({
+  title,
+  image,
+  place,
+  date,
+}: PostEventProps) => {
+  return (
+    <div className="flex w-full flex-col gap-1 rounded-xl border border-solid border-f1-border-secondary bg-f1-background-inverse-secondary p-1 shadow">
+      {image && (
+        <img
```
nlopin, line 25: I think it is fair to consider images here as non-critical resource and apply [`loading=lazy`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/img#loading)

lib/lib/date.ts
```diff
@@ -0,0 +1,13 @@
+import { format } from "date-fns"
+
+export function formatTime(date: Date) {
+  return format(date, "HH:mm")
+}
```
sauldom102, line 5: We should use `p` which is a localized version of the time to use 12h or 24h version when necessary

Screenshot 2024-12-10 at 11 01 08

lib/lib/date.ts
```diff
@@ -0,0 +1,13 @@
...
+
+export function formatMonth(date: Date) {
```
sauldom102, line 7: We should better name this something like `getAbbrMonth`

PR #892, opened 2024-11-25
Only partially match when there's subroutes involved

josepjaume:
## Description

This fixes our default link matching logic to prevent URLs like `/performance_reviews` to match `/performance`.

*None*

### Figma Link

*None*

---

## Type of Change

### DIFF
--- a/lib/lib/linkHandler.spec.tsx
+++ b/lib/lib/linkHandler.spec.tsx
@@ -33,7 +33,7 @@ describe("useLink", () => {
     )
   })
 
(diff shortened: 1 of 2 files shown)


nlopin (approved):
I approve just for the sake of unblocking, but this code cries for Unit Tests

josepjaume:
There are unit tests but not one covering this specific scenario. Now we are!

larizzatg:
Thanks a lot :)

PR #906, opened 2024-11-28
Dropdown and Select items hover zone

dani-moreno:
## Description

_Here's a change that no one asked for and does not drive revenue, yay! But hey we are committed to quality._

This change extends the _hoverable_ zone of each item in a Select and Dropdown to cover the entire row, not just the visible hover background. This makes it easier to click on an item that you, as a user, are visually hovering over, even if you're not exactly on top of it.

_PS: Linear or MacOS use this trick, so if your cursor is on the dropdown, an item is focused, no matter if you're not exactly on top of it._

image

---

## Type of Change


sauldom102 (approved):
Awesome! 🤩

PR #912, opened 2024-11-29
Make blur from private box more restrictive

sauldom102:
## Description

Make blur from private box more restrictive.

- Ensure that we do not "expose" information to screen readers.
- Do not allow text selection when that's enabled.
- Make blur a little less aggressive.


dani-moreno (approved):
Perfect! It looks a lot better with the less aggressive blur!

PR #915, opened 2024-11-29
Update privacy area styling

dani-moreno:
## Description

Makes the blur of a privacy area [a bit more contained](https://factorialteam.slack.com/archives/C07LQSJ5UQY/p1732876277064169?thread_ts=1732870097.350609&cid=C07LQSJ5UQY) in its area itself.

## Screenshots

https://github.com/user-attachments/assets/08c97230-a7fb-4022-a05d-85ae3ce9b3ee

---

## Type of Change

### DIFF
--- a/lib/experimental/Utilities/PrivateBox/index.tsx
+++ b/lib/experimental/Utilities/PrivateBox/index.tsx
@@ -5,5 +5,15 @@ import { FC, PropsWithChildren } from "react"
 export const PrivateBox: FC<PropsWithChildren> = ({ children }) => {
   const { enabled } = usePrivacyMode()
 
-  return <div className={cn(enabled && "blur-md", "inline")}>{children}</div>
+  return (
+    <div
+      className={cn(
+        "inline-flex",
+        enabled &&
+          "overflow-hidden rounded-sm ring-1 ring-inset ring-f1-border-secondary"
+      )}
+    >
+      <div className={cn(enabled && "opacity-30 blur")}>{children}</div>
+    </div>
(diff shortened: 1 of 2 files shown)


sauldom102 (changes requested):
This is not as straightforward, we are using this to hide chart y axis too

dani-moreno:
> This is not as straightforward, we are using this to hide chart y axis too

Ah, damn, thanks for the heads-up! Is there a story where I could test it, or any way to test it in Storybook?

sauldom102 (commented):
I mean this:

Screenshot 2024-11-29 at 15 37 04

lib/experimental/Utilities/PrivateBox/index.tsx
```diff
@@ -1,16 +1,30 @@
...
+          "select-none overflow-hidden rounded-sm ring-f1-border-secondary"
+      )}
       aria-hidden={enabled}
     >
-      {children}
+      <motion.div
+        className="h-full w-full"
+        animate={{
+          opacity: enabled ? 0.3 : 1,
+          filter: enabled ? "blur(5px)" : "blur(0px)",
```
sauldom102, line 22: We should increase this a little bit more, in the screenshot that I attached you could see the number

dani-moreno: Went with a different approach as we discussed, not using blur at all, which is prone to errors, and just hiding the text instead.

sauldom102 (approved):

PR #920, opened 2024-11-29
Force color on header text

dani-moreno:
## Description

Force the token `foreground` to the text on the page header, so it takes that color instead of the default Gamma one.

## Screenshots

image

_This is how it is looking now, with a grey Gamma color._

---

## Type of Change


sauldom102 (approved):
LGTM!

PR #925, opened 2024-12-02
new Carousel for widget strip

Rogermax:
## Description
Improve Carousel component.
This component is need it for new Finance Overview. We can't use Dashboard, as we want to classify widgets per sections, So we want to use N widgetStrips. The problem is widgetStrip expands Widgets too fit all space. We don't want this behaviour. We need to have widgets responsive and let know the user that there are more widgets to see (with Arrow buttons)

## Screenshots

Simplified PR, just improving current Carousel

https://github.com/user-attachments/assets/8f2f606b-9c95-45b8-8218-2e7a11b73284

## Type of Change

- [] New experimental component

**Responsiveness**

  - [x] Verified the component works across various screen sizes

**Testing**

  - [x] Component includes unit tests with sufficient coverage

**Accessibility**

  - [ ] Accessibility standards meet at least AA level requirements
  - [ ] All interactive elements are keyboard-navigable and have focus states
  - [ ] Proper ARIA attributes are used where needed

### DIFF
--- a/lib/experimental/Navigation/Carousel/index.tsx
+++ b/lib/experimental/Navigation/Carousel/index.tsx
@@ -73,7 +73,7 @@ export const Carousel = ({
       onMouseEnter={autoplay ? handleMouseEnter : undefined}
       onMouseLeave={autoplay ? handleMouseLeave : undefined}
     >
-      <div className="flex flex-col gap-3">
+      <div className="flex h-full flex-col gap-3">
         <div className="relative">
           <CarouselContent showFade={showFade}>
             {React.Children.map(childrenArray, (child, index) => (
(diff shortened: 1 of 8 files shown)


lib/experimental/Navigation/Carousel/index.tsx
```diff
@@ -34,22 +35,23 @@ export const Carousel = ({
...
+  showFade = false,
 }: CarouselProps) => {
   const childrenArray = React.Children.toArray(children)
 
-  const plugin = React.useRef(
-    autoplay ? Autoplay({ delay: delay, stopOnInteraction: true }) : undefined
-  )
+  const plugin = useMemo(() => {
+    return autoplay ? Autoplay({ delay, stopOnInteraction: true }) : undefined
+  }, [autoplay, delay])
```
Rogermax, line 44: We need this to reactivate autoplay if the prop autoplay change from `false` to `true`

nlopin:
Roger, we're currently busy with the migration and cannot invest time. Consider moving this code to the main repo and have a Carousel component there. We can return to it at the end of December or early January

Rogermax:
Closed becasuse it has been splited in two:
* #961
* #962

PR #934, opened 2024-12-03
Allow subitems on page header actions

josepjaume:
## Description

Allows for items in the page's subnavigation.

image

### Figma Link

*None*

---

## Type of Change

### DIFF
--- a/lib/experimental/Navigation/Header/PageHeader/index.stories.tsx
+++ b/lib/experimental/Navigation/Header/PageHeader/index.stories.tsx
@@ -33,7 +33,16 @@ export const Default: Story = {
       {
         label: "More options",
         icon: EllipsisHorizontal,
-        href: "/recruitment/settings",
+        actions: [
+          {
+            label: "Profile",
+            href: "/recruitment/profile",
+          },
+          {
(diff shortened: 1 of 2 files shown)


lib/experimental/Navigation/Dropdown/index.tsx
```diff
@@ -61,7 +61,7 @@ const DropdownItem = ({ item }: { item: DropdownItem }) => {
   )
 
   const itemClass = cn(
-    "flex items-start gap-1.5 w-full",
+    "cursor-pointer flex items-start gap-1.5 w-full",
```
dani-moreno, line 64: With the introduction of [this PR](https://github.com/factorialco/factorial-one/pull/914), I think this shouldn't be needed if these header actions are links, right?

josepjaume: LOL we fixed it at the same time. I was debating myself whether to always make it `pointer` regardless of it being a link or not. Are we sure users know the difference? Anyways this will always be a link in this use case, so I imagine no harm done.

dani-moreno (approved):

PR #938, opened 2024-12-04
Add Omnibutton

dani-moreno:
## Description

Adds `Omnibutton`, a floating button on the bottom right corner of the screen with a few links. The links come from Factorial, and this component just accepts a list of options as our [Dropdown](https://one.factorial.dev/?path=/docs/experimental-navigation-dropdown--documentation) component, and display them.

(The idea is to conditionally display this component in Factorial, if the F1 layout is enabled)

https://github.com/user-attachments/assets/02f0c6e0-0d8e-43dc-ad9f-7ecbc89881b6

---

## Type of Change


lib/experimental/Navigation/Omnibutton/index.tsx
```diff
@@ -0,0 +1,62 @@
...
+  description?: string
+  href?: string
+  onClick?: () => void
+  icon?: string
+  target?: string
+}
+
+interface OmniButtonProps {
+  label: string
+  options: Option[]
```
nlopin, line 17: do we always have options or there is also one action scenario?

dani-moreno: As far as I understand, yes, [there is always options](https://github.com/factorialco/factorial/blob/316becfcad90aaf7b10322811cc7c63f70a7bc49/frontend/src/modules/onboarding/components/OmniButton/getOptions.ts#L4).

PR #962, opened 2024-12-11
Let widgets grow if container lets them

Rogermax:
## Description

For that PR: https://github.com/factorialco/factorial-one/pull/961 we need the widget to expand to all available height, now this is not posible as there are som flex flex-col without a h-full that prevents that.

Problem:
Screenshot 2024-12-11 at 15 01 46

With this PR the chart fills the space:
Screenshot 2024-12-11 at 15 02 45

## Type of Change

- [x] Promote component from experimental to stable

### DIFF
--- a/lib/experimental/Navigation/Carousel/index.stories.tsx
+++ b/lib/experimental/Navigation/Carousel/index.stories.tsx
@@ -1,14 +1,17 @@
+import { BarChartProps } from "@/components/Charts/BarChart"
+import BarChartStory from "@/components/Charts/BarChart/index.stories"
+import { BarChartWidget } from "@/experimental/Widgets/Charts/BarChartWidget"
 import { Placeholder } from "@/lib/storybook-utils/placeholder"
 import { Meta, StoryObj } from "@storybook/react"
 import { Carousel } from "."
 
 const meta: Meta<typeof Carousel> = {
   component: Carousel,
-  parameters: {
-    layout: "fullscreen",
-  },
   argTypes: {
     autoplay: { control: "boolean" },
+    showDots: { control: "boolean" },
+    showArrows: { control: "boolean" },
+    showPeek: { control: "boolean" },
   },
 }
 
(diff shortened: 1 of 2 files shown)


sauldom102 (approved):

nlopin:
Roger, do you want to stretch the content of the widget, or the widget itself?

Rogermax:
> Can we always stretch to full height and avoid having a prop?

We could, in fact I did it that way the first time... but the PageLayout and some other places breaks a lot.. so I was not confident on changing it for all cases. If you tell me that PageLayout is not used anywhere I can change it for all cases

nlopin:
> > Can we always stretch to full height and avoid having a prop?
> 
> We could, in fact I did it that way the first time... but the PageLayout and some other places breaks a lot.. so I was not confident on changing it for all cases. If you tell me that PageLayout is not used anywhere I can change it for all cases

It is almost not used now. Let's try and see the outcome?

nlopin:
I didn't noticed the automerge 🤦

Rogermax:
> I didn't noticed the automerge 🤦

I'll make a new PR now to make this without the prop for all

lib/experimental/Navigation/Carousel/index.stories.tsx
```diff
@@ -85,9 +118,59 @@ export const SneakPeek: Story = {
...
+    autoplay: { control: "boolean" },
+    showDots: { control: "boolean" },
+    showArrows: { control: "boolean" },
+    showPeek: { control: "boolean" },
+  },
+  args: {
+    showArrows: true,
+    children: SLIDES_RANDOM,
+    showPeek: true,
+    showDots: false,
```
josepjaume, line 174: Do we really want to have all these props? What's the use case where we want to show arrows or not? Can't we just figure that out automatically? Same for `showPeek` or `showDots`.

Every time we add more choices to the users we create more potential bugs and inconsistencies. We need to learn from our mistakes and keep the interfaces as small as possible.

lib/experimental/Widgets/Widget/index.tsx
```diff
@@ -46,6 +46,7 @@ export interface WidgetProps {
     text: string
     variant: StatusVariant
   }
+  fullHeight?: boolean
```
josepjaume, line 49: I think the right way to make things full height is for the container to actually dictate that instead of the children. We can easily do:

```
<div class="flex flex-row items-stretch">
  <Widget ... />
</div>
```

What do you think?

PR #966, opened 2024-12-12
Add single tab support in `Tabs` component

dani-moreno:
## Description

We have cases (usually due of permissions) were we only display a single tab. In these cases, we want to still display them, but make them look more like a title, rather than a tab.

This PR adds that option. When there is only a single tab in the `tabs` prop, it will be displayed with a different style, and with no link.

## Screenshots

image

### Links

[Link to Figma Design](https://www.figma.com/design/pZzg1KTe9lpKTSGPUZa8OJ/Web-Components?node-id=323-16471&t=xvQ6ftl11zlH8Kuv-4)
[JIRA ticket](https://factorialmakers.atlassian.net/browse/FCT-22056)

---

## Type of Change

### DIFF
--- a/lib/experimental/Navigation/Tabs/index.tsx
+++ b/lib/experimental/Navigation/Tabs/index.tsx
@@ -28,7 +28,7 @@ export const BaseTabs: React.FC<TabsProps> = ({ tabs, secondary = false }) => {
   return (
     <TabNavigation
       secondary={secondary}
-      asChild={!isSingleTab}
+      asChild
       aria-label={secondary ? "primary-navigation" : "secondary-navigation"}
     >
       {isSingleTab ? (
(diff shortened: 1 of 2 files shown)


lib/experimental/Navigation/Tabs/index.tsx
```diff
@@ -30,19 +30,27 @@ export const BaseTabs: React.FC<TabsProps> = ({ tabs, secondary = false }) => {
...
-          active={activeTab?.href === props.href}
-          href={props.href}
-          secondary={secondary}
-          asChild
-        >
-          <Link role="link" {...props}>
-            {label}
-          </Link>
+      {tabs.length === 1 ? (
+        <TabNavigationLink disabled secondary={secondary} className="p-0">
```
nlopin, line 34: Accessibility wise it is not great. It is disabled, hence will be pronounced as a disabled element. However I'm not sure it can be selected at all. 

It is also not copyable, it is counterintuitive when you deal with just a text

dani-moreno: Ah, you're right, it's not semantically correct either, it's not really disabled. Changed it to a li element, so it allows text selection/copying.

PR #970, opened 2024-12-12
Add credit card app icon

lochiego:
Is present in the [Design System](https://www.figma.com/design/2odK5veKs0u9bFDt5bp7sU/Icons?node-id=10-119&node-type=frame&t=1VYOiLX9N8JsngaV-0) but was missing in the repo.

Screenshot 2024-12-12 at 17 00 45


dani-moreno (approved):
Hey, hope you don't mind, but I updated the icon! We need the icons not flattened since we change the stroke width depending on the size :-)

lochiego:
Don't mind at all, I had just exported from Figma and am not sure how best to do it otherwise. Given the flattening, I'd be remiss if I didn't mention that the hearthold app icon/Benefits module svgs recently added are probably flattened.

PR #972, opened 2024-12-13
Get rid of glitch in HomeLayout component

sauldom102:
## Description

Get rid of glitch in HomeLayout component.

Before:

https://github.com/user-attachments/assets/3f64a188-ff00-4cb3-8e59-021a1bc4f75a

After:

https://github.com/user-attachments/assets/e95b4e1f-c12a-448a-9584-14dac239e035

### DIFF
--- a/lib/experimental/PageLayouts/HomeLayout/index.tsx
+++ b/lib/experimental/PageLayouts/HomeLayout/index.tsx
@@ -6,14 +6,14 @@ import {
   useImperativeHandle,
   useRef,
 } from "react"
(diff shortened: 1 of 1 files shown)


nlopin:
what's the difference?

nlopin (approved):

sauldom102:
> what's the difference?

In the first video you can observe how in a specific size it starts to do very weird things, this is because we were using the width of the two different containers rendered to get the available size, instead of using a fixed "reference" to obtain this value

PR #982, opened 2024-12-17
Expose forms

josepjaume:
## Description

We recently got interest from some people to start iterating on forms and help us push things forward. We're a small team with a huge surface to cover, so I think it's fantastic that we get some motivated extra hands to help us with some of the most difficult aspects of the design system.

This exports missing components from forms.

Related conversation: https://factorialteam.slack.com/archives/C06QT46115K/p1734348724908599

*None*

### Figma Link

*None*

---

## Type of Change

### DIFF
--- a/lib/experimental/Forms/Fields/exports.ts
+++ b/lib/experimental/Forms/Fields/exports.ts
@@ -1,23 +1,6 @@
-import { Component } from "@/lib/component"
-import { Input as RawInput } from "./Input"
-import { NumberInput as RawNumberInput } from "./NumberInput"
 export * from "./Calendar"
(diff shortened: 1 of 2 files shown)


nlopin:
I am a bit anxious with exposing an API we haven't discussed at all, but let's go. Let's see what we get

nlopin (approved):

josepjaume:
Well the rules still apply in the sense that we should have a defined group of people using things in `experimental`. It's true that we can't track usage and things have historically gone out of control 👀 . I'll try to be very explicit about this rule when discussing this with people, is that ok?

PR #993, opened 2024-12-19
Unify label styles

sauldom102:
## Description

Unify label styles.

## Screenshots

Screenshot 2024-12-19 at 16 44 27

Screenshot 2024-12-19 at 16 48 34


dani-moreno (approved):
Cool! 💅

nlopin:
Is it fine we have 5 different font configurations? 

I think the fonts for the legend in the LineChart and in the ProgressBar should be the same, but they are slightly different.

sauldom102:
> Is it fine we have 5 different font configurations?
> 
> I think the fonts for the legend in the LineChart and in the ProgressBar should be the same, but they are slightly different.

Probably we should unify those as well, but we need some confirmation from design, I'll update those in a next PR

PR #995, opened 2024-12-19
`CompanySelector` notification & additional options

dani-moreno:
## Description

We are expanding the `CompanySelector` to show additional options below the list of companies. This way, we can have options like "Add new workplace", or the temporary "Upload company avatar" option needed for the new workplace logo.

image

This also adds the option to show a notification badge in the avatar of the `CompanySelector`, to let the user know there is something in the menu that requires their attention.

### Figma Link

[Link to Figma Design](https://www.figma.com/design/KkWrl3CD4YqDpMuT0cTHcC/Company-avatar?node-id=3001-110411&p=f&t=mJT5uuGfzFQq09L9-0)

---

## Type of Change


lib/experimental/Navigation/Sidebar/CompanySelector/index.tsx
```diff
@@ -40,10 +51,13 @@ export function CompanySelector({
     )
   }
 
-  if (companies.length === 1) {
+  if (companies.length + (additionalOptions?.length || 0) === 1) {
```
nlopin, line 54: what do you think about setting `additionalOptions` default value to `[]`? it will eliminate all extra checks

lib/experimental/Navigation/Sidebar/CompanySelector/index.tsx
```diff
@@ -83,14 +103,32 @@ const Selector = ({
           "aria-label": `${company.name} logo`,
         } satisfies AvatarVariant,
       })),
-    [companies]
+      ...(additionalOptions ? ["separator" as const] : []),
+      ...(additionalOptions?.map((option) => ({
```
nlopin, line 107: can't we spread the whole `additionalOptions` object as is? 
```
...additionalOptions
```

lib/experimental/Navigation/Sidebar/CompanySelector/index.tsx
```diff
@@ -18,13 +19,23 @@ export type CompanySelectorProps = {
   selected?: string
   onChange: (value: string) => void
   isLoading?: boolean
+  notification?: boolean
```
nlopin, line 22: maybe `withNotification`? or `withBadge`? 

`notification` hints me I can provide a message, which I cannot

nlopin (approved):

lib/experimental/Navigation/Sidebar/CompanySelector/index.tsx
dani-moreno: Right! That's a lot simpler, thanks 🙏

lib/experimental/Navigation/Sidebar/CompanySelector/index.tsx
dani-moreno: Yes! I went with `withNotification` because using `badge` might imply that any badge can be used, but in this case, it's specifically tied to the notification circle icon.

PR #1011, opened 2025-01-03
Create CommunityPost component

sauldom102:
## Description

Create new CommunityPost component.

## Screenshots

Screenshot 2025-01-03 at 12 18 12

Screenshot 2025-01-03 at 12 18 29

Screenshot 2025-01-03 at 12 18 40

Screenshot 2025-01-03 at 12 19 29

Screenshot 2025-01-03 at 13 57 36

Screenshot 2025-01-03 at 12 18 59

### Figma Link

https://www.figma.com/design/pZzg1KTe9lpKTSGPUZa8OJ/Web-Components?node-id=2333-12691&t=9eBZI2wvIjDCe1Tj-4

### DIFF
--- a/lib/experimental/Information/Communities/Post/CommunityPost/index.tsx
+++ b/lib/experimental/Information/Communities/Post/CommunityPost/index.tsx
@@ -4,6 +4,7 @@ import { PersonAvatar } from "@/experimental/Information/Avatars/PersonAvatar"
 import { Reactions, ReactionsProps } from "@/experimental/Information/Reactions"
 import { Dropdown } from "@/experimental/Navigation/Dropdown"
 import { EllipsisHorizontal } from "@/icons/app"
+import { getAgo } from "@/lib/date"
 import { withSkeleton } from "@/lib/skeleton"
 import { Skeleton } from "@/ui/skeleton"
 import { PostEvent, PostEventProps } from "../PostEvent"
@@ -19,7 +20,7 @@ type CommunityPostProps = {
     title: string
     url: string
   }
-  ago: string
+  createdAt: Date
 
   title: string
   description?: string
@@ -45,7 +46,7 @@ type CommunityPostProps = {
 export const BaseCommunityPost = ({
   author,
   group,
-  ago,
+  createdAt,
   title,
   description,
   onClick,
@@ -59,6 +60,8 @@ export const BaseCommunityPost = ({
   const countersDisplay = [counters.views, counters.comments]
     .filter(Boolean)
     .join(" · ")
+
+  const ago = getAgo(createdAt)
 
   return (
     <div
--- a/lib/experimental/Information/Communities/Post/CommunityPost/index.tsx
+++ b/lib/experimental/Information/Communities/Post/CommunityPost/index.tsx
@@ -190,9 +190,8 @@ export const CommunityPostSkeleton = ({
         <Skeleton className="h-2.5 w-18 rounded-2xs" />
       </div>
       <Skeleton className="mt-3.5 h-3.5 w-1/5 rounded-2xs" />
-      <div className="mt-3 flex flex-col justify-around gap-3 py-2">
-        <Skeleton className="h-2.5 w-1/2 rounded-2xs" />
-        <Skeleton className="h-2.5 w-2/3 rounded-2xs" />
+      <div className="mt-3">
+        <PostDescription.Skeleton />
       </div>
       {withImage && !withEvent && (
         <div className="mt-3 aspect-video w-full overflow-hidden rounded-xl md:w-2/3">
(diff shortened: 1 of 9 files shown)


nlopin:
I want to spin the design discussion. Why do we force users to click a post to read it? Why can't I see it full? I'm curious, how an average community post in factorial looks like

lib/experimental/Information/Communities/Post/CommunityPost/index.tsx
```diff
@@ -0,0 +1,217 @@
...
+type CommunityPostProps = {
+  author: {
+    firstName: string
+    lastName: string
+    avatarUrl?: string
+    url: string
+  }
+  group: {
+    title: string
+    url: string
```
nlopin, line 22: shouldn't URLs be optional?

lib/experimental/Information/Communities/Post/CommunityPost/index.tsx
```diff
@@ -0,0 +1,217 @@
...
+  }
+  createdAt: Date
+
+  title: string
+  description?: string
```
nlopin, line 27: Maybe `content` is a better name. Also if we expect an HTML string there, let's create an alias for this: `type HTMLString = string` and use it here for clarity

lib/experimental/Information/Communities/Post/CommunityPost/index.tsx
```diff
@@ -0,0 +1,217 @@
...
+  onClick: () => void
+  imageUrl?: string
+
+  event?: PostEventProps
```
nlopin, line 31: I'd consider having another component for event posts. There is a high change it will grow in a different direction from the text-and-picture post

lib/experimental/Information/Communities/Post/PostDescription/TextEditorTheme.css
nlopin, line 1: We shouldn't implicitly couple the implementation with third party dependency. It should not leave in the factorial one I think

lib/experimental/Information/Communities/Post/CommunityPost/index.tsx
sauldom102: I wouldn't consider having two different components at least for now, it's only one in Figma

sauldom102:
> I want to spin the design discussion. Why do we force users to click a post to read it? Why can't I see it full? I'm curious, how an average community post in factorial looks like

That's a very interesting discussion to have, let's talk about it internally with the design team!

lib/experimental/Information/Communities/Post/CommunityPost/index.tsx
sauldom102: In this case I would better add an `onClick` that would automatically filter the whole feed using that group

lib/experimental/Information/Communities/Post/PostDescription/TextEditorTheme.css
sauldom102: The thing is that existing posts and new ones that are created in Factorial use these classes and we should use them to give them some styles

lib/experimental/Information/Communities/Post/CommunityPost/index.tsx
nlopin: ok, let's keep an eye on it then

lib/experimental/Information/Communities/Post/PostDescription/TextEditorTheme.css
nlopin: we will rely on the css provided by factorial. it is their responsibility to create the content and manage its style

With the proposed solution we will have to react on any class changes in Factorial content editor, creating a mandatory work to do on our side before the release. We must be agnostic to such issues because it will create time pressure which is bad for any design system.

I propose to keep this styles in Storybook files just to keep things visually clear, but document that the post only provide a shell for the content. The content style itself is the responsibility of the user

sauldom102: Then we should have control on Factorial One over what classes these posts can contain, we should provide these classes with an editor component as well that is on our side, otherwise we would be missing so much control over how these posts look.

We even have a `PostDescription` component in Figma that mandates how mentions should look.

Screenshot 2025-01-03 at 17 09 24

I'd say we should also revisit these styles to guarantee that the colors used are part of the tokens and not some random ones.

lib/experimental/Information/Communities/Post/CommunityPost/index.tsx
sauldom102: Given that in Figma we have the `PostDescription` component I would leave it here as `description`, and `content` in `PostDescription` component not to be redundant there

nlopin (approved):

PR #1017, opened 2025-01-07
Add autogenerated tests and stories with cursor

josepjaume:
## Description

Here's a few one-shot tests generated by cursor just by using our `.cursorrules` file and telling it "can you write some tests for this file?".

image

### Figma Link

*None*
---

## Type of Change

### DIFF
--- a/lib/components/Actions/Button/index.stories.tsx
+++ b/lib/components/Actions/Button/index.stories.tsx
@@ -43,7 +43,6 @@ const meta = {
     },
     label: {
       control: "text",
-      description: "Button text content",
     },
     icon: {
       control: "boolean",
@@ -51,11 +50,9 @@ const meta = {
     },
     disabled: {
       control: "boolean",
-      description: "Disabled state of the button",
     },
     loading: {
       control: "boolean",
-      description: "Loading state of the button",
     },
     hideLabel: {
       control: "boolean",
--- a/lib/components/Actions/Button/index.stories.tsx
+++ b/lib/components/Actions/Button/index.stories.tsx
@@ -3,6 +3,8 @@ import { fn } from "@storybook/test"
 
 import Add from "@/icons/app/Add"
 import Archive from "@/icons/app/Archive"
+import Delete from "@/icons/app/Delete"
+import Save from "@/icons/app/Save"
 import { Button } from "."
 
 const meta = {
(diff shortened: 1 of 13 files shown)


lib/components/Actions/Button/index.stories.tsx
```diff
@@ -19,81 +21,179 @@ const meta = {
...
+      control: "text",
+      description: "Button text content",
+    },
+    icon: {
+      control: "boolean",
+      description: "Whether to show an icon",
+    },
+    disabled: {
+      control: "boolean",
+      description: "Disabled state of the button",
```
nlopin, line 54: I am counter such obvious descriptions. It is like useless javadocs in Java:

```
/**
/* sets value
/**
public void setValue(int x) { ... }
```

image

lib/components/Actions/Button/index.test.tsx
```diff
@@ -28,3 +29,77 @@ test("Button is temporarily disabled when onClick is a promise until the promise
   await new Promise((resolve) => setTimeout(resolve, 100))
   expect(button.attributes.getNamedItem("disabled")).toBeNull()
 })
+
+test("Button renders with different variants", () => {
```
nlopin, line 33: We check visual things with visual tests. Why do we need to test it here?

josepjaume: Probably not? I'll look into them tomorrow. But hey, just two comments in a 641 lines PR that was generated in 10 minutes? Not bad.

josepjaume: Removed this text and the next one, which were too presentational. I've added a note on the `.cursorrules` file so future tests are about behavior instead :)

lib/components/Actions/Link/index.test.tsx
```diff
@@ -0,0 +1,71 @@
...
+    render(<Link href="/test">Click me</Link>)
+    expect(screen.getByText("Click me")).toBeInTheDocument()
+  })
+
+  it("applies link variant by default", () => {
+    render(<Link href="/test">Click me</Link>)
+    expect(screen.getByRole("link")).toHaveClass("text-f1-link")
+  })
+
+  it("applies text variant correctly", () => {
```
nlopin, line 22: This can be dropped, we have visual test for that

lib/experimental/Information/Shortcut/index.tsx
```diff
@@ -37,12 +37,17 @@ function Shortcut({ keys, variant }: ShortcutProps) {
         return (
           <kbd
             key={index}
+            role="term"
```
nlopin, line 40: I don't think we need this role on `kbd` elements. It is used to mark [definitions](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/term_role):
> The term role can be used for a word or phrase with an optional corresponding [definition](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/definition_role). It is equivalent in semantics to the HTML [<dfn>](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dfn) element and the definition term ([<dt>](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dt)) element within a definition list ([<dl>](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dl) ).

This is no definition

lib/experimental/Information/Shortcut/index.tsx
```diff
@@ -37,12 +37,17 @@ function Shortcut({ keys, variant }: ShortcutProps) {
...
             className={cn(
               shortcutVariants({ variant }),
               isIcon ? "w-5 px-0.5" : "min-w-5 px-1"
             )}
           >
-            {isIcon ? <Icon icon={iconMap[lowerKey]} size="sm" /> : key}
+            {isIcon ? (
+              <Icon data-testid="cmd-icon" icon={iconMap[lowerKey]} size="sm" />
```
nlopin, line 47: I'd prefer to avoid using `data-testid` where possible. Do you think we can avoid it here?

lib/experimental/Navigation/Tabs/index.spec.tsx
```diff
@@ -0,0 +1,69 @@
+import { render, screen } from "@testing-library/react"
+import { ComponentProps, PropsWithChildren } from "react"
+import { describe, expect, it, vi } from "vitest"
+import { BaseTabs, TabsSkeleton } from "."
+
+// Mock the linkHandler module
+vi.mock("@/lib/linkHandler", () => ({
```
nlopin, line 7: can't we reuse the mock from `Link/__mocks__`? it looks similar

nlopin (approved):

josepjaume:
Hahaha I now see that I shifted all the workload to the reviewer. We'll keep iterating on cursorrules so the tests they generate are better. In any case, this is clearly speeding some things up.

lib/experimental/Information/Shortcut/index.tsx
josepjaume: image

josepjaume: image

josepjaume: image

This is freaking amazing. I wonder why isn't GitHub adding this to PR comments like yours.

lib/experimental/Navigation/Tabs/index.spec.tsx
josepjaume: image

nlopin: if the component is self-contained, why do we mock the link? 🤯 I'm fine to have it merged though

PR #1018, opened 2025-01-07
Application frame a11y improvements

nlopin:
Among several small improvements, I have two big ones:

- the sidebar is not selectable anymore when collapsed
- when user collapses the sidebar, we move the focus to the open button and vice versa

https://github.com/user-attachments/assets/95afe838-9322-417d-9b1d-b355856621b2


eliseo-juan:
Cool!

PR #1021, opened 2025-01-08
Create ClockInControls component

sauldom102:
## Description

Create `ClockInControls` component.

## Screenshots

Screenshot 2025-01-08 at 17 13 33

Screenshot 2025-01-08 at 17 13 42

Screenshot 2025-01-08 at 17 13 51

Screenshot 2025-01-08 at 17 14 01

### Figma Link

https://www.figma.com/design/060q78tnGIGaRc1sxecTvV/Home-Web?node-id=811-15934&t=fxiFd5JhpwgSfCli-4


lib/icons/app/SolidPause.tsx
```diff
@@ -0,0 +1,33 @@
+import * as React from "react"
```
dani-moreno, line 1: Are these icons auto-generated or did you made them manually?

If you place the svgs under `assets/icons/app` and then `npm run generate-app-icons`, they are created automatically through these svgs and we get to keep the original svgs, which is good if we want to update them in the future.

sauldom102: Ooh I did them manually, I'll better do it the way you explain it, thanks!!

sauldom102: Done! ✅

lib/experimental/Widgets/Content/ClockIn/ClockInGraph/index.spec.tsx
```diff
@@ -0,0 +1,22 @@
...
+    render(<ClockInGraph />)
+    expect(screen.getByText("00:00")).toBeInTheDocument()
+  })
+
+  it("renders with custom dates", () => {
+    const data = [
+      {
+        from: new Date("2024-03-20T09:00:00"),
+        to: new Date("2024-03-20T17:00:00"),
+        variant: "clocked-in" as const,
```
nlopin, line 16: I think we should use `satisfies Typename` or type the `data` variable instead of forcing `as const`

lib/experimental/Widgets/Content/ClockIn/ClockInGraph/index.tsx
```diff
@@ -0,0 +1,149 @@
...
+        value: Math.abs(remainingMinutes),
+        color: CLOCK_IN_COLORS.overtime,
+      }
+    }
+
+    return {
+      value: 1,
+      color: CLOCK_IN_COLORS.empty,
+    }
+  })()
```
nlopin, line 60: IFEE? This function doesn't modify the state, it should be extracted from the component.

lib/experimental/Widgets/Content/ClockIn/ClockInGraph/index.tsx
```diff
@@ -0,0 +1,149 @@
...
+
+  const normalizedData = [
```
nlopin, line 62: should also be a business logic function outside of the component

lib/experimental/Widgets/Content/ClockIn/ClockInGraph/index.tsx
```diff
@@ -0,0 +1,149 @@
...
+
+  const secondaryLabel = (() => {
+    if (remainingMinutes && remainingMinutes < 0) {
+      return lastClockedInEntry
+        ? formatTime24Hours(lastClockedInEntry.to)
+        : EMPTY_LABEL
+    }
+
+    return clockedOutAt ? formatTime24Hours(clockedOutAt) : EMPTY_LABEL
+  })()
```
nlopin, line 90: same. those IFEE should be either outside or `useMemo`

lib/experimental/Widgets/Content/ClockIn/ClockInControls/index.tsx
```diff
@@ -0,0 +1,153 @@
...
+  const lastEntry = data[data.length - 1]
+  const status = lastEntry?.variant || "clocked-out"
+
+  const statusText = {
+    "clocked-out": labels.clockedOut,
+    "clocked-in": labels.clockedIn,
+    break: labels.onBreak,
+  }[status]
+
+  const subtitle = (() => {
```
nlopin, line 54: extract to avoid creating the function on every render

nlopin (changes requested):
My main concern is the amount of immediately called functions in component bodies. It is a signal they do are pure functions transforming input data to another form. Therefore they should be extracted from the component body and complex ones are even tested

nlopin (approved):

PR #1022, opened 2025-01-09
Makes cat deterministic

josepjaume:
## Description

Cats should be deterministic. Schrödinger set a bad example for us all. This gets storybook to always match against the right cat. 

*None*

### Figma Link

*None*

---

## Type of Change


dani-moreno (approved):
My poor, completely random cats 😿 !

nlopin:
Deterministic cats, checked

PR #1025, opened 2025-01-09
Fix not working links and buttons in CommunityPost component

sauldom102:
## Description

When clicking on any link or button within CompanyPost component they were not working because the event was captured by the parent box instead.


eliseo-juan (approved):
Ups!

PR #1026, opened 2025-01-09
Adds "other actions" dropdown to `ResourceHeader`

nlopin:
## Description

Introduces new `otherActions` prop to `ResourceHeader` and `BaseHeader`. It accepts the list of actions and shows a dropdown with them

I also created a new dropdown component for mobile, which shows a bottom sheet as per [Figma design](https://www.figma.com/design/uImdC0GKxH59DkjzBqRJEF/Layouts-%26-Patterns?node-id=2763-45951&p=f&t=dhxmvvUfRFnLBsVI-0), but I'm a bit hesitant to have two components under the same roof of Dropdown. What do you think about the approach?

## Screenshots

https://github.com/user-attachments/assets/9050a475-7288-40f2-ad2f-f69520285e7a

### DIFF
--- a/lib/experimental/Information/Headers/BaseHeader/index.tsx
+++ b/lib/experimental/Information/Headers/BaseHeader/index.tsx
@@ -40,11 +40,11 @@ export function BaseHeader({
         {
           label: status.label,
           value: {
-            type: "status",
+            type: "status" as const,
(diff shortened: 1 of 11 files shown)


eliseo-juan (approved):
Good job!

PR #1027, opened 2025-01-09
Add navigation and improve stories

josepjaume:
## Description

This adds up/down navigation and also reworks the stories of `PageHeader` and `Page` so they are more relevant. Also, there were certain things missing such as `statusLabel`.

image

### Figma Link

https://www.figma.com/design/pZzg1KTe9lpKTSGPUZa8OJ/Web-Components?node-id=2246-4719&t=MeSjTmnyKxJLaOfb-1

---

## Type of Change

- [X] New experimental component


lib/experimental/Navigation/Page/index.test.tsx
nlopin, line 1: I feel a lot of resistance and anger with the AI generated tests of this sort. For my recent PR (#1026) I created them and then decided not to commit them at all. 

They test the basic react functionality, that can be confirmed by scanning the component. For example, _renders children correctly_ tests that React renders children when used. There is no custom logic to check, no if-else, no complex data transformation. I don't see the point in testing those things just because we can generate a test about it. 

If we won't keep tests under control, we soon make our developer experience bad. The only test that to me worth keeping here is the `embedded` check.

nlopin (approved):

lib/experimental/Navigation/Page/index.test.tsx
josepjaume: image

josepjaume: I've actually ended up removing specs, as they add little to no value if we already have storybook snapshots.

PR #1029, opened 2025-01-10
Add support to know added/removed emoji with picker

sauldom102:
## Description

We didn't have any way to know what emoji was added or removed when the user clicked on the reaction buttons. We need that in order to implement emoji reactions.


eliseo-juan (approved):
Needed!

PR #1032, opened 2025-01-10
Prevent add emoji button to look odd

sauldom102:
## Description

When found without emojis, the `Reactions` component looked like this:

Screenshot 2025-01-10 at 17 50 45

Now it should look like this instead:

Screenshot 2025-01-10 at 17 51 10

### DIFF
--- a/lib/experimental/Information/Reactions/picker.tsx
+++ b/lib/experimental/Information/Reactions/picker.tsx
@@ -1,9 +1,8 @@
-import { Icon } from "@/components/Utilities/Icon"
+import { Button } from "@/components/Actions/Button"
 import { Reaction } from "@/icons/app"
 import { useReducedMotion } from "@/lib/a11y"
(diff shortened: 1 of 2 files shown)


lib/experimental/Information/Reactions/picker.tsx
```diff
@@ -23,11 +22,13 @@ export function Picker({ onSelect }: PickerProps) {
...
           variant="outline"
-          className={cn("h-auto px-1.5", isOpen && "border-f1-border-hover")}
-          aria-label="Add reaction"
-        >
-          <Icon icon={Reaction} aria-hidden="true" />
-        </Button>
+          label="Add reaction"
+          icon={Reaction}
+          onClick={(event) => {
+            event.stopPropagation()
```
sauldom102, line 28: This is needed too

nlopin (approved):

PR #1037, opened 2025-01-13
Loading states for breadcrumbs

josepjaume:
## Description

Adds loading states to breadcrumbs

image

### Figma Link

None, I made that up mehehehehe

---

## Type of Change

### DIFF
--- a/lib/experimental/Navigation/Header/Breadcrumbs/index.test.tsx
+++ b/lib/experimental/Navigation/Header/Breadcrumbs/index.test.tsx
@@ -1,14 +1,7 @@
 import { Home, Settings } from "@/icons/app"
 import { render, within } from "@testing-library/react"
 import { beforeEach, describe, expect, it, vi } from "vitest"
-import Breadcrumbs, { BreadcrumbItemType } from "."
-
-const mockBreadcrumbs: BreadcrumbItemType[] = [
-  { label: "Home", href: "/" },
-  { label: "Products", href: "/products" },
-  { label: "Electronics", href: "/products/electronics" },
-  { label: "Laptops", href: "/products/electronics/laptops" },
-]
+import Breadcrumbs from "."
 
 // Mock ResizeObserver
 global.ResizeObserver = vi.fn().mockImplementation(() => ({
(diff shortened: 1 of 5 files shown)


lib/experimental/Navigation/Header/Breadcrumbs/index.tsx
```diff
@@ -34,32 +73,25 @@ interface BreadcrumbItemProps {
 type DropdownItemWithoutIcon = Omit<DropdownItemObject, "icon">
 
 function BreadcrumbItem({ item, isLast }: BreadcrumbItemProps) {
-  // eslint-disable-next-line @typescript-eslint/no-unused-vars
-  const { label, ...props } = item
+  const isLoading = "loading" in item
+  const content = isLast ? (
```
nlopin, line 77: It is not clear and together with chained ternary operators have also nested ones. Can you extract it to a component with a simpler control flow?

josepjaume: Yeah, I think I will. I just found out of this scenario by accident, there's a bit of coupling. I'm thinking that the "embedded" scenario should be dealt with from the breadcrumbs directly, but maybe that's for another PR.

lib/experimental/Navigation/Header/Breadcrumbs/index.test.tsx
```diff
@@ -114,18 +118,21 @@ describe("Breadcrumbs", () => {
...
+    const firstItem = mockBreadcrumbs[0]
+    const lastItem = mockBreadcrumbs[mockBreadcrumbs.length - 1]
 
-    // Last item should not be a link
-    const lastItem = within(nav!).getByRole("link", { current: "page" })
-    expect(lastItem).toHaveTextContent(
-      mockBreadcrumbs[mockBreadcrumbs.length - 1].label
-    )
-    expect(lastItem.closest("a")).toBeNull()
+    if (!("loading" in firstItem) && !("loading" in lastItem)) {
```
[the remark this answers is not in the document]
josepjaume: The culprit of all that was the strong typing and the way we were mocking breadcrumbs with a common set of data. I rewrote it so it's more explicit and allows avoiding type assertions and weird "convincing" of TS 👍

lib/experimental/Navigation/Header/Breadcrumbs/index.tsx
```diff
@@ -20,10 +19,52 @@ import { cn, focusRing } from "@/lib/utils"
...
+    })
+  | {
+      loading: true
+    }
+
+function BreadcrumbSkeleton() {
+  return (
+    <div
+      className="h-4 w-24 animate-pulse rounded-xs bg-f1-background-secondary"
+      aria-label="Loading breadcrumb"
```
josepjaume, line 36: Dammit, forgot about this hardcoded copy. These are the kinds of things I feel should be provided with a context @nlopin .

nlopin (approved):

PR #1038, opened 2025-01-13
Add missing icons for community dropdown

sauldom102:
## Description

Add missing icons for community dropdown.

## Screenshots

Example story that I wrote locally to test if we were capable of replicating Figma with this component.

https://github.com/user-attachments/assets/4b8edae9-c154-484a-a21f-f43ba249f48a


lib/experimental/Forms/Fields/Select/index.tsx
```diff
@@ -26,7 +26,7 @@ type SelectItemObject<T> = {
 type SelectItemProps<T> = SelectItemObject<T> | "separator"
 
 type SelectProps<T> = {
-  placeholder: string
```
sauldom102, line 29: `placeholder` shouldn't be mandatory when we provide a `children`

nlopin: we should have different type then. Try 
```
type SelectProps<T> = {...} && ({placeholder: string, children?: React.ReactNode} || {placeholder?: string, children: React.ReactNode})
```

nlopin (approved):

PR #1043, opened 2025-01-14
Introduce full emoji selector

sauldom102:
## Description

Introduce full emoji selector.

## Screenshots

Screenshot 2025-01-14 at 13 26 04


sauldom102 (commented):
It's not identical to Figma, but it's a robust functional solution that looks good as well after some customization

lib/experimental/Information/Reactions/Picker/index.css
```diff
@@ -0,0 +1,16 @@
...
+  --color-border-over: hsl(var(--neutral-10));
+  --color-border: hsl(var(--neutral-10));
+  --font-size: 14px;
+  --rgb-accent: 1, 22, 55;
+  --rgb-background: 255, 255, 255;
+  --rgb-color: 1, 22, 55;
+  --rgb-input: 255, 255, 255;
+
+  width: 372px !important;
+  max-height: 518px !important;
```
nlopin, line 14: what happens if this is not set?

package.json
nlopin, line 1: those are huge dependencies 😬

lib/experimental/Information/Reactions/Picker/index.css
sauldom102: It has a default max-height by default, I just decided to set it according to what we see in figma instead

nlopin (approved):

PR #1049, opened 2025-01-15
Fix Select hover bg color

dani-moreno:
## Description

Elements inside a `Select` item were being drawn behind the background color, so they were _tainted_ with the background color. This makes these elements to be drawn on top of it.

## Screenshots

image

_Not easy to notice in a single screen, but the elements in the select item are being drawn on top of the background color now._

---

## Type of Change


lib/ui/select.tsx
```diff
@@ -107,7 +107,7 @@ const SelectItem = React.forwardRef<
...
+      "relative grid w-full cursor-default select-none grid-cols-[1fr_20px] gap-x-1.5 rounded px-3 py-2 outline-none transition-colors after:absolute after:inset-x-1 after:inset-y-0 after:z-0 after:h-full after:rounded after:bg-f1-background-hover after:opacity-0 after:transition-opacity after:duration-75 after:content-[''] first:pt-3 first:after:top-1 first:after:h-[calc(100%-0.25rem)] last:pb-3 last:after:bottom-1 last:after:h-[calc(100%-0.25rem)] hover:after:opacity-100 focus:after:bg-f1-background-hover focus:after:text-f1-foreground focus:after:opacity-100 data-[disabled]:pointer-events-none data-[disabled]:opacity-50 [&_*]:z-10",
```
nlopin, line 110: What happen if we replace `:after` with `:before` this should fix the issue without the need of adding any z-indices

dani-moreno: I tried that at first but the `before` element is still drawn on top of the content without changing any `z-index`. I've been testing a few things and I'm not sure why, though. Something must be creating a new stacking context?

nlopin: ah, that's because it's absolutely positioned

nlopin (approved):

PR #1059, opened 2025-01-16
Automatically collapse ClockInControls component using css

sauldom102:
## Description

We were using a `collapse` prop for that, which wasn't very trustful for doing this. Now, purely using css we just show the controls in a column way when the component doesn't have space left to show it in row.

## Screenshots

https://github.com/user-attachments/assets/e929e939-ce09-41bf-bc1a-121b6290300a

### DIFF
--- a/lib/experimental/Widgets/Content/ClockIn/ClockInControls/index.tsx
+++ b/lib/experimental/Widgets/Content/ClockIn/ClockInControls/index.tsx
@@ -1,6 +1,5 @@
 import { Button } from "@/components/Actions/Button"
 import { SolidPause, SolidPlay, SolidStop } from "@/icons/app"
-import { cn } from "@/lib/utils"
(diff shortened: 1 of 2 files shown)


lib/experimental/Widgets/Content/ClockIn/ClockInControls/index.tsx
```diff
@@ -47,92 +44,91 @@ export function ClockInControls({
...
-                  repeatDelay: 1,
-                }}
-              />
-              <div
-                className="absolute inset-[3px] rounded-full"
-                style={{
-                  backgroundColor: statusColor,
-                }}
-              />
+    <div className="@container">
```
sauldom102, line 47: We are using container queries to do this ✨

dani-moreno (approved):
Cool to see container queries being used 🙌

PR #1063, opened 2025-01-16
Enable Header metadata actions in mobile

dani-moreno:
## Description

Metadata actions were hidden on smaller screens. This PR makes them accessible via the `MobileDropdown` component.

Also updated the Drawer design to match [our Figma specs](https://www.figma.com/design/uImdC0GKxH59DkjzBqRJEF/Design-Patterns?node-id=4706-319156&t=82ExqQte3BHa84pW-4). Worked through a bunch of live fixes with Mario.

## Screenshots

https://github.com/user-attachments/assets/3023471c-3ee0-4ffd-8367-6687b4f5879a

_Metadata actions_

image

_How "other actions" look with the Drawer changes_

### Figma Link

[Link to Figma Design](https://www.figma.com/design/uImdC0GKxH59DkjzBqRJEF/Design-Patterns?node-id=4706-319156&t=82ExqQte3BHa84pW-4)

---

## Type of Change


CarlosBM18 (approved):
I'm curious, how this look like on the web? I mean, the bottom sheet.

lib/experimental/Information/Headers/Metadata/index.tsx
```diff
@@ -53,6 +54,7 @@ function renderMetadataValue(item: MetadataItem) {
 function MetadataItem({ item }: { item: MetadataItem }) {
   const [isActive, setIsActive] = useState(false)
   const isAction = item.actions?.length
+  const metadataValue = renderMetadataValue(item)
```
josepjaume, line 57: I would consider using a component for this. Components are just functions in the end. Why not just:

```
<MetadataValue item={item} />
```

You can even ask cursor to do this :)

josepjaume: If you're thinking you're saving up resources because you're reusing the same variable, that's not how react works. React creates the tree and *then* renders it, so both approaches would result on nearly the same amount of computation (minus the "case" thing which is negligible).

josepjaume (approved):
Cool!

nlopin:
We show this horizontal handler bar. On mobile they mean you can pull it up for the drawer to take the full screen. We don't have that functionality. Why do we need it then?

PR #1066, opened 2025-01-17
Fix truncate company name and skip to content button position

dani-moreno:
## Description

Fixes a couple of reported visual bugs in the layout:

- Company name not getting truncated [[#1](https://factorialteam.slack.com/archives/C082ZNKS403/p1737025100422449), [#2](https://factorialteam.slack.com/archives/C082ZNKS403/p1737107896894529)]
- Skip to content button visible when a banner is visible [[#1](https://factorialteam.slack.com/archives/C082ZNKS403/p1736954615364759)]

### DIFF
--- a/lib/experimental/Navigation/ApplicationFrame/index.tsx
+++ b/lib/experimental/Navigation/ApplicationFrame/index.tsx
@@ -29,7 +29,7 @@ const SkipToContentButton = ({ contentId }: { contentId?: string }) => {
     <a
       href={`#${contentId}`}
       className={focusRing(
-        "absolute z-50 -translate-y-full translate-x-4 rounded-md bg-f1-background px-4 py-2.5 text-base font-medium text-f1-foreground no-underline transition-transform duration-200 focus-visible:translate-y-4"
+        "absolute z-50 -translate-y-[250%] translate-x-4 rounded-md bg-f1-background px-4 py-2.5 text-base font-medium text-f1-foreground no-underline transition-transform duration-200 focus-visible:translate-y-4"
       )}
     >
       Skip to content
--- a/lib/experimental/Navigation/ApplicationFrame/index.tsx
+++ b/lib/experimental/Navigation/ApplicationFrame/index.tsx
@@ -29,7 +29,7 @@ const SkipToContentButton = ({ contentId }: { contentId?: string }) => {
     <a
       href={`#${contentId}`}
       className={focusRing(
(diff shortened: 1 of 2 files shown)


lib/experimental/Navigation/Sidebar/CompanySelector/index.tsx
```diff
@@ -53,7 +53,7 @@ export function CompanySelector({
 
   if (companies.length + (additionalOptions?.length || 0) === 1) {
     return (
-      <div className="p-1.5">
+      <div className="max-w-[calc(240px - 72px)] p-1.5">
```
dani-moreno, line 56: There has to be a better solution than this. Just forcing it for now as it looks bad for a few clients.

nlopin: Why does this fix the issue? Could you show before/after?

dani-moreno: It's forcing a width on the avatar + label container. Not totally sure why it's failing now, works fine in Storybook. Probably needs a better fix, but wanted to prevent customers from seeing a broken view.

I tried the fix live doing a become on a company with a long name:

**Before:**
image

**After:**
image

lib/experimental/Navigation/ApplicationFrame/index.tsx
```diff
@@ -29,7 +29,7 @@ const SkipToContentButton = ({ contentId }: { contentId?: string }) => {
     <a
       href={`#${contentId}`}
       className={focusRing(
-        "absolute z-50 -translate-y-full translate-x-4 rounded-md bg-f1-background px-4 py-2.5 text-base font-medium text-f1-foreground no-underline transition-transform duration-200 focus-visible:translate-y-4"
+        "absolute z-50 -translate-y-[250%] translate-x-4 rounded-md bg-f1-background px-4 py-2.5 text-base font-medium text-f1-foreground no-underline transition-transform duration-200 focus-visible:translate-y-4"
```
nlopin, line 32: let's move it up 1000px to be sure

nlopin (approved):

lib/experimental/Navigation/ApplicationFrame/index.tsx
dani-moreno: Good idea, more than one banner can be displayed now that I think of it.

PR #1072, opened 2025-01-17
Ensure that texts in CommunityPost have right color

sauldom102:
## Description

Just realized that on production the texts in title and description don't have the right color. This PR fixes that.

Screenshot 2025-01-17 at 17 07 13

### DIFF
--- a/lib/experimental/Information/Communities/Post/CommunityPost/index.tsx
+++ b/lib/experimental/Information/Communities/Post/CommunityPost/index.tsx
@@ -164,7 +164,7 @@ export const BaseCommunityPost = ({
               </div>
             </div>
           </div>
-          <div className="flex flex-col gap-1">
+          <div className="flex flex-col gap-1 text-f1-foreground">
             <p className="text-xl font-semibold">{title}</p>
             {description && <PostDescription content={description} collapsed />}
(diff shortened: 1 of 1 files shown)


lib/experimental/Widgets/Content/CalendarEvent/index.tsx
```diff
@@ -125,7 +125,7 @@ export const CalendarEvent = forwardRef<HTMLDivElement, CalendarEventProps>(
                   {label}
                 </p>
               )}
-              <p className="line-clamp-3 font-medium">
```
sauldom102, line 128: Happens the same thing with other components where we don't set the color explicitly

Screenshot 2025-01-17 at 17 15 31

nlopin: does it mean our default color set incorrectly?

nlopin (approved):

lib/experimental/Widgets/Content/CalendarEvent/index.tsx
sauldom102: Indeed, at least in our frontend. Within F1 it just looks good.

PR #1084, opened 2025-01-21
add checkbox example to factorial form

DavidRodriguezHidalgo:
## Description

Recently we added Checkbox field to form fields but we didn't add it to the form example.

Captura de pantalla 2025-01-21 a las 14 01 01

---

## Type of Change

### DIFF
--- a/lib/experimental/Forms/Fields/Checkbox/index.tsx
+++ b/lib/experimental/Forms/Fields/Checkbox/index.tsx
@@ -5,6 +5,7 @@ type CheckboxProps = Omit<
   "asChild"
 > & {
   title?: string
+  onCheckedChange?: (checked: boolean) => void
 }
 
 const Checkbox: React.FC<CheckboxProps> = ShadcnInput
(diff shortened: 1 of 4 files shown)


lib/experimental/Forms/lib/useForm.tsx
```diff
@@ -12,6 +12,7 @@ import { z, ZodType } from "zod"
 
 export const buildFormSchema = z.object
 export const stringField = z.string
+export const booleanField = z.boolean
```
DavidRodriguezHidalgo, line 15: I'm not fully sure of this approach for this. This obscures the use of the Zod library, I think I would prefer to know what I'm using. @josepjaume Any thoughts of getting rid of this in a future PR? I'm thinking of doing an example of nested schemas and I think this approach is not the best for it.

josepjaume: I think you might be right and I was aware it could cause issues with nested schemas. Feel free to simplify the API and make the dependencies more explicit :)

lib/experimental/Forms/Fields/Checkbox/index.tsx
```diff
@@ -5,6 +5,7 @@ type CheckboxProps = Omit<
   "asChild"
 > & {
   title?: string
+  onCheckedChange?: (checked: boolean) => void
```
nlopin, line 8: what about built in `onChange`? doesn't it work?

nlopin (approved):

PR #1094, opened 2025-01-23
feat: add onClick to BarChart

rubenmoya:
## Description

In the Talent Analytics page we want to allow filtering the table by clicking the chart bars:

image

To do so we need to have an `onClick` handler in the `BarChart` that allows us to know what have been clicked.

The typing is awful, but `activePayload` is `any[]` in `recharts`, any feedback and improvements are welcomed!

image

### DIFF
--- a/lib/components/Charts/BarChart/index.tsx
+++ b/lib/components/Charts/BarChart/index.tsx
@@ -27,11 +27,24 @@ import { fixedForwardRef } from "../utils/forwardRef"
 import { prepareData } from "../utils/muncher"
 import { ChartPropsBase } from "../utils/types"
 
+type ChartDataPoint<K extends ChartConfig> = {
+  label: string
+  values: {
+    [key in keyof K]: number
+  }
+}
+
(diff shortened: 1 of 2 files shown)


lib/components/Charts/BarChart/index.tsx
```diff
@@ -72,6 +86,22 @@ const _BarChart = <K extends ChartConfig>(
...
+          if (!onClick || !data.activeLabel || !data.activePayload) {
+            return
+          }
+
+          const chartData = {
+            label: data.activeLabel,
+            values: {},
+          } as ChartDataPoint<K>
+
+          for (const payload of data.activePayload as ActivePayload<K>) {
```
nlopin, line 99: is there a way to avoid casting?

nlopin (approved):
vamooooos

lib/components/Charts/BarChart/index.tsx
rubenmoya: I'm open to suggestions, I couldn't find a way :(

PR #1099, opened 2025-01-24
`ResourceHeader` new props: disabled, tooltip and isVisible

nlopin:
## Description

Adds several features to cover use cases:
- `disabled` prop for primary and secondary action. It will render the button in a disabled state
- `tooltip` for primary and secondary actions
- `isVisible` prop to make the use of the component simpler. Right now if I have to check for permissions before rendering an action I have to manually build the array I pass to one of the action props. It is cumbersome, the declarative approach works better and give less friction


nlopin:
I also simplified some logic regarding here and there

josepjaume:
Nice job! Convenience FTW

PR #1104, opened 2025-01-27
Add support to show video in CommunityPost component

sauldom102:
## Description

When using this component in our frontend we could see that there were some cases of posts that couldn't show media properly and that was due to some items having videos that weren't supported.

## Screenshots

Screenshot 2025-01-27 at 10 50 17


lib/experimental/Information/Communities/Post/CommunityPost/helpers.ts
```diff
@@ -0,0 +1,13 @@
+const videoExt = ["avi", "mkv", "mov", "mpeg", "mp4", "webm", "wmv"]
+
+export const isVideo = (path: string | undefined | null): boolean => {
```
sauldom102, line 3: We had this small utility in frontend, I think it makes sense having it here

lib/experimental/Information/Communities/Post/CommunityPost/helpers.ts
nlopin, line 1: `helpers` tend to become a bucket of unrelated stuff. I propose to call it `video.ts`

nlopin (approved):

PR #1110, opened 2025-01-27
Remove unwanted fill-opacity in Bullet icon

sauldom102:
## Description

This unwanted fill opacity here is not allowing us to give this icon the proper color we want.


nlopin (changes requested):
fix it in the `svg` too or `fillOpacity` will return on next icon generation

nlopin (approved):

PR #1112, opened 2025-01-28
Fix build

nlopin:
several things done:
- npm → pnpm
- class-variance-authority → cva@beta 
- vite build → tsup


arnaub (approved):

josepjaume (approved):
Great job! The only thing that bothers me is that shadcn probably doesn't work with this beta version of cva, no? Not a deal breaker, but we'll lose the ability to (easily) add new shadcn components 👀

PR #1116, opened 2025-01-28
Add support to hide break button in ClockInControls

sauldom102:
## Description

We need to be able to hide the break button. Depending on permissions we might not be able to show this button to the user.


eliseo-juan (approved):
LGTM!

PR #1117, opened 2025-01-28
Fix some display stuff in clock-in graph and controls

sauldom102:
## Description

There are a couple of things that I needed to fix regarding the clock-in graph and controls:
- When it's clocked-out, we still need to show the amount of time that the user has clocked-in previously.
- We shouldn't show any "ending time" unless the user has no remaining time left.

What we're seeing while using this component that needs to be fixed:
Screenshot 2025-01-28 at 15 29 45

### DIFF
--- a/lib/experimental/Widgets/Content/ClockIn/ClockInControls/index.stories.tsx
+++ b/lib/experimental/Widgets/Content/ClockIn/ClockInControls/index.stories.tsx
@@ -119,3 +119,21 @@ export const WithLongProjectName: Story = {
     projectName: "Bolt’s project with a very long name",
   },
 }
+
+export const ClockedOutWithSomeTime: Story = {
+  args: {
+    remainingMinutes: 320,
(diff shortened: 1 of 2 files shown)


lib/experimental/Widgets/Content/ClockIn/ClockInGraph/helpers.spec.ts
```diff
@@ -120,5 +120,35 @@ describe("ClockInGraph helpers", () => {
...
+      ]
+
+      const result = getLabels({ data: mockData, remainingMinutes: 0 })
+
+      expect(result).toEqual({
+        primaryLabel: "09:00",
+        secondaryLabel: "12:00",
+        time: "03:00",
+      })
+    })
```
eliseo-juan, line 137: Good job! We need to evolve our tests at the same time!

eliseo-juan (approved):

PR #1123, opened 2025-01-31
lets try to run it!

Rogermax:
## Description

This PR is to let any branch to run the script deploy.yaml to can have a build with the dist folder and take that commit on factorial (in development) to can use and test the components in factorial before merging to main.

We need this to start working with the new EmployeeSelector on factorial without merging the component in main.


nlopin (approved):
let's allow this, but be careful when merging main to your branch because we update factorial-one frequently

PR #1127, opened 2025-01-31
Add location picker to ClockInControl component.

sauldom102:
## Description

Add support to select a location within the `ClockInControl` component.

Screenshot 2025-01-31 at 18 56 24

### DIFF
--- a/lib/experimental/Widgets/Content/ClockIn/ClockInControls/index.tsx
+++ b/lib/experimental/Widgets/Content/ClockIn/ClockInControls/index.tsx
@@ -1,54 +1,18 @@
 import { Button } from "@/components/Actions/Button"
-import { Icon, IconType } from "@/components/Utilities/Icon"
+import { IconType } from "@/components/Utilities/Icon"
+import { Select } from "@/experimental/exports"
 import { RawTag } from "@/experimental/Information/Tags/RawTag"
 import {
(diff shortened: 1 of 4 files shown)


lib/experimental/Widgets/Content/ClockIn/ClockInControls/index.tsx
```diff
@@ -107,6 +82,26 @@ export function ClockInControls({
...
 
+  const canSelectLocation =
+    showLocationAndProjectSelectors &&
+    locations.length &&
+    !locationSelectorDisabled &&
+    canShowLocation
+  const canSelectProject =
+    showLocationAndProjectSelectors &&
+    !projectSelectorDisabled &&
+    canShowProject
```
sauldom102, line 93: Additionally, I have added support to hide selectors and badges whenever necessary as per needs in our frontend

arnaub (approved):

sergiocarracedo:
I think the select component should provide the icon functionality to avoid need to  write trigger components for those use cases (I'm sure it will happen again)

sauldom102:
> I think the select component should provide the icon functionality to avoid need to write trigger components for those use cases (I'm sure it will happen again)

As we talked privately, agree with you on unifying as much as possible this component with the one from the form, it would make it easier

PR #1135, opened 2025-02-03
feat: make AvatarNameListTag remove button keyboard focusable

rubenmoya:
## Description

This PR changes the `AvatarNameListTag` component so the `remove` action uses a `Button` instead of an `onClick` in the `Icon` directly. This improves a11y since otherwise it can be focused with keyboards, and it won't be read to screen readers as it is not interactive.


lib/experimental/Forms/AvatarNameSelector/AvatarNameListTag/index.tsx
```diff
@@ -28,12 +28,16 @@ const AvatarNameListTagContent = <T extends { avatar?: string; name: string }>({
...
-            className="ml-auto cursor-pointer text-f1-icon-secondary"
-            onClick={() => onRemove?.(entity)}
-          />
+          <div className="ml-auto">
+            <Button
+              variant="ghost"
+              size="sm"
+              onClick={() => onRemove?.(entity)}
+              icon={Cross}
+              label="Remove"
```
rubenmoya, line 37: We need to think how to solve this

lib/experimental/Forms/AvatarNameSelector/AvatarNameListTag/index.tsx
```diff
@@ -28,12 +28,16 @@ const AvatarNameListTagContent = <T extends { avatar?: string; name: string }>({
           />
         }
         right={
-          <Icon
-            icon={Cross}
-            size="sm"
-            className="ml-auto cursor-pointer text-f1-icon-secondary"
-            onClick={() => onRemove?.(entity)}
-          />
+          <div className="ml-auto">
```
rubenmoya, line 31: I'm not sure if this should be moved to the tag itself tbh

sergiocarracedo (approved):

PR #1141, opened 2025-02-04
Add AvatarList

dani-moreno:
## Description

Adds the `AvatarList`, a component to display a list of avatars (well, duh).

## Screenshots

https://github.com/user-attachments/assets/8ddf5771-b20e-4b01-8bf4-2dd35694bfa3

### Figma Link

[Link to Figma Design](https://www.figma.com/design/pZzg1KTe9lpKTSGPUZa8OJ/Web-Components?node-id=6-886&t=iGiMBOakmfUkdVHi-4)

---

## Type of Change


nlopin:
Let's set the min size for the avatars
CleanShot 2025-02-05 at 11 34 28@2x

lib/experimental/Information/Avatars/AvatarList/utils.ts
nlopin, line 1: have you explored setting a [`drop-shadow`](https://developer.mozilla.org/en-US/docs/Web/CSS/filter-function/drop-shadow) without a blur instead of having a fixed set of masks? drop-shadow follows the geometry of the image

nlopin:
we don't export this to public api, correct?

lib/experimental/Information/Avatars/AvatarList/utils.ts
dani-moreno: Yes! But that only allows a solid color, which you'd need to tweak when used over a container that might change background color on hover, for example. With these masks we can use the avatars in any background and they'll look good, even on gradients or pattern bgs.

lib/experimental/Information/Avatars/AvatarList/utils.ts
```diff
@@ -0,0 +1,32 @@
+import { type as avatarType, sizes } from "@/ui/avatar"
+
+export const getMask = {
```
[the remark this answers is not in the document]
dani-moreno: This looks way better, thanks! Gotta shake off my habit of reaching for functions all the time, hah.

lib/experimental/Information/Avatars/AvatarList/index.tsx
```diff
@@ -0,0 +1,82 @@
...
+      large: "-space-x-2",
+      xlarge: "-space-x-3",
+    } satisfies Record<(typeof sizes)[number], string>,
+  },
+  defaultVariants: {
+    size: "medium",
+  },
+})
+
+const CLIP_MASK: Record<
```
nlopin, line 23: it was worth extracting to a file though 😢

PR #1145, opened 2025-02-06
F1SearchBox component

sergiocarracedo:
- **feat: F1SearchBox component**

## Description

Adds the F1SearchBox component to provide a searchbox functionality including, debounce, clearable, threshold, icon etc
Modifies internal `Input` component to allow to prepend an icon

## Screenshots 

image

### Figma Link

---

## Type of Change

---

## Experimental Component Checklist

- [ ] Component added to the

### DIFF
--- a/lib/experimental/Forms/Fields/F1SearchBox/index.test.tsx
+++ b/lib/experimental/Forms/Fields/F1SearchBox/index.test.tsx
@@ -1,0 +1,143 @@
+import { act, fireEvent, render, screen } from "@testing-library/react"
+import { beforeEach, describe, expect, it, vi } from "vitest"
+import { F1SearchBox } from "./index"
+
+describe("F1SearchBox", () => {
+  beforeEach(() => {
+    vi.useFakeTimers()
+  })
+
+  describe("threshold behavior", () => {
+    it("does not trigger onChange when input length is below threshold", () => {
+      const onChange = vi.fn()
+      render(<F1SearchBox threshold={3} onChange={onChange} />)
+
+      const input = screen.getByRole("searchbox")
+      fireEvent.change(input, { target: { value: "ab" } })
+
+      expect(onChange).not.toHaveBeenCalled()
+    })
+
+    it("triggers onChange when input length reaches threshold", () => {
+      const onChange = vi.fn()
+      render(<F1SearchBox threshold={3} onChange={onChange} />)
+
+      const input = screen.getByRole("searchbox")
+      fireEvent.change(input, { target: { value: "abc" } })
+
+      act(() => {
+        vi.runAllTimers()
+      })
+
+      expect(onChange).toHaveBeenCalledWith("abc")
+    })
+
+    it("always triggers onChange when clearing the input", () => {
+      const onChange = vi.fn()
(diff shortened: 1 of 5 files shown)


lib/experimental/Forms/Fields/F1SearchBox/index.test.tsx
```diff
@@ -0,0 +1,158 @@
+import { act, fireEvent, render, screen } from "@testing-library/react"
+import { beforeEach, describe, expect, it, vi } from "vitest"
+import { F1SearchBox } from "./index"
+
+describe("F1SearchBox", () => {
```
nlopin, line 5: I think we can review the tests and keep only the ones that test the functionality we added. Most of the things we delegate to the input and we can rely it does its job

lib/experimental/Forms/Fields/F1SearchBox/index.tsx
```diff
@@ -0,0 +1,58 @@
...
+  debounceTime?: number
+  clearable?: boolean
+  onChange?: (value: string) => void
+}
+
+const F1SearchBox = ({
+  value,
+  threshold = 0,
+  onChange,
+  debounceTime = 0,
```
nlopin, line 19: I'm not sure about the debounce. We can use [`useDeferredValue`](https://18.react.dev/reference/react/useDeferredValue) from react, to keep the value "lagging behind" the actual state in case of load. Could it be a good way to do the debounce automagically without people controlling the debounce time?

It could be abused

nlopin (approved):
Great work, Sergio! 🙌

lib/experimental/Forms/Fields/F1SearchBox/index.tsx
sergiocarracedo: Not sure if this hook do the same I expect, but will check.
I know they can do it by themselves, I just trying to give the feature out of the box to avoid each one implement it in different ways, but we can remove it if you think it could be abused

lib/experimental/Forms/Fields/F1SearchBox/index.test.tsx
sergiocarracedo: sure

sergiocarracedo: I was checking the tests, and there 3 that con be attached to the input component, but the thing they are testing under the hood is if the link with the input is correct. But we can remove them if you want. TBH I created the tests with cursor and a couple of extra instructions to get them more complete

lib/experimental/Forms/Fields/F1SearchBox/index.tsx
sergiocarracedo: I was checking the useDeferredValue hook and it not the same as a debounce, the hooks defers the value until the new one is ready, by here I'm trying to give developers a functionality to avoid to a lot of network requests

lib/experimental/Forms/Fields/F1SearchBox/index.test.tsx
nlopin: The tests I'd remove:
- it renders (tested in storybook and screenshot tests)
- is renders disabled and the rest of the props we delegate

the rest is great

PR #1147, opened 2025-02-06
feat: SearchBox in Select

sergiocarracedo:
## Description

Add a search box to the `Select`. 
The searchbox functionality is enabled via prop `showSearchBox` and placeholder and default value can pass as prop too

Select will filter the items using the user input in the search box. For data need to be filtered in the backend the component provides a prop `externalSearch` that delegate the filtering in the parent component

image

---

## Type of Change

---

### DIFF
--- a/.storybook/main.ts
+++ b/.storybook/main.ts
@@ -19,6 +19,10 @@ const config: StorybookConfig = {
             directory: "../src/playground",
             titlePrefix: "Playground",
           },
+          {
+            directory: "../lib/ui",
+            titlePrefix: "🔒 Internal",
+          },
         ]),
   ],
   staticDirs: ["../public"],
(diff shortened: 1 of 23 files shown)


bbay:
cool @sergiocarracedo !!

May I suggest an improvement, when you click on the Select, auto-focus on the search input so you can start typing right away. Otherwise it requires a double click: first open the selector, second click on the input to start typing. wdyt?

sergiocarracedo:
> cool @sergiocarracedo !!
> 
> May I suggest an improvement, when you click on the Select, auto-focus on the search input so you can start typing right away. Otherwise it requires a double click: first open the selector, second click on the input to start typing. wdyt?

Sure, I can do that. Thanks for the suggestion

PR #1155, opened 2025-02-07
Update default example button

desiree-np:
## Description
Change the actions of the main example in the resource header documentation.

## Screenshots

Screenshot 2025-02-07 at 13 53 49

### DIFF
--- a/lib/experimental/Information/Headers/ResourceHeader/index.stories.tsx
+++ b/lib/experimental/Information/Headers/ResourceHeader/index.stories.tsx
@@ -68,7 +68,7 @@ export const Default: Story = {
     },
 
     primaryAction: {
-      label: "publish",
+      label: "Publish",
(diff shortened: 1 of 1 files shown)


lib/experimental/Information/Headers/ResourceHeader/index.stories.tsx
```diff
@@ -66,19 +66,20 @@ export const Default: Story = {
         },
       ],
     },
+
+    primaryAction: {
+      label: "publish",
```
dani-moreno, line 71: Seems weird that the label is all lowercase. What about "Publish"?

image

desiree-np: updated!

dani-moreno (approved):

PR #1166, opened 2025-02-10
Make more avatars deterministic

dani-moreno:
## Description

If I learned something from JJ, is that cats has to be deterministic. Or something like that. I wasn't paying attention, I must admit.

Related to [this](https://github.com/factorialco/factorial-one/pull/1128), we still had some random avatars in a couple of stories that were always triggering snapshot tests.

---

## Type of Change


lib/experimental/Forms/AvatarNameSelector/avatar-name.factory.tsx
```diff
@@ -1,149 +1,151 @@
+import avatar from "../../../../storybook-assets/avatar.jpeg"
 import { AvatarNamedEntity, AvatarNamedSubEntity } from "./types"
 
 export const famousEmployees: AvatarNamedEntity[] = [
   {
     id: 1000,
     name: "Albert Einstein",
-    avatar: "https://i.pravatar.cc/301",
```
sergiocarracedo, line 7: I think the problem is the misleading on the url path `/301`, `/302` etc. This is the avatar size, not the id of the avatar
You could use https://i.pravatar.cc/150?u=[X] replacing [X] with any value (ex: https://i.pravatar.cc/150?u=1) and for the same X value the image will be the same

Check: https://pravatar.cc/

sergiocarracedo (changes requested):

sergiocarracedo (approved):

PR #1171, opened 2025-02-11
Fix focus order & ring in `ResourceHeader`

dani-moreno:
## Description

Fix the focus order of the actions in the ResourceHeader. It was reversed due to CSS. Also fixed the focus being cut because of an overflow.

## Screenshots

https://github.com/user-attachments/assets/a04d4f79-04a7-474a-abbd-65dcc2c006ea

---

## Type of Change


nlopin (approved):
😍

PR #1173, opened 2025-02-11
Qa employee selector

Rogermax:
## Description

Changes to improve UX

https://github.com/user-attachments/assets/e2bc0a36-1299-4c12-8864-3d47cfe8996c

### QA link

[Jira](https://factorialmakers.atlassian.net/browse/FCT-24129)

## Type of Change


lib/components/Actions/Button/index.tsx
```diff
@@ -14,6 +14,7 @@ export type ButtonProps = Pick<
   loading?: boolean
   icon?: IconType
   hideLabel?: boolean
+  tabIndex?: number
```
Rogermax, line 17: I need this to can avoid the tabIndex [here](https://github.com/factorialco/factorial-one/pull/1173/files#diff-fc32636dfa006ea563d428160168b5c6e66d9a5731fc95475b5e8435e053a29dR271) this makes navigation with tabs more fluent

nlopin: not good. Do you have other options?

Rogermax: I can replace the button for other html tag not indexable

nlopin (approved):

PR #1185, opened 2025-02-14
Storybook headers layout section

desiree-np:
## Description

After interacting with some teams, I still saw doubts about the arrangement of the navigation elements, what goes first, what goes after, so we need to add a section on storybook about the resource and the section header to explain the **layout**. What is the correct order, what is mandatory and optional in those elements, in short, to see the headers contextualized and understand the reasoning behind these decisions.

## Screenshots 
Screenshot 2025-02-14 at 12 29 34
Screenshot 2025-02-14 at 12 19 19


dani-moreno (approved):
You're becoming a master in this! 🚀

PR #1194, opened 2025-02-17
Use index for days of the week for the `activatedDays` prop

nlopin:
## The issue

When passing the string values in activated days and weekdays, we can face the collision because we compare them by value. For example, if `daysOfTheWeek` is ["M", "T", "W", "T", "F", "S", "S"] and `activatedDays` are ["T", "S"] will lead to four values being selected because there is no difference between them.

We have this bug in production. An employee has working days set to Tuesday and Friday, but we select both Ts in the component because we cannot distinguish them:
CleanShot 2025-02-17 at 18 54 36@2x

## The fix

In the fix `activedDays` now expects indices which will then generate a unique value index+day to avoid collisions

### DIFF
--- a/lib/experimental/PageLayouts/Utils/DetailsItemsList/index.stories.tsx
+++ b/lib/experimental/PageLayouts/Utils/DetailsItemsList/index.stories.tsx
@@ -36,7 +36,7 @@ const meta: Meta = {
         title: "Workable days",
         content: {
           type: "weekdays",
-          activatedDays: ["Monday", "Tuesday", "Wednesday"],
+          activatedDays: [0, 1, 2],
         },
         spacingAtTheBottom: true,
       },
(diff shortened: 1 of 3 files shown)


eliseo-juan (approved):
Lgtm!

lib/experimental/PageLayouts/Utils/DetailsItemsList/index.stories.tsx
```diff
@@ -36,7 +36,7 @@ const meta: Meta = {
         title: "Workable days",
         content: {
           type: "weekdays",
-          activatedDays: ["Monday", "Tuesday", "Wednesday"],
+          activatedDays: [0, 1, 2],
```
sergiocarracedo, line 39: Why not using the Date standard were 0 is sunday?
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/getDay

nlopin: the current solution is unaware it shows dates. Technically, it shows a list of strings and we pick values from this list according indices passed. 

The next step in the evolution of this component will be making it date aware and use the standard

nlopin:
Despite a very valid comment from Sergio, I merge the current solution to resolve the issue we have in live.

PR #1196, opened 2025-02-18
Hide project selector in ClockInControls for now

sauldom102:
## Description

As we won't be using our new project selector yet for the ClockInControls, we should hide it for now, because we will provide a custom implementation in the frontend for it. I just leave it commented because it's something we should recover later on.


nlopin:
can we delete it for the time being instead of commenting it out?

PR #1197, opened 2025-02-18
Fix issue with video in storybook stories changing all the time

sauldom102:
## Description

This story where a video is showing up is changing all the time in different PRs and that is caused by the video being preloaded and having a different progress every time it gets rendered. With this change, it won't have this behavior when we're within storybook.

### DIFF
--- a/lib/experimental/Information/Communities/Post/CommunityPost/index.tsx
+++ b/lib/experimental/Information/Communities/Post/CommunityPost/index.tsx
@@ -6,7 +6,6 @@ import { Dropdown, DropdownItem } from "@/experimental/Navigation/Dropdown"
 import { EllipsisHorizontal } from "@/icons/app"
 import { getAgo } from "@/lib/date"
 import { withSkeleton } from "@/lib/skeleton"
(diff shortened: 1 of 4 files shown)


lib/experimental/Information/Communities/Post/CommunityPost/index.tsx
```diff
@@ -71,10 +72,16 @@ export const BaseCommunityPost = ({
...
 
+  const isStorybook = useIsStorybook()
+
   const handleClick = () => {
     onClick(id)
   }
 
+  const handleVideoClick = (event: React.MouseEvent<HTMLVideoElement>) => {
+    event.stopPropagation()
+  }
```
sauldom102, line 83: This is an extra thing we need for the frontend to stop navigating when clicking on the video

nlopin:
With this approach we'll ship unrelated code to users. As an alternative, I propose to add a 'noPreload' property to the component and has it set for all the stories

sauldom102:
> With this approach we'll ship unrelated code to users. As an alternative, I propose to add a 'noPreload' property to the component and has it set for all the stories

Done! ✅

nlopin (approved):

PR #1198, opened 2025-02-18
Add max counter on `AvatarList`

dani-moreno:
## Description

Add a new prop to the `AvatarList` component, `max`, that sets a limited number of avatars to display. The remaining ones are moved to a counter at the end of the list. If `showTooltip` is set to true, it shows a popover with the list of the rest of avatars.

https://github.com/user-attachments/assets/5e228e15-c7e3-451c-ab5e-313ed3332991

Also modifies the metadata component to show tooltips on each avatar and this max counter.

## Type of Change

### DIFF
--- a/lib/experimental/Information/Avatars/AvatarList/index.tsx
+++ b/lib/experimental/Information/Avatars/AvatarList/index.tsx
@@ -2,6 +2,7 @@ import { Tooltip } from "@/experimental/Overlays/Tooltip"
 import { sizes } from "@/ui/avatar"
 import { cva } from "cva"
 import { Avatar, AvatarVariant } from "../Avatar"
+import { MaxCounter } from "./MaxCounter"
 
 type AvatarType = AvatarVariant["type"]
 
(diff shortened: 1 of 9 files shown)


lib/experimental/Information/Avatars/AvatarList/index.tsx
```diff
@@ -61,17 +62,29 @@ type Props = {
...
+   * @default 3
+   */
+  max?: number
 }
 
 export const AvatarList = ({
   avatars,
   size = "medium",
   type,
   showTooltip = false,
```
nlopin, line 77: i'd replace `showTooltip` to `noTooltip` or something similar to keep things shorter:
`<AvatarList noTooltip/>` vs `<AvatarList showTooltip={false} />`

nlopin (approved):

lib/experimental/Information/Avatars/AvatarList/index.tsx
dani-moreno: Ah yes, much better! Showing tooltips is going to be the default setting anyway. Will do that in another PR now :-)

PR #1201, opened 2025-02-18
Ignore video element

josepjaume:
## Description

Ignores video elements on chromatic snapshots, as the loading bar is flaky and gets different snapshots.

*None*

### Figma Link

*None*


sauldom102 (approved):
Oh didn't know we had this attribute available! I solved it in a different way https://github.com/factorialco/factorial-one/pull/1197

nlopin:
let's keep fix from Saul?

josepjaume:
Yeah, Saul's fix is probably better. That said, I was trying to implement a fix that strips out `data-chromatic` attributes from JSX through a babel plugin, which is probably a good idea anyways.

PR #1210, opened 2025-02-19
Fix/page header is embbeded

sergiocarracedo:
## Description

Simplifies the breadcrumbs type, (now accepts any subtype as all contains the same basic elements)

---

## Type of Change


lib/experimental/Navigation/Header/Breadcrumbs/index.tsx
```diff
@@ -218,7 +218,10 @@ const BreadcrumbContent = React.forwardRef<
     ),
     link: (
       <ShadBreadcrumbLink asChild className="p-0">
-        <Link {...("href" in item ? item : {})} className="block">
+        <Link
+          {...("href" in item ? (item as BreadcrumbNavItemType) : {})}
```
josepjaume, line 222: Do we really need a type assertion here? They are usually dangerous, and it looks like `"href" in item` should already be able to discriminate, no?

PR #1211, opened 2025-02-19
Organize storybook content index

desiree-np:
## Description

I’ve been working on unifying the storybook content index with the goal of making it more accessible and easier to consume for everyone. I’ve streamlined the content by removing unnecessary sections and prioritizing what’s most useful.

What was done:
- Removed some stories pages that were considered irrelevant for everyone to see.
- Renamed and reorganized some elements for better clarity.
- Improved the overall structure for easier navigation.

414745967-b2e5b277-bb95-4db1-971a-0b40e21be7ec


nlopin:
why do we delete stories? have we considered keep them, but for internal use only?

sergiocarracedo:
I was exploring a way to hide stories with a a tag (internal) in storybook 7 seems easy, were we should create a function, but i think we should do it

nlopin:
Have you tried [`!dev` tag](https://storybook.js.org/docs/writing-stories/tags#docs-only-stories)?

dani-moreno:
> Have you tried [`!dev` tag](https://storybook.js.org/docs/writing-stories/tags#docs-only-stories)?

We tried that, but it was still showing on the sidebar :(

PR #1220, opened 2025-02-20
Simplify collection types

josepjaume:
## Description

This PR greatly simplifies collection types by completely getting rid of the schema and all its magic and instead deriving the types from the `fetchData` function. This is quite cool, as gives us total rendering freedom and plays especially well with graphql and fragments when it comes to using components to render properties.

As you can see, we're deleting ~500 lines and adding 200, which is great.

*None*

### Figma Link

*None*

---

## Type of Change


lib/experimental/Collections/types.ts
```diff
@@ -1,57 +1,40 @@
...
 import type { FiltersDefinition, FiltersState } from "./Filters/types"
-import type {
-  ExtractPropertyOptions,
-  ExtractPropertyType,
-  PropertySchema,
-} from "./properties"
 
 /**
  * Defines the structure and configuration of a data source for a collection.
  * @template Schema - The schema defining the properties of the collection
```
nlopin, line 6: update this please

lib/experimental/Collections/types.ts
```diff
@@ -1,57 +1,40 @@
...
+ * Represents a record type with string keys and unknown values.
+ * This type is used to represent the data structure of a collection.
+ */
+export type RecordType = Record<string, unknown>
+
 /**
  * Extracts the property keys from a collection schema and its filters.
  * @template Schema - The schema defining the properties of the collection
  * @template Filters - The available filter configurations for the collection
  */
```
nlopin, line 36: the comment looks obsolete

nlopin (approved):
Looks great! Update the comments and it is ready to merge

nlopin:
We should start testing complex types, wdyt?

lib/experimental/Collections/types.ts
josepjaume: Freaking comments

lib/experimental/Collections/Card/index.spec.tsx
```diff
@@ -2,19 +2,11 @@ import { render, screen, waitFor } from "@testing-library/react"
...
 import type { DataSource } from "../types"
 
-type TestSchema = {
-  id: Omit<NumberPropertySchema, "value">
-  name: Omit<StringPropertySchema, "value">
-  email: Omit<StringPropertySchema, "value">
-  role: Omit<StringPropertySchema, "value">
-}
-
 type TestFilters = FiltersDefinition
```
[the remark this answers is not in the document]
josepjaume: You're right, it's a leftover!

lib/experimental/Collections/types.ts
josepjaume: Done!

lib/experimental/Collections/types.ts
josepjaume: Fixed!

josepjaume:
> We should start testing complex types, wdyt?

@nlopin what do you mean? How would you test them? We indirectly test them via `@ts-expect-error` in places like this:

```tsx
<Filters
  schema={
    {
      status: {
        // @ts-expect-error - Invalid filter type in definition
        type: "invalid",
        label: "Status",
      },
    } as const
  }
  filters={{}}
  onChange={() => {}}
/>
```

Is there another good practice I don't know about?

PR #1222, opened 2025-02-21
Add `OnePagination` component

dani-moreno:
## Description

Adds the [Pagination component](https://www.figma.com/design/pZzg1KTe9lpKTSGPUZa8OJ/Web-Components?node-id=308-39589&t=5X2R7FGGBLKBLmfU-4).

Takes the total number of pages and current page, handling display of previous and next pages with controls.
Started working on this in the table, but it's generic and useful enough to be a standalone component.

## Screenshots

image

---

## Type of Change


lib/experimental/OnePagination/index.mdx
sergiocarracedo, line 1: I <3<3<3<3 extra docs!!

lib/experimental/OnePagination/index.tsx
```diff
@@ -0,0 +1,157 @@
...
+
+    const addPageRange = (start: number, end: number) => {
+      for (let i = start; i <= end; i++) {
+        pages.push(i)
+      }
+    }
+
+    pages.push(1) // Always show first page
+
+    if (currentPage <= 3) {
```
sergiocarracedo, line 74: do you think makes sense to let developers to customize the max number of visible items via prop?

lib/experimental/OnePagination/index.tsx
```diff
@@ -0,0 +1,157 @@
...
+  PaginationEllipsis,
+  PaginationItem,
+  PaginationLink,
+  PaginationNext,
+  PaginationPrevious,
+  Pagination as PaginationRoot,
+} from "@/ui/pagination"
+import { useCallback, useMemo } from "react"
+
+interface OnePaginationProps {
```
sergiocarracedo, line 13: I think we should provide a "indeterminate" version of the pagination. In this case we dont know the total items (or total pages) just if exists "nextPage"

sergiocarracedo:
do you think makes sense to provide a version without the direct page buttons?

sergiocarracedo:
should we provide a way to select the items page in certain cases?

lib/experimental/OnePagination/index.tsx
dani-moreno: Yes! Done! Was hesitant at first but given this is a pretty generic component, I think it's a good idea.

lib/experimental/OnePagination/index.tsx
dani-moreno: Good idea! I added an indeterminate state when the user `totalPages` as `0`, and there's another prop, `hasNextPage`, so the user can specify if there are more pages.

dani-moreno:
> Is it planned to allow to jump to a random page? I mean, for example an input, or combobox to select the exact page number? image

Not planned for now. Might be interesting in the future for some cases, but not exactly needed for the data collection. I'll discuss this with Mario when he's back.

sergiocarracedo (approved):

PR #1265, opened 2025-02-26
Add `Chip` component

dani-moreno:
## Description

[Chip](https://www.figma.com/design/pZzg1KTe9lpKTSGPUZa8OJ/Web-Components?node-id=2991-2267) is one of the components we're going to use for Data Collections. They represent user-supplied information, like filters.

## Screenshots

image

### Figma Link

[Link to Figma Design](https://www.figma.com/design/pZzg1KTe9lpKTSGPUZa8OJ/Web-Components?node-id=2991-2267)

---

## Type of Change

### DIFF
--- a/lib/experimental/OneChip/index.tsx
+++ b/lib/experimental/OneChip/index.tsx
@@ -1,0 +1,79 @@
+import { Icon, IconType } from "@/components/Utilities/Icon"
+import {
+  Avatar,
+  AvatarVariant,
+} from "@/experimental/Information/Avatars/Avatar"
+import { CrossedCircle } from "@/icons/app"
+import { cn, focusRing } from "@/lib/utils"
+import { cva, type VariantProps } from "cva"
+
+export const chipVariants = cva({
+  base: "flex items-center gap-1 rounded-full border border-solid border-f1-border px-2 py-0.5 font-medium",
+  variants: {
+    variant: {
+      default: "",
+      selected: "bg-f1-background-selected-secondary border-f1-border-selected",
+    },
+  },
+  defaultVariants: {
+    variant: "default",
+  },
+})
+
+interface ChipProps extends VariantProps<typeof chipVariants> {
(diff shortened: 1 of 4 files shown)


lib/experimental/OneChip/index.tsx
```diff
@@ -0,0 +1,80 @@
...
+  /**
+   * If defined, an avatar will be displayed in the chip
+   */
+  avatar?: AvatarVariant
+
+  /**
+   * If defined, an icon will be displayed in the chip
+   */
+  icon?: IconType
+}
```
sergiocarracedo, line 44: If icon and avatar can not be defined at same time, this type ensures that
```suggestion
interface ChipProps extends VariantProps<typeof chipVariants> {
  /** The label of the chip */
  label: string
  
  /** If defined, the close icon will be displayed and the chip will be clickable */
  onClose?: () => void
  
} & (
  | { /** If defined, an avatar will be displayed in the chip */
      avatar: AvatarVariant
      icon?: never
    }
  | { /** If defined, an icon will be displayed in the chip */
      icon: IconType
      avatar?: never
    }
  | { 
      avatar?: never
      icon?: never
    }
)
```

sergiocarracedo (approved):

lib/experimental/OneChip/index.tsx
dani-moreno: Ah wow, this is magical, thanks! 🫶

nlopin:
what about a very long text? how the component behaves?

dani-moreno:
> what about a very long text? how the component behaves?

These chips will wrap in the data collection, so I guess we don't need to limit it.

nlopin (approved):

dani-moreno:
> We need some extra padding on narrow screens

Ah good catch! I'll update the examples with flex-wrap as that's how they'll actually be displayed.

PR #1290, opened 2025-02-28
Support async options in filters

josepjaume:
## Description

Support async options in filters

Support async options in filters

### Figma Link

*None*


josepjaume:
I'm merging this without approval - it's just a way for filter options to accept async callbacks. Very useful for lists, and I would like to be able to showcase this in the app. Open to opening a PR to correct any comments!

PR #1296, opened 2025-03-03
Enable links in tables row

dani-moreno:
## Description

Adds the option to define a link for each row in the Data Collection (also adding it to the `OneTable` component).

Followed [this article](https://robertcooper.me/post/table-row-links#solution-3-use-css-grid-to-create-a-table) to make accessible links:

- Each row is focusable, also interactive elements inside.
- Interactive elements has a higher z-index, making it not fire the link that is behind.
- Works well with tab navigation.

There are a few cons with this approach, though.

- For every interactive element that we want to be focusable, we need to select it through css and add a `relative` and a `z-index ` class.

## Screenshots

image

---

## Type of Change

### DIFF
--- a/lib/experimental/Collections/Table/index.tsx
+++ b/lib/experimental/Collections/Table/index.tsx
@@ -81,7 +81,14 @@ export const TableCollection = <
                 </TableCell>
               ))}
               {source.actions && (
-                <TableCell key="actions">
+                <TableCell
+                  key="actions"
+                  href={
+                    "href" in item && typeof item.href === "string"
+                      ? item.href
+                      : undefined
+                  }
+                >
                   <ActionsDropdown item={item} actions={source.actions} />
                 </TableCell>
               )}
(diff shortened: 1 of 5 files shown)


lib/experimental/Collections/Table/index.tsx
```diff
@@ -67,13 +67,28 @@ export const TableCollection = <
...
           {data.map((item, index) => (
             <TableRow key={`row-${index}`}>
-              {columns.map((column) => (
-                <TableCell key={String(column.label)}>
+              {columns.map((column, cellIndex) => (
+                <TableCell
+                  key={String(column.label)}
+                  firstCell={cellIndex === 0}
+                  href={
+                    "href" in item && typeof item.href === "string"
```
nlopin, line 75: we can extract the repetitive code into a variable

nlopin (approved):

PR #1297, opened 2025-03-03
Add Table grouping

dani-moreno:
## Description

Add basic component for grouping in tables. (Only in the `OneTable` for now)

## Screenshots 

https://github.com/user-attachments/assets/caf04915-0297-4c5d-b494-cb5ded23f446

**Next steps:**

- Animation (Motion does not get along really well with tables)
- Avoid content jumps when there's no column width specified.

### Figma Link

[Link to Figma Design](https://www.figma.com/design/pZzg1KTe9lpKTSGPUZa8OJ/Web-Components?node-id=4209-53388&t=O2gFsRTICUj11exo-4)

---

## Type of Change

### DIFF
--- a/lib/experimental/OneTable/TableGroup/index.tsx
+++ b/lib/experimental/OneTable/TableGroup/index.tsx
@@ -1,0 +1,9 @@
+import { TableCell as TableCellRoot } from "@/ui/table"
+
+interface TableCellProps {
+  children: React.ReactNode
+}
+
+export function TableGroup({ children }: TableCellProps) {
+  return <TableCellRoot>{children}</TableCellRoot>
+}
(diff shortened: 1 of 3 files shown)


lib/experimental/OneTable/TableGroup/index.tsx
```diff
@@ -0,0 +1,43 @@
...
+
+export function TableGroup({
+  label,
+  number,
+  open = false,
+  onClick,
+}: TableCellProps) {
+  return (
+    <TableCellRoot
+      colSpan={100}
```
sergiocarracedo, line 23: Why a fixed number?

sergiocarracedo (approved):

lib/experimental/OneTable/TableGroup/index.tsx
dani-moreno: Not happy with it, because it's a bit hacky. As I don't know the number of columns, if you set a high number, it stills take only the columns the table has.

But 100% open to change this 🙌

sergiocarracedo: maybe get the total number of cols from the parent as prop

PR #1301, opened 2025-03-04
Hide internal tags in Storybook filter menu

dani-moreno:
## Description

Hide the internal tags that are not used to depict a component status in the Storybook filter menu. Only applies in production builds.

## Screenshots

image

_Before_

---

image

_After_


eliseo-juan (commented):
Cleaner!

PR #1308, opened 2025-03-04
Add maturity table to documentation

nlopin:
## Description

Instead of having 99% of the library marked experimental, let's provide granular information about the state of components.

## How

Use `OneMeta` component instead of built-in `Meta`. It requires you to pass a maturity object and renders the table based on it.

https://github.com/user-attachments/assets/733eae6a-3414-4550-a687-5b00e3fcc407

CleanShot 2025-03-04 at 18 09 22@2x


sergiocarracedo (approved):
I love it!
🚀

PR #1314, opened 2025-03-05
Remove actions generic

josepjaume:
## Description

`Actions` was a generic, which cluttered our interface and wasn't really valuable.

*None*

### Figma Link

*None*

---

## Type of Change


lib/experimental/Collections/README.md
```diff
@@ -0,0 +1,110 @@
+# Collections Component Library
```
sergiocarracedo, line 1: This docs are nice, i think we should show them in storybook

PR #1319, opened 2025-03-05
Feat/select virtual list

sergiocarracedo:
## Description

Adds tanstack virtual to virtualize the list render for performance
- feat: Allow tu use virtual list in internal select component via items props (need to define the item height)
- Select component uses virtual list. Defines 2 different sizes depending if the item has description or not
- Select Viewport only rendered when open for better performance
- select options allow to pass an `item` of any type as in the SelectItemObject and onChange returns it as second argument. Useful to avoid do a find in the parent after select a value. For example you want to select an user you can pass the user in the item for each element and when an item is selected you can get all the props of the user without exra code 

> Includes the PR https://github.com/factorialco/factorial-one/pull/1313 to test it in the monorepo without merge

The story uses 10000 options, without virtual list takes about 5s to open, with it is instant

Screenshot 2025-03-05 at 17 20 11

## Type of Change


docs/development/development.md
```diff
@@ -0,0 +1,48 @@
+# Development
```
josepjaume, line 1: Nice! Although out of scope of this PR (doesn't bother me that much)

lib/ui/Select/select.tsx
```diff
@@ -0,0 +1,132 @@
+"use client"
```
josepjaume, line 1: We don't really need this, right?

lib/ui/Select/components/SelectItemsVirtual.tsx
```diff
@@ -0,0 +1,83 @@
...
+  items,
+  parentRef,
+  positionIndex,
+}: Props) => {
+  const virtualizer = useVirtualizer({
+    count: items.length,
+    getScrollElement: () => parentRef.current,
+    estimateSize: (i: number) => items[i].height,
+    overscan: 5,
+  })
```
josepjaume, line 27: This is very nice!

josepjaume (approved):
Great job! I like it 👌 . I'm starting to think we need to support virtualization for tables as well.

Maybe you should consider breaking this PR in smaller PRs, as there seems to be a lot of unrelated docs in the branch.

lib/ui/Select/select.tsx
sergiocarracedo: yes, i know, but i included this PR https://github.com/factorialco/factorial-one/pull/1313 to be able to test the select in the monorepo without need to merge this branch to main

vite/build-sync.plugin.ts
```diff
@@ -0,0 +1,83 @@
...
+  sources?: string[]
+  target: string
+}
+
+/**
+ * @description Sync plugin
+ * This plugin will syncronize the build files between the local and the remote server (coder)
+ * The list of the files and folders to sync is defined in the package.json file in the attribute "files"
+ * The target is the remote server where the files will be copied
+ */
```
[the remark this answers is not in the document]
sergiocarracedo: yes, to be able to test the code in monorepo without merge it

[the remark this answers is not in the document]
josepjaume: Yeah yeah, I meant that maybe this could go on different PRs but no worries!

PR #1321, opened 2025-03-05
Dynamic dots in Carousel + small fixes

dani-moreno:
## Description

Make the dots in the Carousel a bit more dynamic:

- If there are more than 5 pages in the carousel, [the dots display a different design](https://www.figma.com/design/pZzg1KTe9lpKTSGPUZa8OJ/Web-Components?node-id=7369-4197&t=ixCg1og6rh4vJJHr-4), to avoid displaying a huge list of dots.
- Dots can be clicked to go to a specific page.
- Left/Right arrows are only shown when the user has the cursor over the carousel.
- Left/Right arrows have a container with a background to avoid merging the buttons colors with the background.

## Screenshots

https://github.com/user-attachments/assets/6718dd55-cabd-4de8-af5a-5b6c5b51cf43

---

## Type of Change

### DIFF
--- a/lib/ui/carousel.tsx
+++ b/lib/ui/carousel.tsx
@@ -139,7 +139,7 @@ const Carousel = React.forwardRef<
         <div
           ref={ref}
           onKeyDownCapture={handleKeyDown}
-          className={cn("relative", className)}
+          className={cn("group/carousel relative", className)}
           role="region"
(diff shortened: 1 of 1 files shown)


sauldom102 (approved):
LGTM! I just tested it as well and it works like a charm! 🔥 Loved it!

PR #1326, opened 2025-03-07
fix: Use `primary-navigation` as `aria-label` when it is primary navigation

eliseo-juan:
## Description

I have asked Claude for a Haiku to celebrate my first PR in Factorial one

> Cursor guides with care
> Sometimes perfect, sometimes not
> Perhaps not its fault

## Type of Change

### DIFF
--- a/lib/experimental/Navigation/Tabs/index.tsx
+++ b/lib/experimental/Navigation/Tabs/index.tsx
@@ -33,7 +33,7 @@ export const BaseTabs: React.FC<TabsProps> = ({
     <TabNavigation
       secondary={secondary}
       asChild
-      aria-label={secondary ? "primary-navigation" : "secondary-navigation"}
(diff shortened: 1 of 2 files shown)


nlopin:
Those names must be translated

lib/experimental/Navigation/Tabs/index.tsx
```diff
@@ -33,7 +33,7 @@ export const BaseTabs: React.FC<TabsProps> = ({
     <TabNavigation
       secondary={secondary}
       asChild
-      aria-label={secondary ? "primary-navigation" : "secondary-navigation"}
+      aria-label={secondary ? "secondary-navigation" : "primary-navigation"}
```
nlopin, line 36: Let's then update the language file and request proper translations for the labels

sergiocarracedo:
Does this PR makes sense now? If not, could you  close it? if yes, can we merge it?

PR #1329, opened 2025-03-10
Add `OverflowList` internal component

dani-moreno:
## Description

The `OverflowList` is a responsive component that manages lists of items when space is limited. It solves the common problem of displaying a horizontal list of elements that might not all fit within their container.

As the available space changes (like when a user resizes their browser window), the component dynamically recalculates which items should be visible and which should be in the overflow dropdown.

Useful for making responsive [tabs](https://www.figma.com/design/Drd7hhxgof79A7zeftSYMo/Patterns-Catalog?node-id=1-84382&t=4dhsXugM0YCFpyfw-4), [presets in a DataCollection](https://www.figma.com/design/pZzg1KTe9lpKTSGPUZa8OJ/Web-Components?node-id=4230-61642&t=5nW68MStEmlUGzai-4), [breadcrumbs](https://www.figma.com/design/pZzg1KTe9lpKTSGPUZa8OJ/Web-Components?node-id=38-711&t=5nW68MStEmlUGzai-4)...

## Screenshots

https://github.com/user-attachments/assets/9ec23cce-b9af-4e8d-8e7e-476e0c62dfdc

### DIFF
--- a/lib/ui/OverflowList/index.tsx
+++ b/lib/ui/OverflowList/index.tsx
@@ -28,7 +28,6 @@ function useOverflowCalculation<T>(items: T[], gap: number) {
   const containerRef = useRef<HTMLDivElement>(null)
   const overflowButtonRef = useRef<HTMLButtonElement>(null)
   const measurementContainerRef = useRef<HTMLDivElement>(null)
-  const isCalculatingRef = useRef(true)
 
   // Combined state for visible and overflow items
   const [itemsState, setItemsState] = useState<{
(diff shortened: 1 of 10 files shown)


lib/ui/OverflowList/index.tsx
```diff
@@ -0,0 +1,346 @@
...
+  ])
+
+  // Initial calculation and initialization
+  useEffect(() => {
+    // Run the calculation
+    calculateVisibleItems()
+
+    if (!isInitializedRef.current) {
+      isInitializedRef.current = true
+      setIsInitialized(true)
```
nlopin, line 170: why do we need both ref and state to store the state of initialization?

dani-moreno: Ah, sorry! That was dead code that I forgot to remove.

sergiocarracedo (approved):

PR #1333, opened 2025-03-11
Update Dialog component

dani-moreno:
## Description

Updates the [Dialog](https://www.figma.com/design/pZzg1KTe9lpKTSGPUZa8OJ/Internal-Components?node-id=1382-14258) component according to what we have in Figma.

The current Dialog is a generic modal that's become obsolete (still using Gamma styles) and was pretty open-ended. Although we'll have a generic modal (still not design-ready), the Dialog modal in Figma has a much more specific purpose.

- The `OneDialog` component is transformed to a specific use case (confirmation dialogs) instead of a generic modal.
- The `Dialog` inside the ui folder now acts as a generic modal-like component, serving as a base for other components (like this one, or any other modals we create).

## Screenshots

image

_Critical type dialog_

### Figma Link

[Link to Figma Design](https://www.figma.com/design/pZzg1KTe9lpKTSGPUZa8OJ/Internal-Components?node-id=1382-14258)

### DIFF
--- a/lib/experimental/Overlays/Dialog/index.tsx
+++ b/lib/experimental/Overlays/Dialog/index.tsx
@@ -1,37 +1,42 @@
 import { Button, ButtonProps } from "@/components/Actions/Button"
-import { Icon, IconType } from "@/components/Utilities/Icon"
+import {
+  AlertAvatar,
+  type Props as AlertAvatarProps,
+} from "@/experimental/Information/Avatars/AlertAvatar"
 import {
   Dialog,
   DialogContent,
   DialogDescription,
   DialogFooter,
   DialogHeader,
-  DialogIcon,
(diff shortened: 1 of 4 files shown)


lib/experimental/Overlays/Dialog/index.tsx
```diff
@@ -1,37 +1,42 @@
 import { Button, ButtonProps } from "@/components/Actions/Button"
-import { Icon, IconType } from "@/components/Utilities/Icon"
+import {
+  AlertAvatar,
+  type Props as AlertAvatarProps,
```
nlopin, line 4: let's export better named props instead of generic `Props` from the `AlertAvatar`

nlopin (approved):

PR #1336, opened 2025-03-11
feat: allow no actions in a row in data collections

sergiocarracedo:
## Description

- feat:Allow no action in some rows
- feat: allow no url link
- chore: add dropdown test

---

## Type of Change

### DIFF
--- a/lib/experimental/Collections/Actions/Dropdown.tsx
+++ b/lib/experimental/Collections/Actions/Dropdown.tsx
@@ -16,8 +16,14 @@ export const ActionsDropdown = <
 }) => {
   if (!actions || actions.length === 0) return null
 
+  const items = filterActions(actions, item)
+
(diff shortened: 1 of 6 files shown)


eliseo-juan (approved):
Make sense

lib/experimental/Collections/Actions/Dropdown.tsx
```diff
@@ -16,8 +16,10 @@ export const ActionsDropdown = <
 }) => {
   if (!actions || actions.length === 0) return null
 
-  return (
-    <Dropdown items={filterActions(actions, item)}>
+  const items = filterActions(actions, item)
+
+  return !items.length ? null : (
```
nlopin, line 21: early return will be harder to miss:

```
if (items.length === 0) {
  return null;
}
```

PR #1342, opened 2025-03-14
fix-multiple-small-bugs-employee-selector

Rogermax:
## Description

Employee Selector has problems leaving inside a form

https://github.com/user-attachments/assets/32c60536-6443-4295-bb5f-343f8dda2437

## Type of Change

### DIFF
--- a/lib/experimental/Forms/AvatarNameSelector/AvatarNameListItem/index.tsx
+++ b/lib/experimental/Forms/AvatarNameSelector/AvatarNameListItem/index.tsx
@@ -7,6 +7,7 @@ import LogoAvatar from "@/icons/app/LogoAvatar"
 import { cn } from "@/lib/utils"
 import { Checkbox } from "@/ui/checkbox"
 import { ChevronDown, ChevronRight } from "lucide-react"
+import { useState } from "react"
 import { HighlightText } from "../AvatarNameHighLightText"
 import { AvatarNamedEntity } from "../types"
 
@@ -73,7 +74,7 @@ export const AvatarNameListItemSingleContent = ({
   const firstName = nameParts[0] || ""
   const lastName = nameParts.slice(1).join(" ")
 
-  const handleLabelClick = (ev: React.MouseEvent<HTMLLabelElement>) => {
+  const handleLabelClick = (ev: React.MouseEvent<HTMLButtonElement>) => {
     ev.preventDefault()
     ev.stopPropagation()
     if (disabled) return
@@ -84,10 +85,10 @@ export const AvatarNameListItemSingleContent = ({
     }
   }
 
-  const handleKeyDown = (ev: React.KeyboardEvent<HTMLLabelElement>) => {
+  const handleKeyDown = (ev: React.KeyboardEvent<HTMLButtonElement>) => {
+    ev.preventDefault()
+    ev.stopPropagation()
     if (ev.key === "Enter" || ev.key === " ") {
-      ev.preventDefault()
-      ev.stopPropagation()
       if (disabled) return
       if (!selected) {
         onSelect(entity)
@@ -95,12 +96,8 @@ export const AvatarNameListItemSingleContent = ({
         onRemove(entity)
       }
     } else if (ev.key === "ArrowDown") {
-      ev.preventDefault()
-      ev.stopPropagation()
       focusNextFocusable(ev.currentTarget, goToFirst)
     } else if (ev.key === "ArrowUp") {
-      ev.preventDefault()
-      ev.stopPropagation()
       focusPreviousFocusable(ev.currentTarget, goToLast)
     }
   }
@@ -108,10 +105,7 @@ export const AvatarNameListItemSingleContent = ({
   return (
     <div className="w-full pl-1 pr-1">
       <label
-        onClick={handleLabelClick}
-        onKeyDown={handleKeyDown}
         aria-label={entity.name}
-        data-avatarname-navigator-element="true"
         className={cn(
           marginLeft,
           "flex flex-row flex-wrap items-center gap-2 rounded-[10px] border p-2 hover:cursor-pointer",
@@ -135,11 +129,11 @@ export const AvatarNameListItemSingleContent = ({
         </div>
 
         <Checkbox
+          data-avatarname-navigator-element="true"
           checked={selected}
           disabled={disabled}
-          onClick={(ev) => {
-            ev.preventDefault()
-          }}
+          onClick={handleLabelClick}
+          onKeyDown={handleKeyDown}
           className={cn(
             "pointer-events-none ml-auto h-[20px] w-[20px] rounded-xs border-[1px] data-[state=checked]:text-f1-foreground-inverse",
             singleSelector ? "opacity-0" : ""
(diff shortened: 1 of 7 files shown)


lib/experimental/Forms/AvatarNameSelector/AvatarNameListItem/index.tsx
```diff
@@ -84,34 +85,27 @@ export const AvatarNameListItemSingleContent = ({
     }
   }
 
-  const handleKeyDown = (ev: React.KeyboardEvent<HTMLLabelElement>) => {
+  const handleKeyDown = (ev: React.KeyboardEvent<HTMLButtonElement>) => {
+    ev.preventDefault()
+    ev.stopPropagation()
```
nlopin, line 90: do you need both?

lib/experimental/Forms/AvatarNameSelector/index.tsx
```diff
@@ -64,63 +24,52 @@ export const AvatarNameSelector = ({
...
+              subItems: entities.find((e) => e.id === entity.id)?.subItems,
+            }
+          : e
       )
-      return singleSelector ? [entity] : newSelectedEntities
-    })
+    }
+    onSelect(
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      (singleSelector ? entity : newSelectedEntities) as any
```
nlopin, line 59: what's the issue with `any`? Can't we have correct types here?

lib/experimental/Forms/AvatarNameSelector/AvatarNameListItem/index.tsx
```diff
@@ -84,34 +85,27 @@ export const AvatarNameListItemSingleContent = ({
...
-      ev.stopPropagation()
       focusPreviousFocusable(ev.currentTarget, goToLast)
     }
   }
 
   return (
     <div className="w-full pl-1 pr-1">
       <label
-        onClick={handleLabelClick}
-        onKeyDown={handleKeyDown}
```
Rogermax, line 112: This two lines are one of the main problems

lib/experimental/Forms/AvatarNameSelector/AvatarNameListItem/index.tsx
```diff
@@ -287,6 +284,11 @@ const AvatarNameListItem = ({
             disabled={disabled}
             onClick={handleGroupClick}
             indeterminate={partialSelected}
+            onPointerDown={(ev) => {
+              ev.preventDefault()
+              ev.stopPropagation()
+              setPressingLabel(false)
+            }}
```
Rogermax, line 291: I didn't find another way to do that... the problem is that I want that clicking on a label expand the group (Team, workspace, whatever) But I don't want to select all the group. 

The are two problems here.. I can't have a onClick listner in the label and antoher in the checkbox inside a form.. this produces the original problem of a loop of clicks.. Then clicks on the label are automatically send to the checkbox, this make virtuially imposible to know if the click is from the label or the checkbox

Any recomendations or alternatives are welcome

lib/experimental/Forms/AvatarNameSelector/AvatarNameSelectorContent/AvatarNameSelectorMainContent/AvatarNameSelectorSearcher.tsx
```diff
@@ -29,6 +29,9 @@ export const AvatarNameSelectorSearcher = ({
       ev.preventDefault()
       ev.stopPropagation()
       focusPreviousFocusable(ev.currentTarget, goToLast)
+    } else if (ev.key === "Enter") {
+      ev.preventDefault()
+      ev.stopPropagation()
```
Rogermax, line 34: Enter key, is interpreted as a click in a form.. to avoid that behaviour we must stop the event

lib/experimental/Forms/AvatarNameSelector/index.tsx
```diff
@@ -64,63 +24,52 @@ export const AvatarNameSelector = ({
...
   width,
   loading = false,
   singleSelector = false,
   disabled = false,
   children,
   ...props
 }: AvatarNameSelectorProps & { children?: React.ReactNode }) => {
-  const [selectedEntities, setSelectedEntities] = useState<AvatarNamedEntity[]>(
-    selectedAvatarName ?? []
-  )
```
Rogermax, line 77: As this component should be used as a stateless one... I removed this unnecassy state (this implies update all stories to know handle the state change)

lib/experimental/Forms/AvatarNameSelector/AvatarNameSelectorContent/AvatarNameSelectorMainContent/index.tsx
```diff
@@ -392,6 +405,7 @@ export const AvatarNameSelectorMainContent = ({
                 size="sm"
                 onClick={onSelectAll}
                 title={selectAllLabel + ` (${totalFilteredEntities})`}
+                type="button"
```
Rogermax, line 408: This change is super important!
In a form.. if a button has no property type, then the form things it's a submit button..
So all buttons of this component now has type="button" to avoid submiting the wrapper form

Rogermax: Maybe we should have type="button" a default in our button component?

lib/experimental/Forms/AvatarNameSelector/AvatarNameListItem/index.tsx
Rogermax: Not sure.. after 10000 proves and errors.. when I found it was working I stopped changing things.. 
I have to make a second iteration to that PR to clean a bit

lib/experimental/Forms/AvatarNameSelector/index.tsx
Rogermax: it's not trivial, but ok! I'll try

lib/experimental/Forms/AvatarNameSelector/index.tsx
```diff
@@ -10,83 +10,21 @@ import {
...
-  selectedLabel,
-  notFoundTitle,
-  notFoundSubtitle,
-  selectedAvatarName,
-  onSelect: onSelectProp,
-  width,
-  loading = false,
-  singleSelector = false,
-  disabled = false,
-  children,
```
Rogermax, line 72: To help typescript infer that if singleSelector is true then onSelect can just accept an element (not an array) all props must be inside props.. we can't deconstruct...

lib/experimental/Forms/AvatarNameSelector/index.tsx
Rogermax: solved! But I had to remove the props deconstruction

lib/experimental/Forms/AvatarNameSelector/AvatarNameSelectorContent/AvatarNameSelectorMainContent/index.tsx
nlopin: yes, totally make sense

nlopin (approved):

PR #1343, opened 2025-03-14
Add Data collection search component

dani-moreno:
## Description

Adds the search box to the Data Collection component. This adds the visual part, and the component with needed props to make the search actually work. It features:

- An expanding searchbox as in the Figma designs.
- Responsive list of presets, as the searchbox pushes the content.
- Keyboard navigation.
- A "clear" button.
- A "loading" state.

**Things to consider while implementing the search function:**

- It should be optional.

## Screenshots

https://github.com/user-attachments/assets/edd25390-23e6-479e-9f76-d38726b7ab0e

_Searchbox motion_

---

https://github.com/user-attachments/assets/423ad152-bf89-4597-bd50-1e2af33c9b12

_How the presets work when expanding the searchbox_

---

https://github.com/user-attachments/assets/52ef8eca-608f-4926-b605-c5df5cf41265

_Loading state_

---

### Figma Link

[Link to Figma Design](https://www.figma.com/design/RBKhrXjevsqIHcLWAvhV3I/Internal-Patterns?node-id=5903-63315&t=R2rVb6yBfMQ3Qgd3-4)


josepjaume:
I really like it! How do you imagine the search happening though - do we search as we type? In that case, could we prevent the search from collapsing if not empty?

dani-moreno:
> I really like it! How do you imagine the search happening though - do we search as we type? In that case, could we prevent the search from collapsing if not empty?

Yes! It would search as you type. And yes! That's how it works now. If the search component has a value, it does not collapse. It goes to a closed state displaying the value and a clear icon.

image

Can't be easily tested here as I'm not 'saving' the value as you type (as I just worked on the visual part), but if you set a value to the Search component you can check for yourself.

josepjaume (approved):

josepjaume:
Awesome! Let's do this!

dani-moreno:
> Actually - do you mind if I build on top of this PR to add the rest of the support?

Not at all, go ahead! 🙌

PR #1350, opened 2025-03-17
fix: SelectItem hoverf padding

sergiocarracedo:
## Description

Select last item has no bottom padding, add the correct calculation to after hover element

**Before**
image

---

## Type of Change


eliseo-juan (approved):
LGTM!

PR #1397, opened 2025-03-18
fix: dropdown button action on click

sergiocarracedo:
## Description

Dropdown button must execute the action in the dropdown on click, not select the action

## Type of Change

### DIFF
--- a/lib/components/Actions/OneDropdownButton/OneDropdownButton.tsx
+++ b/lib/components/Actions/OneDropdownButton/OneDropdownButton.tsx
@@ -36,7 +36,7 @@ const OneDropdownButton = ({
   value,
   ...props
 }: OneDropdownButtonProps) => {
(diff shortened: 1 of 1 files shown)


nlopin (approved):

lib/components/Actions/OneDropdownButton/OneDropdownButton.tsx
```diff
@@ -36,7 +36,7 @@ const OneDropdownButton = ({
   value,
   ...props
 }: OneDropdownButtonProps) => {
-  const [localValue, setLocalValue] = useState(value || items[0].value)
+  const [localValue] = useState(value || items[0].value)
```
nlopin, line 39: do we need the local value?

sergiocarracedo: i can be a memo instead of a state as it will not change

PR #1425, opened 2025-03-19
chore/workspaces

sergiocarracedo:
## Description

- chore: Move factorial one react library to its own workspace to be able to make the publish pipeline work again as repo now is a monorepo
- chore: Rename github action workflows to have the workspace name for easy understading
- chore: use paths in worksflows to only trigger the workflows for the specific workspaces (Seems github is not able to trigger it as the workflow is not in main yet)
- chore: uppdate release-please to use the correct workspace
- chore: move react example app to packages/examples folder to keep the same criteria than other packages (use packages folder for workspaces)


CarlosBM18 (approved):
LGTM! I think we should move the icons into the core package, but I'll do it in another PR, as I'll create the integration to make them work on react-native.

PR #1429, opened 2025-03-19
feat(PageHeader): add product update dropdown

nlopin:
Follows the design form [Figma](https://www.figma.com/design/566K8KYuC1p19UUTk7TqF6/In-app-Updates-Notifications---25Q1?node-id=13082-5161&t=7oBu1SZLkQWG5Q1u-0)


sergiocarracedo (approved):

lib/experimental/Navigation/Header/ProductUpdates/index.tsx
```diff
@@ -0,0 +1,329 @@
...
+  imageURL: UrlString
+  updated: string
+  unread?: boolean
+}
+
+type ProductUpdatesProp = {
+  label: string
+  moreUpdatesLabel: string
+  updatesPageUrl: UrlString
+  getUpdatesQuery: () => Promise<Array<ProductUpdate>>
```
josepjaume, line 32: Isn't the word `query` a bit weird? Why not simply `getUpdates`?

packages/react/src/global.d.ts
```diff
@@ -1 +1,3 @@
 /// <reference types="user-agent-data-types" />
+
+type UrlString = `http://${string}` | `https://${string}`
```
josepjaume, line 3: Nice

packages/react/src/experimental/Navigation/Header/ProductUpdates/index.tsx
```diff
@@ -0,0 +1,329 @@
...
+  href,
+  imageURL,
+  unread,
+  updated,
+}: ProductUpdate) => {
+  const itemClass = "flex flex-col items-stretch gap-3 w-full"
+  return (
+    <DropdownMenuItem asChild>
+      <Link
+        href={href}
```
josepjaume, line 148: I think forcing it to be an external URL might be a bit too much. The `Link` element is clever enough to understand if this is a relative path or an absolute path. If it's relative, it uses React Router. If not, it sets `target="_blank"`. You can find this on the FactorialOneProvider implementation on the monorepo.

josepjaume (approved):
Good job! I like how this is all about semantics now. Easy to understand, limited, but extendible.

packages/react/src/experimental/Navigation/Header/ProductUpdates/index.tsx
nlopin: what would be your solution? just pass an `href`

packages/react/src/experimental/Navigation/Header/ProductUpdates/index.tsx
```diff
@@ -0,0 +1,329 @@
...
+  DropdownMenuPortal,
+  DropdownMenuSeparator,
+  DropdownMenuTrigger,
+} from "@/ui/dropdown-menu"
+import { Skeleton } from "@/ui/skeleton"
+import { ReactElement, useCallback, useState } from "react"
+
+type ProductUpdate = {
+  title: string
+  href: UrlString
```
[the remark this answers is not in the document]
nlopin: not really

packages/react/src/experimental/Navigation/Header/ProductUpdates/index.tsx
josepjaume: Oh! What I mean is that we're forcing the type to start with `http://` or `https://` so we won't ever allow internal links, if we want. We could just pass the `href` as a `string` or maybe consider including `/{string}` as a valid type?

PR #1431, opened 2025-03-19
chore(library): export icons

josepjaume:
## Description

Up to this moment, we were importing icons like `factorialco/factorial-one/icons/app` with the hopes that we would only end up compiling icons that we have imported (and thus get a smaller footprint). This didn't end up happening and it's inconvenient.

This changes the import to become just `AppIcons`, `ModuleIcons` and `AnimatedIcons` which is more convenient and provides a better DX as well.

*None*

### Figma Link

*None*

---

## Type of Change


josepjaume:
Hey @sergiocarracedo @nlopin sorry if I'm moving fast with this - only have a couple days left and I need to move forward!

josepjaume:
Okay, on second thought, I think I would rather keep the old behavior. The scope of the change on the app is way too big.

PR #1450, opened 2025-03-21
chore(react-native): update imports and package configuration for better module resolution

CarlosBM18:
## Description

Fix import and build of react-native package example app

### DIFF
--- a/packages/react-native/package.json
+++ b/packages/react-native/package.json
@@ -2,8 +2,19 @@
   "name": "@factorialco/factorial-one-react-native",
   "version": "0.0.1",
   "description": "React Native components for Factorial One Design System",
+  "main": "dist/index.js",
   "module": "dist/index.mjs",
   "types": "dist/index.d.ts",
+  "exports": {
+    ".": {
+      "types": "./dist/index.d.ts",
+      "import": "./dist/index.mjs",
+      "require": "./dist/index.js"
+    }
+  },
+  "files": [
+    "dist"
+  ],
   "scripts": {
     "build": "tsup",
     "dev": "tsup --watch",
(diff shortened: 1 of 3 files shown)


sergiocarracedo (approved):

packages/react-native/tsup.config.ts
```diff
@@ -1,11 +1,17 @@
-import { defineConfig } from "tsup"
+import { defineConfig } from "tsup";
 
 export default defineConfig({
   entry: ["src/index.ts"],
-  format: ["esm"],
+  format: ["esm", "cjs"],
```
nlopin, line 5: have you decided to build cjs too?

packages/react-native/package.json
nlopin, line 1: don't we need `"type": "module"` here?

packages/react-native/tsup.config.ts
nlopin: take a look whether or not your consumer uses ESM

packages/react-native/package.json
CarlosBM18: which type could it be instead?

nlopin: By default the value is `"commonjs"`, hence all files without are considered a CJS module ([docs](https://nodejs.org/api/packages.html#type)).

CarlosBM18: ok, and what happens if we don't type it as a module?

nlopin: Good question which made me realize that in the case of the library the change won't cause any major changes, but since we use ESM, better have it

packages/react-native/tsup.config.ts
CarlosBM18: Yes, we need it for react-native.

PR #1468, opened 2025-03-25
feat: Data collection property renderers

sergiocarracedo:
## Description

- feat: added col align property for tables
- refactor: Rename to OneDataCollection
- feat: Provides a type-safe cell or card renderer definition allowing only to use the defined renderers

To add a new renderer:

1 .Edit file  `packages/react/src/experimental/OneDataCollection/visualizations/property`
2. Add the name and the render function. NOTE: the render function must receive all the properties in a single argument to allows us to add more metadata in the future to renderers

Renderer gets the current visualizacion and the property definition to adapt the result toi the context

image

image

### Figma Link

https://www.figma.com/design/RBKhrXjevsqIHcLWAvhV3I/Internal-Patterns?node-id=10120-50501&t=C3AwNIHRxmGquxBC-0

---

## Type of Change


packages/react/src/experimental/OneDataCollection/visualizations/collection/Table/index.tsx
```diff
@@ -65,6 +77,8 @@ export const TableCollection = <
     Sortings
   >(source)
 
+  console.log(data)
```
nlopin, line 80: 🙅

sergiocarracedo: agggg!! :(

PR #1470, opened 2025-03-25
feat(pageHeader): track function on mount

ChloeDumit:
## Description

Adding a `trackView` property in product updates page header in order to track whenever the button is being shown


nlopin:
We found a solution outside of the factorial one

PR #1486, opened 2025-03-26
Fct 26144 fixed columns

sergiocarracedo:
## Description

Allow to define fixed (sticky columns) in the table

NOTE: the sticky cols need a fixed width to work

https://github.com/user-attachments/assets/531c76a2-5b0e-40cf-bd2e-dd35cd164b2c

### Figma Link

---

## Type of Change


packages/react/src/experimental/OneDataCollection/index.stories.tsx
```diff
@@ -256,9 +256,11 @@ const createPromiseDataFetch = (delay = 500) => {
 const ExampleComponent = ({
   useObservable = false,
   usePresets = false,
+  fixedCols = 0,
```
nlopin, line 259: I suggest to name it _frozenCols_ or _freezeCols_ because this is the term used in all data sheet processors like excel and google spreadsheets.

Fixed columns could be interpreted as "fixed width column" which brings ambiguity

packages/react/src/experimental/OneDataCollection/utils.ts
```diff
@@ -0,0 +1,89 @@
...
+export const renderValue = <RecordType>(
+  item: RecordType,
+  property: PropertyDefinition<RecordType>,
+  visualization: VisualizationType
+): ReactNode => {
+  const renderIsRendererDefinition = (
+    renderDef: RendererDefinition | string | number | undefined
+  ): renderDef is RendererDefinition => {
+    return renderDef !== undefined && typeof renderDef === "object"
+  }
```
nlopin, line 63: let's move it out of the function to avoid recreation of the function on every call. It should be safe since you don't work with hooks here

packages/react/src/experimental/OneDataCollection/utils.ts
```diff
@@ -0,0 +1,89 @@
...
+
+  const renderDefinition = property.render(item)
+
+  const { type, value } = renderIsRendererDefinition(renderDefinition)
+    ? renderDefinition
+    : ({
+        type: "text" as const,
+        value: renderDefinition ?? "-",
+      } as const)
```
nlopin, line 72: maybe `satisfies RendererDefinition` instead of `as const`?

packages/react/src/experimental/OneTable/TableHead/index.tsx
```diff
@@ -118,15 +124,28 @@ export function TableHead({
     </>
   )
 
+  const isNumber = (width: ColumnWidth): width is number =>
```
nlopin, line 127: should we have this function shared? there is a duplication in the `TableCell` file

nlopin (approved):

packages/react/src/experimental/OneDataCollection/index.stories.tsx
sergiocarracedo: makes sense

PR #1489, opened 2025-03-27
feat: datacolletion selectable rows

sergiocarracedo:
### Figma Link

---

## Type of Change

---

- [ ] Component added to `experimental` folder
- [ ] Component is documented with basic stories
- [ ] Component added to the

## Promotion to Stable Checklist (if applicable)

- [ ] **Documentation**

  - [ ] The component has a dedicated documentation page
  - [ ] Includes description and clear use cases
  - [ ] Includes Storybook stories covering all variations
  - [ ] Component updated on the

- [ ] **Responsiveness**

  - [ ] Verified the component works across various screen sizes

- [ ] **Testing**

  - [ ] Component includes unit tests with sufficient coverage

- [ ] **Accessibility**
  - [ ] Accessibility standards meet at least AA level requirements
  - [ ] All interactive elements are keyboard-navigable and have focus states
  - [ ] Proper ARIA attributes are used where needed


packages/react/src/experimental/OneDataCollection/useSelectable.ts
```diff
@@ -0,0 +1,133 @@
...
+        selected.set(id, value.item)
+      } else {
+        unselected.set(id, value.item)
+      }
+    })
+    return [selected, unselected]
+  }, [itemsState])
+
+  const selectedCount: number = selectedItems.size
+  const unselectedCount: number = unselectedItems.size
```
nlopin, line 41: If I'm not wrong, there is no need to type variables here:
```suggestion
  const selectedCount = selectedItems.size
  const unselectedCount = unselectedItems.size
```

packages/react/src/experimental/OneDataCollection/useSelectable.ts
```diff
@@ -0,0 +1,133 @@
...
+
+  const handleItemStateChange = (item: Record, checked: boolean) => {
+    const id = source.selectable && source.selectable(item)
+    if (id === undefined) {
+      return
+    }
+
+    itemsState.set(id, { item, checked })
+
+    setItemsState(new Map(itemsState))
```
nlopin, line 53: it is usually better to use function as an argument to read the current value of the state and modify it:

```suggestion
    setItemsState(current => new Map(current))
```

packages/react/src/experimental/OneDataCollection/useSelectable.ts
```diff
@@ -0,0 +1,133 @@
...
+  }
+
+  const handleSelectAll = (checked: boolean) => {
+    setAllSelectedCheck(checked)
+    if (checked) {
+      data.forEach((item) => {
+        handleItemStateChange(item, true)
```
nlopin, line 60: I am worried that we instantiate a new `Map` on every item in the table and throw them away. Is there a better way to organize it? maybe calling `setItemsState` from here after the loop?

packages/react/src/experimental/OneDataCollection/useSelectable.ts
```diff
@@ -0,0 +1,133 @@
...
+  const isAllSelected = allSelectedCheck || areAllKnownItemsSelected
+
+  // If the filters change, we need to reset the selected items
+  useEffect(() => {
+    handleSelectAll(false)
+    setItemsState(new Map())
+  }, [source.currentFilters])
+
+  // If the data changes, we need to update new items that are added to the data
+  // If the data changes, we need to update new items that are added to the data
```
nlopin, line 92:
```suggestion
```

packages/react/src/experimental/OneDataCollection/useSelectable.ts
```diff
@@ -0,0 +1,133 @@
...
+
+  useEffect(() => {
+    // Notify the parent component about the selected items
+    onSelectItems?.(
+      allSelectedCheck && unselectedCount == 0,
+      Array.from(selectedItems.values())
+    )
+  }, [onSelectItems, allSelectedCheck, selectedItems, unselectedCount])
+
+  return {
```
nlopin, line 126: I usually wrap the return object with `useMemo` to avoid it's recreation on every call. if this data lands as a dependency in `useEffect` it will run it on every render

packages/react/src/experimental/OneDataCollection/useSelectable.ts
```diff
@@ -0,0 +1,133 @@
...
+
+  // Control the allSelectedCheck state
+  // If all items are selected, we need to set the allSelectedCheck state to true
+  // If there are no selected items, we need to set the allSelectedCheck state to false
+  // If some items are selected, we need to keep the state as is to know if we will need to check the next page items
+  useEffect(() => {
+    if (areAllKnownItemsSelected) {
+      setAllSelectedCheck(true)
+    }
+    if (selectedCount === 0) {
```
nlopin, line 113: should we have it as `else if`? Could it be that we set true and then false in some circumstances?

packages/react/src/experimental/OneDataCollection/useSelectable.ts
```diff
@@ -0,0 +1,133 @@
...
+      setAllSelectedCheck(false)
+    }
+  }, [areAllKnownItemsSelected, selectedCount])
+
+  useEffect(() => {
+    // Notify the parent component about the selected items
+    onSelectItems?.(
+      allSelectedCheck && unselectedCount == 0,
```
nlopin, line 121: isn't `allSelectedCheck` sufficient?

packages/react/src/experimental/OneDataCollection/useSelectable.ts
sergiocarracedo: Sure i can refactor it. I thought react queue the changes for a while before update the DOM like Vue

packages/react/src/experimental/OneDataCollection/useSelectable.ts
sergiocarracedo: yes, agree, clearer code

packages/react/src/experimental/OneDataCollection/useSelectable.ts
sergiocarracedo: No, as in the cases not all the data is loaded (async pagination) allSelectCheck means we want to select all but the user could deselect an item

PR #1509, opened 2025-03-31
chore: standardise properties across charts

joerivett:
We have varying different prop experiences across line, bar and vertical bar charts at the moment. There are also props that can be passed down into these components but have no impact on the component at all.

This commit standardises a number of things:

- Ensures all charts support the option to hide the tooltip
- Ensures all charts support the option to hide the grid
- Ensures all charts support the domain property in recharts (previously we supported passing this in the wrapper but it wasn't passed to recharts)


packages/react/src/components/Charts/BarChart/index.tsx
```diff
@@ -122,15 +122,15 @@ const _BarChart = <K extends ChartConfig>(
           onClick(chartData)
         }}
       >
-        {hideTooltip && (
```
joerivett, line 125: This was an existing bug

nlopin (approved):
Thank you for giving some love to charts Joe!

PR #1510, opened 2025-03-31
Cursor rules and integration test combobox

Rogermax:
## Description

Added cursor rules + integration test on AvatarNameSelector (combobox)
* Also added an example component to use as templae for new components

Error: 

https://github.com/user-attachments/assets/93d33a9a-56d5-48e1-a3ae-11cdbb2043b5

## Type of Change


nlopin:
why does the popover shifts in such a way on the second 4 in the video?

.gitignore
```diff
@@ -25,4 +25,5 @@ node_modules
 *storybook.log
 
 dist
-/packages/react/icons
\ No newline at end of file
+/packages/react/icons
+.nyc_output
```
nlopin, line 29: what is `nyc_output`?

.cursor/rules/00-project-overview.mdc
```diff
@@ -0,0 +1,68 @@
...
+
+# Factorial One Design System - Overview
+
+## Key Principles
+- Write modular, reusable components following atomic design principles.
+- Use TypeScript with strict type safety.
+- Follow functional and declarative programming patterns.
+- Implement comprehensive tests and ensure accessibility (a11y) compliance in all components.
+- Support both web and mobile platforms with consistent design patterns.
+- All code generated and comments shoud be in English
```
nlopin, line 15:
```suggestion
- All code generated and comments should be in English
```

.cursor/rules/02-typescript-teting.mdc
nlopin, line 1: fix the name plz

packages/react/src/experimental/Forms/AvatarNameSelector/index.mdx
nlopin, line 1: is it AI generated?

.gitignore
Rogermax: I may have some extension in VSC that's tries to run coverage test, and this is the folder for the generated data. If that boder you I can remove it and try to find which of my extensions is adding files there...

packages/react/src/experimental/Forms/AvatarNameSelector/index.mdx
Rogermax: yep.. In two weeks if this works I may show this in tech all hands. Try to bring our documentation to this kind of files.. to let cursor work for use on typicall developments

Rogermax:
> why does the popover shifts in such a way on the second 4 in the video?

The problem is that table is updating on every filter change, and the change of state changes the table content and the summary totals, so I think the shift is because of that...

packages/react/src/experimental/Forms/AvatarNameSelector/index.stories.tsx
```diff
@@ -140,6 +140,75 @@ export const Default = {
       />
     )
   },
+  play: async ({ canvasElement }: { canvasElement: HTMLCanvasElement }) => {
```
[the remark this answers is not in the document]
Rogermax: yep on story book directly. I thought pipeline run this test also? may not?

https://github.com/user-attachments/assets/a8afe437-851c-4f30-8085-d37add87de4a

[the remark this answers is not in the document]
nlopin: looks cool. Let's see if our pipeline runs it

packages/react/src/experimental/Forms/AvatarNameSelector/index.mdx
nlopin: Reach Desireé to polish the docs further. I think we have a defined document structure every documentation must follow.

nlopin:
> > why does the popover shifts in such a way on the second 4 in the video?
> 
> The problem is that table is updating on every filter change, and the change of state changes the table content and the summary totals, so I think the shift is because of that...

hmm, maybe we should remember the biggest item or add some animation to make it looks smoother. Not in the scope of this PR for sure, but those things are usually game changers in term of UI perception

nlopin (approved):

Rogermax:
> * Do we need all those rules? For example, build one looks redundant to me
> * why do we need an example component? we have simple components in the repo they can be used as examples

I added all these rules because before already exists a big one ".cursorrules" file, but I'm with you, I don't think we need all of them.

I will use this `Example` component as example to discuss with Desireé the perfect documentation, then I will move this example to the cursorrules and delete from here.

PR #1516, opened 2025-03-31
Fix(page header)  adding as child property

ChloeDumit:
## Description

The feature product update in page header dropdown was not redirecting to the link correctly, 
## Type of Change


packages/react/src/experimental/Navigation/Header/ProductUpdates/index.tsx
```diff
@@ -94,7 +94,9 @@ const ProductUpdates = ({
         if (open && updates === null) {
           invokeGetUpdates()
         }
-        onOpenChange(open)
+        if (open) {
+          onOpenChange(open)
+        }
```
nlopin, line 99: `onOpenChange` should trigger on any change state - opening and closing. That's why we pass a boolean to convey what is the change open→closed or closed→open

nlopin (approved):

PR #1519, opened 2025-04-01
SectionHeader: adjust separator based on the layout

nlopin:
## Description

CleanShot 2025-04-01 at 09 29 00@2x

When `SectionHeader` is used together with the `StandardLayout`, the separator does not cover the full surface of the page.

## Solution

I added a new **internal provider** so we can understand the layout in which a component is rendered. Based on this knowledge, I adjust the width of the block using the negative margin trick

I also added a story for the issue so we can keep track on it with screenshot tests


eliseo-juan (approved):
We should do the same with Data Collections!

PR #1528, opened 2025-04-01
fix(DataCollection): Ensure seamless row height in table view

dani-moreno:
## Description

Different data types were making the row height inconsistent. While we want the height to grow if needed, we want to ensure a minimum of 48px. Some cell types were making this height larger, as the content inside made the cells grow.

To avoid this, the padding top and bottom of the cell now depends on the type of data in the table visualization, ensuring that larger elements like tags don't unnecessarily increase the row size.

## Screenshots

image

### Figma Link

[Link to Figma Design](https://www.figma.com/design/RBKhrXjevsqIHcLWAvhV3I/Internal-Patterns?node-id=10388-167106&t=M80YbIZDpttoL3eg-4)

---

## Type of Change

### DIFF
--- a/packages/react/src/experimental/OneDataCollection/visualizations/collection/Table/index.tsx
+++ b/packages/react/src/experimental/OneDataCollection/visualizations/collection/Table/index.tsx
@@ -23,6 +23,16 @@ import {
 import { CollectionProps, RecordType } from "../../../types"
 import { useData } from "../../../useData"
 import { useSelectable } from "../../../useSelectable"
+import { PropertyRendererType } from "../../../visualizations/property"
+
+// Define padding values for different property types
+const PROPERTY_TYPE_PADDINGS: Partial<Record<PropertyRendererType, number>> = {
+  avatarList: 8,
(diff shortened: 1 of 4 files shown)


packages/react/src/experimental/OneDataCollection/visualizations/collection/Table/index.tsx
```diff
@@ -23,6 +23,16 @@ import {
 import { CollectionProps, RecordType } from "../../../types"
 import { useData } from "../../../useData"
 import { useSelectable } from "../../../useSelectable"
+import { PropertyRendererType } from "../../../visualizations/property"
+
+// Define padding values for different property types
+const PROPERTY_TYPE_PADDINGS: Partial<Record<PropertyRendererType, number>> = {
```
sergiocarracedo, line 29: I think each cell type should provide its own padding, or at least return it to avoid couple the visualization and the cell types code

sergiocarracedo (changes requested):

PR #1531, opened 2025-04-02
Fix: card link navigation on widget

luchhh:
## Description

Fixed navigation of CardLink on Widget component. The component was rendering a `<Button>` inside `<Link>`, causing a full-page reload instead of internal navigation.  I'm proposing to use an `<Icon>` instead of a `<Button>`:

- Simplify event handling
- The previous approach is against W3C [recommendation](https://html.spec.whatwg.org/#the-a-element) of not having interactive elements nested on a Link

## Screenshots

image

## Type of Change

### DIFF
--- a/packages/react/src/ui/card.tsx
+++ b/packages/react/src/ui/card.tsx
@@ -111,7 +111,6 @@ const CardLink = React.forwardRef<
       )}
       aria-label={title}
       {...props}
-      role="link"
     >
       <Icon size="sm" icon={icon} className="text-f1-icon-bold" />
     </Link>
--- a/packages/react/src/ui/card.tsx
+++ b/packages/react/src/ui/card.tsx
@@ -7,7 +7,6 @@ import InfoCircleLine from "../icons/app/InfoCircleLine"
 
 import { Icon, IconType } from "@/components/Utilities/Icon"
 import { Link } from "@/lib/linkHandler"
-import { Button } from "../components/Actions/Button"
 import {
   Tooltip,
   TooltipContent,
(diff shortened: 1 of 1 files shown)


packages/react/src/ui/card.tsx
```diff
@@ -101,14 +100,20 @@ const CardLink = React.forwardRef<
...
-        hideLabel
-      />
+    <Link
+      ref={ref}
+      className={cn(
+        "group inline-flex aspect-square h-6 items-center justify-center gap-1", //layout
+        "rounded-sm border border-solid border-f1-border bg-f1-background-inverse-secondary", //appearance
+        "whitespace-nowrap px-0 text-base font-medium text-f1-foreground", //typography
+        "cursor-pointer transition-colors hover:border-f1-border-hover focus-visible:outline-none focus-visible:ring-1 focus-visible:ring-f1-ring focus-visible:ring-offset-1", //interaction
+        className
```
luchhh, line 110: This is heavily using the same styles as a button. I was wondering what is the recommended approach to avoid duplication. Maybe having a Clickable list of variants on the [button](https://github.com/factorialco/factorial-one/blob/main/packages/react/src/ui/button.tsx) that can be exported?

nlopin: yep, we should create a function that brings all those styles at once, like `focusRing` function from `utils` does

packages/react/src/ui/card.tsx
```diff
@@ -101,14 +100,20 @@ const CardLink = React.forwardRef<
...
+      )}
+      aria-label={title}
+      {...props}
+      role="link"
```
nlopin, line 114: do we need a role if there is a link inside?

nlopin (approved):
great work Lucho 🙌

packages/react/src/ui/card.tsx
luchhh: Yeah, maybe not, I'm removing it

PR #1536, opened 2025-04-02
Create OneModal component

sauldom102:
## Screenshots

Screenshot 2025-04-02 at 15 09 06

Screenshot 2025-04-02 at 15 08 46


packages/react/src/experimental/Modals/OneListModal/ListModal.spec.tsx
```diff
@@ -0,0 +1,35 @@
...
+    fireEvent.click(screen.getByRole("presentation"))
+    expect(defaultProps.onClose).toHaveBeenCalled()
+  })
+
+  it("calls onClose when clicking close button", () => {
+    render(<ListModal {...defaultProps} />)
+    fireEvent.click(screen.getByLabelText("Close modal"))
+    expect(defaultProps.onClose).toHaveBeenCalled()
+  })
+})
```
sauldom102, line 35: I'm gonna move these specs to the new play functionality for running tests inside storybook

packages/react/src/experimental/Modals/OneListModal/OneListModal.tsx
```diff
@@ -0,0 +1,106 @@
...
+  const handleClose = () => {
+    setOpen(false)
+    onClose()
+  }
+
+  const isSmallScreen = useMediaQuery("(max-width: 440px)", {
+    initializeWithValue: false,
+  })
+
+  const Content = () => (
```
nlopin, line 56: this looks like a separate component

packages/react/src/experimental/Modals/OneListModal/OneListModal.tsx
nlopin, line 1: Overall, let's split the modal in atomic chunks for future exposure. I imagine that in most situations the user will use `OneModal`, but in some, I just want a modal I want to fill myself. I imagine the following structure of this component:
- OneModal
- OneModal.Header
- OneModal.Dialog (the container itself)

nlopin:
why is it called  `OneListModal` isn't it a generic modal you place a list component into?

packages/react/src/experimental/Modals/OneModalHeader/OneModalHeader.tsx
```diff
@@ -0,0 +1,34 @@
+import { ButtonInternal } from "@/components/Actions/Button/internal"
+import {
+  DropdownInternal,
+  DropdownInternalProps,
+} from "@/experimental/Navigation/Dropdown/internal"
+import CrossIcon from "@/icons/app/Cross"
+
+export type OneModalHeaderProps = {
+  title: string
+  dropdownItems?: DropdownInternalProps["items"]
```
nlopin, line 10: we usually call those things `otherActions` to convey the semantics, not implementation details

sauldom102:
> why is it called `OneListModal` isn't it a generic modal you place a list component into?

I just named it that way because I thought not every modal should look like a bottom sheet in its mobile version and this would only be the case for modals with list content. We have apart from this the dialogs which always show up like that both in mobile and desktop https://one.factorial.dev/?path=/docs/components-dialog--documentation

sauldom102 (commented):
@nlopin now we have it defined in a way so that you use it like this:

```
<OneModal>
    <OneModal.Header {...} />
    <OneModal.Content {...} />
</OneModal>
```

PR #1548, opened 2025-04-03
fix: Text editor buttons and lists errors

siguenzaraul:
## Description

I changed a couple of styles and added avoid defaults to the toolbar buttons as they were somehow submitting forms.

## Type of Change

---

- [ ] Component added to the


packages/react/src/experimental/RichText/RichTextEditor/Enhance/index.tsx
```diff
@@ -41,7 +41,8 @@ const EnhanceActivator = ({
   const enhanceButtonRef = useRef<HTMLButtonElement>(null)
   const [open, setOpen] = useState(false)
 
-  const handleEnhanceClick = () => {
+  const handleEnhanceClick = (e?: React.MouseEvent) => {
```
nlopin, line 44: I think there is always an event passed to the handler, no need to make it optional

nlopin (approved):

PR #1565, opened 2025-04-07
fix: remove section header padding top

rodrigo-picanco:
## Description
So, here's where the usage in our app kinda leaks into a PR/feature request. 

The section header has a padding top that looks all right on the screenshots we have on the storybook. 

Buuuut, in our app, since we already have a padding for the ResponsivePage component, we end up with this "double padding top" which is making designers very sad with the final result of our implementation 😢 

I've tough about suggesting removing the padding top form the ResponsivePage, but thenI realized this component is optional. 

So, I come here with this PR removing the padding top of the section header, as it's a very specific component planned to be used on the top of sections that at least right now already have a padding. 

100% open to other solutions tough. 

### How it looks on storybook
Screenshot 2025-04-07 at 11 49 50

---

## Type of Change


nlopin (commented):
Agree we need to tackle it. This fix solves the issue only for the top padding, the bottom padding will keep the same double padding behavior. Have you tried [:first-child](https://tailwindcss.com/docs/hover-focus-and-other-states#first-last-odd-and-even) or something similar? This could be the ultimate solution

rodrigo-picanco:
> Agree we need to tackle it. This fix solves the issue only for the top padding, the bottom padding will keep the same double padding behavior. Have you tried [:first-child](https://tailwindcss.com/docs/hover-focus-and-other-states#first-last-odd-and-even) or something similar? This could be the ultimate solution

I haven't but it's worth a shot. I'll get back to u.

packages/react/src/experimental/Information/Headers/SectionHeader/index.tsx
```diff
@@ -38,10 +38,10 @@ export const SectionHeader = ({
   return (
     <div
       className={cn(
-        "border-1 flex flex-col justify-between gap-4 border-dashed border-transparent px-6 py-5 md:flex-row md:gap-2",
+        "border-1 flex flex-col justify-between gap-4 border-dashed border-transparent px-6 py-5 md:flex-row md:gap-2 first:py-0",
```
nlopin, line 41: shouldn't we split  ̀py` into top and bottom? I'm thinking about something like  ̀pt-5 pb-5 first:pt-0 last:pb-0`?

nlopin (approved):

packages/react/src/experimental/Information/Headers/SectionHeader/index.tsx
rodrigo-picanco: Sure, makes sense!

rodrigo-picanco:
I'm not sure this is needed anymore, as the padding seems to have been removed from the responsive page on the app 

🤔 

Screenshot 2025-04-07 at 15 12 13

PR #1570, opened 2025-04-09
fix: make widgets more transparent

sauldom102:
## Description

According to our final Figma designs, widgets background opacity should be 0.4 instead of 0.6.

Screenshot 2025-04-09 at 17 18 30


packages/react/src/experimental/Widgets/Widget/index.tsx
```diff
@@ -93,6 +93,9 @@ const Container = forwardRef<
         fullHeight ? "h-full" : "",
         "relative flex gap-3 border-f1-border-secondary"
       )}
+      style={{
+        backgroundColor: "hsl(var(--white-40))",
```
sauldom102, line 97: We don't have this color in F1 with a semantic name, that's why it has been added specified here this way

PR #1573, opened 2025-04-10
chore: move icons svgs to core package

sergiocarracedo:
## Description

Move svg icons to core package to make them available for `react-native` package

---

## Type of Change


eliseo-juan (approved):
Should we need to change anything in our monorepo?

PR #1575, opened 2025-04-10
fix(DataCollection): Reported changes from teams

dani-moreno:
## Description

A bunch of small changes to the DataCollection or related components, mostly reported from teams [here](https://www.notion.so/factorialco/Data-Collection-Reports-Improvements-1cb5e6e051ee8093b478def1dccea18a).

- Backdrop blur added to the `OneActionBar` so what's behind can't be seen through it.
- Disabled prop for `OnePagination`.
- Pagination in DataCollection is disabled when there is only a single page.
- Force text color in data types (so Gamma colors are not used).
- Fixed text size in `OneActionBar`.
- Head table aligned to the right when the column is right aligned.

## Screenshots

image

---

## Type of Change


sergiocarracedo (approved):

sergiocarracedo:
I id some changes in the pagination before to see this PR, please when you resolve the conflicts priorize yours

PR #1581, opened 2025-04-11
chore: notify new released on slack

sergiocarracedo:
## Description

Add notificacion to slack step on publish a package to npm


eliseo-juan (approved):
Vamooos ;)

Should we add this at any place of the documentation?

sergiocarracedo:
True!

PR #1597, opened 2025-04-13
fix: set proper HighlightBanner button

sauldom102:
## Description

The button in `HighlightBanner` component suddenly started to show up this way:

Screenshot 2025-04-13 at 13 08 58


packages/react/src/experimental/Information/Communities/HighlightBanner/index.tsx
```diff
@@ -1,4 +1,4 @@
-import { Button } from "@/ui/button"
+import { Button } from "@/components/Actions/Button"
```
sauldom102, line 1: We are not using the `Button` from `ui` anymore as that's the base one.

PR #1602, opened 2025-04-14
fix: start break when the amount of break types is only one

sauldom102:
## Description

When we pass an array of break types it will probably be more usual to pass multiple of them, however, when we are passing only one break type we are also showing the picker when the user clicks on the break button, which is a little bit redundant.

## Screenshots

This is what we want to avoid:

Screenshot 2025-04-14 at 10 08 54

Now, if only one break type is available, that one will be selected by default and the break will start automatically.

### DIFF
--- a/packages/react/src/experimental/Widgets/Content/ClockIn/ClockInControls/index.tsx
+++ b/packages/react/src/experimental/Widgets/Content/ClockIn/ClockInControls/index.tsx
@@ -94,19 +94,20 @@ export function ClockInControls({
 
   const showLocationAndProjectSelectors = status === "clocked-out"
 
-  const breakTypeOptions = breakTypes?.map((breakType) => ({
-    value: breakType.id,
-    label: breakType.duration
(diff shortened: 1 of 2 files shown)


packages/react/src/experimental/Widgets/Content/ClockIn/ClockInControls/index.tsx
```diff
@@ -106,10 +106,13 @@ export function ClockInControls({
   const [breakTypePickerOpen, setBreakTypePickerOpen] = useState(false)
 
   const handleClickBreakButton = () => {
-    if (breakTypeOptions?.length && !breakTypePickerOpen) {
-      setBreakTypePickerOpen(true)
-    } else if (!breakTypeOptions?.length) {
-      onBreak?.()
+    if ((breakTypeOptions?.length ?? 0) > 1) {
```
nlopin, line 109: is there a way to always have breaktypeOptions set? do you use the empty state somewhere?

nlopin (approved):

PR #1606, opened 2025-04-14
fix: storybook toc width

sergiocarracedo:
## Description

Fix the toc width to avoid overflow in small screens

Before
image

[

https://github.com/user-attachments/assets/edadaf8e-5afe-437d-a273-62690ef0516a

](url)

### DIFF
--- a/packages/react/.storybook/preview-head.html
+++ b/packages/react/.storybook/preview-head.html
@@ -99,9 +99,11 @@
   }
 
   .sbdocs-wrapper {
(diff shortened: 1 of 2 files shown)


nlopin (approved):

nlopin:
I'd rather hide it on small screens

sergiocarracedo:
> I'd rather hide it on small screens

Yes, we are do it (check the video i added) but the padding it hardcoded and used a value not enough before hide it

PR #1628, opened 2025-04-15
feat: add new banner component

Nicotelo:
## Description

I'm adding a new banner component, that has two bottons: one meant for an upsell request, the other for redirection.
The banner will have 2 variants as well, depending on the usage of the feature

Warning:
image

Info:
image

Critical:
image

Options: 
image

### Figma Link

[Link to Figma Design](https://www.figma.com/design/Vfp0dMcoBEM2f8vcoEMEUq/Invoices-in-Projects?node-id=2163-29933&t=yzxcNk9KHyLQ58lC-4)

Slack conversation: https://factorialteam.slack.com/archives/C06QT46115K/p1744712639367679
---

## Type of Change

---

- [ ] Component added to the

### DIFF
--- a/packages/react/src/experimental/Alert/index.tsx
+++ b/packages/react/src/experimental/Alert/index.tsx
@@ -1,5 +1,7 @@
 import { Button } from "@/components/Actions/Button"
-import Share from "@/icons/app/Share"
+import { Icon } from "@/factorial-one"
+import { ExternalLink } from "@/icons/app"
+import { cn, focusRing } from "@/lib/utils"
 import { cva, type VariantProps } from "cva"
 import { AlertAvatar } from "../Information/Avatars/AlertAvatar"
 
@@ -42,20 +44,23 @@ const buttonVariants: Record<AlertVariant, "outline" | "promote" | "critical"> =
 interface AlertProps extends VariantProps<typeof alertVariants> {
   title: string
   description: string
-  buttonPrimaryLabel: string
-  buttonSecondaryLabel: string
-  onRequestClick?: () => void
-  onSeeClick?: () => void
+  primaryAction: {
+    label: string
+    onClick?: () => void
+  }
+  secondaryAction: {
+    label: string
+    href: string
+  }
+
   variant: AlertVariant
 }
 
 export const Alert = ({
   title,
   description,
-  buttonPrimaryLabel,
-  buttonSecondaryLabel,
-  onRequestClick,
-  onSeeClick,
+  primaryAction,
+  secondaryAction,
   variant = "info",
 }: AlertProps) => {
   return (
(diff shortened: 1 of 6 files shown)


packages/react/src/experimental/Alert/index.tsx
```diff
@@ -0,0 +1,106 @@
...
+  }
+
+interface AlertProps extends VariantProps<typeof alertVariants> {
+  title: string
+  description: string
+  primaryAction: {
+    label: string
+    onClick?: () => void
+  }
+  secondaryAction: {
```
nlopin, line 51: We always have a link here, not an action. Let's call it `link`. 

I see it is marked as mandatory here, but in the code you always check whether or not it exists. It must be optional prop then

packages/react/src/experimental/Alert/index.tsx
```diff
@@ -0,0 +1,106 @@
...
+  {
+    info: "outline",
+    warning: "promote",
+    critical: "outline",
+  }
+
+interface AlertProps extends VariantProps<typeof alertVariants> {
+  title: string
+  description: string
+  primaryAction: {
```
nlopin, line 47: let' call it just `action` since there is no secondary action on the page. From the code, it looks like this field it optional

nlopin (commented):
One thing is that we already have an [`Alert` component](https://one.factorial.dev/?path=/docs/components-alert--documentation). How is this different? why do we need another one?

Nicotelo:
hey @nlopin , context for another Alert, in this conversation
https://factorialteam.slack.com/archives/C06QT46115K/p1744712639367679
I was told it's going to be deprecated

packages/react/src/experimental/Alert/index.tsx
Nicotelo: Hmm, in the [design](https://www.figma.com/design/pZzg1KTe9lpKTSGPUZa8OJ/Components?node-id=932-7586&t=zyr9jwIim2y3M00P-0) they don't allude to it being option at any point, would prefer to leave it as mandatory until further notice
I'll follow the same with the link, taking off the checkings

nlopin:
@dani-moreno @Nicotelo we need to find another name at least for some time. The current `Alert` is used in 5 places in the app, we cannot release a component with the same name. I suggest we go with the prefix and call it `OneAlert`

Nicotelo:
Hey @nlopin I talked to @dani-moreno about it
He would be in charge of deprecating and changing the uses of the old alert for the new one
I didn't add the deprecation of the old alert in this PR to be able to do the changes more calmly and not in a hurry
But if we all agree, I can implement the prefix idea yes ^^

nlopin:
As I understand, you want to merge this PR in the following days. It won't be possible without the migration. The prefix enables you to proceed

dani-moreno (approved):

PR #1638, opened 2025-04-16
feat: export filters from F1

sauldom102:


sergiocarracedo:
If this component will be used outside the Datacollection we should move it outside dataconnection folder and create stories for it

PR #1670, opened 2025-04-23
feat(DataCollection): Eq filter

siguenzaraul:
## Description

Added a new EqFilter component to the OneDataCollection system that allows users to select exactly one option from a list, with the ability to deselect their choice.

### Key Features
- Single-selection model (one option at a time)
- Clean UI with checkbox visuals
- Support for both static and async options
- Animated transitions for filter chips

### Following Existing Patterns
I've carefully matched the existing component structure, maintaining consistency with naming conventions, animations, TypeScript patterns, and general design language.

### Perfect for Boolean GraphQL Filters
This component was necessary for building interfaces that connect to GraphQL queries with boolean filters:

```graphql
{ 
 ats {
  questionTemplatesConnection(
    filter: {
      autoDisqualify: { eq: true }
      mandatory: { eq: false }
    }
  ) {
    nodes {
      id
      label
      ...
    }
  }
 }
}
```

The EqFilter provides a natural UI for these true/false filters, creating a seamless connection between interface and API.

### Use Cases
Ideal for status selectors, boolean filters, type selectors, and any scenario requiring single-option selection from a defined list.

https://github.com/user-attachments/assets/65e92b77-fd44-4000-8963-2377fd60268f

https://github.com/user-attachments/assets/b5c5e9b6-feae-4f7e-92e5-42fdda71356d

https://github.com/user-attachments/assets/7bf68077-9585-4e31-8cee-54888793f1b1

---

## Type of Change

---


sergiocarracedo:
As the filtering type system will change in https://github.com/factorialco/factorial-one/pull/1743 i guess we should recreate this feature with the new one.
I was also thinking that is not a new filter, is the 'in' filter with an option to make the selection single

PR #1677, opened 2025-04-28
feat: add activity button in Sidebar

sauldom102:
## Description

We need to add support to show the activity button in the Sidebar.

## Screenshots

Screenshot 2025-04-28 at 10 28 37


sergiocarracedo (approved):

packages/react/src/experimental/Navigation/Sidebar/Footer/index.tsx
```diff
@@ -0,0 +1,67 @@
...
+import { Badge } from "@/experimental/exports"
+import { PersonAvatar } from "@/experimental/Information/Avatars/PersonAvatar"
+import { Circle as CircleIcon, Feed as FeedIcon } from "@/icons/app"
+import { cn, focusRing } from "@/lib/utils"
+import { Dropdown, DropdownItem } from "../../Dropdown"
+
+interface SidebarFooterProps {
+  user: {
+    firstName: string
+    lastName: string
```
nlopin, line 11: I now start to question this decision we took. We also have preferred name in the system and I think we should take one string containing the full name desired

packages/react/src/experimental/Navigation/Sidebar/Footer/index.tsx
```diff
@@ -0,0 +1,67 @@
...
+              lastName={user.lastName}
+              size="xsmall"
+            />
+            <span className="min-w-0 truncate text-f1-foreground">
+              {user.firstName} {user.lastName}
+            </span>
+          </button>
+        </Dropdown>
+      </div>
+      {showActivityButton && (
```
nlopin, line 49: what is the plan? use it with a FF? if so, we need to delete this prop after the release

nlopin (approved):

packages/react/src/experimental/Navigation/Sidebar/Footer/index.tsx
sauldom102: Yes! That's the plan, after it's fully released it we are going to delete this prop and condition from here

packages/react/src/experimental/Navigation/Sidebar/Footer/index.tsx
sauldom102: We need to pass `firstName` and `fullName` separately for the avatars, because we show the initials there in case the user doesn't have an avatar image set, we could actually try to change the approach there in the avatar to take the initials from the full name though

PR #1683, opened 2025-04-29
Add support to fill container with AvatarList component

sauldom102:
## Description

We need to support an alternative way of visualizing a list of avatars that consists of the entire parent container to be filled with avatars. We need it for this widget that we're about to create:

Screenshot 2025-04-29 at 09 51 23

## Screenshots

Screenshot 2025-04-29 at 09 51 50

### DIFF
--- a/packages/react/src/experimental/Information/Avatars/AvatarList/index.tsx
+++ b/packages/react/src/experimental/Information/Avatars/AvatarList/index.tsx
@@ -76,10 +76,10 @@ type Props = {
   remainingCount?: number
 
   /**
-   * Whether to fill the container with avatars.
-   * @default false
+   * The layout of the avatar list.
+   * @default "compact"
    */
-  fillContainer?: boolean
+  layout?: "fill" | "compact"
 }
 
 export const AvatarList = ({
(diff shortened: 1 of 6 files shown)


packages/react/src/experimental/Information/Avatars/AvatarList/index.tsx
```diff
@@ -73,6 +74,12 @@ type Props = {
    * The remaining number to display.
    */
   remainingCount?: number
+
+  /**
+   * Whether to fill the container with avatars.
+   * @default false
+   */
+  fillContainer?: boolean
```
nlopin, line 82: what do you think about having this prop a string with possible layouts? 

For example, `layout: "default" | "fill"`. The default is a bad name and we can come up with a clear one. This will keep the API size untouched in case we need to add another layout for the component

packages/react/src/ui/OverflowList/index.tsx
```diff
@@ -277,6 +284,8 @@ const OverflowList = function OverflowList<T>({
 
   const showOverflowDropdown = overflowItems.length > 0
 
+  console.log({ overflowItems: overflowItems.length })
```
nlopin, line 287: ❌

sauldom102 (commented):
Thank you for your review @nlopin! Applied all the suggestions you told me, it looked a bit confusing indeed

packages/react/src/experimental/Information/Avatars/AvatarList/index.tsx
```diff
@@ -73,6 +74,12 @@ type Props = {
    * The remaining number to display.
    */
   remainingCount?: number
+
+  /**
+   * The layout of the avatar list.
```
nlopin, line 79: let's add a short description what each means?

nlopin (approved):

PR #1684, opened 2025-04-29
Add support to show emoji in widget highlight items

sauldom102:
## Description

We need to support showing an emoji in widget highlight items for one new widget that we'll be creating.

Screenshot 2025-04-29 at 09 51 23

## Screenshots

Screenshot 2025-04-29 at 10 22 49

### DIFF
--- a/packages/react/src/ui/indicator.tsx
+++ b/packages/react/src/ui/indicator.tsx
@@ -1,3 +1,4 @@
+import { EmojiImage } from "@/lib/emojis"
 import { forwardRef } from "react"
 import { Icon, IconType } from "../components/Utilities/Icon"
 import { cn } from "../lib/utils"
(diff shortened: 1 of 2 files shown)


packages/react/src/ui/indicator.tsx
```diff
@@ -21,11 +23,12 @@ export const Indicator = forwardRef<HTMLDivElement, IndicatorProps>(
           >
             {label}
           </p>
-          {icon && (
+          {icon && !emoji && (
             <span className={cn("flex", color)}>
               <Icon icon={icon} />
             </span>
           )}
+          {emoji && !icon && <EmojiImage emoji={emoji} size="md" />}
```
nlopin, line 31: if i pass both emoji and icon, I get nothing. Let's make impossible states impossible. For that, we have to modify the prop:
```
type IndicatorProps = {
  content: string
  label: string
  color?: string
  emoji: string
} | {
  content: string
  label: string
  color?: string
  icon: IconType
}
```

We explicitly allow only `icon` or `emoji` prop, making the situation when both things are passed impossible.

PR #1688, opened 2025-04-29
Rename avatar name selector and some details

Rogermax:
## Description

This PR do 3 things:
* Refactor -> rename AvatarNameSelector to ComboBox
* Fix -> a small fix to remove unnecsary scroll on meta actions
* Feature -> Add a predefined action for ResouceHeader actions metadata (Copy)

### New copy action!

https://github.com/user-attachments/assets/c302e812-12a7-4a7c-88bc-9e3c918c5331

### Fix remove unnecesary scroll

Before:

Screenshot 2025-04-29 at 12 53 58

### ComboBox

Screenshot 2025-04-29 at 12 55 10

## Type of Change


packages/react/src/experimental/Information/Headers/Metadata/index.tsx
```diff
@@ -55,6 +74,26 @@ function MetadataItem({ item }: { item: MetadataItem }) {
   const isAction = item.actions?.length
   const hasHover = isAction || isList
 
+  const getValueToCopy = (value: MetadataItemValue): string => {
```
nlopin, line 77: I think we should not assume what users want to copy, especially for complex cases like lists

Rogermax: But then do not use this default.. use a custom copy action no?

packages/react/src/experimental/Forms/ComboBox/index.tsx
nlopin, line 1: After a discussion with the team our recommendation is to use name `EntitySelect` (see more in the PR summary)

I'd also consider naming folders here without the `ComboBox` prefix (e.g. `ListTag` instead of `ComboBoxListTag`) since they are contextual and not exported to the user

packages/react/src/experimental/Information/Headers/Metadata/index.tsx
nlopin: On the type level you do not allow to create a custom copy action. If you set `type: copy`, you can't set anything else. an optional `onClick` prop will be a good middle ground

packages/react/src/experimental/Information/Headers/Metadata/index.tsx
```diff
@@ -55,6 +74,26 @@ function MetadataItem({ item }: { item: MetadataItem }) {
...
+      case "dot-tag":
+        return value.label
+      case "date":
+        return value.formattedDate
+      case "tag-list":
+        return value.tags.join(", ")
+      case "data-list":
+        return value.data.join(", ")
+      default:
+        return "" // Should not happen for copyable types
```
nlopin, line 93: you can do TS exhaustive check:
`const _exhaustiveCheck: never = value.type`

packages/react/src/components/Actions/Button/copy.tsx
nlopin, line 1: what is the difference from the previous implementation we had in the Metadata?

nlopin (dismissed):
It would be better to have a PR for each feature. With the current approach we will have to maintain several completely unrelated discussions.

We fully agree that the `AvatarNameSelector` does not convey that the component is generic and can be used with any entities. At the same time `ComboBox` is too generic and convey that the component is a simple atomic component. The issue is that the component is not simple - it has filters, groups, etc. We consider `EntitySelect` is a better name for it

packages/react/src/components/Actions/Button/copy.tsx
Rogermax: The effect of the icon change.. plus the feature of copying in the window.navigator.clipboard.writeText(valueToCopy) this is the reason why we don't expose a onClick.. because this component is doing all we need. if you need to to do custom things.. simply do not use this type

packages/react/src/experimental/Forms/ComboBox/index.tsx
Rogermax: @danigonza @smalonso fyi it will be called `EntitySelect` not `ComboBox`

packages/react/src/experimental/Information/Headers/Metadata/index.tsx
nlopin: After your answer to the copy button thread, I see why the `onClick` is missing. Can user at least pass custom value they want to copy? I think it is crucial to have this part open for customization, we cannot see all possible situations

packages/react/src/experimental/Information/Headers/Metadata/index.tsx
Rogermax: nice hack! done

sergiocarracedo (approved):

PR #1729, opened 2025-05-02
feat(DataCollection): support multiple dotTags

siguenzaraul:
## Description

Enhance dotTag rendering to support multiple tags and update story examples

Captura de pantalla 2025-05-06 a las 20 47 50

Captura de pantalla 2025-05-02 a las 11 09 16

### Figma Link

[Link to Figma Design](https://www.figma.com/design/nWV0d6jAjXRhW5JrT9HK22/Sandbox---Data-Collection?node-id=285-157440&t=8y53svlNwyumG4re-4)

---

## Type of Change


sergiocarracedo:
We need to use the component <OverflowList> to ensure the dotTags dont wraps to the next line.
The cell type needs also an option to define the `max` dotTags to render (Overflow list has priority)

This solution will apply to other cell types like Status, and Tag

packages/react/src/experimental/Information/Tags/Tag.tsx
```diff
@@ -0,0 +1,91 @@
...
+  | ({
+      type: "balance"
+      status: "positive" | "neutral" | "negative"
+      text: string
+      onClick?: () => void
+    } & BalanceTagProps &
+      WithTooltipDescription)
+
+export const Tag = ({ tag }: { tag: TagVariant }): ReactNode => {
+  switch (tag.type) {
```
sergiocarracedo, line 47: Switch is not usually a good practice, it is better to use a something like

```
const tagTypes = {
  dotTag: <DotTag />,
  person: <Person />
  ....
} as const
```

and with that you can do it more generic, something like to make the tag component accept anything related to the tagTypes without need to use conditionals or define the types manually

```
export const Tag = <T extends keyof typeof tagTypes>({ type, ...rest  }: { type: T} & ComponentProps<(typeof tagTypes)[T]> }): ReactNode => {
  const Component = tagTypes[type]
if (!Component) {
   return 'Invalid tag type
}
  return <Component  {...rest} />
}

File: packages/react/src/experimental/Information/Tags/TagList/TagCounter.tsx [hunk 2]
```diff
@@ -0,0 +1,88 @@
...
+    default:
+      return ""
+  }
+}
+
+const getTagDescription = (tag: TagVariant): string | undefined => {
+  if ("description" in tag && typeof tag.description === "string") {
+    return tag.description
+  }
+  return undefined
```
sergiocarracedo, line 85: IF you need this you can define it in the tagTypes

const tagTypes = {
  dot: {
     component: <Dot/>
     displayName: (tag) -=> tag.text
     description: (tag) => tag.description     
  }
 ....

The underlaying idea is to put all the things related to a tag type togheter and if we need to add a new one need to review all the code and add ifs with the risk of forgeting something

sergiocarracedo (changes requested):

packages/react/src/experimental/Information/Tags/TagList/TagCounter.tsx
siguenzaraul: I just removed the display name because it didn't make sense to repeat the label twice. Now we'll keep the description, as it's necessary and useful to explain the data, but duplicating the data itself doesn't add value

packages/react/src/experimental/Information/Tags/Tag.tsx
siguenzaraul: I made some changes to make it more generic, but I wasn't able to implement exactly what you suggested 😢

sergiocarracedo (changes requested):

sergiocarracedo (approved):

PR #1743, opened 2025-05-05
Datacollection modular filters and date filter

sergiocarracedo:
## Description

Rewrite (refactor) the filter types to decouple the code (dont include specific behavior of a filter in the common code via ifs)
The new system uses a declarative way to define filter types (check filters.mdx story for more info)

e.g.
```
export const dateFilter: FilterTypeDefinition<string, DateFilterOptions> = {
  emptyValue: "",
  render: (props) => <DateFilter {...props} />,
  isEmpty: (value) => !value || value === "",
  chipLabel: (value) => value,
  options: {},
}
```
-feat: adds the date filter 
-docs: adds tecnical documentation on storybook about who filters, presets and filter types works

### Figma Link

---

## Type of Change

---


packages/react/src/experimental/OneDataCollection/Filters/Components/FilterChipButton.tsx
```diff
@@ -0,0 +1,73 @@
...
+import { filterTypes } from "../FilterTypes"
+import type { FilterValue, FiltersDefinition } from "../types"
+
+/**
+ * Animated chip component that displays an active filter with its current value.
+ * Uses discriminated unions to properly type the filter values based on filter type.
+ *
+ * Note: This component uses minimal type assertions that are necessary due to
+ * TypeScript limitations with generics across discriminated unions. These assertions
+ * are type-safe because they are guarded by the filter.type check.
```
nlopin, line 16: do you want to keep the comments?

PR #1786, opened 2025-05-09
Feat/date navigation

sergiocarracedo:
## Description

- feat: Add navigation filters modular system to data collection
- feat: Implements DateNavigation filter based in OneDatePicker
- docs: Adds date navigation filters docs
- docs: Adds docs about how to add a new filter\

https://github.com/user-attachments/assets/c4b43bfd-6eba-4f8c-8774-f5dcb433efd0

### Figma Link

---

## Type of Change


eliseo-juan (approved):
Everything makes sense for me!

packages/react/.storybook/preview.tsx
```diff
@@ -40,7 +40,7 @@ export const FactorialOne = (Story, { parameters }) => {
       }}
       l10n={{
         l10n: {
-          locale: parameters.l10n?.locale ?? "en",
+          locale: "es-ES", //parameters.l10n?.locale ?? "en",
```
nlopin, line 43: this should be rolled back, no?

packages/react/src/experimental/OneDataCollection/index.tsx
```diff
@@ -152,28 +197,46 @@ export const OneDataCollection = <
...
+    // Actions
     primaryActions,
     secondaryActions,
-    presets,
+    // Summary
+    totalItemSummary = (totalItems: number | undefined) => {
+      if (totalItems === undefined) {
+        return null
+      }
+      return `${totalItems} ${i18n.collections.itemsCount}`
```
nlopin, line 238: I'd consider extracting this default setter to a function. 

Or maybe compressing it down to `return totalItems !== undefined ? `${totalItems} ${i18n.collections.itemsCount}` : null`

nlopin (approved):

packages/react/.storybook/preview.tsx
sergiocarracedo: yes, my mistake :(

packages/react/src/experimental/OneDataCollection/index.tsx
sergiocarracedo: I can not extract it to a function as depends in i18n which is defined inside the hook, but i can compress it in a single line

nlopin: true, my bad

PR #1802, opened 2025-05-12
feat(dataCollection): implement single select functionality in inFilter

siguenzaraul:
### Added a singleSelect property to the InFilterOptions interface that restricts users to selecting only one option at a time
Modified the InFilter component to:
- Handle single selection behavior (deselecting previous choice when a new one is selected)
- Disable the "Select all" button when in single-select mode
- Updated the role filter in the stories to use this new single-select functionality

https://github.com/user-attachments/assets/c55e9ebe-955f-4d03-9c64-9817ef01ef48

https://github.com/user-attachments/assets/613bcf01-e767-4914-af03-c9722e3f1d46

### Figma Link

---

## Type of Change

### DIFF
--- a/packages/react/src/experimental/OneDataCollection/Filters/FilterTypes/InFilter/InFilter.tsx
+++ b/packages/react/src/experimental/OneDataCollection/Filters/FilterTypes/InFilter/InFilter.tsx
@@ -46,6 +46,7 @@ type InFilterComponentProps<T = unknown> = FilterTypeComponentProps<
  *         { value: "active", label: "Active" },
  *         { value: "inactive", label: "Inactive" }
  *       ]
+ *       singleSelect: false,
  *     }
  *   }}
  *   value={["active"]}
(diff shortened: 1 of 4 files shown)


packages/react/src/experimental/OneDataCollection/Filters/FilterTypes/InFilter/InFilter.tsx
```diff
@@ -136,6 +137,11 @@ export function InFilter<T extends string>({
   const showSearch = options.length > 0 && !isLoading
 
   const handleSelectAll = () => {
+    if (schema.options.singleSelect) {
```
nlopin, line 140: we shouldn't show the button at all

nlopin (changes requested):
The feature requires visual changes as well as accessibility updates.

Saying "Single select" we imply radio button behavior, therefore UI elements must look like radio buttons and have corresponding role (currently the role is set to "checkbox")

packages/react/src/experimental/OneDataCollection/Filters/FilterTypes/InFilter/InFilter.tsx
siguenzaraul: we are not showing the button, this was just a double check

siguenzaraul:
> The feature requires visual changes as well as accessibility updates.
> 
> Saying "Single select" we imply radio button behavior, therefore UI elements must look like radio buttons and have corresponding role (currently the role is set to "checkbox")

I’ll double-check this with the design team. I was originally told to match the ‘inFilter’ design, but I think using radio buttons makes more sense

PR #1809, opened 2025-05-13
feat: internal component select asList mode

sergiocarracedo:
## Description

- feat: Add `asList` mode in the select. This mode removes the trigger and shows always the list

image

---

## Type of Change

---


dani-moreno (approved):

sergiocarracedo:
> Have you considered creating a new component? I see they intersect in functionality, however we have to manipulate visuals with _if_ which makes me think it could be a different component. Accessibility wise it should be a list of checkboxes or radio buttons, since there is no "select" interaction like open/close

I think about, I tried to avoid create a new component that which behavior is mostly the same and visual speaking is the same

Regarding the select interaction, the HTML native select has no insteraction to open/close when is in multiple mode https://www.w3schools.com/tags/tryit.asp?filename=tryhtml_select_multiple

But we can refactor it and create a `SelectList` (or a better name) component

nlopin (approved):

PR #1835, opened 2025-05-16
fix: pageHeader prev next button reload page

sergiocarracedo:
## Description

Fixes navigation and back button that reload the page after click them

---

## Type of Change


sergiocarracedo:
To discuss: I think the Button component should accept the `href` prop and internally manage if the html element rendered (button or anchor). I know the rol of a link and a button are different but we are using buttons to navigate in this case the behavior is like an anchor (it should open in a new tab using the middle button, etc)

CarlosBM18 (approved):

PR #1842, opened 2025-05-19
feat: add new activity list component and icon avatar

CarlosBM18:
## What?

Adding the new ActivityList component, as well as its Skeleton.

Also, the new IconAvatar component is needed for the ActivityList.

## Demo

https://github.com/user-attachments/assets/dcf4c326-7832-4cc9-9af7-cec54f678452


packages/react-native-playground/components/Activity/ActivityItem/index.stories.tsx
```diff
@@ -0,0 +1,182 @@
...
+  component: ActivityItem,
+  argTypes: {
+    isUnread: {
+      control: "boolean",
+    },
+    onPress: { action: "pressed" },
+  },
+  args: {
+    id: "1",
+    date: formatDate(new Date()),
```
sergiocarracedo, line 40: Please don't use variable values as they will cause false positive in visual tests

packages/react-native-playground/components/Activity/ActivityItem/index.stories.tsx
```diff
@@ -0,0 +1,182 @@
...
+} satisfies Meta<typeof ActivityItem>;
+
+export default meta;
+
+type Story = StoryObj<typeof meta>;
+
+export const Default: Story = {
+  args: {
+    id: "1",
+    date: formatDate(new Date()),
```
sergiocarracedo, line 63: Please don't use variable values as they will cause false positive in visual tests

packages/react-native-playground/components/Activity/ActivityItem/index.stories.tsx
```diff
@@ -0,0 +1,182 @@
...
+    id: "4",
+    description: undefined,
+  },
+};
+
+export const ActivityItemList: Story = {
+  args: {
+    // Adding default args to fix type error
+    id: "1",
+    date: formatDate(new Date()),
```
sergiocarracedo, line 100: Please don't use variable values as they will cause false positive in visual tests

packages/react-native-playground/components/Activity/ActivityItem/index.stories.tsx
```diff
@@ -0,0 +1,182 @@
...
+    title: "Default Title",
+    category: "Default",
+    onPress: () => {},
+  },
+  render: () => {
+    const activityItems = [
+      {
+        id: "1",
+        date: formatDate(new Date(Date.now() - 1000 * 60 * 30)), // 30 minutes ago
```
sergiocarracedo, line 109: Please don't use variable values as they will cause false positive in visual tests

sergiocarracedo (changes requested):

packages/react-native-playground/components/Activity/ActivityItem/index.stories.tsx
CarlosBM18: Sure, I'll update those.

CarlosBM18: Done!

sergiocarracedo (approved):

PR #1855, opened 2025-05-21
fix: add babel runtime and remove unecessary type for storybook

dmaza81:
## Description

There are a couple of errors trying to up the storybook for react-native-one if there is nothing cached

### First error:
Screenshot 2025-05-21 at 13 19 16

### Second error:
Screenshot 2025-05-21 at 13 21 45

https://github.com/facebook/metro/issues/1047#issuecomment-1657200840

---

## Type of Change


sergiocarracedo (approved):

packages/react-native-playground/.storybook/main.ts
```diff
@@ -1,6 +1,4 @@
-import { StorybookConfig } from "@storybook/react-native";
-
-const main: StorybookConfig = {
```
CarlosBM18, line 3: Why are we removing this? This is just a type to enforce the structure of this object.

CarlosBM18 (changes requested):

packages/react-native-playground/.storybook/main.ts
dmaza81: Yep, it seems there is a problem with the node version and the pnpm, I will do a doble check ASAP 🙏

dmaza81:
We are going to close. We are unsing the monorepo installing the dep inside of the individual package, so that was the mistake

PR #1874, opened 2025-05-23
feat: add actions to Select

Rogermax:
## Description

Add the posibility to have actions in the Select component

Screenshot 2025-05-23 at 10 14 38

Also added plus icon:
Screenshot 2025-05-23 at 12 18 13

### Figma Link

[Link to Figma Design](https://www.figma.com/design/Ue0qdvB0z8aVax0zWP84yy/Sales?node-id=8697-42052&t=mDcLPxBXuQHQTkVW-0)

---

## Type of Change

Maintenance / Bug Fix / Other


packages/react/src/ui/Select/components/SelectContent.tsx
```diff
@@ -121,7 +121,7 @@ const SelectContent = forwardRef<
           asList || virtualReady ? "" : "opacity-0"
         )}
         style={{
-          height: virtualizer.getTotalSize(),
+          height: virtualizer.getTotalSize() + 8, // Add 8px to the height to account for the padding
```
dmaza81, line 124: Magic number.
Would you consider to set this in a constant with a proper name then we can remove the comment as well and doing the code more readable

packages/react/src/experimental/Forms/Fields/Select/index.tsx
```diff
@@ -226,9 +238,27 @@ export const Select = forwardRef<HTMLButtonElement, SelectProps<string>>(
...
+                        onClick={action.onClick}
+                        icon={action.icon}
+                        label={action.label}
+                        disabled={action.disabled}
+                      />
+                    ))}
+                  </div>
+                </>
+              )
+            }
```
dmaza81, line 258: Would you consider to move this to a component with the action logic? or at least in a function?

packages/react/src/experimental/Forms/Fields/Select/index.tsx
```diff
@@ -9,7 +11,7 @@ import {
   VirtualItem,
 } from "@/ui/Select"
 import { forwardRef, useEffect, useMemo, useRef, useState } from "react"
-import { Icon } from "../../../../components/Utilities/Icon"
+import { Icon, IconType } from "../../../../components/Utilities/Icon"
```
dmaza81, line 14: I guess we are mapping components with parth or alias so we could do:
`import { Icon, IconType } from "@components/Utilities/Icon"`

dmaza81 (commented):
Good work @Rogermax ! just some comments

packages/react/src/experimental/Forms/Fields/Select/index.tsx
Rogermax: not sure why we are not doing it.. let me check if alias is set, in that case I'll change it

packages/react/src/experimental/Forms/Fields/Select/index.stories.tsx
```diff
@@ -2,6 +2,7 @@ import type { Meta, StoryObj } from "@storybook/react"
 import { fn } from "@storybook/test"
 import { Select, SelectProps } from "./index"
 
+import { Plus } from "lucide-react"
```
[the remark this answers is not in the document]
Rogermax: Ok! doing it.

Why don't we upload all that icons in a single commit to have it available in the code?

packages/react/src/experimental/Forms/Fields/Select/index.tsx
```diff
@@ -226,20 +292,8 @@ export const Select = forwardRef<HTMLButtonElement, SelectProps<string>>(
...
-                    clearable
-                    value={searchValue}
-                    key="search-input"
-                    ref={searchInputRef}
-                  />
-                </div>
-              )
-            }
+            bottom={selectContentBottom}
+            top={selectContentTop}
```
nlopin, line 296: it's not a react way of doing things. We either write a component or use inline logic. I don't see a win of creating a memoized value here

Rogermax: you prefer as it was before? https://github.com/factorialco/factorial-one/pull/1874#discussion_r2104007770
It's just to make it more readble I suppose

Rogermax: You prefer something more <Component..>?

Rogermax: done

packages/react/src/experimental/Forms/Fields/Select/index.stories.tsx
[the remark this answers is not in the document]
nlopin: We do, but there are always changes. We need to automate it

nlopin (approved):

PR #1887, opened 2025-05-26
feat: add emojiAvatar component

dmaza81:
## Description

Adding the EmojiAvatar without motion on Factorial-one mobile

### DIFF
--- a/packages/react-native/src/components/EmojiAvatar/__snapshots__/index.spec.tsx.snap
+++ b/packages/react-native/src/components/EmojiAvatar/__snapshots__/index.spec.tsx.snap
@@ -1,35 +1,35 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
-exports[`Button Snapshot - different sizes 1`] = `
+exports[`Button Snapshot - different sizes text emoji and diferent size 1`] = `
 <View
   className="flex aspect-square items-center justify-center border border-solid border-f1-border-secondary bg-f1-background dark:bg-f1-background-inverse-secondary w-6 h-6 rounded-sm"
-/>
+>
+  <Text
+    className="text-xs"
+  >
+    test
+  </Text>
+</View>
 `;
 
-exports[`Button Snapshot - different sizes 2`] = `
-<View
-  className="flex aspect-square items-center justify-center border border-solid border-f1-border-secondary bg-f1-background dark:bg-f1-background-inverse-secondary w-9 h-9 rounded-md"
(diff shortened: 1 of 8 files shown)


packages/react-native/src/components/EmojiAvatar/__snapshots__/index.spec.tsx.snap
CarlosBM18, line 1: 💅

packages/react-native/src/components/EmojiAvatar/index.tsx
```diff
@@ -0,0 +1,37 @@
...
+  sm: "w-6 h-6 rounded-sm",
+  md: "w-9 h-9 rounded-md",
+  lg: "w-10 h-10 rounded-lg",
+};
+
+const emojiSize = {
+  sm: { height: 16, width: 16, text: "text-xs" },
+  md: { height: 20, width: 20, text: "text-sm" },
+  lg: { height: 24, width: 24, text: "text-md" },
+} as const;
```
CarlosBM18, line 21: I think we should use NativeWind here, for the height and width

packages/react-native/src/lib/emojis.tsx
```diff
@@ -0,0 +1,45 @@
...
+    <SvgUri
+      onError={() => setError(true)}
+      width={size?.width}
+      height={size?.height}
+      uri={emojiEntity.url}
+    />
+  ) : (
+    <Text className={cn(size?.text, className)} key={emoji}>
+      {emoji}
+    </Text>
```
CarlosBM18, line 34: nice! I like to see this controlled as a fallback

packages/react-native/src/components/EmojiAvatar/index.tsx
dmaza81: I'm afraid we can not use it on the SVGUrl, that why I have set as width and height

packages/react-native/src/components/EmojiAvatar/__snapshots__/index.spec.tsx.snap
dmaza81: I would like to have an snapshot but it is almost imposible with the Error from the network 😢

packages/react-native/src/components/EmojiAvatar/index.tsx
dmaza81: Forget about it, I have created a wrap to do the w- and h-

PR #1890, opened 2025-05-26
fix: add support to hide reactions button in CommunityPost component via prop

sauldom102:


packages/react/src/experimental/Information/Communities/Post/CommunityPost/index.tsx
```diff
@@ -54,6 +54,8 @@ export type CommunityPostProps = {
 
   onClick: (id: string) => void
 
+  canReact?: boolean
```
nlopin, line 57: what do you think about following the same pattern we have for `noVideoPreload`? like, `noReactions` or something

nlopin (approved):

PR #1924, opened 2025-05-28
feat: manage open status in Entity Selector

RafaMellado:
## Description

We want to unify the open status in Entity Selector. As you can see, in Factorial, we're managing in the EntitySelector wrapper the open status https://github.com/factorialco/factorial/blob/main/frontend/src/modules/core/components/EntitySelectWrapper/index.tsx#L75

As this should be within the component, we're implementing this here

## Type of Change

### DIFF
--- a/packages/react/src/experimental/Forms/EntitySelect/index.tsx
+++ b/packages/react/src/experimental/Forms/EntitySelect/index.tsx
@@ -13,6 +13,23 @@ import {
 export const EntitySelect = (
   props: EntitySelectProps & { children?: React.ReactNode }
 ) => {
+  const [open, setOpen] = useState(
+    (props.alwaysOpen || props.defaultOpen) ?? false
+  )
+
+  const onOpenChange = (open: boolean) => {
+    setOpen(open)
+    props.onOpenChange?.(open)
+  }
+
+  useEffect(() => {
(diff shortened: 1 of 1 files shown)


packages/react/src/experimental/Forms/EntitySelect/index.tsx
```diff
@@ -13,6 +13,22 @@ import {
...
+  )
+
+  const onOpenChange = (open: boolean) => {
+    setOpen(open)
+    props.onOpenChange?.(open)
+  }
+
+  useEffect(() => {
+    if (props.defaultOpen && open) {
+      props.onOpenChange?.(true)
```
nlopin, line 27: why do we need this effect? is it to run the callback after the component is mount?

nlopin (approved):

packages/react/src/experimental/Forms/EntitySelect/index.tsx
RafaMellado: Yes, I want to run this only when the component is rendered

nlopin: could you add a comment?

nlopin (approved):

PR #1943, opened 2025-05-29
fix: change upsell icon color

ChloeDumit:
## Description

Fixing upsell icon color to be the same as promote

Screenshot 2025-05-29 at 11 30 24 AM

## Type of Change


nlopin (changes requested):
The solution will break after the next icon generation

PR #1948, opened 2025-05-29
fix(DataCollection): Truncate In Filter label in filters

dani-moreno:
## Description

Labels in the In Filter type are truncated only when the text has spaces between words. But if the label does not have spaces, it is not truncated, breaking the filters popover. Now it uses OneEllipsis so a tooltip is shown if the text is truncated.

### Before

image

### After

Screenshot 2025-05-30 at 17 53 56

---

## Type of Change

### DIFF
--- a/packages/react/vite/vitest.setup.ts
+++ b/packages/react/vite/vitest.setup.ts
@@ -26,3 +26,12 @@ if (typeof window !== "undefined") {
   window.HTMLElement.prototype.releasePointerCapture = () => {}
   window.HTMLElement.prototype.scrollIntoView = () => {}
 }
+
+// Mock ResizeObserver
+vi.stubGlobal(
+  "ResizeObserver",
+  class MockedResizeObserver {
+    observe = vi.fn()
+    disconnect = vi.fn()
+  }
+)


dani-moreno:
> Why not using OneEllipsis?

Ah gosh, you're right, I forgot about it. Just updated it to use `OneEllipsis` instead. It works like a charm!

packages/react/src/experimental/OneDataCollection/Filters/FilterTypes/InFilter/InFilter.tsx
```diff
@@ -193,8 +194,8 @@ export function InFilter<T extends string>({
                 )
               }}
             >
-              <span className="line-clamp-1 w-fit text-left">
-                {option.label}
+              <span className="max-w-[250px] flex-1 whitespace-nowrap">
```
nlopin, line 197: I tend to use _rem_ for widths because it grows with font size, making more space for text. Not sure if it is applicable here, but it could be a nice polishing touch

nlopin (approved):

PR #1951, opened 2025-05-30
refactor: remove tokens from react package

nlopin:
I noticed we still have color and spacing tokens in the React package codebase and it leaks to the code. 

The PR deletes those files and switches the usages to the `factorial-one-core` package


sauldom102 (approved):
Great! 🔥

PR #1954, opened 2025-05-30
Fix responsive layout for ClockInGraph

somebody32:
## Summary
- make `ClockInGraph` responsive so the pie chart scales with its container

## Testing
- `pnpm --filter @factorialco/factorial-one-react run vitest` *(fails: vitest not found)*


sergiocarracedo:
The ResponsiveContainer needs to mock ResizeObserver in tests

somebody32:
that is just me playing, feel free to ignore for now

PR #1970, opened 2025-06-02
feat: add create action to EntitySelect component

Rogermax:
## Description

add create action to EntitySelect component

https://github.com/user-attachments/assets/6395e150-757e-46ef-8a6c-774bad234832

Screenshot 2025-06-12 at 15 51 06


packages/react/src/experimental/Information/Tags/DotTag/index.tsx
```diff
@@ -1,5 +1,5 @@
-import { baseColors } from "@factorialco/factorial-one-core"
 import { forwardRef } from "react"
+import { baseColors } from "../../../../../../core/src/tokens/colors"
```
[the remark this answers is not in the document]
Rogermax: 😮 

good to know

Rogermax:
Closed because designer and mario decided to tackle this in a future in a different way.

packages/react/src/experimental/Forms/EntitySelect/Content/index.tsx
```diff
@@ -57,6 +59,8 @@ export const Content = ({
   disabled?: boolean
   hiddenAvatar?: boolean
   actions?: Action[]
+  onCreate?: (partialName: string) => void
+  onCreateLabel?: string
```
nlopin, line 63: Two thoughts here:
- those two props will be set together in most of the cases. It could be better to either fuse those two props into one or split the type definition and make the impossible state (only one of those two is set) impossible
- is creation the only way to use this top item? shouldn't we use more generic prop to keep the usage open?

nlopin (approved):

packages/react/src/experimental/Forms/EntitySelect/index.stories.tsx
```diff
@@ -702,3 +702,60 @@ export const WithActions = {
     )
   },
 }
+
+export const WithCreate = {
```
nlopin, line 706: can we make it open by default in the story? this way we can capture a screenshot with the item

packages/react/src/experimental/Forms/EntitySelect/CreateItem/index.stories.tsx
```diff
@@ -0,0 +1,30 @@
+import type { Meta, StoryObj } from "@storybook/react-vite"
+
+import { ComponentProps } from "react"
+import { fn } from "storybook/test"
+import { CreateItem } from "./index"
+
+const meta: Meta = {
```
nlopin, line 7: disable snapshot test either for this one (if you make the `WithCreate` story open by default) or the `WithCreate` story

this way we don't create several snapshots of the same thing

PR #1988, opened 2025-06-03
refactor/module avatar semantic

sergiocarracedo:
## Description

Module avatar is a component to render module icons, not arbitrary icons. This refactor deprecated the `icon` prop to use `module` prop that has a semantic meaning

## Implementation details
This refactor deprecated the `icon` prop to use `module` prop that has a semantic meaning


packages/react/src/experimental/Information/ModuleAvatar/ModuleAvatar.tsx
```diff
@@ -0,0 +1,86 @@
...
+ * @description A component that displays a module avatar
+ * @experimental
+ * @returns
+ */
+export function ModuleAvatar({ size = "md", ...props }: ModuleAvatarProps) {
+  if ("icon" in props) {
+    console.warn(
+      "ModuleAvatar:The `icon` prop is deprecated. Use the `module` prop instead."
+    )
+  }
```
nlopin, line 60: should we make a helper for deprecated props?

nlopin (approved):

PR #2026, opened 2025-06-05
feat: create new component avatar for mobile

dmaza81:
## Description

We have added the Avatar & Badge & RawTag & DotTag components to our F1 Mobile. 

Things we should improve:

1. There is no support for using mask when the badge is shown 
2. We should create the list components for Avatar
3. We should review all classes we are using. Im pretty sure some of them are irrelevant for Mobile


packages/react-native/src/components/Tags/RawTag/__snapshots__/index.spec.tsx.snap
```diff
@@ -23,7 +23,7 @@ exports[`RawTag Snapshot 1`] = `
       }
     }
     accessible={true}
-    className="flex flex-row items-center justify-start gap-0.5 rounded-full py-0.5 pr-2 text-base font-medium text-f1-foreground pl-2 border-[1px] border-solid border-f1-border-secondary"
+    className="flex flex-row items-center justify-start gap-0.5 rounded-full py-0.5 pr-2 pl-2 border-[1px] border-solid border-f1-border-secondary"
```
CarlosBM18, line 26: Could you try `border`? Without the [1px], I think it should do the same.

packages/react-native/src/components/Tags/RawTag/__snapshots__/index.spec.tsx.snap
```diff
@@ -113,7 +113,7 @@ exports[`RawTag Snapshot with icon 1`] = `
       }
     }
     accessible={true}
-    className="flex flex-row items-center justify-start gap-0.5 rounded-full py-0.5 pr-2 text-base font-medium text-f1-foreground pl-1 border-[1px] border-solid border-f1-border-secondary"
+    className="flex flex-row items-center justify-start gap-0.5 rounded-full py-0.5 pr-2 pl-1 border-[1px] border-solid border-f1-border-secondary"
```
CarlosBM18, line 116: same here

packages/react-native/src/components/Tags/DotTag/index.tsx
```diff
@@ -0,0 +1,54 @@
...
+  const backgroundColor =
+    "color" in props && props.color
+      ? `hsl(${baseColors[props.color][50]})`
+      : "customColor" in props && props.customColor;
+
+  if (!backgroundColor) return null;
+
+  return (
+    <BaseTag
+      classNameContainer="border-[1px] border-solid border-f1-border-secondary"
```
CarlosBM18, line 39: Could you try border? Without the [1px], I think it should do the same.

CarlosBM18 (changes requested):

PR #2036, opened 2025-06-06
fix: better grouping in ActivityItemList

sauldom102:
## Description

Set correct grouping for `ActivityItemList`. In this PR I'm also setting the same sorting for react-native too.

### DIFF
--- a/packages/react/src/lib/date.ts
+++ b/packages/react/src/lib/date.ts
@@ -1,8 +1,7 @@
 import {
+  differenceInDays,
   format,
   formatDistanceToNowStrict,
-  isThisMonth,
-  isThisWeek,
   isToday,
   isYesterday,
   startOfDay,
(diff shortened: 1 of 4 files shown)


packages/react/src/lib/date.ts
```diff
@@ -64,22 +63,23 @@ export const categorizeItemsByDate = <
...
-    } else if (isThisWeek(date)) {
+    } else if (diffDays <= 7) {
       groups.lastWeek.push(item)
-    } else if (isThisMonth(date)) {
+    } else if (diffDays <= 30) {
       groups.lastMonth.push(item)
     } else {
-      groups.other.push(item)
+      groups[date.getFullYear()] = [...(groups[date.getFullYear()] || []), item]
     }
```
sauldom102, line 83: It should be grouped by year number as well

dmaza81 (approved):
Good one, let use the same sort for those list 👍

PR #2048, opened 2025-06-10
feat: disable drag and drop in sidebar on touch screens

sergiocarracedo:
## Description

Drag and drop was triggered when the user tries to scroll the menu in touch screens

https://factorialmakers.atlassian.net/browse/FCT-31175

## Implementation details

add the useTouchScreen hook to detect that and remove the drag and drop on those cases


nlopin:
why? works badly?

sergiocarracedo:
> why? works badly?

it's something known time ago. is about the user experience. If you try to scroll the menu you do involuntary drag and drop. For simplicity we (design team) decided just to disable it on touch screens

PR #2051, opened 2025-06-10
fix(DataCollection): display tooltip with info icon & custom icon

dani-moreno:
## Description

The info prop for the table headers isn't working. Even though the icon is shown, no tooltip is displayed on hover. Also, some teams need to change the icon to depict something about the label without taking up much space, like this case of Compensations:

image

_PD: That's not the final design, but you get the idea, they need a different icon other than the info circle._

## Screenshots

image

_Info icon with a working tooltip_

image

_Custom icon_

## Implementation details

- Now, the tooltip is displayed only while hovering the icon. It was not working before as the tooltip was wrapping the entire header, and there are some interactive elements inside that was breaking the tooltip.
- Added a new prop to define a custom info icon.


nlopin (approved):
Can I activate the tooltip from the keyboard or hear it with screen-reader?

PR #2057, opened 2025-06-11
FCT 31289 allow to define default selected items in datacollections

sergiocarracedo:
## Description

Allow to pass defaultSelected items and groups in datacollection

### DIFF
--- a/packages/react/src/experimental/OneDataCollection/__stories__/selectable-bulk-actions.mdx
+++ b/packages/react/src/experimental/OneDataCollection/__stories__/selectable-bulk-actions.mdx
@@ -79,7 +79,7 @@ The callback receives the an object with the following properties:
 The `defaultSelectedItems` prop is an object that contains the default selected
 items.
 
-````tsx
+```tsx
 const { source, currentSelectable, setCurrentSelectable } = useDataCollection({
   ...
   selectable: (item) => item.id,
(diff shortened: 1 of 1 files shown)


packages/react/src/experimental/OneDataCollection/useSelectable.ts
```diff
@@ -66,6 +68,32 @@ export function useSelectable<
     new Map()
   )
 
+  /**
+   * Set the default selected items and groups
+   */
+  useEffect(() => {
+    if (defaultSelectedItems) {
+      if (isGrouped) {
+        ;(defaultSelectedItems.groups || []).forEach((defaultGroup) => {
```
nlopin, line 77: what do you think on using `for..of` loop here? I prefer if over callback driven `forEach` for readability. Also it will eliminate the need of `;`

sergiocarracedo: sure, and for..of is faster that foreach

packages/react/src/experimental/OneDataCollection/useSelectable.ts
```diff
@@ -66,6 +68,32 @@ export function useSelectable<
...
+      }
+
+      ;(defaultSelectedItems.items || []).forEach((item) => {
+        const record = data.records.find((record) => record.id === item.id)
+        if (record) {
+          handleSelectItemChange(record, item.checked)
+        }
+      })
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps -- we are checking deeply the defaultSelectedItems
```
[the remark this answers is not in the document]
sergiocarracedo: Nice one, thanks!!!

PR #2064, opened 2025-06-11
feat: Onclick dropdown footer

9Shuck:
## Description

We added onDropdownClick as a prop to sidebar footer. On frontend this function is not reaching to the parent so it's not working the prop.

## Implementation details

Adapt Sidebar main component to get footer onClick function.


packages/react/src/experimental/Navigation/Sidebar/Sidebar.tsx
```diff
@@ -1,11 +1,12 @@
 import { ScrollArea } from "@/ui/scrollarea"
-import { AnimatePresence, motion } from "motion/react"
-import { ReactNode } from "react"
+import { AnimatePresence, motion } from "framer-motion"
```
sergiocarracedo, line 2: please use 'motion/react' as we are migration to motion@12

sergiocarracedo (changes requested):

PR #2070, opened 2025-06-12
feat: sort dropdown items alphabetically between separators

rodrigo-picanco:
## Description
We've received some feedback from users that they wanted their dropdown alphabetically sorted. Sound crazy first, but I guess it's a nice little detail for the more detail oriented/OCD haver people (like me). 

Before
Screenshot 2025-06-12 at 13 44 03

After
Screenshot 2025-06-12 at 13 43 44

## Implementation details
Sorts items within each separator

### DIFF
--- a/packages/react/CHANGELOG.md
+++ b/packages/react/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Changelog
 
+## [1.92.3](https://github.com/factorialco/factorial-one/compare/factorial-one-react-v1.92.2...factorial-one-react-v1.92.3) (2025-06-11)
+
+
+### Bug Fixes
+
+* whitespace prewrap added to add wrapping and avoid issues when pre code block is added ([#2053](https://github.com/factorialco/factorial-one/issues/2053)) ([3099da6](https://github.com/factorialco/factorial-one/commit/3099da62c880efacb5a564ce847454a4bdb2732f))
+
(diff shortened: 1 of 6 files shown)


packages/react/src/experimental/Navigation/Dropdown/utils.ts
```diff
@@ -0,0 +1,24 @@
...
+      sortedItems.push(...currentGroup)
+      sortedItems.push(item) // Add the separator
+      currentGroup = [] // Reset for the next group
+    } else {
+      currentGroup.push(item)
+    }
+  }
+
+  // Add any remaining items in the last group
+  if (currentGroup.length > 0) {
```
nlopin, line 18: do we need the check? 

it is guaranteed it will be an empty array, we can safely run sort and push without the check

nlopin (approved):
I don't think sorted items is the best default behavior. 

It is usually better to have the default behavior as straightforward as possible. I'd be surprised to see a dropdown sorting the values I pass without explicit command.

I suggest we add `sorted` prop to the dropdown and make it opt-in

packages/react/CHANGELOG.md
sergiocarracedo, line 1: This file is autogenerated in the release process https://github.com/factorialco/factorial-one/blob/main/packages/react/docs/development/release-and-versioning.md

rodrigo-picanco:
Hey @sergiocarracedo @nlopin 

Sorry for the autogenerated files, I've run the build on local to test it directly on the app as well, just to see it. 

Anyway, If u don't feel confident about the sorting RN, this was a 5 min PR. 

I'll close it and we can revisit it later!

PR #2084, opened 2025-06-13
chore: mock images in storybook

sergiocarracedo:
## Description

Use local mock images in storybook stories for speed and to avoid changes in visual tests.


nlopin:
> Give me a day and I'll do it! (2 weeks ago)

@dani-moreno is it a blocker? we can replace images later. Moreover they will grind your eyes in production

PR #2123, opened 2025-06-20
fix: set correct spacing and flex for data collection list item description

sauldom102:
## Description

The description in data collection list row items wasn't shown properly.

## Screenshots

Screenshot 2025-06-20 at 12 47 53


packages/react/src/experimental/OneDataCollection/visualizations/collection/List/components/ItemTeaser.tsx
```diff
@@ -26,7 +26,7 @@ export const ItemTeaser = ({ title, avatar, description }: ItemTeaserProps) => {
           {description && description.length > 0 && (
             <div className="flex w-full flex-col text-base font-normal text-f1-foreground-secondary md:flex-row md:gap-1">
               {description.map((item, index) => (
-                <div key={index} className="flex min-w-0 flex-1 flex-row gap-1">
+                <div key={index} className="flex min-w-0 flex-row gap-1">
```
dmaza81, line 29: Should we remove `flex-row`? I guess flex-row is the value by default in web, isn't It?

dmaza81 (changes requested):

PR #2145, opened 2025-06-24
feat(RichTextEditor): add AIBlock extension and integrate into BasicT…

siguenzaraul:
…extEditor

### DIFF
--- a/packages/react/src/experimental/RichText/BasicTextEditor/extensions.tsx
+++ b/packages/react/src/experimental/RichText/BasicTextEditor/extensions.tsx
@@ -16,12 +16,20 @@ import {
   TypographyExtension,
   UnderlineExtension,
 } from "@/experimental/RichText/CoreEditor"
+import { AIBlockExtension } from "@/experimental/RichText/CoreEditor/Extensions/AIBlock"
 import { SlashCommandGroupLabels } from "@/experimental/RichText/CoreEditor/Extensions/SlashCommand"
+
+interface AIBlockConfig {
+  buttons: { type: string; emoji: string; label: string }[]
+  onClick: (type: string) => Promise<any>
+  title: string
+}
 
 export const createBasicTextEditorExtensions = (
   placeholder: string,
   toolbarLabels: ToolbarLabels,
-  groupLabels?: SlashCommandGroupLabels
+  groupLabels?: SlashCommandGroupLabels,
+  aiBlockConfig?: AIBlockConfig
 ) => [
   StarterKitExtension,
   UnderlineExtension,
@@ -34,8 +42,9 @@ export const createBasicTextEditorExtensions = (
   TextAlignExtension,
   LinkExtension,
   MoodTrackerExtension,
+  AIBlockExtension,
   PersistSelection,
   createPlaceholderExtension(placeholder),
   createAccessibilityExtension(placeholder),
-  createSlashCommandExtension(toolbarLabels, groupLabels),
+  createSlashCommandExtension(toolbarLabels, groupLabels, aiBlockConfig),
 ]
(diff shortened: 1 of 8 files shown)


packages/react/src/experimental/RichText/CoreEditor/Extensions/MoodTracker/index.tsx
```diff
@@ -114,49 +170,55 @@ export const MoodTrackerView: React.FC<NodeViewProps> = ({ node }) => {
...
+              <motion.div
+                className="text-f1-text-primary"
+                initial={{ opacity: 0, y: -10 }}
+                animate={{ opacity: 1, y: 0 }}
+                transition={{ delay: 0.1, duration: 0.3 }}
+              >
+                <div className="flex flex-col gap-2">
+                  {data.days.map((day, index) => {
+                    return (
+                      <motion.div
```
sauldom102, line 183: As we're not doing anything else here we could get rid of the `return`

```suggestion
                  {data.days.map((day, index) => (
                      <motion.div
```

packages/react/src/experimental/RichText/BasicTextEditor/extensions.tsx
```diff
@@ -16,26 +17,60 @@ import {
...
-  MoodTrackerExtension,
-  PersistSelection,
-  createPlaceholderExtension(placeholder),
-  createAccessibilityExtension(placeholder),
-  createSlashCommandExtension(toolbarLabels, groupLabels),
-]
+  groupLabels?: SlashCommandGroupLabels,
+  aiBlockConfig?: AIBlockConfig,
+  aiBlockLabels?: AIBlockLabels,
+  moodTrackerLabels?: MoodTrackerLabels
```
sauldom102, line 37: Are these labels going to be the same for every text editor? Maybe we should think of moving them to `FactorialOneProvider`

packages/react/src/experimental/RichText/BasicTextEditor/index.tsx
```diff
@@ -49,7 +58,10 @@ const BasicTextEditorComponent = forwardRef<
     extensions: createBasicTextEditorExtensions(
       placeholder,
       toolbarLabels,
-      slashCommandGroupLabels
+      slashCommandGroupLabels,
+      aiBlockConfig,
+      aiBlockLabels,
+      moodTrackerLabels
```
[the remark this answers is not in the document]
siguenzaraul: yeah we can have like a group called labels with all of them inside, instead of having a million different props, what do u think?

packages/react/src/experimental/RichText/CoreEditor/Extensions/SlashCommand/AvailableCommands.tsx
```diff
@@ -31,21 +33,149 @@ interface SlashCommandGroupLabels {
...
   labels: ToolbarLabels,
-  groupLabels: SlashCommandGroupLabels
+  groupLabels: SlashCommandGroupLabels,
+  aiBlockConfig?: AIBlockConfig
 ): CommandGroup[] => [
+  // Only include AI Block group if config is provided
+  ...(aiBlockConfig
+    ? [
+        {
+          title: aiBlockConfig.title || "AI Block",
```
sauldom102, line 59: What about this copy? We should have that localized

packages/react/src/experimental/RichText/CoreEditor/Extensions/SlashCommand/AvailableCommands.tsx
```diff
@@ -31,21 +33,149 @@ interface SlashCommandGroupLabels {
...
+          commands: [
+            // Add individual commands for each AI button using emoji
+            ...aiBlockConfig.buttons.map((button) => ({
+              title: button.label ?? "Generate",
```
sauldom102, line 63: Same here

sauldom102 (approved):

PR #2160, opened 2025-06-25
FCT-31574-datacollection-date-filter-calendar-overflows-the-space

sergiocarracedo:
## Description

Add the height option to avoid vertical scroll in calendar views


packages/react/src/experimental/OneDataCollection/Filters/Components/FilterContent.tsx
```diff
@@ -77,7 +78,7 @@ export function FilterContent<Definition extends FiltersDefinition>({
   }
 
   return (
-    <div className="relative flex w-full flex-col gap-1">
+    <div className={cn("relative flex w-full flex-col gap-1")}>
```
sauldom102, line 81: We don't need this here 👀

sergiocarracedo: not really

sauldom102 (approved):

PR #2174, opened 2025-06-27
fix: fix more modal issues

sauldom102:
## Description

A couple of issues regarding modals are fixed here:

- "Jumping" modal on production when showing it up because of re-render.
- Horizontal scroll because of long texts with no spaces that we were not cutting.

### DIFF
--- a/packages/react/src/ui/dialog.tsx
+++ b/packages/react/src/ui/dialog.tsx
@@ -35,11 +35,13 @@ const DialogContent = React.forwardRef<
     withTraslateAnimation?: boolean
   }
 >(({ className, children, withTraslateAnimation = true, ...props }, ref) => {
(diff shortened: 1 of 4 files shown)


packages/react/src/ui/dialog.tsx
```diff
@@ -35,12 +35,14 @@ const DialogContent = React.forwardRef<
...
   }
 >(({ className, children, withTraslateAnimation = true, ...props }, ref) => {
-  const [container, setContainer] = useState<HTMLElement | null>(null)
+  const [container, setContainer] = useState<HTMLElement | null>()
 
   useEffect(() => {
     setContainer(document.getElementById("content"))
   }, [])
 
+  if (container === undefined) return null
```
sauldom102, line 44: With this we are waiting for `setContainer` call to be done above, avoiding the modal to be rendered twice

PR #2197, opened 2025-06-30
fix(datacollection): deps on empty state

sergiocarracedo:
## Description

- fix: datacollection were not able to check for updates after enter in 'empty mode'
- fix: datacollection were not using the deps enter in 'empty mode'

https://github.com/user-attachments/assets/4099eb9a-b292-471b-aebf-7607331c9d4a

## Implementation details

The datacollection's visualizacion must be always mounted as the load and observable subscription happens in the visualizacion.
The datacolection emptyState useMemo must use the dataProvider as dep to check

### DIFF
--- a/packages/react/src/experimental/OneDataCollection/__stories__/item-actions.stories.tsx
+++ b/packages/react/src/experimental/OneDataCollection/__stories__/item-actions.stories.tsx
@@ -184,7 +184,7 @@ export const BasicActionsExample: Story = {
 // Example showing how actions can be used with card visualization
 export const CardActionsExample: Story = {
   parameters: {
-    i11y: {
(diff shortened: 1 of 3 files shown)


packages/react/src/experimental/OneDataCollection/__stories__/item-actions.stories.tsx
```diff
@@ -196,7 +196,6 @@ export const CardActionsExample: Story = {
       dataAdapter: {
         fetchData: () => Promise.resolve({ records: mockUsers }),
       },
-      //secondaryActions: buildActions(),
```
nlopin, line 199: should it be commented out?

PR #2221, opened 2025-07-04
feat(Icons): new icons missing from Gamma

dani-moreno:
## Description

Add a bunch of icons to F1. Most of them as a replacement for Gamma icons, so we can deprecate them.


sauldom102 (commented):
@dani-moreno can you add this sort icon too? 🙏🏼 
Screenshot 2025-07-04 at 15 48 19

nlopin (approved):

PR #2265, opened 2025-07-17
fix: input type extends string

sergiocarracedo:
## Description

Input should always extend string as is the native type, this removes the union type with number to avoid misleadings


dmaza81 (approved):
💯

PR #2272, opened 2025-07-17
feat(Card): use DataCollection's metadata

dani-moreno:
## Description

`Metadata` is a concept we're using in several components: DataCollection, Cards, ResourceHeader... and we're repeating the same pattern in each one. The idea is to make a generic metadata renderer that we'll use in these components, but for now, we'll use the DataCollection ones to pave the way for this transition.

Instead of using them directly, we have a list of available properties, because not all of them makes sense in a Card:

```jsx
const cardPropertyRenderers = {
  text: propertyRenderers.text,
  number: propertyRenderers.number,
  date: propertyRenderers.date,
  amount: propertyRenderers.amount,
  person: propertyRenderers.person,
  company: propertyRenderers.company,
  team: propertyRenderers.team,
  status: propertyRenderers.status,
  tag: propertyRenderers.tag,
  avatarList: propertyRenderers.avatarList,
  tagList: propertyRenderers.tagList,
  alertTag: propertyRenderers.alertTag,
  dotTag: propertyRenderers.dotTag,
}
```


sergiocarracedo (approved):
AS we mentioned we should extract the celltypes outside the datacollection. But we can do it later

PR #2280, opened 2025-07-21
fix(RichTextDisplay): enhance content processing to support line breaks

siguenzaraul:
## Description

We need to support line-breaks if they provide us a plain text instead of an html content to be retro compatible with old content, for example: performance reviews


nlopin:
@siguenzaraul what is the current behaviour?  can you share screenshots?

siguenzaraul:
> @siguenzaraul what is the current behaviour? can you share screenshots?

Of course! There are times when we use the display in places where plain text generated by a textarea was previously used. This text often has `\n` that act as line breaks. We need to be backward compatible, so if the content we are given is not HTML, we need to transform those `\n` to `<br>` so that the display makes the line breaks.


this is the content:
Captura de pantalla 2025-07-21 a las 17 12 16

this is without the fix:
Captura de pantalla 2025-07-21 a las 17 13 12

and this with the fix:


Captura de pantalla 2025-07-21 a las 17 14 33

PR #2282, opened 2025-07-22
fix(PulseAvatar): icons not showing the right color

rubenmoya:
## Description

Currently, in the [factorial-one(soon-to-be-zero)]() storybook, the `PulseAvatar` icons are broken, this PR fixes them.

I have also fixed a problem with the `color` prop in the `Icon` component. When we introduced it in https://github.com/factorialco/factorial-one/pull/2239 the type only took into account the first level keys in the object, while the icon colors had nested keys (critical, critical-bold)

Icons before:
image

Icons after:
image

Icon color autocomplete before:
image

Icon color autocomplete after:
image

## Implementation details

To be honest I'm not sure about adding "mood" related colors to the icon colors, feels to coupled to a feature, we could _almost_ reuse the colors we already had, but we are missing one for the "mood-positive", which uses `colors.flubber.50`, maybe we can come up with a name for it and add it.

### DIFF
--- a/packages/core/src/tokens/colors.ts
+++ b/packages/core/src/tokens/colors.ts
@@ -113,9 +113,7 @@ export const baseColors = {
   },
 } as const
 
-
 export type BaseColor = keyof typeof baseColors
-
 
 export const f1Colors = {
   foreground: {
@@ -231,6 +229,13 @@ export const f1Colors = {
       DEFAULT: "hsl(var(--selected-50))",
       hover: "hsl(var(--selected-60))",
     },
+    mood: {
+      "super-negative": "hsl(var(--mood-super-negative))",
+      negative: "hsl(var(--mood-negative))",
+      neutral: "hsl(var(--mood-neutral))",
+      positive: "hsl(var(--mood-positive))",
+      "super-positive": "hsl(var(--mood-super-positive))",
+    },
   },
   ring: "hsl(var(--ring))",
   link: "hsl(var(--link))",
@@ -238,6 +243,4 @@ export const f1Colors = {
   "special-highlight": "hsl(var(--special-highlight))",
 } as const
 
-
 export type F1Color = keyof typeof f1Colors
-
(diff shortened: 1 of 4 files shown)


packages/core/src/tokens/colors.ts
```diff
@@ -231,13 +229,18 @@ export const f1Colors = {
       DEFAULT: "hsl(var(--selected-50))",
       hover: "hsl(var(--selected-60))",
     },
+    mood: {
+      "super-negative": "hsl(var(--mood-super-negative))",
+      negative: "hsl(var(--mood-negative))",
+      neutral: "hsl(var(--mood-neutral))",
+      positive: "hsl(var(--mood-positive))",
+      "super-positive": "hsl(var(--mood-super-positive))",
+    },
```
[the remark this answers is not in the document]
sergiocarracedo: I think this is a good solution as the mood colors could be used in other places than the mood tracker. The important thing for me is the semantic meaning, and this is perfect

packages/react/src/components/Utilities/Icon/index.tsx
```diff
@@ -23,12 +23,33 @@ const iconVariants = cva({
...
+}[keyof T & string]
+
 export interface IconProps
   extends SVGProps<SVGSVGElement>,
     VariantProps<typeof iconVariants> {
   icon: IconType
   state?: "normal" | "animate"
-  color?: Lowercase<keyof typeof f1Colors.icon> | `#${string}` | "currentColor"
+  color?:
+    | "default"
```
sergiocarracedo, line 49: What `default` color does?

sergiocarracedo (approved):

packages/react/src/components/Utilities/Icon/index.tsx
rubenmoya: It was already there in the `Icon` component: 

```tsx
  const getColorClass = (colorValue: string) => {
    if (colorValue === "currentColor") return "text-current"
    if (colorValue === "default") return "text-f1-icon"
    if (colorValue.startsWith("#")) return ""
    return `text-f1-icon-${colorValue}`
  }
```

PR #2288, opened 2025-07-22
feat(Avatar): new `large` size

dani-moreno:
## Description

A few months ago we added a new size in avatars, `large`, which has a size of 40px. The current sizes become:

- `large` → `xlarge`
- `xlarge` → `xxlarge`

Changes:
- Adds that new size to avatars, and their respective masks (for abdges and module icons).
- Filter the sizes available in the `AvatarList` component, as we only allow up to the `medium` size (32px) in design, while we were providing every size in code. As large sizes are not a use-case we have, the sizes are now filtered.

## Screenshots

image

Screenshot 2025-07-22 at 13 50 21

[Link to Figma Design](https://www.figma.com/design/pZzg1KTe9lpKTSGPUZa8OJ/Components?node-id=3-19&t=aDa8DUK58tmOGKzg-4)


dani-moreno:
> will you migrate all the sizes in the monorepo?

Yes, there is only one usage to change

PR #2304, opened 2025-07-25
feat(Card): compact mode

dani-moreno:
## Description
Adds a prop to the ⁠`Card` component that sets the layout to ⁠`compact`, which means some changes to styling and how the data is presented to reduce vertical space. Useful for specific use cases, like the Spaces module.

## Screenshots

Screenshot 2025-07-25 at 16 07 58

_Card in compact layout_

Screenshot 2025-07-25 at 16 07 40

_Same with an image_

---

[Link to Figma Design](https://www.figma.com/design/pZzg1KTe9lpKTSGPUZa8OJ/Components?node-id=4598-75751&t=aWWYWgekM59cevuB-4)


nlopin:
Dani, I have a design question. On a card we have actions in top right and bottom down corner (favorite?). What is the reason to split them such drastically?

PR #2309, opened 2025-07-28
feat(BarChart): allow to have a different color per bar

rubenmoya:
## Description

There are different parts of Factorial where we need to be able to color each BarChart bar independently, for example:

<details><summary>Performance</summary>
image
</details> 

<details><summary>Employee pulse</summary>
image
</details> 

This is currently not supported, so those parts of the product still use Nivo Charts and are using a custom implementation. If want to migrate to factorial-~one~zero we need to be able to achieve this.

image

## Implementation details

This PR allows developers to add a `color` to each `data` and it will be used if present.

### DIFF
--- a/packages/react/src/components/Charts/BarChart/index.tsx
+++ b/packages/react/src/components/Charts/BarChart/index.tsx
@@ -79,8 +79,16 @@ const _BarChart = <K extends ChartConfig>(
       }
     }
 
+    if (item.color) {
+      return {
+        ...item,
+        fill: item.color,
+      }
+    }
+
     return item
   })
+
   const maxLabelWidth = Math.max(
     ...preparedData.flatMap((el) =>
       bars.map((key) =>
@@ -185,7 +193,6 @@ const _BarChart = <K extends ChartConfig>(
               : undefined
           }
         />
-
         {bars.map((key, index) => (
           <Bar
             key={`bar-${key}`}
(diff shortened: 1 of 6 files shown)


packages/react/src/components/Charts/BarChart/index.tsx
```diff
@@ -196,11 +203,7 @@ const _BarChart = <K extends ChartConfig>(
                 ? "stack"
                 : undefined
             }
-            fill={
-              highlightLastBar
-                ? (((data: { fill: string }) => data.fill) as unknown as string)
-                : (dataConfig[key].color ?? autoColor(index))
-            }
+            fill={dataConfig[key].color ?? autoColor(index)}
```
rubenmoya, line 206: The type for this is `string | undefined`, I'm not sure why we were passing a function and casting it as string 🤔

packages/react/src/components/Charts/VerticalBarChart/index.tsx
```diff
@@ -20,6 +19,7 @@ import {
   yAxisProps as yAxisConfigureProps,
 } from "../utils/elements"
 import { fixedForwardRef } from "../utils/forwardRef"
+import { prepareData } from "../utils/muncher"
```
rubenmoya, line 22: It was the same function duplicated, and just used here, so I removed the file

sergiocarracedo (approved):

rubenmoya:
> Does this PR makes sense now? If not, could you close it? if yes, could we merge it?

Let's close it, we can tackle it in the future if needed.

PR #2313, opened 2025-07-28
feat: file and folder cell types

sergiocarracedo:
## Description

- feat: Add file and folder type 
- chore: allow to pass simple struct to fileavatar
- feat: units in number field 
- feat: currency symbol in amount field

image

[Link to Figma Design](https://www.figma.com/design/pZzg1KTe9lpKTSGPUZa8OJ/Components?node-id=247-3691&t=gejNnfrgyX8htkkP-0)


packages/react/src/experimental/OneDataCollection/__stories__/cell-types.mdx
```diff
@@ -605,3 +650,67 @@ render: (item) => ({
 `,
   }}
 />
+
+### file
+
+Renders a file avatarwith a name
```
mrcasals, line 656: Wrong description?

sergiocarracedo: fixed, good catch

PR #2314, opened 2025-07-28
fix(Checkbox): ensure size is a11y compliant

dani-moreno:
## Description

Our current `Checkbox` target area does not meet [WCAG 2.2 minimum size](https://www.w3.org/WAI/WCAG22/Understanding/target-size-minimum.html).

This PR changes the checkbox to have a bigger target area (24 x 24px, which is the minimum), while keeping the visual size the same (20 x 20 px).

There are a few changes in how the Checkbox is used in several places to accommodate this size change.

## Screenshots

https://github.com/user-attachments/assets/f80e65f2-a2c2-4619-b99b-35dbccceaaf0

### DIFF
--- a/packages/react/src/ui/checkbox.tsx
+++ b/packages/react/src/ui/checkbox.tsx
@@ -1,5 +1,5 @@
 import * as CheckboxPrimitive from "@radix-ui/react-checkbox"
-import { AnimatePresence, motion } from "motion/react"
+import { AnimatePresence } from "motion/react"
 import * as React from "react"
 import { useId } from "react"
 import { Icon } from "../components/Utilities/Icon"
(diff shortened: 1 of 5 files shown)


packages/react/src/ui/checkbox.tsx
```diff
@@ -41,23 +41,12 @@ const Checkbox = React.forwardRef<
         disabled={disabled}
       >
         <AnimatePresence>
-          <CheckboxPrimitive.Indicator className="flex items-center justify-center text-current transition-none">
-            <motion.div
```
dani-moreno, line 45: Removed this animation entirely as it was doing nothing. It is enough with the AnimatePresence, so the check fades in/out.

sergiocarracedo (approved):

PR #2319, opened 2025-07-29
feat(Card): testing and structure

dani-moreno:
## Description

- Adds a testing file for the Card component, checking if it renders things correctly, and actions work as expected.
- Move stories to a `__stories__` folder.
- Move every 'piece' to a `components` folder.


packages/react/src/experimental/OneCard/__tests__/OneCard.test.tsx
```diff
@@ -0,0 +1,185 @@
...
+import userEvent from "@testing-library/user-event"
+import React from "react"
+import { describe, expect, it, vi } from "vitest"
+import { defaultTranslations, I18nProvider } from "../../../lib/providers/i18n"
+import type { CardSecondaryLink } from "../components/CardActions"
+import { OneCard } from "../OneCard"
+
+const TestWrapper = ({ children }: { children: React.ReactNode }) => (
+  <I18nProvider translations={defaultTranslations}>{children}</I18nProvider>
+)
```
nlopin, line 12: don't we have a custom render that adds translations automatically? Let's add it to avoid duplicating the code. There is an [example in the documentation](https://testing-library.com/docs/react-testing-library/setup/#custom-render)

nlopin (approved):

PR #2323, opened 2025-07-29
feat: name prop in inputs

sergiocarracedo:
## Description

Add the name prop to input fields


dani-moreno (approved):

sergiocarracedo:
> Shouldn't we require it?

It is not a required prop in the native input, so i think should be the same

PR #2335, opened 2025-07-30
feat(BasicTextEditor): ScrollArea 

siguenzaraul:
## Description

Added scrollArea component to improve the scrolling experience


siguenzaraul:
> what's the difference?

now with the scrollArea if you are not hover the "area" the scrollbar disappears instead of been there all the time 


https://github.com/user-attachments/assets/37f4e3a3-ef91-4018-bc5c-1adfa8c7e46a

PR #2342, opened 2025-07-31
feat(EntitySelect): enhance match scoring with word normalization

siguenzaraul:
## Description

After using the entity select we have realized that users, especially in languages like Spanish, search for names without using accents or special characters, for convenience or sometimes it happens that I write "siguenza raul" and the result I expect is "Raúl Sigüenza", so I have added a better search system that allows you to clean the special characters and search by words, even if they are disordered

https://github.com/user-attachments/assets/42e532ae-f931-4fc1-bacd-592c4ab4901e

### DIFF
--- a/packages/react/src/experimental/Forms/EntitySelect/index.tsx
+++ b/packages/react/src/experimental/Forms/EntitySelect/index.tsx
@@ -506,12 +506,47 @@ export const EntitySelect = (
   )
 }
 
+function normalizeText(text: string): string {
+  return text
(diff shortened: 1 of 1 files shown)


nlopin:
should we consider fuzzy search at some point?

siguenzaraul:
> should we consider fuzzy search at some point?

I think is a good idea, sometimes if you work with people from other countries is a little bit hard to spell some of their names and last names

PR #2356, opened 2025-08-01
feat(ChipList): add chip list component

rubenmoya:
## Description

This PR adds a `ChipList` component, that we need in the new surveys design for the editors section:

⚠️ I'm not sure about the name, since the chip is called `Chip`, but the folder `OneChip`. Also I'm not sure if both of them should be inside a common folder or not, or if the `Chip` should also be exported, so any feedback is welcomed.

image

image

## Implementation details

It follows the same implementation as the `TagList`


sauldom102 (approved):
LGTM! 🚀

sergiocarracedo (approved):
Before merge it as @dani-moreno about the naming, I'm not sure about chip
And please add the prefix `F0` and move it to the components folder marking it as experimental with the `experimentalComponent` function

packages/react/src/components/F0ChipList/index.stories.tsx
```diff
@@ -0,0 +1,72 @@
...
+  parameters: {
+    layout: "centered",
+  },
+  tags: ["autodocs", "experimental"],
+} satisfies Meta<typeof F0ChipList>
+
+export default meta
+type Story = StoryObj<typeof meta>
+
+const mockAvatarUrl = "https://i.pravatar.cc/150"
```
dani-moreno, line 16: We have mockup avatars in `@storybook-static/avatars/`, so we don't rely on external sources.

PR #2368, opened 2025-08-05
fix: add support to show warning message for bulk actions in data collection

sauldom102:
## Description

We need to show a warning message in the action bar of the data collection depending on some logic that we need to determine in our frontend.

## Screenshots

Screenshot 2025-08-05 at 12 05 14

### DIFF
--- a/packages/core/src/tokens/colors.ts
+++ b/packages/core/src/tokens/colors.ts
@@ -128,7 +128,7 @@ export const f1Colors = {
     accent: "hsl(var(--accent-70))",
     critical: "hsl(var(--critical-70))",
     info: "hsl(var(--info-70))",
-    warning: "hsl(var(--warning-50))",
+    warning: "hsl(var(--warning-70))",
     positive: "hsl(var(--positive-70))",
     selected: "hsl(var(--selected-70))",
   },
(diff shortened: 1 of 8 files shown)


packages/core/src/tokens/colors.ts
```diff
@@ -128,7 +128,7 @@ export const f1Colors = {
     accent: "hsl(var(--accent-70))",
     critical: "hsl(var(--critical-70))",
     info: "hsl(var(--info-70))",
-    warning: "hsl(var(--warning-70))",
+    warning: "hsl(var(--warning-50))",
```
sauldom102, line 131: This color wasn't right at all as per design

dani-moreno: The token seems to be right, it is warning-70 in figma. What is wrong is:

```
--warning-50: theme("colors.orange.50");
--warning-60: theme("colors.orange.60");
--warning-70: theme("colors.orange.70");
```

Inside the `base.css` file, inside the `dark` object. They should be:

```
--warning-50: theme("colors.orange.70");
--warning-60: theme("colors.orange.60");
--warning-70: theme("colors.orange.50");
```

That way, it'll be exactly the same as in Figma:
image

Can you fix it here?

dani-moreno (changes requested):

dani-moreno (approved):

PR #2371, opened 2025-08-05
fix: get rid of wrong selected count on data collection

sauldom102:
## Description

We were showing an incorrect number of elements selected in the data collection under some circumstances.

https://github.com/user-attachments/assets/ad8ce4eb-4e7e-4915-bae8-f981788c9036


sauldom102:
> I have an impression pagination info was here for a reason. We need tests for hooks

Totally

PR #2373, opened 2025-08-05
feat: add F0ButtonToggle component 

siguenzaraul:
## Description

Button without label and selected status

https://github.com/user-attachments/assets/dbccfd19-22ff-4eaf-a56b-f07a48dba5c7

[Link to Figma Design](https://www.figma.com/design/pZzg1KTe9lpKTSGPUZa8OJ/Components?m=auto&node-id=517-19698&t=HEsZdD32pCUXJlGt-1)

### DIFF
--- a/packages/react/src/experimental/Actions/F0ButtonToggle/index.tsx
+++ b/packages/react/src/experimental/Actions/F0ButtonToggle/index.tsx
@@ -11,7 +11,7 @@ interface F0ButtonToggleProps {
   label: string
   disabled?: boolean
   icon: IconType
-  size?: "sm" | "md"
+  size?: "sm" | "md" | "lg"
 }
 
 export const F0ButtonToggle = forwardRef<
@@ -30,21 +30,13 @@ export const F0ButtonToggle = forwardRef<
     },
     ref
   ) => {
-    const sizeClasses = cva({
-      base: "aspect-square px-0",
+    const toggleButtonClasses = cva({
+      base: "aspect-square px-0 data-[state=on]:bg-f1-background-selected data-[state=on]:text-f1-icon-selected hover:data-[state=on]:bg-f1-background-selected-hover hover:data-[state=on]:text-f1-icon-selected-hover",
       variants: {
         size: {
           sm: "h-6",
           md: "h-8",
-        },
-      },
-    })
-
-    const toggleButtonVariants = cva({
-      variants: {
-        variant: {
-          default:
-            "data-[state=on]:bg-f1-background-selected data-[state=on]:text-f1-icon-selected hover:data-[state=on]:bg-f1-background-selected-hover hover:data-[state=on]:text-f1-icon-selected-hover",
+          lg: "h-10",
         },
       },
     })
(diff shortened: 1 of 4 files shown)


nlopin:
@siguenzaraul we have [`pressed` state on a regular button](https://one.factorial.dev/?path=/docs/components-button--documentation#states)

siguenzaraul:
> @siguenzaraul we have [`pressed` state on a regular button](https://one.factorial.dev/?path=/docs/components-button--documentation#states)

Factorial-zero design team told me to create a separated componente with this state, maybe we should talk with them

packages/react/src/experimental/Actions/F0ButtonToggle/index.tsx
```diff
@@ -0,0 +1,57 @@
...
+      label,
+      disabled = false,
+      icon,
+      size = "md",
+      ...props
+    },
+    ref
+  ) => {
+    return (
+      <Button
```
nlopin, line 32: I'm not sure it is should be a button semantically. Since it has two states "on" and "off", it should be a checkbox under the hood

siguenzaraul: Yes, we can do that, maybe it makes more sense

nlopin: @siguenzaraul I researched the topic a bit and [it is fine to use button](https://www.w3.org/TR/wai-aria-1.1/#button):

> Buttons support the optional [attribute](https://www.w3.org/TR/wai-aria-1.1/#dfn-attribute) [aria-pressed](https://www.w3.org/TR/wai-aria-1.1/#aria-pressed). Buttons with a non-empty aria-pressed attribute are toggle buttons. When aria-pressed is true the button is in a "pressed" [state](https://www.w3.org/TR/wai-aria-1.1/#dfn-state), when aria-pressed is false it is not pressed. If the attribute is not present, the button is a simple command button.

We only need to have `aria-pressed` value set

packages/core/src/tokens/colors.ts
```diff
@@ -177,6 +177,7 @@ export const f1Colors = {
     selected: {
       DEFAULT: "hsl(var(--selected-50) / 0.1)",
       secondary: "hsl(var(--selected-50) / 0.05)",
+      hover: "hsl(var(--selected-50) / 0.2)",
```
nlopin, line 180: @dani-moreno do we have new token in Figma?

nlopin (approved):

packages/react/src/experimental/Actions/F0ButtonToggle/index.tsx
rubenmoya: Shadcn already has a `Toggle` component, if we're using shadcn as the base for other components why not for this one too? https://ui.shadcn.com/docs/components/toggle

nlopin: 100% agree

dani-moreno (approved):

PR #2384, opened 2025-08-07
Small UI fixes for data collection

sauldom102:
## Description

We want to polish a few things for the data collection:
- Add support to hide certain filters from the selectors above the data collection.
- Add support to hide grouping selector above the data collection.
- Some spacings adjusted to match our design.


dmaza81 (approved):
Nice!

PR #2393, opened 2025-08-07
Add IT Management module icon

alejandromav:
## Description
Add new Module and icon for IT Management.

image

After creating the new page in https://github.com/factorialco/factorial/pull/74079, we get this error:

image

And the icon in the title is not shown:

image

So we assumed the changes proposed in this PR are needed.


dani-moreno:
The icon seems to be wrong, as we don't use stroke icons for the Module avatar, we use filled ones.

I can help you with the icon tomorrow, but please wait until we have the correct one.

dani-moreno (approved):

PR #2398, opened 2025-08-08
Support metadata in cards

eliseo-juan:
## Description

Use F0Card metadata instead of adding content to the card.

Before
image

After:
image

### DIFF
--- a/packages/react/src/experimental/OneDataCollection/visualizations/collection/Card/index.tsx
+++ b/packages/react/src/experimental/OneDataCollection/visualizations/collection/Card/index.tsx
@@ -1,4 +1,7 @@
 import { F0Card } from "@/components/F0Card"
+import { cardPropertyRenderers } from "@/components/F0Card/components/CardMetadata"
+import { CardMetadataProperty } from "@/components/F0Card/types"
+import { IconType } from "@/components/Utilities/Icon"
 import { AvatarVariant } from "@/experimental/Information/Avatars/Avatar"
 import { GroupHeader } from "@/experimental/OneDataCollection/components/GroupHeader/GroupHeader"
 import { NavigationFiltersDefinition } from "@/experimental/OneDataCollection/navigationFilters/types"
@@ -7,6 +10,7 @@ import {
   useGroups,
 } from "@/experimental/OneDataCollection/useGroups"
 import { useSelectable } from "@/experimental/OneDataCollection/useSelectable"
+import { Placeholder } from "@/icons/app"
 import { Card, CardContent, CardHeader, CardTitle } from "@/ui/Card"
 import { Skeleton } from "@/ui/skeleton"
 import { AnimatePresence, motion } from "motion/react"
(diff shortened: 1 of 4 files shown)


packages/react/src/experimental/OneDataCollection/visualizations/collection/Card/index.tsx
```diff
@@ -196,18 +195,43 @@ const GroupCards = <
...
+                      type: "number" as const,
+                      value: renderResult,
+                    }
+                  } else {
+                    // Check if renderResult is a valid CardMetadataProperty type
+                    if (renderResult.type in cardPropertyRenderers) {
+                      cardProperty = renderResult as CardMetadataProperty
+                    } else {
+                      return null
+                    }
```
eliseo-juan, line 225: I don't like this cast at all. But I didn't find a better way to handle with a RendererDefinition that should be CardPropertyType, without making metadata something generic.

dmaza81: What about this:

```
<F0Card
  key={index}
  title={title(item)}
  selectable={selectable}
  description={description?.(item)}
  avatar={avatar?.(item)}
  selected={selectable && selectedItems.has(id)}
  onSelect={(selected) => handleSelectItemChange(item, selected)}
  secondaryActions={secondaryActions}
  primaryAction={primaryAction}
  otherActions={otherActions}
  onClick={itemOnClick}
  link={itemHref}
  metadata={getMetadata(item, cardProperties)}
/>
```

Then Create some functions to cover a

```
function getMetadata(
  item: Record,
  properties: CardPropertyDefinition<Record>[]
): Array<{ icon: React.ElementType; property: CardMetadataProperty }> {
  return properties
    .map((property) => {
      const result = property.render(item)
      if (result === undefined) return null

      const cardProperty = convertToCardMetadataProperty(result)
      if (!cardProperty) return null

      return {
        icon: property.icon ?? Placeholder,
        property: cardProperty,
      }
    })
    .filter((item): item is { icon: React.ElementType; property: CardMetadataProperty } => item !== null)
}

function convertToCardMetadataProperty(value: unknown): CardMetadataProperty | null {
  if (typeof value === "string") {
    return { type: "text", value }
  }

  if (typeof value === "number") {
    return { type: "number", value }
  }

  if (isCardMetadataProperty(value)) {
    return value
  }

  return null
}

function isCardMetadataProperty(value: unknown): value is CardMetadataProperty {
  if (
    typeof value !== "object" ||
    value === null ||
    !("type" in value)
  ) {
    return false
  }

  const typeValue = (value as Record<string, unknown>).type
  return typeof typeValue === "string" && typeValue in cardPropertyRenderers
}
```

1. Type-Safe Code

By using explicit type guards (isCardMetadataProperty), TypeScript can fully infer the types without unsafe assertions like as.
This helps catch errors at compile time and avoids runtime surprises.

2. More Maintainable

Extracting logic into helper functions (convertToCardMetadataProperty, getMetadata) keeps your JSX clean and focused only on rendering.
Changes to the property logic can be made in one place without touching your UI layer.

3. Improved Readability

The JSX block is now concise and easy to read.
All complex logic is moved into well-named, testable functions.

PR #2421, opened 2025-08-19
Update calendar weekly view

davidFactorial:
## Description

We want to add a new format string for "Weekly view" in Date navigator component. We want to change it from:
`W22 2025 `to `Week of Aug 2025`. 

Now:

image

Before:
image

## Implementation details

To achieve this new behavior we add a new function in the granularities formatters, and use it when needed. In our case, as we only want to use it for week granularity we check the granularity type.

Also, we need to use translations for the "Week of" part, so we need to add a new key in the translation file for it.


sergiocarracedo:
https://github.com/factorialco/factorial-one/pull/2479

PR #2423, opened 2025-08-19
feat: Tokens for spacing between elements

dani-moreno:
## Description

Adds new spacing tokens for `gap` and `space`, with a simpler list of options (`sm`, `md`, `lg` and `xl`), documented [here](https://zeroheight.com/6cc527675/p/94c391-spacing/t/3409209197).

Each token is aimed at different combination of elements.

## Why?

After interviewing a few designers, it's clear they're not sure how to use the spacing tokens. This is part of an effort to reduce decision fatigue and cognitive overload.

The idea is to have tokens on top of the _foundational_ ones (which are tailwind spacings), specific for different use cases and synced with Figma:

image

Also, reducing the options available, we ensure more consistency and meaning in our spacing usage.

So, instead of writing `gap-2` or `space-y-4`, you'd write `gap-sm` and `space-y-md`. Numbers confuse designers (since they think of 2 as 2px, not as a spacing unit), and we shorten the list of options, making it easier to avoid mistakes.

## Let's apply this to every component!

Quieto parao 😔✋ Let me try out how this works and iterate on it before launching. But having the tokens ready and making new things with them is fine.

## Next steps

1. Try them, get some designers feedback.
2. Replace current gaps and spaces with these tokens.
3. Make these the only available options for gap and space. For now, the foundational tokens are also available (so you can keep using gap-2).

### DIFF
--- a/packages/core/src/tokens/spacing.ts
+++ b/packages/core/src/tokens/spacing.ts
@@ -87,3 +87,15 @@ export const relativeSpacing: ThemeConfig["spacing"] = {
   80: "20rem",
   96: "24rem",
 }
+
+/*
+ * Spacing between elements
+ * This special scale is used for the gap between common elements.
+ * Docs: https://zeroheight.com/6cc527675/p/94c391-spacing/t/3409209197
+ */
+export const betweenSpacing: ThemeConfig["spacing"] = {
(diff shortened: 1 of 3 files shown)


dmaza81 (approved):
🥇

PR #2424, opened 2025-08-20
feat: add tracking to data collection

sauldom102:
## Description

We want to provide a unified way of tracking events related to our design system and more specifically for this pull request related to data collection. With this approach, discussed with @sergiocarracedo who was the one that had this great idea, we'll just pass a tracking function to our main provider and it will be able to be called from all of our components with the correct event name and needed params.


sergiocarracedo (changes requested):

packages/react/.storybook/preview.tsx
```diff
@@ -45,6 +45,12 @@ export const FactorialOne = (Story, { parameters }) => {
           locale: parameters.l10n?.locale ?? "en",
         },
       }}
+      trackingFunction={(
+        eventName: string,
+        params: Record<string, unknown>
+      ) => {
+        console.log(eventName, params)
```
dmaza81, line 52: Should we need this console.log?

sauldom102: It's only for development purposes and it will only be used in storybook

packages/react/src/lib/providers/tracking/tracking-provider.tsx
```diff
@@ -0,0 +1,39 @@
...
+      {children}
+    </TrackingContext.Provider>
+  )
+}
+
+export function useTracking() {
+  const context = useContext(TrackingContext)
+
+  if (context === null) {
+    throw new Error("useTracking must be used within a TrackingProvider")
```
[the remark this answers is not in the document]
sauldom102: Yeah you're right, I just removed, anyway this provider is included as part of OneProvider so we wouldn't reach this at all

sauldom102:
> If I understood right, the proposal is to have a `trackingFunction` prop in `FactorialOneProvider` where users can catch events they are interested in.
> 
> I like the approach, we avoid exposing props just for the sake of tracking. One change I propose is to expose a separate provider just for tracking. This will allow users to wrap only the component they want to track with a provider and better control the scope.
> 
> For example if I have two DataCollections, I can choose to treat them as separate sources of events:
> 
> ```js
> <ZeroEventCatcherProvider onEvent={(event: string) => { // only events from one }}>
>   <OneCollection id="one" />
> </ZeroEventCatcherProvider>
> <ZeroEventCatcherProvider onEvent={(event: string) => { //only events from two }}>
>   <OneCollection id="two" />
> </ZeroEventCatcherProvider>
> ```
> 
> or as one:
> 
> ```js
> <ZeroEventCatcherProvider onEvent={(event: string) => { // events from one and two }}>
>   <OneCollection id="one" />
>   <OneCollection id="two" />
> </ZeroEventCatcherProvider>
> ```
> 
> This gives a better semantic on what is being tracked and creates isolated functions instead of one god-like function in `FactorialOneProvider`

Having `ZeroEventCatcherProvider` wrapping certain components in our code looks a bit cumbersome to me, while it is true that it makes it more explicit and probably gives us more control, I think we can gain that control just by introducing an `enabled` attribute in `trackingMeta` that is `false` by default if we don't pass it.

Another thing I wanted to avoid is having different event handling done in different places.

nlopin:
My main issue with tracking as a product engineer is that spreads over all codebase, like a very thin piece of butter over a big piece of bread. Having it in one place makes it easier to grasp tracking as a whole

packages/react/src/lib/providers/events/event-catcher-provider.tsx
```diff
@@ -0,0 +1,59 @@
...
+const EventCatcherContext = createContext<ContextType | null>(null)
+
+export interface EventCatcherProviderProps {
+  children: ReactNode
+  onEvent: EventCatcherFunction
+  enabled?: boolean
+  catchEvents?: string[]
+}
+
+export function EventCatcherProvider({
```
nlopin, line 24: should it be `F0...`?

packages/react/src/lib/providers/events/event-catcher-provider.tsx
```diff
@@ -0,0 +1,59 @@
...
+  onEvent: EventCatcherFunction
+  enabled?: boolean
+  catchEvents?: string[]
+}
+
+export function F0EventCatcherProvider({
+  children,
+  onEvent,
+  enabled = true,
+  // If not provided, all events will be caught
```
nlopin, line 28: if moved to line 21, it will be shown in documentation

PR #2429, opened 2025-08-21
Improve f0 folder structure

eliseo-juan:


packages/react/eslint.config.mjs
```diff
@@ -196,4 +196,40 @@ export default [
...
+              message:
+                "Components in src/components cannot import from sds or kits folders. Use components from components/ or other allowed directories instead.",
+            },
+          ],
+        },
+      ],
+    },
+  },
+  // Restrict imports from sds and other kits folders in kits
+  {
```
sergiocarracedo, line 218: i think this rule will block the imports inside the same kit

sergiocarracedo:

PR #2434, opened 2025-08-21
Fix(DateNavigator): pass the right value to onCompareToChange

FaithAdekunle:
## Description
This issue came up when trying to implement data comparisons for different dates in financial workspace. Upon using the DateNavigator component and passing relevant props to it, it turned out the `onCompareToChange` function prop was not getting the appropriate parameters passed in when called from within the component. See implementation in this PR.

Screenshot 2025-08-21 at 12 09 56

## Implementation details
```
const compareToOptions: ({
    label: string;
    value: string;
    description: string;
    dateValue: Required<DateRange> | Required<DateRange>[];
} | {
    label: string;
    value: string;
    description: string;
    dateValue: undefined;
})[]
```
The current implementation of compareToOptions  assumes value is a stringified number (e.g. "0", "1", etc.) that can be coerced into an array index using the unary +. When building the array of options, value is set to the index position of the option object. This however does not take into account that a default `none` option will later be placed in the first position as first option, thereby invalidating the value previously set for the other options.

This PR fixes this by incrementing the option values by 1 when building the options array. Thereby leaving the 0 index spot vacant for the default none option.

### DIFF
--- a/packages/react/src/ui/DatePickerPopup/OneDatePickerPopup.tsx
+++ b/packages/react/src/ui/DatePickerPopup/OneDatePickerPopup.tsx
@@ -160,7 +160,7 @@ export function OneDatePickerPopup({
 
       return {
         label: compare.label,
-        value: index.toString(),
+        value: (index + 1).toString(), // This leaves index 0 spot vacant for the 'none' option.
         description,
         dateValue: value,
       }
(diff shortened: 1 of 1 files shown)


sergiocarracedo (approved):
LGTM Thanks for fixing it!!

PR #2440, opened 2025-08-22
chore: report coverage

sergiocarracedo:


packages/react/src/lib/exports.ts
```diff
@@ -8,4 +8,4 @@ export type { TranslationsType } from "./providers/i18n"
 export { defaultTranslations } from "./providers/i18n/i18n-provider-defaults"
 export { FactorialOneProvider } from "./providers/one/one-provider"
 export { useXRay } from "./xray"
-// 6
+// 11
```
arnaub, line 11: what is this number?

arnaub (approved):

packages/react/src/lib/exports.ts
sergiocarracedo: a trick to force the react pipeline to work

PR #2446, opened 2025-08-26
Make secondary options nillable so we dont render the footer when the…

rodrigo-picanco:
## Description
After some changes to the card collection, some cards started rendering strangely. 

The two main things is that now we force an icon, which makes file properties have two icons. 

The other is that the card footer is always displayed

Screenshot 2025-08-26 at 10 54 29

### DIFF
--- a/packages/react/src/experimental/OneDataCollection/visualizations/collection/Card/index.tsx
+++ b/packages/react/src/experimental/OneDataCollection/visualizations/collection/Card/index.tsx
@@ -156,6 +156,9 @@ const GroupCards = <
 
         const cardProperty = convertToCardMetadataProperty(result)
         if (!cardProperty) return null
+        if (cardProperty.type === 'file') return {
+          property: cardProperty,
+        }
 
         return {
           icon: property.icon ?? Placeholder,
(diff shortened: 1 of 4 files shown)


packages/react/src/experimental/OneDataCollection/visualizations/collection/Card/index.tsx
```diff
@@ -158,11 +158,17 @@ const GroupCards = <
         if (!cardProperty) return null
```
nlopin, line 159: maybe just adding this special case here and avoid creating a new function?

```
if (cardProperty.type === 'file') {
  return { property: cardProperty }
}

File: packages/react/src/experimental/OneDataCollection/visualizations/collection/Card/index.tsx [hunk 2]
```diff
@@ -220,7 +226,7 @@ const GroupCards = <
         const primaryAction =
           itemActions.find((action) => action.type === "primary") || undefined
         const secondaryActions =
-          itemActions.filter((action) => action.type === "secondary") || []
+          itemActions.filter((action) => action.type === "secondary")
```
nlopin, line 229: I'd rather kept the invariant that `secondaryActions` is always an array and checked it's size in the place we render

nlopin (approved):

PR #2447, opened 2025-08-26
New kanban card for kanban view

Rogermax:
### F0Card: opt‑in draggable + headless DnD infra (Atlaskit adapter) + intent stories

- Why: enable drag on `F0Card` without coupling UI to a specific DnD lib; let containers (Lane/Kanban) decide business rules (e.g., “same/adjacent only”).

- What’s included
  - F0Card
    - Optional `drag` prop to make a card draggable-capable (no semantics inside the card).
    - Keeps `forwardRef` and wires it to the root element.
  - Headless DnD infra
    - `DndProvider({ driver })` with `createAtlaskitDriver(instanceId)`.
    - Hooks: `useDraggable`, `useDroppableList`, `useDndEvents`.
    - Safe by default: if no provider or `disabled`, hooks do nothing.
  - Stories
    - `IntentsShowcase`: one page showing:
      - Reorder lane: in-lane reordering with top/bottom indicators.
      - Enter container lane: highlight and log when hovering the lane (no card target).
      - Cancel lane: “not-allowed” feedback and no drop behavior.

- Public API (summary)
  - `F0Card drag`:
    - `drag={{ id: string, type?: string, data?: Record<string, unknown> }}`
  - DnD setup:
    - `DndProvider({ driver: createAtlaskitDriver(Symbol('...')) })`
    - `useDraggable({ ref, payload, disabled?, handleRef? })`
    - `useDroppableList({ ref, id, accepts })`
    - `useDndEvents(({ phase, source }) => {})` // phases: start | over | drop | cancel

- Backward compatibility
  - No breaking changes. Without `drag` and without `DndProvider`, everything behaves as before.
  - Hooks are no-ops without a provider.

- Minimal usage example
  - In a card:
    - `useDraggable({ ref, payload: { kind: 'kanban-card', id } })`
  - In a lane:
    - `useDroppableList({ ref, id: 'lane-1', accepts: ['kanban-card'] })`
    - On drop (container-level), apply strategy: reorder within lane or move across lanes per policy.

- Library swap
  - UI code does not import Atlaskit directly.
  - To change engines later, implement a new driver and pass it to `DndProvider`.

- Not in this PR
  - Cross-lane move and domain policies (e.g., adjacent-only) in production components; shown conceptually in stories.
  - High-level “intent” hooks (we can add them later if we want to hide monitoring details).

- Next steps
  - Headless `KanbanLane` with adjacent-only policy, cross-lane move, autoscroll, focus restore.
  

https://github.com/user-attachments/assets/3b5d2f52-5289-40d9-aa37-37711847fc64


packages/react/src/components/F0Card/__stories__/DraggableStoryCard.tsx
Rogermax, line 1: This is component is just for the story, will be replaced in a future PR for a KanbanLane.
Also the story of drag and drop will be not necessary when I add the KanbanLane

packages/react/src/components/F0Card/__stories__/DropLaneEnter.tsx
Rogermax, line 1: Same as above.. this compoennt will be removed in the next PR

packages/react/src/components/F0Card/CardInternal.tsx
```diff
@@ -103,149 +111,207 @@ export interface CardInternalProps {
...
+            "hover:border-f1-border",
+          link &&
+            "focus-within:border-f1-border-hover focus-within:shadow-md hover:border-f1-border-hover hover:shadow-md",
+          selected &&
+            "border-f1-border-selected bg-f1-background-selected-secondary",
+          drag && "cursor-grab select-none active:cursor-grabbing"
+        )}
+        onClick={onClick}
+        data-testid="card"
+        ref={setRefs}
```
Rogermax, line 199: `F0Card` has forwardRef and `Card` also have it... but if this CardInternal do not propagate the forwardRef from one to another then it's meanless. F0Card -> CardInternal -> Card

This is the reason to add forwardRef to this component and pass it here

packages/react/src/components/F0Card/CardInternal.tsx
Rogermax, line 1: Here seems there are a lot of changes but this is not true... we just added card prop and forwardRef (this one shift all 4 spaces to the left and that makes crazy the difference algorithm)

packages/react/src/components/F0Card/F0Card.tsx
```diff
@@ -6,13 +6,13 @@ const privateProps = ["forceVerticalMetadata"] as const
 
 export type F0CardProps = Omit<CardInternalProps, (typeof privateProps)[number]>
 
-const F0CardBase = forwardRef<HTMLDivElement, F0CardProps>((props) => {
+const F0CardBase = forwardRef<HTMLDivElement, F0CardProps>((props, ref) => {
```
Rogermax, line 9: adding ref also remove a warning from the console.log

packages/react/src/components/F0Card/CardInternal.tsx
```diff
@@ -1,6 +1,7 @@
 import { Link } from "@/components/Actions/Link"
 import { Image } from "@/components/Utilities/Image"
 import { DropdownItem } from "@/experimental/Navigation/Dropdown"
+import { useDraggable } from "@/lib/exports"
```
sergiocarracedo, line 4: please dont import from exports to avoid circular imports

packages/react/src/components/F0Card/CardInternal.tsx
```diff
@@ -103,149 +111,207 @@ export interface CardInternalProps {
    * Private prop
    */
   forceVerticalMetadata?: boolean
+
+  /**
+   * Optional drag configuration to make the card draggable
+   */
+  drag?: {
```
sergiocarracedo, line 118: why not use an external wrapper to decouple the card from drag as the dragging logic is not related to the card

sergiocarracedo (approved):

PR #2448, opened 2025-08-26
update-sectionheader

desiree-np:
## Description

I replaced the button under the title of the SectionHeader with the Link component, to keep the component updated with the new UI definition of the link.

## Screenshots 
Before
Screenshot 2025-08-26 at 17 47 28

After
Screenshot 2025-08-26 at 17 44 58


sergiocarracedo (approved):
Is the old prop `supoortButton` in use in the monorepo? if yes we need to create a PR to migrate it

PR #2452, opened 2025-08-27
chore: move avatar to components, renamed with new convention and folder structure

eliseo-juan:
Several changes for Avatar component
* Moved to components
* Adapted to new folder structure for components
* Renamed to match new name convention


sergiocarracedo (approved):
🚀

PR #2454, opened 2025-08-27
feat: add hover actions to table

juliangaspar:
## Description

Following this [Figma](https://www.figma.com/design/v1YD01TajEDThOX2BKzXee/Time-tracking---Data-collections?node-id=4001-14459&t=XJjNaqcgOVnIWbuj-0) and this [thread](https://factorialteam.slack.com/archives/C099YVA8E1J/p1756194673826619) we are adding an on hover button to the `<Table/>` row similar to the `<List/>` component.

https://github.com/user-attachments/assets/14088faa-f17f-4c11-b0d9-823ec6c5bb87

[Link to Figma Design](https://www.figma.com/design/v1YD01TajEDThOX2BKzXee/Time-tracking---Data-collections?node-id=4001-14459&t=XJjNaqcgOVnIWbuj-0)

## Implementation details

Added on hover actions to the `<Table/>`

### DIFF
--- a/packages/react/src/experimental/OneDataCollection/visualizations/collection/Table/components/Row.tsx
+++ b/packages/react/src/experimental/OneDataCollection/visualizations/collection/Table/components/Row.tsx
@@ -1,8 +1,11 @@
 import { ItemActionsDropdown } from "@/experimental/OneDataCollection/ItemActions/ItemActionsDropdown"
-import { forwardRef } from "react"
+import { forwardRef, useState } from "react"
 
+import { Button } from "@/components/Actions/Button"
 import { FiltersDefinition } from "@/components/OneFilterPicker/types"
+import { DropdownItemSeparator } from "@/experimental/Navigation/Dropdown/internal"
 import {
+  ActionDefinition,
   filterItemActions,
   ItemActionsDefinition,
 } from "@/experimental/OneDataCollection/item-actions"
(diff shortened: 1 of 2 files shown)


juliangaspar:
> why it overflows on the video? it shouldn't behave like this

Hey @nlopin it behaves like this [Playbook](https://one.factorial.dev/?path=/docs/components-data-collection-visualizations-table--documentation) when you haven't scrolled to the end of the row. I haven't changed that behaviour.

packages/react/src/experimental/OneDataCollection/visualizations/collection/Table/components/Row.tsx
```diff
@@ -94,8 +97,55 @@ const RowComponentInner = <
 
   const itemActions = filterItemActions(source.itemActions, item)
 
+  // gets the primary item actions (max 2)
+  const primaryItemActions = itemActions
```
sergiocarracedo, line 101: As the actions logic is the same as in the List view we should extract it and the render of the actions to a component used by the table and the view

packages/react/src/experimental/OneDataCollection/visualizations/collection/Table/components/Row.tsx
```diff
@@ -147,13 +197,35 @@ const RowComponentInner = <
...
+                <Button
+                  key={action.label}
+                  label={action.label}
+                  variant="outline"
+                  onClick={action.onClick}
+                  icon={action.icon}
+                />
+              ))}
+
+              <ItemActionsDropdown
```
sergiocarracedo, line 222: In the List view the ItemActionsDropdown has a mobile version https://github.com/factorialco/factorial-one/blob/e78b12c3adb0538956c51d5d231323b7c54de5bd/packages/react/src/experimental/OneDataCollection/visualizations/collection/List/components/Row.tsx:209
We should have it here too

packages/react/src/experimental/OneDataCollection/visualizations/collection/Table/components/Row.tsx
```diff
@@ -147,13 +197,35 @@ const RowComponentInner = <
           sticky={{
```
sergiocarracedo, line 197: Makes no sense to show the table cell as the actions are now only shown on hover (only makes sense to have then in the mobile version for the fixed dropdown

sergiocarracedo (changes requested):

packages/react/src/experimental/OneDataCollection/visualizations/collection/Table/components/Row.tsx
juliangaspar: I saw it, but I assumed it wasn't expected. How do I test it in mobile? How can I see what's the expected behaviour?

juliangaspar:
Sorry for the delay, we've been running a bit with the alignment. Could I get some 👀 again @sergiocarracedo?

.gitignore
```diff
@@ -11,6 +11,7 @@ coverage
 node_modules
 *.local
 storybook-static
+.husky
```
arnaub, line 14: Just checking — is this change related to the PR?

arnaub (approved):
Approved! I left one small comment. Also, packages/react/src/experimental/OneDataCollection/components/itemActions/index.ts is empty — do we want to keep this file?

sergiocarracedo (approved):

PR #2460, opened 2025-08-27
fix: onclick propagation

sergiocarracedo:
## Description

Select were emiting in click event when the value changes causing. if the select were in a modal, to close it on click the modal

## Implementation details

It can be relatsd to this BUG in radix-ui **https://github.com/radix-ui/primitives/pull/3183/files**, but as we wrap the radix primitive I capture the click event to avoid the propagation

### DIFF
--- a/packages/react/src/experimental/Forms/Fields/Select/__stories__/index.stories.tsx
+++ b/packages/react/src/experimental/Forms/Fields/Select/__stories__/index.stories.tsx
@@ -23,12 +23,7 @@ const SelectWithHooks = (props: SelectProps<string>) => {
   }
 
   return (
-    <div
-      className="w-48"
-      onClick={() => {
-        console.log("clicked")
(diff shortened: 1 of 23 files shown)


eliseo-juan (approved):

packages/react/src/experimental/Forms/Fields/Select/__stories__/index.stories.tsx
```diff
@@ -23,7 +23,12 @@ const SelectWithHooks = (props: SelectProps<string>) => {
   }
 
   return (
-    <div className="w-48">
+    <div
+      className="w-48"
+      onClick={() => {
+        console.log("clicked")
```
arnaub, line 29: do you want this console.log here?

arnaub (approved):
Approved but I left a comment

PR #2465, opened 2025-08-28
feat: upgrade copilot kit 1.10.2

sergiocarracedo:


nlopin:
it's other way around I think. We require 1.10.2 and monorepo uses 1.9.3

sergiocarracedo:
yes, but i will do it in the same PR i want to update the f0-react dependency version

PR #2484, opened 2025-08-29
fix: show presets itemCount when value is 0

sergiocarracedo:
## Description

Fix: Show presetItem count when the value is 0

image


eliseo-juan (approved):
LGTM!

PR #2485, opened 2025-08-29
feat: hide filter chips if the filter matches a preset

sergiocarracedo:
## Description

Hide filter chips when they match a preset

image


packages/react/src/components/OneFilterPicker/OneFilterPicker.tsx
```diff
@@ -201,12 +201,20 @@ const FiltersChipsList = () => {
...
     filters,
     setIsFiltersOpen,
+    presets,
     removeFilterValue,
     setFiltersValue,
   } = useContext(FiltersContext)
 
+  const isPresetFilter = useMemo(() => {
+    return (presets || []).some((preset) => {
+      return JSON.stringify(value) === JSON.stringify(preset.filter)
```
eliseo-juan, line 211: Claude said that maybe the order is not the same in both, but they can have the same filters, make sense to sort them before make them a json?

eliseo-juan (approved):
Check the comment, maybe I'm super wrong

PR #2516, opened 2025-09-04
feat(ValueDisplay): add percentage cell type

rubenmoya:
## Description

In Talent Engagement we want to show the survey participation rate in the data collection. To do so we are showing the amount of people and a percentage chart:

image

| With value | With custom label | Placeholder |
|--------|--------|--------|
| image | image | image |

[Link to Figma Design](https://www.figma.com/design/zEzrsNcGNKYj9JXTQa3moF/%F0%9F%93%8B-Engagement-Surveys-Q3?node-id=2875-178568&t=CoAk6TjzAth4FHYr-0)

## Implementation details

I'd loved to reuse the `RadialProgressChart` but to achieve this design we'd need to:

- Allow a background
- Allow to hide the "empty" path (the grey one)
- Allow to change the path size, since when making it smaller (`28px`) it looked too skinny. 

I decided to go forward with a new implementation because I can see a future where we add different status colors to this percentage chart, for example a chart with a low amount could be shown as red so signify there is something wrong, and the background, etc, helps with that.

### DIFF
--- a/packages/react/src/components/value-display/__stories__/ValueDisplay.mdx
+++ b/packages/react/src/components/value-display/__stories__/ValueDisplay.mdx
@@ -736,3 +736,79 @@ render: (item) => ({
 `,
   }}
 />
+
+### percentage
+
+Renders a percentage with a chart
+
+#### Value type
+
+```
+type value = number | undefined
+// or
+type value = {
+    percentage: number | undefined
+}
+```
+
+#### PercentageInputAsObject
+
+<Canvas
+  of={ValueDisplayStories.PercentageInputAsObject}
+  source={{
+    code: `
+render: (item) => ({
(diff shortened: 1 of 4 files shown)


packages/react/src/components/value-display/__stories__/ValueDisplay.mdx
```diff
@@ -736,3 +736,79 @@ render: (item) => ({
...
+
+Renders a percentage with a chart
+
+#### Value type
+
+```
+type value = number | undefined
+// or
+type value = {
+    percentage: number | undefined
```
sergiocarracedo, line 750: I miss the label and placeholder prop here

sergiocarracedo (approved):

PR #2534, opened 2025-09-05
fix: auto-select first filter by default when opening the filters picker

sauldom102:
## Description

When the user opens the filters picker and there is no active filter applied before, we should auto-select the first one on the list, this way we save the user a click, and it is really handy when there is only a single filter there.

https://github.com/user-attachments/assets/75e4940f-e91f-481d-bca7-a37f5978594d

### DIFF
--- a/packages/react/src/components/OneFilterPicker/components/FiltersControls.tsx
+++ b/packages/react/src/components/OneFilterPicker/components/FiltersControls.tsx
@@ -70,12 +70,10 @@ export function FiltersControls<Filters extends FiltersDefinition>({
     }
 
     if (isOpen) {
-      // First try to find a filter that has a value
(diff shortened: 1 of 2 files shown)


packages/react/src/components/OneFilterPicker/components/FiltersControls.tsx
```diff
@@ -70,7 +70,15 @@ export function FiltersControls<Filters extends FiltersDefinition>({
     }
 
     if (isOpen) {
-      setSelectedFilterKey(getFirstFilterNotEmpty()?.[0] as keyof Filters)
+      // First try to find a filter that has a value
```
nlopin, line 73: let's clean comments because the code below speaks for itself

nlopin (approved):

PR #2576, opened 2025-09-12
feat: 2 small fixes in kanban view

Rogermax:
Fix padding + first loading

### DIFF
--- a/packages/react/src/experimental/RichText/NotesTextEditor/index.stories.tsx
+++ b/packages/react/src/experimental/RichText/NotesTextEditor/index.stories.tsx
@@ -252,9 +252,8 @@ export const Default: Story = {
       },
       {
         type: "text",
-        label: "Text",
-        content: "Metadata",
         label: "Metadata",
+        content: "Metadata",
       },
     ],
   },


packages/react/src/experimental/OneDataCollection/visualizations/collection/Kanban/Kanban.tsx
```diff
@@ -210,7 +201,71 @@ export const KanbanCollection = <
...
+            } else {
+              // Moving to an empty lane - use the source record as reference
+              destinyRecord = {
+                record: sourceRecord as R,
+                position: "below",
+              }
+            }
+          }
+
+          await onMove(fromLaneId, toLaneId, sourceRecord as R, destinyRecord)
```
Rogermax, line 268: All this block will be edited in a new PR (I'm changing onMove) don't take it into consideration

packages/react/src/ui/Lane/Lane.tsx
```diff
@@ -9,7 +9,7 @@ import { LaneHeader } from "./components/LaneHeader"
 import { LoadingSkeleton } from "./components/LoadingSkeleton"
 import { LaneProps } from "./types"
 
-const DEFAULT_MAX_LANE_HEIGHT = 700
+const DEFAULT_MAX_LANE_HEIGHT = 600
```
sergiocarracedo, line 12: Im thinking the lane height should be inherited from the datacollection height

sergiocarracedo (approved):

packages/react/src/ui/Lane/Lane.tsx
Rogermax: Yes.. this is just to add a max height in case there is no limit set in the datacollection. Factorial web let objects (like datatables) to grow with no limit, this make problems with infinit-scroll.

PR #2603, opened 2025-09-16
fix: emit preset event with filter values

sauldom102:
## Description

We were emitting the preset onClick event with the label instead of with the filter value it triggers, which is way more useful for event tracking.

### DIFF
--- a/packages/react/src/experimental/OneDataCollection/useEventEmitter.ts
+++ b/packages/react/src/experimental/OneDataCollection/useEventEmitter.ts
@@ -35,17 +35,23 @@ export const useEventEmitter = <Sortings extends SortingsDefinition>({
 
   const emitFilterChange = useCallback(
     (filters: FiltersState<FiltersDefinition>) => {
-      const newFilter = Object.entries(filters ?? {}).find(
+      if (!filters) return
+
+      const changedFilter = Object.entries(filters).find(
         ([field, value]) => latestFilters.current?.[field] !== value
       )
 
-      if (!newFilter || !isScalar(newFilter[1])) return
+      if (!changedFilter) return
+
+      const [field, value] = changedFilter
+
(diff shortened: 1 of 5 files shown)


packages/react/src/experimental/OneDataCollection/useEventEmitter.ts
```diff
@@ -72,9 +72,18 @@ export const useEventEmitter = <Sortings extends SortingsDefinition>({
...
-    (preset: string) => {
+    (filters: FiltersState<FiltersDefinition>) => {
+      const newFilter = Object.entries(filters ?? {}).find(
+        ([field, value]) => latestFilters.current?.[field] !== value
+      )
+
+      if (!newFilter || !isScalar(newFilter[1])) return
+
+      latestFilters.current = filters
+
```
dmaza81, line 83:
```suggestion
(filters: FiltersState<FiltersDefinition>) => {
    if (!filters) return

    const changedFilter = Object.entries(filters).find(
      ([field, value]) => latestFilters.current?.[field] !== value
    )

    if (!changedFilter) return

    const [field, value] = changedFilter

    if (!isScalar(value)) return

    latestFilters.current = filters

```

packages/react/src/experimental/OneDataCollection/useEventEmitter.ts
```diff
@@ -72,9 +72,18 @@ export const useEventEmitter = <Sortings extends SortingsDefinition>({
...
       onEvent("datacollection.preset-click", {
-        name: preset,
+        name: newFilter?.[0],
+        value: newFilter?.[1],
```
dmaza81, line 86:
```suggestion
     name: field, value
```

dmaza81 (approved):
Small suggestion but not mandatory at all

PR #2637, opened 2025-09-22
feat/datacollection sort hide cols

sergiocarracedo:
- **feat: datacollection visualization settings**
- **feat: datacollection visualization settings logic**
- **feat: datacollection sort and hide**
- **chore: fix duplicated object key**
- **feat: order and hide settings component**
- **feat: order and hide settings component**
- **feat: datacollection settings**
- **feat: initial settings**


packages/react/.storybook/main.ts
```diff
@@ -42,8 +39,8 @@ const config: StorybookConfig = {
     getAbsolutePath("@storybook/addon-a11y"),
     getAbsolutePath("@storybook/addon-themes"),
     getAbsolutePath("@vueless/storybook-dark-mode"),
+    // getAbsolutePath("storybook-addon-tag-badges"),
```
dmaza81, line 42: Should we remove this line commented?

packages/react/src/components/OneFilterPicker/OneFilterPicker.tsx
```diff
@@ -108,7 +108,7 @@ const FiltersRoot = <Definition extends FiltersDefinition>({
   useEffect(() => {
     setLocalFiltersValue(value)
     // eslint-disable-next-line react-hooks/exhaustive-deps -- We deep compare the filters object
-  }, [JSON.stringify(filters)])
+  }, [JSON.stringify(filters), JSON.stringify(value)])
```
dmaza81, line 111: What about :
```
const filters = useMemo(() => buildFilters(rawFilters), [rawFilters]);
const value = useMemo(() => calculateValue(rawValue), [rawValue]);

useEffect(() => {
  setLocalFiltersValue(value);
}, [filters, value]);
```

Of course `buildFilters` and `calculateValue` will be our own functions

Or maybe we can use [use-deep-compare-effect](https://github.com/kentcdodds/use-deep-compare-effect)

packages/react/.storybook/main.ts
sergiocarracedo: Yes, that is the add-on was causing the build issue, I commented intead of removing it just in case this has side effects I didnt detect

packages/react/src/components/OneFilterPicker/OneFilterPicker.tsx
sergiocarracedo: oh i didnt know [use-deep-compare-effect](https://github.com/kentcdodds/use-deep-compare-effect). It is something we need a lot!!
I will install it a start to refactor out code

dmaza81 (approved):
Pretty awesome work!

PR #2638, opened 2025-09-22
fix: item click on Kanban View

Rogermax:
## Description

Adding the capability to click on items in Kanban View (missing feature done!)

https://github.com/user-attachments/assets/964722df-32c9-40fa-a97a-8453d38f711e

## Implementation details

As KanbanCards can drag and drop or be clicked we had to add in F0Card a prop to disable the Link that captures clicks and interface with the drag and drop.


packages/react/src/experimental/OneDataCollection/visualizations/collection/Kanban/index.spec.tsx
```diff
@@ -0,0 +1,122 @@
...
+type Person = RecordType & {
+  id: number | string
+  name: string
+}
+
+type TestFilters = FiltersDefinition
+type TestNavigationFilters = NavigationFiltersDefinition
+
+const Wrapper: React.FC<{ children: React.ReactNode }> = ({ children }) => (
+  <I18nProvider translations={defaultTranslations}>{children}</I18nProvider>
```
sergiocarracedo, line 26: use zeroRender from @/testing/utils.tsx that includes the i18n provider

packages/react/src/components/F0Card/CardInternal.tsx
```diff
@@ -108,6 +108,12 @@ export interface CardInternalProps {
    * Whether the card should have a full height
    */
   fullHeight?: boolean
+
+  /**
+   * When true, disables the full-card overlay link so parent components
+   * can manage drag-and-drop while still allowing click navigation via onClick
+   */
+  disableOverlayLink?: boolean
```
sergiocarracedo, line 116: add this property as private in the F0Card.tsx to avoid using it from outside f0

sergiocarracedo (changes requested):

PR #2652, opened 2025-09-22
fix(OneAlert): Action prop conditional 

siguenzaraul:
## Description

This PR fixes two issues in the **Alert** component:

1. **Support for alerts without actions**  
   - Based on the Figma design, alerts should be allowed without an action button.  

2. **Button height misalignment with multi-line links**  
   - When an alert included both a link and an action, if the link text wrapped into two lines, the button’s height was incorrectly stretched.  
   - We might want to consider using a `OneEllipsis` component to prevent multi-line wrapping in the future.  

---

### Before
Alert before fix

### After
Alert after fix 1
Alert after fix 2
Alert after fix 3
Alert after fix 4

---

## Design Reference

[Figma – Alert component](https://www.figma.com/design/pZzg1KTe9lpKTSGPUZa8OJ/Components?node-id=921-3949&t=t41w8IEX3jaEBlXa-4)

### DIFF
--- a/packages/react/src/experimental/OneAlert/index.tsx
+++ b/packages/react/src/experimental/OneAlert/index.tsx
@@ -44,7 +44,7 @@ const buttonVariants: Record<AlertVariant, "outline" | "promote" | "critical"> =
 interface AlertProps extends VariantProps<typeof alertVariants> {
   title: string
   description: string
-  action: {
+  action?: {
     label: string
     onClick?: () => void
   }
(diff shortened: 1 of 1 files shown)


packages/react/src/experimental/OneAlert/index.tsx
```diff
@@ -44,7 +44,7 @@ const buttonVariants: Record<AlertVariant, "outline" | "promote" | "critical"> =
 interface AlertProps extends VariantProps<typeof alertVariants> {
   title: string
   description: string
-  action: {
+  action?: {
     label: string
     onClick?: () => void
```
miriam-mr90, line 49: Does it make sense for `onClick` to be optional? I’d say it should be required if there’s an `action`.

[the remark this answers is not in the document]
siguenzaraul: It doesn't make any sense. I think we should eliminate optionality

packages/react/src/experimental/OneAlert/index.tsx
```diff
@@ -77,27 +77,31 @@ export const OneAlert = ({
...
+                href={link.href}
+                target="_blank"
+                rel="noreferrer"
+                className={cn(
+                  "flex items-center gap-1 rounded-sm px-2 py-0.5 text-base font-medium text-f1-foreground no-underline transition-colors hover:bg-f1-background-secondary-hover [&>svg]:text-f1-foreground-secondary",
+                  focusRing()
+                )}
+              >
+                {link.label}
+                <F0Icon icon={ExternalLink} size="sm" />
```
sergiocarracedo, line 93: please dont use an anchor use <OneLink>  instead (this component provides also the icon when target is blank

siguenzaraul: Oh, I didn't realize, this component probably predates the <Link /> one, I'll change it now!

packages/react/src/experimental/OneAlert/index.tsx
```diff
@@ -61,43 +60,43 @@ export const OneAlert = ({
...
-            <F0AvatarAlert type={variant} />
+            {variant === "neutral" ? (
+              <F0AvatarIcon icon={icon || Placeholder} size="sm" />
+            ) : (
+              <F0AvatarAlert type={variant} size="sm" />
+            )}
           </div>
           <div className="flex flex-col gap-0.5">
-            <h3 className={titleVariants({ variant })}>{title}</h3>
+            <p className={titleVariants({ variant })}>{title}</p>
```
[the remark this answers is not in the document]
siguenzaraul: Figma uses a text size of 14px and those are the default paragraphs, I can try forcing heading 3 to use a text size of 14px to make it more accessible

PR #2689, opened 2025-09-26
fix: avoid re-renders in useDataSource

Rogermax:
## Description

Fix to avoid unnecesary renders when using DataSource


sergiocarracedo:

PR #2702, opened 2025-09-26
F0AvatarFlag

denjalonso:


sergiocarracedo:
> @sergiocarracedo we discussed creating a package for flags. do you see it makes sense? it requires a setup to build and publish it

I found that makes the things (like build, publish, etc ) more complex, and it's not worthy as the idea is to move the flag to a CDN as soon as possible

sergiocarracedo (approved):

PR #2708, opened 2025-09-29
feat: searchbox in tableofcontent

sergiocarracedo:
## Description

Add searchbox in `F0TableOfContent`
Allow to hide children counters

image

[Link to Figma Design](https://www.figma.com/design/pZzg1KTe9lpKTSGPUZa8OJ/Components?node-id=15062-20382&t=Xzh5ZN1e5Ttia9Xy-0)

### DIFF
--- a/packages/react/src/lib/providers/i18n/i18n-provider-defaults.ts
+++ b/packages/react/src/lib/providers/i18n/i18n-provider-defaults.ts
@@ -52,6 +52,9 @@ export const defaultTranslations = {
     failedToLoadOptions: "Failed to load options",
     retry: "Retry",
   },
+  toc: {
+    search: "Search",
+  },
   collections: {
     sorting: {
       noSorting: "No sorting",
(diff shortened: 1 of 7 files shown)


packages/react/src/lib/providers/i18n/i18n-provider-defaults.ts
```diff
@@ -52,6 +52,9 @@ export const defaultTranslations = {
     failedToLoadOptions: "Failed to load options",
     retry: "Retry",
   },
+  toc: {
+    search: "Search",
```
nlopin, line 56: should they provide this value to the ToC itself? It could be a different message depending on the use

nlopin (approved):

packages/react/src/lib/providers/i18n/i18n-provider-defaults.ts
sergiocarracedo: they can provide a custom placeholder, this is the placeholder by default

PR #2719, opened 2025-09-30
Feat/datacollection customize height

Rogermax:
## Description

Improve Kanban View and Datacollection maxHeight

## Screenshots

https://github.com/user-attachments/assets/4864d698-bd7f-4bf0-8233-fccc9c1afa8d

## Implementation details

* Let pass a maxHeight prop to dataCollecion
* Fix Kanban onClick issue that didn't allow to select cards
* Improve drop logic to stick to the nearest card (even drop is in the lane)


sergiocarracedo:
We dont want to control de datacollection height in the datacollection, it should be done in the parent. At some point the fullheight props will be removed and that will be the default behavior

Rogermax:
> We dont want to control de datacollection height in the datacollection, it should be done in the parent. At some point the fullheight props will be removed and that will be the default behavior

mmm... but I don't see the problem on havingo a maxHeight prop.. why it's better having: 
```react
<div className="max-h-56">
    <OneDataCollection />
</div>
```

Than this:
```react
<OneDataCollection maxHeight="500" />
```

Moreover.. if the user wrap to handle the size.. the scroll will be the default one... 
If we let maxHeight be handled by the OneDataCollection some visualizations (like Kanban) can use the ScrollArea wrapper that unifies scrollbars in MAcos and Windows

PR #2730, opened 2025-09-30
feat: create sdm callout from the AiBanner component

josiola95:
## Description

In order to make the SDM available on the F0 inbox, we need a callout with tooltip. This PR add new callout component to F0 based on the AiBanner component.

image

[Link to Figma Design](figma.com/design/1ofJHGsqfUuEl75ZB0atfo/✨-Automations-and-smart-decision-making?node-id=4612-131008&t=ag4PZstpgUeedn19-1)

## Implementation details

[Slack conversation](https://factorialteam.slack.com/archives/C09DRQRAZEC/p1758632352897809)

### DIFF
--- a/packages/react/src/experimental/Banners/F0Callout/CalloutInternal.tsx
+++ b/packages/react/src/experimental/Banners/F0Callout/CalloutInternal.tsx
@@ -12,8 +12,7 @@ const calloutVariants = cva({
   base: "flex w-full flex-col rounded-lg p-[1px]",
   variants: {
     variant: {
-      default:
-        "bg-gradient-to-l from-[#A1ADE51F] via-[#E519431F] to-[#E556191F]",
+      ai: "bg-gradient-to-l from-[#A1ADE51F] via-[#E519431F] to-[#E556191F]",
       critical: "bg-f1-background-critical",
       positive: "bg-f1-background-positive",
       info: "bg-f1-background-info",
@@ -21,7 +20,7 @@ const calloutVariants = cva({
     },
   },
   defaultVariants: {
-    variant: "default",
+    variant: "ai",
   },
 })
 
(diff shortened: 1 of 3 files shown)


sauldom102 (commented):
Let's just better rename it to `F0Callout` as it is the name that we have for the component in figma

packages/react/src/experimental/Banners/F0SdmCallout/SdmCalloutInternal.tsx
```diff
@@ -0,0 +1,166 @@
...
+        >
+          <RichTextDisplay content={content} />
+          {hasTooManyActions && (
+            <div className="mt-2 rounded-md border border-f1-border-warning bg-f1-background-warning/10 px-3 py-2">
+              <p className="text-sm text-f1-foreground-warning">
+                ⚠️ The limit of callout actions is 2. Only the first 2 actions
+                will be displayed.
+              </p>
+            </div>
+          )}
```
sauldom102, line 99: We don't want to show any of this

packages/react/src/experimental/Banners/F0SdmCallout/SdmCalloutInternal.tsx
```diff
@@ -0,0 +1,166 @@
...
+
+export const SdmCalloutInternal = forwardRef<
+  HTMLDivElement,
+  SdmCalloutInternalProps
+>(function SdmCalloutInternal(
+  { title, onClose, content, calloutActions = [], variant = "default" },
+  ref
+) {
+  // Limit actions to maximum 2
+  const limitedActions = calloutActions.slice(0, 2)
```
sauldom102, line 49: Instead of slicing the array I would just throw an error here in the case of this array containing more than 2 elements

packages/react/src/experimental/Banners/F0SdmCallout/SdmCalloutInternal.tsx
```diff
@@ -0,0 +1,166 @@
...
+            <div>
+              {limitedActions[1] && (
+                <Button
+                  label={limitedActions[1].label}
+                  onClick={limitedActions[1].onClick}
+                  variant="outline"
+                  icon={limitedActions[1].icon}
+                />
+              )}
+            </div>
```
sauldom102, line 122: We could use a `map` here to improve the code, right?

PR #2745, opened 2025-10-02
feat: accept special icons in product updates modules

ChloeDumit:
## Description

Making the modules inside the product card accept not only module icons but special icons like One

Screenshot 2025-10-02 at 8 37 12 AM
Screenshot 2025-10-02 at 8 37 17 AM

### DIFF
--- a/packages/react/src/components/UpsellingKit/ProductCard/index.tsx
+++ b/packages/react/src/components/UpsellingKit/ProductCard/index.tsx
@@ -1,5 +1,10 @@
 import { Button } from "@/components/Actions/Button"
-import { F0AvatarModule, ModuleId } from "@/components/avatars/F0AvatarModule"
+import {
+  F0AvatarModule,
+  ModuleId,
+  modules,
+} from "@/components/avatars/F0AvatarModule"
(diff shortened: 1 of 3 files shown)


packages/react/src/components/UpsellingKit/ProductCard/index.tsx
```diff
@@ -11,7 +16,7 @@ export type ProductCardProps = {
   isVisible: boolean
   dismissable?: boolean
   trackVisibility?: (open: boolean) => void
-  module: ModuleId
+  module: ModuleId | IconType
```
sergiocarracedo, line 19: I prefer to pass the icon in another prop for example
```
Visible: boolean
  dismissable?: boolean
  trackVisibility?: (open: boolean) => void 
} & (| 
{ 
    type: "one-campaign"
    module?: never
} | {
    module: ModuleId
    type?: never
})

sergiocarracedo: doit in that way we are sure no other icons are used in an incorrect way

PR #2748, opened 2025-10-02
feat: add itemFilter in useData

sauldom102:
## Description

For infinite pagination especially we need a way to filter our items locally based on a condition (e.g. data present in the Apollo cache), for those cases we will be able to define a filtering function that will be iterated on all of our raw data elements and discard those that don't match our condition, updating the total count accordingly.


sergiocarracedo (approved):

nlopin:
I have an impression that the surface of the data collection is so big, I won't be able to understand what to use. Is there a way to have one function that does the filtering job?

PR #2755, opened 2025-10-03
fix: select remove option on clear

sergiocarracedo:
## Description

- fix(select): remove selected option when the the clear button is used to set the value (to '')
- feat(inputfield): clear button as button a focusable. Stops propagation
- chore(select): add clear test

image

https://github.com/user-attachments/assets/85eb82ba-f54d-4e1c-9eee-f664230c2359


sergiocarracedo:
> Isn't the button too bold? it has more visual weight than the selected value

its the focus ring, not the real button

PR #2775, opened 2025-10-07
feat/select primitive multiselect

sergiocarracedo:
- **feat: multiselect primitive**
- **feat: ui select component multiselection**
- **feat: ui select component multiselection**


packages/react/package.json
```diff
@@ -177,8 +177,26 @@
...
+    "@radix-ui/react-id": "^1.1.1",
+    "@radix-ui/react-popper": "^1.2.7",
+    "@radix-ui/react-portal": "^1.1.9",
+    "@radix-ui/react-primitive": "^2.1.3",
+    "@radix-ui/react-switch": "^1.2.6",
+    "@radix-ui/react-use-callback-ref": "^1.1.1",
+    "@radix-ui/react-use-controllable-state": "^1.2.2",
+    "@radix-ui/react-use-layout-effect": "^1.1.1",
+    "@radix-ui/react-use-previous": "^1.1.1",
+    "@radix-ui/react-visually-hidden": "^1.2.3",
```
sauldom102, line 199: Do we need to add all of this? 🤔

packages/react/src/ui/Select/components/radix-ui/select.tsx
nlopin, line 1: anything specific to check here or it is a full copy of radix ui?

packages/react/src/ui/Select/components/Select.tsx
```diff
@@ -1,19 +1,27 @@
...
 }
+
 /**
  * Select Root component
  */
-const Select = (props: SelectProps) => {
-  // If open prop is not provided, we'll manage it internally
+
+const Select = <T extends string = string>(props: SelectProps<T>) => {
+  type Value = Exclude<typeof props.value, undefined>
```
nlopin, line 22: you can use `NonNullable<typeof props.value>`

packages/react/package.json
sergiocarracedo: yes, to create the custom radix select primitive replacement

packages/react/src/ui/Select/components/radix-ui/select.tsx
sergiocarracedo: it's a modified copy to allow multiselection

nlopin: > it's a modified copy to allow multiselection

could you point me to the lines you modified?

PR #2809, opened 2025-10-13
fix: allow to truncate tags

rubenmoya:
## Description

We've got a report about a data table showing information that overlaps and makes it impossible to read, you can see it here:

image

I've wrapped the `OverflowList` component in a `300px` `div` with a red border so we can see the overflow.

Before:
image

After:
image

## Implementation details

The were a few problems. The first one is that the `ValueDisplay` renderer for `tagList` is not using the `fill` prop, so any tags inside it won't try and fit the available content.

The second problem was that in the `calculateVisibleItemCount`, we were allowing to return no visible elements, but we don't want that, we want to show at least one of them, and that item should be truncated.

Then we needed to make the `BaseTag` truncate the text, and add a `min-w-0` to avoid the overflowing.

### DIFF
--- a/packages/react/src/components/tags/BaseTag/index.tsx
+++ b/packages/react/src/components/tags/BaseTag/index.tsx
@@ -37,7 +37,7 @@ export const BaseTag = forwardRef<HTMLDivElement, Props>(
     >
       {left}
       {!!text && (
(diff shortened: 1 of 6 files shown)


sergiocarracedo:
Tags stabilization that includes that was merged

PR #2810, opened 2025-10-13
feat: duplicate one sidepanel without AI

AmaliaGlez:
## Description

We’re introducing a version of the current AI chat interface to display a waitlist promotion for users who don’t have access to the feature. This allows us to promote One directly in-app without enabling any AI-related functionality.

| Users without One and NOT in waitlist | Users without One but on waitlist |
| --- | --- |
| Screenshot 2025-10-14 at 01 18 50 | Screenshot 2025-10-14 at 01 19 31 |

| Users with One | Users with One but disabled on purpose |
| --- | --- |
| Screenshot 2025-10-14 at 01 24 11 | Screenshot 2025-10-14 at 15 23 27 |

🎨 [Figma](https://www.figma.com/design/6Zg3LO7NxYpDDpJlLgVFXW/One-GTM---25Q3?node-id=6325-27268&t=04uVjCOXFATThvkX-0)

## Implementation details

- Duplicate existing sidepanel component to display promotional content only.
- Remove AI logic and dependencies to avoid unnecessary requests when the feature is inactive.

This approach was chosen as a temporary solution, copying the component is faster and safer than refactoring the existing chat logic, which could introduce regressions.

The promotional component will be removed once the waitlist campaign ends.

### DIFF
--- a/packages/react/src/experimental/AiPromotionChat/providers/AiPromotionChatStateProvider.tsx
+++ b/packages/react/src/experimental/AiPromotionChat/providers/AiPromotionChatStateProvider.tsx
@@ -101,7 +101,7 @@ export const AiPromotionChatStateProvider: FC<
       ).matches
       setShouldPlayEntranceAnimation(!prefersReducedMotion)
     }
-  }, [open])
+  }, [open, onShow])
 
   return (
     <AiPromotionChatStateContext.Provider


sergiocarracedo:
Please dont forget to remove it alter the campaign

sergiocarracedo (approved):

PR #2824, opened 2025-10-15
fix(Dialog): loading spinner size

rubenmoya:
## Description

When the `Dialog` component has attached an async action the button shows a spinner. Since we're overriding the `Button` styles inside the `Dialog` to make them full size it was breaking the icon.

Before:
spinner before

After:
spinner after

## Implementation details

I just changed the selector to target direct childs, and put it in the `div` previous to the `Button`.


sergiocarracedo (approved):

nlopin:
that was the fastest merge in history

PR #2855, opened 2025-10-21
feat(DataCollection): add list navigation filter

rubenmoya:
## Description

In Engagement, we need a new navigation filter to move through the different cycles for each survey.

image

Keep in mind, that even if you see dates there, we're not selecting date ranges from a calendar by a given granularity like we do in the date navigation. In this case is a list of cycles, and the granularity could be different, or non-existen if the survey is sent manually.

More information [in this slack thread](https://factorialteam.slack.com/archives/C082ZNKS403/p1759318985977889)

Closed:
image

Open:
image

With different list:
image

Use in the DataCollection:
image

## Implementation details

This PR implements a new `F0ListNavigation` component that allows users to navigate through a given list using arrows. This component is not exported, right now it's only used internally for the `DataCollection`.


sergiocarracedo (approved):

sergiocarracedo:
Does this PR makes sense now? If not, could you  close it? if yes can you merge it?

PR #2878, opened 2025-10-24
feat: use dropdown groups

sergiocarracedo:
## Description

Remove { type: "separator" } to show the items in groups in the dropdown button, use a group structure instead allowing us to add other props like the group title, and making simple the manage the data

For simplicity the item prop accept an array of items (like before add the groups), a group or an array of groups

e.g.

// Same as before (no groups or in another way of see it, one group)
```
items={[{
        value: "1",
        label: "Item 1",
        description: "New creation process",
        icon: Add,
      },
      {
        value: "2",
        label: "Item 2",
        description: "Edit item's information",
        icon: Pencil,
      }]
}

// Groups

items={[
      {
        label: "Group 1",
        items: [
          {
            value: "1",
            label: "Item 1",
            description: "New creation process",
            icon: Add,
          },
          {
            value: "2",
            label: "Item 2",
            description: "Edit item's information",
            icon: Pencil,
          },
        ],
      },
      {
        label: "Group 2",
        items: [
          {
            value: "3",
            label: "Item 3",
            description: "Save changes",
            icon: Save,
          },
        ],
      },
    ]}
```

## Implementation details

Refactor datacollection bulkactions and primary actions to pass the groups structure to keep the datacollection api the same, but we should migrate it to the group prop type in the future

### DIFF
--- a/packages/react/src/components/Actions/Button/internal.tsx
+++ b/packages/react/src/components/Actions/Button/internal.tsx
@@ -1,5 +1,6 @@
 import { Button as ShadcnButton } from "@/ui/button"
 
+import { DataAttributes } from "@/global.types"
 import { EmojiImage } from "@/lib/emojis.tsx"
 import { cn } from "@/lib/utils"
 import { cva } from "cva"
--- a/packages/react/src/components/Actions/Link/OneLink.tsx
+++ b/packages/react/src/components/Actions/Link/OneLink.tsx
@@ -1,3 +1,4 @@
+import { DataAttributes } from "@/global.types"
 import { cva, type VariantProps } from "cva"
 import { forwardRef } from "react"
 import ExternalLink from "../../../icons/app/ExternalLink"
--- a/packages/react/src/components/F0Checkbox/F0Checkbox.tsx
+++ b/packages/react/src/components/F0Checkbox/F0Checkbox.tsx
@@ -1,3 +1,4 @@
+import { DataAttributes } from "@/global.types"
 import { experimentalComponent } from "@/lib/experimental"
 import { Checkbox as CheckboxRoot } from "@/ui/checkbox"
 
--- a/packages/react/src/experimental.ts
+++ b/packages/react/src/experimental.ts
@@ -1,3 +1,1 @@
-/// <reference types="./global" />
-
 export * from "./experimental/exports"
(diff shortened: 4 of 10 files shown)


packages/react/src/components/Actions/OneDropdownButton/OneDropdownButton.tsx
```diff
@@ -1,66 +1,85 @@
...
+ * Normalize the items to an array of DropdownButtonGroup
+ */
+const normalizeItems = (
+  items:
+    | OneDropdownButtonItem[]
+    | OneDropdownButtonGroup[]
+    | OneDropdownButtonGroup
+) => {
+  if (Array.isArray(items)) {
+    // DropdownButtonItem[]
```
nlopin: AI alert 👮

packages/react/src/experimental/OneActionBar/index.tsx
```diff
@@ -139,49 +204,34 @@ export const OneActionBar = ({
...
+                      items={primaryActionsDropdownItems}
                       onClick={(value) => {
-                        const action = primaryActions.find(
-                          (a) => isActionType(a) && a.label === value
-                        )
-                        if (action && isActionType(action)) {
-                          action.onClick?.()
-                        }
+                        const action = getActionByValue(value)
+                        ;(action as ActionType)?.onClick?.()
```
nlopin: something is wrong here. why do we have semicolon at the beginning?

packages/react/src/components/Actions/OneDropdownButton/OneDropdownButton.tsx
sergiocarracedo: 😓 I dont know how to interpret that. I wrote every single line of the normalizeFunction and added the comments to see easily which type we are converting to DropdownButtonGroup 😓

packages/react/src/experimental/OneActionBar/index.tsx
sergiocarracedo: semicolon is needed as the line starts with a (, if no semicolon parser will interpret this is code for the previous line 
https://standardjs.com/rules#semicolons

packages/react/src/components/Actions/OneDropdownButton/OneDropdownButton.tsx
nlopin: I'm sorry! This is type of comments AI loves to add. I'm wondering that maybe we can create separate functions to address each type? Not mandatory though

PR #2944, opened 2025-11-07
fix(AiChat): targetBlank to MessageSources and consistency in loading states

siguenzaraul:
## Description

This PR adds:
- The option to select from target _self and target _blank, the options we want to handle in the sources depend on the context. 
- The description of the icon prop 
- Action item in loading state with the word thinking, to improve the UX

https://github.com/user-attachments/assets/d717971c-c451-4b8d-bf03-bde4ebdbcddd

### DIFF
--- a/packages/react/src/experimental/AiChat/ActionItem.tsx
+++ b/packages/react/src/experimental/AiChat/ActionItem.tsx
@@ -1,7 +1,6 @@
 import { F0Icon } from "@/components/F0Icon"
 import OutlineCircle from "@/icons/animated/CheckCircleLine"
 import DottedCircle from "@/icons/app/DottedCircle"
-import { motion } from "motion/react"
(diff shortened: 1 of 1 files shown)


packages/react/src/lib/providers/i18n/i18n-provider-defaults.ts
```diff
@@ -202,6 +202,7 @@ export const defaultTranslations = {
     sendMessage: "Send message",
     thoughtsGroupTitle: "Reflection",
     resourcesGroupTitle: "Resources",
+    thinking: "Thinking...",
```
sauldom102: How validated is this copy? Just asking because I can imagine we might had another one before and I'm not sure if we're having some back and forth regarding this

sauldom102 (approved):

PR #2947, opened 2025-11-10
feat: add tooltip on F0Icon and CardMetadata

Rogermax:
## Description

Add tooltip on F0Icon + Card

https://github.com/user-attachments/assets/35f5346f-7623-43e8-bb8a-072c25a490d7

Screenshot 2025-11-10 at 10 13 38

### DataCollection Card View

Screenshot 2025-11-10 at 12 27 16

### DataCollection Kanban View

Screenshot 2025-11-10 at 12 27 03


Rogermax:
> @Rogermax add the tooltip to the icon can be problematic as a lot of component use it. I think is better to add the tooltip in the card component when we generate card icon and value

But TooltipWrapper is nothing if you don't set that prop... shouldn't have any impact

PR #2957, opened 2025-11-10
fix: support Select component as list with scroll

sauldom102:
## Description

We need to support the `Select` component to be rendered without its input, being always open and scrollable.

https://github.com/user-attachments/assets/9fd136be-1a36-4bb2-834a-d70f1f00f03f


sergiocarracedo (approved):

sergiocarracedo:

PR #2974, opened 2025-11-17
feat: loading and disable in datacollection actions

sergiocarracedo:
## Description

Allow to set the loading and disabled state in the Datacollection actions (ONLY in the buttons not in Button dropdown or dropdown as they don't supp ort it yet)

image

image


dmaza81 (approved):
💯

PR #3013, opened 2025-11-21
fix: button when label is undefined

sergiocarracedo:
## Description

Avoid error when label is undefined (it should not be, but there are some legacy cases)


sauldom102 (approved):

developerdanx (approved):
Nice comments by Copilot :)

PR #3044, opened 2025-11-26
fix: add support to only allow certain question types in co-creation form

sauldom102:
## Description

Per product need we need to be able to only allow certain question types in our form. This has been mentioned [here](https://factorialteam.slack.com/archives/C09MNTWCNPQ/p1764152719162219?thread_ts=1764144020.149489&cid=C09MNTWCNPQ).

We could for example only allow select and multi-select question types in our form now:

Screenshot 2025-11-26 at 12 44 07

### DIFF
--- a/packages/react/src/experimental/CoCreationForm/Form/index.stories.tsx
+++ b/packages/react/src/experimental/CoCreationForm/Form/index.stories.tsx
@@ -32,7 +32,6 @@ type Story = StoryObj<typeof CoCreationForm>
 export const Default: Story = {
   args: {
     isEditMode: true,
-    allowedQuestionTypes: ["text", "select"],
     elements: [
       {
         type: "question",


packages/react/src/experimental/CoCreationForm/Form/index.tsx
```diff
@@ -91,11 +91,13 @@ export const CoCreationForm = ({
   elements,
   isEditMode,
   onChange,
+  allowedQuestionTypes,
 }: CoCreationFormProps) => (
   <CoCreationFormProvider
```
sauldom102: Although we shouldn't receive questions here with not allowed types I would like to filter out here those questions too. I will do that once we have the changes from [this other PR](https://github.com/factorialco/f0/pull/3043) merged to avoid causing some conflicts.

PR #3090, opened 2025-12-04
fix: export dashboard and layout

sergiocarracedo:
## Description

Export Dashboard and Layouts

### DIFF
--- a/.release-please-manifest.json
+++ b/.release-please-manifest.json
@@ -1,5 +1,5 @@
 {
-  "packages/react": "1.302.3",
+  "packages/react": "1.303.0",
   "packages/react-native": "0.20.1",
   "packages/core": "1.42.0"
 }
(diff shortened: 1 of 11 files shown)


packages/react/src/lib/experimental.test.tsx
```diff
@@ -0,0 +1,344 @@
...
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest"
+import { experimentalComponent } from "./experimental"
+import { UserPlatformProvider } from "./providers/user-platafform/UserPlatformProvider"
+
+const _ = React
+
+describe("experimentalComponent", () => {
+  beforeEach(() => {
+    vi.clearAllMocks()
+    vi.useFakeTimers()
```
dmaza81: Maybe we should move this to general setup, if it is not there

PR #3101, opened 2025-12-09
fix: prevent unwanted behavior when having select questions with same values in the options

sauldom102:
## Description

We were throwing an error when we found select questions with options with same values as they were causing a buggy behavior. What we are doing in this PR is handling that more properly so that if this case happens, we modify the values of the options so that they are not the same and it behaves properly.


packages/react/src/experimental/CoCreationForm/SelectQuestion/index.tsx
```diff
@@ -116,6 +156,10 @@ export const SelectQuestion = ({ options, ...props }: SelectQuestionProps) => {
     })
   }
 
+  if (someOptionsWithSameValue) {
+    return null
+  }
```
sauldom102: We can reach this point for the first render before modifying the option values, but it won't be noticiable for the user

PR #3115, opened 2025-12-10
fix: add support to get isThinking from useAiChat hook

sauldom102:
## Description

In order to be able to show the ["applying changes" state](https://github.com/factorialco/f0/pull/3106) in the co-creation form in the frontend we need to be able to know when One is thinking, we need to get that exposed in `useAiChat` hook. It could be something useful for other initiatives as well.

### DIFF
--- a/packages/react/src/experimental/AiChat/providers/AiChatStateProvider.tsx
+++ b/packages/react/src/experimental/AiChat/providers/AiChatStateProvider.tsx
@@ -13,7 +13,13 @@ import {
 } from "react"
 import { WelcomeScreenSuggestion } from "../components/WelcomeScreen"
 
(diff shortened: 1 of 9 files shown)


packages/react/src/experimental/AiChat/providers/AiChatStateProvider.tsx
```diff
@@ -177,31 +190,53 @@ export const AiChatStateProvider: FC<PropsWithChildren<AiChatState>> = ({
...
+}
+
+/**
+ * Internal hook that provides access to internal state setters.
+ * Not part of the public API.
+ * @internal
+ */
+export function useAiChatInternal(): AiChatInternalContextValue {
+  const context = useContext(AiChatStateContext)
+
```
sauldom102: This is because we don't want to expose `setIsThinking` outside of F0

PR #3118, opened 2025-12-10
chore: cycle dependencies detection

sergiocarracedo:
## Description

Checks new cycle imports in the codebase on precomit and on ci

If the cycle imports are increased, the PR or commit will be not accepted
If is the same as the previous commit or less nothing happens

## Implementation details

We store the previous state in a .cache folder to speedup multiple runs
Uses dpdm to get the cycle imports (and uses f0 and experimental as entrypoints)

### DIFF
--- a/packages/react/.scripts/check-cycle-dependencies.ts
+++ b/packages/react/.scripts/check-cycle-dependencies.ts
@@ -269,6 +269,10 @@ function runDpdmOnCommit(
   const tempReactPath = join(tempDir, reactPackageRelativePath)
 
   try {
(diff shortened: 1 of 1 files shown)


.github/workflows/size.yml
```diff
@@ -1,8 +1,9 @@
 name: "size"
 on:
-  pull_request:
-    branches:
-      - "main"
+  workflow_dispatch:
+  # pull_request:
```
developerdanx: This should be commented?

developerdanx (changes requested):

PR #3131, opened 2025-12-14
feat: introduce alwaysOpen prop in F0Select for inline selection lists

siguenzaraul:

### DIFF
--- a/packages/react/src/components/F0Select/types.ts
+++ b/packages/react/src/components/F0Select/types.ts
@@ -48,6 +48,10 @@ type F0SelectBaseProps<T extends string, R = unknown> = {
    * Useful for inline selection lists or embedded selectors.
    */
   alwaysOpen?: boolean
+  /**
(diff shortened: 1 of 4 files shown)


packages/react/src/components/F0Select/types.ts
```diff
@@ -42,6 +42,16 @@ type F0SelectBaseProps<T extends string, R = unknown> = {
...
+  /**
+   * When true, displays the select as an always-visible list without a trigger.
+   * The component renders only the content area (options list) without the input field.
+   * Useful for inline selection lists or embedded selectors.
+   */
+  alwaysOpen?: boolean
+  /**
+   * Height of the list container when alwaysOpen is true.
+   */
+  alwaysOpenHeight?: "sm" | "md" | "lg"
```
sergiocarracedo: Why we limit the height in the component? The component should use the max height and the parent should be the one that set the height limit

PR #3142, opened 2025-12-16
fix: add support to show video for event community post

sauldom102:
## Description

We got reported [this ticket](https://factorialmakers.atlassian.net/browse/FCT-40838) in which we can find a video being attached to a community post event and we need to support this case.

Screenshot 2025-12-16 at 11 27 36


packages/react/src/experimental/Information/Communities/Post/PostEvent/index.tsx
```diff
@@ -1,39 +1,56 @@
 import { Skeleton } from "@/ui/skeleton"
-import { f1Colors } from "@factorialco/f0-core"
+import { f1Colors } from "../../../../../../../core/src/tokens/colors"
 import { formatTime } from "../../../../../lib/date"
```
sergiocarracedo: cloud we use absolute path? @/lib/...

sauldom102: Sure!

PR #3144, opened 2025-12-16
feat: add f0bignumber component

sergiocarracedo:
## Description

Adds the big number component

image

## Implementation details

- add numeric type and utilities to represent numeric values and format then
- refactor: F0TagBalance to use new numeric types

### DIFF
--- a/packages/react/src/lib/numeric/__tests__/numericFormatter.test.ts
+++ b/packages/react/src/lib/numeric/__tests__/numericFormatter.test.ts
@@ -444,7 +444,7 @@ describe("numericFormatter", () => {
         unitsPosition: "prepend",
       }
       const result = numericFormatter(value, { unitsSpaced: true })
-      expect(result).toBe(" $123.45")
+      expect(result).toBe("$ 123.45")
     })
 
     it("should add space after units when unitsSpaced is true and unitsPosition is append", () => {


packages/react/src/components/F0BigNumber/internal/defaults.ts
```diff
@@ -0,0 +1,7 @@
+import { NumericFormatterOptions } from "@/lib/numeric"
+
+export const defaultFormatterOptions: NumericFormatterOptions = {
+  locale: "en-US",
```
eliseo-juan: We need to make sure we set this properly in the monorepo

packages/react/src/components/tags/F0TagBalance/F0TagBalance.tsx
```diff
@@ -1,46 +1,52 @@
...
+// ) => {
+//   const space = spaceBetweenUnits ? " " : ""
+//   return [
+//     value.unitsPosition === "left" && value.units ? value.units : "",
+//     numberFormat(value.number, value.decimalPlaces, value.locale),
+//     value.unitsPosition === "right" && value.units ? value.units : "",
+//   ]
+//     .filter(Boolean)
+//     .join(space)
+// }
```
eliseo-juan: We can remove this right?

packages/react/src/components/F0BigNumber/internal/defaults.ts
sergiocarracedo: ok!

sergiocarracedo: We already are setting it in l10nContext, but i realized we were not using it a defauilt value, I created hook to manage that

PR #3163, opened 2025-12-18
chore: oxlint

sergiocarracedo:
## Description

Replaces eslint by oxlint as is MUCH faster

### oxfmt
❯ time p lint

> @factorialco/f0-react@1.312.0 lint /Users/sergio.carracedo/works/f0-vitest-upgrade/packages/react
> oxlint src/**/*.{ts,tsx} --max-warnings 0

WARNING: JS plugins are experimental and not subject to semver.
Breaking changes are possible while JS plugins support is under development.
Found 0 warnings and 0 errors.
Finished in 42ms on 65 files using 12 threads.
corepack pnpm lint  0.20s user 0.09s system 49% cpu 0.584 total

### eslint
❯ time p lint

> @factorialco/f0-react@1.312.0 lint /Users/sergio.carracedo/works/f0/packages/react
> eslint -c eslint.config.mjs src/**/*.{ts,tsx} --report-unused-disable-directives --max-warnings 0

corepack pnpm lint  1.70s user 0.45s system 155% cpu 1.381 total


packages/react/.oxlintrc.json
```diff
@@ -0,0 +1,1261 @@
...
+            "varsIgnorePattern": "^_",
+            "args": "after-used",
+            "argsIgnorePattern": "^_"
+          }
+        ],
+        "react/react-in-jsx-scope": "off",
+        "react/prop-types": "off",
+        "react/display-name": ["error", {}]
+      },
+      "globals": {
```
connorshea: FWIW you can probably replace this huge list by just adding `browser` or `node` to the env option, depending on the application.

I thought I had fixed this problem in the @oxlint/migrate script a while ago, but maybe not 🤔

developerdanx:
> Hi @developerdanx , yes the goal es tu use the same tooling (and maybe the same of most of the same rules) in all the packages, but I have no idea about how you are structuring the RN package I prefered to limit the scope
> 
> Regarding the experimental, for the react use case works well, I wasnt able to find any limitation https://oxc.rs/docs/guide/usage/linter/js-plugins.html#api-support. But in any case you can use oxlint as eslint plugins, which replaces the native rules with oxc and left the other ones in eslint
> 
> Yes i know biome, but I prefer to use oxc as is part of the vite ecosystem https://voidzero.dev/

Thanks for the clarification, Sergio — really appreciate the explanation 🙌

Totally makes sense to limit the scope initially while the React Native package structure is still evolving. I also like the idea of reusing Oxlint as ESLint plugins where needed — that seems like a pragmatic way to cover rules that aren’t fully supported yet.

I did come across an interesting real-world example that may be useful when we expand: the react-native-permissions library already uses .oxlintrc.json in a production RN project:
[https://github.com/zoontek/react-native-permissions/blob/master/.oxlintrc.json](https://github.com/zoontek/react-native-permissions/blob/master/.oxlintrc.json?utm_source=chatgpt.com)

This shows that Oxlint can be applied successfully in React Native codebases, which could be helpful as we scale linting across the monorepo.

I also see your point about staying in the Oxc/Vite ecosystem — makes sense to keep things aligned there. I’m happy to help test the RN/Expo packages with Oxlint when we decide to expand the coverage.

developerdanx (changes requested):
Thank you for PR Sergio. 

Can we validate this checklist before merging?


- [ ] Update lefthook.yml to include both format commands

- [ ] Test committing files from React package (should use oxfmt)

- [ ] Test committing files from React Native package (should use prettier)

- [ ] Test committing files from both packages simultaneously

- [ ] Verify no conflicts occur

PR #3164, opened 2025-12-18
chore: dpdm-fast

sergiocarracedo:
## Description

Uses dpdm-fast to check cycle imports

### dpdm
corepack pnpm cycle-deps:delta --ci  14.89s user 5.43s system 126% cpu *16.051 total*

### dpdm-fast
corepack pnpm cycle-deps:delta --ci  0.42s user 0.85s system 77% cpu *1.630 total*

10x faster

### DIFF
--- a/packages/react/package.json
+++ b/packages/react/package.json
@@ -144,7 +144,6 @@
     "chromatic": "^11.28.0",
     "consola": "^3.4.0",
     "dotenv": "^16.4.7",
-    "dpdm": "^3.14.0",
     "dpdm-fast": "^1.0.14",
     "emoji-mart": "^5.6.0",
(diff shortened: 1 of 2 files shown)


packages/react/package.json
```diff
@@ -144,7 +144,7 @@
     "chromatic": "^11.28.0",
     "consola": "^3.4.0",
     "dotenv": "^16.4.7",
-    "dpdm": "^3.14.0",
+    "dpdm-fast": "^1.0.14",
```
eliseo-juan: I can not wait for `dpdm-super-fast`

sergiocarracedo: dpdm-fast-pro-max

PR #3168, opened 2025-12-19
feat: image fit options for card view

leonid-kovalenko-factorial:
## Description

Added imageFit and imageHeight to F0Card and OneDataCollection/card subtype to have control over image visualisation.

[Link to Figma Design](https://www.figma.com/design/OapOAF2BXqpyWgwt0jxYCt/Device---Procurement?node-id=8008-40581&p=f&t=hH14vDBKlcoczg6l-0)

## Implementation details

While developing a device catalog (also known as a marketplace), we encountered limitations in image visualization control. Most of the images in the catalog are square, so they get cropped and look unattractive.
This is what it looks like now:
image

Now is precise control of how to visualise the image:
image
image
and many more

### DIFF
--- a/packages/react/src/experimental/OneDataCollection/visualizations/collection/Card/index.tsx
+++ b/packages/react/src/experimental/OneDataCollection/visualizations/collection/Card/index.tsx
@@ -45,6 +45,7 @@ export type CardVisualizationOptions<
   image?: (record: T) => string
   imageFit?: CardImageFit
   imageSize?: CardImageSize
+  blurredBackground?: boolean
   compact?: boolean
 }
 
@@ -129,6 +130,7 @@ type GroupCardsProps<
   image?: (record: Record) => string
   imageFit?: CardImageFit
   imageSize?: CardImageSize
+  blurredBackground?: boolean
   compact?: boolean
   tmpFullWidth?: boolean
 }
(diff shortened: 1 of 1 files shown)


packages/react/src/components/F0Card/CardInternal.tsx
```diff
@@ -117,12 +135,27 @@ export interface CardInternalProps {
...
+    case "contain":
+      return "object-contain h-full w-full"
+    case "fit-width":
+      return "object-cover h-full w-full"
+    case "fit-height":
+      return "object-contain h-full w-auto"
+    case "scale-down":
+      return "object-scale-down h-full w-full"
+  }
+}
```
sergiocarracedo: please use a map instead of a switch, with the correct type makes imposible to forget a map, something you can not control in a switch

```suggestion
const imageFitClassMap: Record<CardImageFit, string> = {
   contain: "object-contain h-full w-full",
   "fit-width": "object-cover h-full w-full",
   "fit-height": "object-contain h-full w-auto".
   "scale-down": "object-scale-down h-full w-full"
  }
function getImageFitClassName(imageFit: CardImageFit): string {
   return imageFitClassMap[imageFit]
}
```

sergiocarracedo (changes requested):

PR #3171, opened 2025-12-20
feat: add label prop and DropdownMenuLabel to dropdown 

adomingc:
## Description

Add label prop to Dropdown and also a new item type "label" using DropdownMenuLabel component from dropdown-menu

image

[Link to Figma Design](https://www.figma.com/design/DzGbJkv4p3IClmiQfhRGw5/%F0%9F%91%A9%F0%9F%8F%BB%E2%80%8D%F0%9F%8F%AB-LMS?node-id=13600-214238&t=g2dY9XFkB0Ahg8J1-4)


sauldom102 (approved):
Maybe we could add some min width here, I know in this case it's just because we only have short labels but we could establish a minimum for such cases

Image

PR #3174, opened 2025-12-23
feat: useSelectable add pageOnlySelection feature 

siguenzaraul:
## Description

Add `pageOnlySelection` feature to the `useSelectable` hook and OneDataCollection component to control selection behavior across paginated data. This enhancement allows developers to choose between two selection modes:

- **Page-only selection (default)**: Selection state resets when navigating between pages
- **Cross-page selection**: Selection state persists across all pages

This feature is particularly useful for bulk actions that should operate either on the current page only or across the entire dataset.

### What have you changed?

1. **Added `pageOnlySelection` property** to:
   - `DataSourceDefinition` type (`datasource.typings.ts`)
   - `UseSelectableProps` type (`useSelectable/typings.ts`)
   - Propagated through `ExampleComponent` and its props

2. **Enhanced `useSelectable` hook** (`useSelectable.ts`):
   - Added logic to track current page/cursor using `previousPageIdentifierRef`
   - Implemented `useEffect` hook that clears selections when page changes (only when `pageOnlySelection` is enabled)
   - Modified `itemsStatus` and `selectedIds` computations to filter items based on current page when in page-only mode
   - Updated `totalKnownItemsCount` to reflect current page count in page-only mode
   - Adjusted selection logic to respect the page-only constraint

3. **Added comprehensive documentation** (`selectable-bulk-actions.mdx`):
   - Explained both selection modes with clear examples
   - Documented the differences in behavior
   - Added usage examples with code snippets

4. **Created two Storybook stories** (`index.stories.tsx`):
   - `WithPageOnlySelection`: Demonstrates default behavior
   - `WithCrossPageSelection`: Demonstrates cross-page selection with `pageOnlySelection: false`

5. **Updated F0Select component** to use the default `pageOnlySelection: false` for backward compatibility

### DIFF
--- a/packages/react/src/components/F0Select/F0Select.tsx
+++ b/packages/react/src/components/F0Select/F0Select.tsx
@@ -361,7 +361,7 @@ const F0SelectComponent = forwardRef(function Select<
     selectedState: initialSelectedState,
     disableSelectAll: disableSelectAll,
     isSearchActive: !!currentSearch,
-    pageOnlySelection: false,
+    allPagesSelection: true,
   })
 
   /**
--- a/packages/react/src/experimental/OneDataCollection/OneDatacollection.tsx
+++ b/packages/react/src/experimental/OneDataCollection/OneDatacollection.tsx
@@ -333,6 +333,16 @@ const OneDataCollectionComp = <
 
   const [selectedItemsCount, setSelectedItemsCount] = useState(0)
 
+  /**
+   * All-pages selection state tracking
+   */
+  const [isAllCurrentPageSelected, setIsAllCurrentPageSelected] =
+    useState(false)
+  const [isAllItemsSelected, setIsAllItemsSelected] = useState(false)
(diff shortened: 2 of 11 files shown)


sergiocarracedo:
Why we need to create ux inconsistencies across app?

siguenzaraul:
> > This is likely a UX inconsistency, but currently all cases are handled solely from a page-level perspective. As long as this remains the case, we should have the option to use a prop to gradually modify the behavior.
> 
> 
> 
> In my opinion is a big inconsistency for the users, on a page the behavior is a real select all and in other a page select all, without a different UI (a different way of doing the select, a different button, etc) the users can no notice the difference causing errors and frustration
> 
> 
> 
> Or at least that's what they told me from product, and from my point of view, makes sense
> 
> 
> 
> Well if product is agree, I don't have anything to say

Let me check with all the products, as it might only be necessary for the product that caused the complaint, and it's not a common issue. In addition, we should improve the design of this so the user clearly understands what's happening

PR #3181, opened 2025-12-30
fix: prevent restoring selections after filter change

siguenzaraul:
## Description

We found a bug where when we selected something while using filters, selecting everything and then removing the filter wasn't clearing the selection.


sauldom102 (approved):
LGTM!

PR #3255, opened 2026-01-16
fix(Popover): add container prop to fix rendering inside dialogs

siguenzaraul:
## Description

Fix portal container issue in FiltersControls popover when used inside dialogs. The popover component now supports a `container` prop that allows rendering the portal content into a specific container element, preventing clipping and z-index issues when the component is used within F0Dialog components.

When FiltersControls with its popover was used inside F0Dialog components, the popover would render in the default body portal. This caused z-index stacking issues and potential clipping problems, especially in center and fullscreen dialogs that have focus traps. By detecting the dialog context and using its portal container, the popover now renders correctly within the dialog's scope, maintaining proper layering and preventing visual issues.

Before:
https://github.com/user-attachments/assets/9984224f-88d4-49e6-8457-97ef28acf60b

After:
https://github.com/user-attachments/assets/08fe3109-1e4d-4f5f-a46e-12e1d812e106

### DIFF
--- a/packages/react/src/components/OneFilterPicker/components/FiltersControls.tsx
+++ b/packages/react/src/components/OneFilterPicker/components/FiltersControls.tsx
@@ -1,11 +1,12 @@
 import { F0Button } from "@/components/F0Button"
 import { ButtonInternal } from "@/components/F0Button/internal"
+import { F0DialogContext } from "@/components/F0Dialog"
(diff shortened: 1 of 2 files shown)


sergiocarracedo:
@siguenzaraul could you merge it? or are you waiting for something?

PR #3259, opened 2026-01-19
feat(AiChat): Add file attachment UI with conditional visibility

jcornudella:
## Summary
Add file attachment support to AiChat with the ability to conditionally show/hide the upload UI based on `fileValidation` prop.

## Changes
- Add `fileValidation` and `onFilesRejected` props to `AiChatProviderProps`
- Pass `fileUploadsEnabled` boolean through context (derived from `fileValidation !== undefined`)
- Conditionally render paperclip button in `ChatTextarea` based on `fileUploadsEnabled`
- Export new types from package

## Usage
```tsx
<ApplicationFrame
  ai={{
    // Only show file upload UI when fileValidation is provided
    fileValidation: shouldEnableUploads
      ? { acceptedTypes: ['application/pdf'], maxFileSize: 50 * 1024 * 1024 }
      : undefined,
  }}
/>
```

## Test plan
- [ ] When `fileValidation` is undefined, paperclip button should not appear
- [ ] When `fileValidation` is provided, paperclip button should appear
- [ ] File validation should reject files that don't match accepted types

### DIFF
--- a/packages/react/src/components/F0Icon/F0Icon.tsx
+++ b/packages/react/src/components/F0Icon/F0Icon.tsx
@@ -84,7 +84,12 @@ export const F0Icon = forwardRef<SVGSVGElement, F0IconProps>(function F0Icon(
         ref={ref}
         {...props}
         animate={state}
-        className={cn(iconVariants({ size }), "select-none", colorClass)}
+        className={cn(
+          iconVariants({ size }),
+          "select-none",
+          colorClass,
+          props.className
+        )}
         style={colorStyle}
         data-has-color={color !== "currentColor" ? "true" : undefined}
       />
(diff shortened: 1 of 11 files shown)


packages/react/src/components/F0Icon/F0Icon.tsx
```diff
@@ -84,7 +84,12 @@ export const F0Icon = forwardRef<SVGSVGElement, F0IconProps>(function F0Icon(
         ref={ref}
         {...props}
         animate={state}
-        className={cn(iconVariants({ size }), "select-none", colorClass)}
+        className={cn(
+          iconVariants({ size }),
+          "select-none",
+          colorClass,
+          props.className
```
sergiocarracedo: We we not providing className prop for the exported component intentionally. Please dont add it. Why do you need it?

jcornudella: @sergiocarracedo disculpa, esta tiene que estar cerrada!

PR #3328, opened 2026-01-30
fix(useSelectable): maintain selection

siguenzaraul:
## Description
We can mantain selection when we disable the selectAll, as we dont use the filters for the backend


eliseo-juan (approved):
LGTM!

PR #3409, opened 2026-02-11
fix: keep preset counters consistent in overflow filter picker

xavikp:
Use preset itemsCount for visible and collapsed presets so badge values stay stable when presets overflow.

Add Storybook width controls and a regression spec to verify counter consistency in expanded and collapsed states.

## Description

Preset counters showed inconsistent values when presets collapsed into the overflow menu, which made the UI unreliable and confusing during filtering. The goal was to keep counter values stable regardless of whether a preset is visible inline or inside the collapsed dropdown.
  
Current status
image

Proposed fix
Captura de pantalla 2026-02-11 a les 15 31 04

Testing changes with the fix
Captura de pantalla 2026-02-16 a les 10 30 43

## Implementation details

Updated FiltersPresets so both visible and overflow presets read from the same itemsCount source (including async values via Await) instead of deriving overflow counts from preset filter key length. Added a regression test for collapsed/expanded consistency and a Storybook story with a width control to manually validate overflow behavior.

### DIFF
--- a/packages/react/src/components/OneFilterPicker/__stories__/OneFilterPicker.stories.tsx
+++ b/packages/react/src/components/OneFilterPicker/__stories__/OneFilterPicker.stories.tsx
@@ -154,20 +154,25 @@ const PresetsOverflowCounterConsistencyComponent = ({
   )
 }
 
-export const WithPresetsOverflowCounterConsistency: StoryObj = {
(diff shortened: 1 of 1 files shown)


sauldom102 (approved):
Nice one! ✅

PR #3419, opened 2026-02-12
feat: Add card interactiveChildren prop to make children components clickable

aaroncaballero-f:
## Description

Make children components clickable by adding new interactiveChildren prop.
Limit the clickable components to clickable default ones.
Add stopPropagation to not enable card onClick when clicking any component or actions inside it.

image

### DIFF
--- a/.release-please-manifest.json
+++ b/.release-please-manifest.json
@@ -1,5 +1,5 @@
 {
-  "packages/react": "1.358.0",
+  "packages/react": "1.362.3",
   "packages/react-native": "0.24.1",
   "packages/core": "1.44.0"
 }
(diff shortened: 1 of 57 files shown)


packages/react/src/components/F0Card/__stories__/Card.stories.tsx
```diff
@@ -43,6 +46,32 @@ const SlotComponent = () => {
...
+          onClick={(e) => e.stopPropagation()}
+        >
+          Click me (goes to Google)
+        </F0Link>
+        <Switch
+          title="Toggle"
+          checked={checked}
+          onCheckedChange={(value) => setChecked(value)}
+          onClick={(e) => e.stopPropagation()}
+        />
```
sauldom102: Would this work without putting e.stopPropagation() here?

PR #3421, opened 2026-02-12
fix: add cursor-pointer to dropdown

aaroncaballero-f:
## Description

Add cursor pointer to dropdown

image


sauldom102 (commented):
We shouldn't do this, cursor-pointer should be reserved for links, we shouldn't always use it. Link component already has it so no need to change anything regarding that

sauldom102 (approved):

PR #3436, opened 2026-02-13
feat: adding custom tooltip to F0OneSwitch

ChloeDumit:
## Description

This PR adds support for custom tooltips in One Switch, allowing us to display different messaging depending on the page context. For example, we can now show a specific tooltip when users enter the marketplace. It also introduces an option to auto open the tooltip for some seconds without hovering it 

Screenshot 2026-02-13 at 10 56 13 AM

https://github.com/user-attachments/assets/192bd933-0b0c-43f5-af38-a836508a2762

[Link to Figma Design](https://www.figma.com/design/n3iMu3amKFTrt5ve4l5EWn/ONE---Sales-Agent?node-id=2074-51649&t=7OHdwVL2pmvR7MvI-0)

## Implementation details

- Added support for passing a custom tooltip to One Switch
- Introduced an option to auto open the tooltip without hovering
- Preserves the default behavior when no custom configuration is provided


packages/react/src/sds/ai/F0OneSwitch/types.ts
```diff
@@ -5,4 +5,9 @@ import * as SwitchPrimitive from "@radix-ui/react-switch"
  */
 export type F0OneSwitchProps = React.ComponentPropsWithoutRef<
   typeof SwitchPrimitive.Root
->
+> & {
+  /** Custom text shown in the tooltip when the chat is closed */
+  customTooltip?: string
+  /** When true, the tooltip text is shown as a persistent label next to the switch (always in layout, no hover) */
+  tooltipAlwaysVisible?: boolean
```
siguenzaraul: I'm not sure if this is a good idea, what's the use case?

ChloeDumit: behavior was changed for auto open lasting 3 secs

PR #3461, opened 2026-02-18
Increase NotesTextEditor image default limit to 50MB

jcornudella:
## Summary
- increase NotesTextEditor image upload default max size from 15MB to 50MB
- keep behavior unchanged when consumers pass an explicit maxFileSize value

## Validation
- pnpm -C packages/react lint src/experimental/RichText/CoreEditor/Extensions/Image/index.tsx
- pnpm -C packages/react tsc


sauldom102 (approved):
I think we shouldn't set up this internally here in F0, accepting it though to not block you

PR #3475, opened 2026-02-18
refactor(one-upselling-kit): update components to match design tokens

daviz-fct:
## Description

I’ve updated the style of some of the components we’ve created for the Sales skill focused on upselling to make them more similar to the Cards or other patterns that already exist in the design system.

## Screenshots 
f0-pr-1


eliseo-juan:

PR #3484, opened 2026-02-19
feat: add support to `PushPinSolid` icon on widget component

lucianosousa:
## Description

Add support to `pushPinnedSolid` icon in the `AvatarModule`, which is being used on `WidgetInboxList`

We need that to make a UI adjustment of adding this icon to the pinned posts from a community

Current PR running on factorial

Screenshot 2026-02-19 at 15 56 26


packages/react/src/icons/modules/index.ts
```diff
@@ -24,6 +24,7 @@ export { default as Payroll } from "./Payroll"
 export { default as Performance } from "./Performance"
 export { default as Profile } from "./Profile"
 export { default as Projects } from "./Projects"
+export { default as PushPinSolid } from "../app/PushPinSolid"
```
lucianosousa: I am not creating a new icon here, just reusing the same from `app` path.
Don't see a need for that. Let me know if it does makes sense

PR #3491, opened 2026-02-20
fix: some ui changes in ai chat in fullscreen mode

sauldom102:
## Description

Three things done here in AI chat in full-screen mode:
- Increased text size for title
- Removed left padding so that it's vertically aligned with text area input
- Removed greeting message

## Screenshots

Screenshot 2026-02-20 at 15 18 44

### DIFF
--- a/packages/react/src/sds/ai/F0AiChat/components/WelcomeScreen.tsx
+++ b/packages/react/src/sds/ai/F0AiChat/components/WelcomeScreen.tsx
@@ -3,7 +3,9 @@ import { Message, randomId } from "@copilotkit/shared"
 import { AnimatePresence, motion } from "motion/react"
 import { useMemo } from "react"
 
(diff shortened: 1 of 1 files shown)


packages/react/src/sds/ai/F0AiChat/components/WelcomeScreen.tsx
```diff
@@ -87,35 +96,37 @@ export const WelcomeScreen = ({
...
-                    content: suggestion.prompt || suggestion.message,
-                  })
-                }
-              />
-            </motion.div>
-          ))}
-        </div>
+        {!!pickedSuggestions.length && (
+          <div className="flex flex-col items-start gap-[6px]">
+            {pickedSuggestions.map((suggestion, index) => (
```
sauldom102: This was necessary to have an empty div here causing an extra unwanted gap between the text area and the title

PR #3495, opened 2026-02-23
fix: improve feedback when submitting forms

sauldom102:
## Description

Previously, we weren't showing any feedback to the user apart from the loading indicator in the submit button. We were handling a toast showing up in the frontend/monorepo code. Now, by default, we'll display an action bar with the message set by the dev returned on the onSubmit callback.

https://github.com/user-attachments/assets/a33d46fa-db92-4e01-9ea7-aca346ff76c1

Screenshot 2026-02-23 at 09 58 56

[Figma link](https://www.figma.com/design/PbfFcQ5TK7zvPFWOoObQF0/Surveys---Q4?node-id=21050-17883&t=v17KlQalr23150oh-4)

### DIFF
--- a/packages/react/src/components/F0Form/components/ActionBar.tsx
+++ b/packages/react/src/components/F0Form/components/ActionBar.tsx
@@ -1,0 +1,122 @@
+import { F0Button } from "@/components/F0Button"
+import { F0Icon, IconType } from "@/components/F0Icon"
+import { F0ActionBar, ActionBarStatus } from "@/experimental/F0ActionBar"
+import { AlertCircle, ChevronDown, ChevronUp } from "@/icons/app"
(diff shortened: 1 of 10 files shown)


packages/react/src/components/F0Form/components/ActionBar.tsx
```diff
@@ -0,0 +1,122 @@
...
+  goToPreviousError: () => void
+  goToNextError: () => void
+}
+
+export function FormActionBar({
+  isActionBar,
+  isDirty,
+  actionBarStatus,
+  hasErrors,
+  errorCount,
```
sauldom102: Just moving all of this here into a separate component for more code clarity

PR #3550, opened 2026-02-27
fix(data-collection): disable row hover when no interactive elements

sergiocarracedo:
## Description

Rows in the Table visualization were always showing a hover highlight, even when the row had no interactive elements (no `itemUrl`, no `itemOnClick`, and no item actions). This made non-interactive rows visually misleading.

This PR introduces `localDisabledHover` — a derived boolean that disables hover when none of the interactive triggers are present — and `showItemActions` to consolidate the item actions visibility check.

N/A — behavioral change only.

## Implementation details

- Added `useMemo`-backed `showItemActions` (replaces inline `hasItemActions && !loading && !nestedRowProps?.onLoadMoreChildren` guard)
- Added `useMemo`-backed `localDisabledHover`: is `true` when `disableHover` prop is set, **or** when the row has no `itemUrl`, `itemOnClick`, and no item actions
- Updated `className` to use `localDisabledHover` instead of `disableHover`
- Updated the item actions render guard to use `showItemActions`
- Added 9 unit tests in `Row.test.tsx` covering all hover behaviour cases


eliseo-juan:

PR #3551, opened 2026-02-27
feat: add new improved text component

aaroncaballero-f:
## Description

Add new F0TextV2 component as a clean replacement for F0Text. This is the first step of a migration plan: create new → migrate usages → delete old → rename back.

F0TextV2 introduces a simplified two-prop architecture (variant + size) that is fully self-contained — no dependency on ui/Text/. It supports 8 semantic variants (title, heading, subtitle, body, label, description, caption, code), 7 size overrides, and 12 foreground color tokens. The old F0Text is marked @deprecated.

image
image

## Implementation details

- Created F0TextV2/ with a clean 3-file structure: index.tsx (component + Component() wrapper), types.ts (token types), utils.ts (class maps)
- Each variant defines its own default font-size, font-weight, HTML tag, and color (secondary for description, subtitle, caption, label)
- Uses CVA for variant/size class resolution, cn() for merging
- Follows F0Box architecture pattern with Component({ name, type: "info" })
- Props: variant, size, color, decoration, transform, align, ellipsis, noEllipsisTooltip
- No as prop, no markdown prop — intentionally simpler API surface
- 72 unit tests covering all variants, sizes, colors, decorations, transforms, alignment, ellipsis, ref forwarding, and edge cases
- Full Storybook stories included
- F0Text and ui/Text/ remain 100% unmodified (only @deprecated JSDoc added to F0Text exports)
- Exported via components/exports.ts


eliseo-juan:

PR #3574, opened 2026-03-03
feat: add delta display cell

FaithAdekunle:
## Description
Adds a new delta value-display renderer to the React value-display system, enabling rendering of positive/negative deltas with directional icons for use in data-collection/table contexts.

Screenshot 2026-03-03 at 12 51 32

[Link to Figma Design](https://www.figma.com/file/iGqeNj7ahbQ1qlGEiz3gHM?type=design&node-id=8077%3A10271&mode=dev)

## Implementation details
- Introduces DeltaCell renderer and registers it under the delta renderer type.
- Adds Storybook stories and MDX docs for the new delta value type.
- Extends shared Storybook mock data with positive/negative delta examples.

### DIFF
--- a/packages/react/src/ui/value-display/__stories__/shared.tsx
+++ b/packages/react/src/ui/value-display/__stories__/shared.tsx
@@ -29,6 +29,14 @@ export const mockItem = {
       src: "/avatars/person02.jpg",
     },
   ],
+  positiveDelta: {
+    label: "10%",
+    deltaStatus: "positive" as const,
+  },
+  negativeDelta: {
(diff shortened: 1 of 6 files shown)


sauldom102 (changes requested):
Could we better use TagBalance and rename delta to balance? Just to reuse what we already have

https://ds.factorial.dev/?path=/docs/components-tags-tagbalance--documentation

As an example, you can see how we reused it in DetailsItemsList https://ds.factorial.dev/?path=/docs/components-list-detailsitemslist--documentation

PR #3576, opened 2026-03-03
feat: add onAddRow button to EditableTable and nested rows

laurafg:
## 🚪 Why?

The `EditableTable` visualization in f0 needs an "Add row" affordance so consumers can let users insert new rows directly from the table UI — both at the root level and within nested/expandable rows. Additionally, the generic `TableCollection` API should not be coupled to editable-table-specific props.

## 🔑 What?

- Add an optional `onAddRow` callback to `EditableTable` that renders a ghost "Add row" button in the table footer and at the bottom of each expanded nested row. The callback receives the parent item when triggered from a nested row (`undefined` at root level) and supports async functions for automatic loading state.
- Add `addRowButtonLabel` and `nestedAddRowButtonLabel` props to customize the button text at root and nested levels, falling back to an i18n default.
- Introduce `AddRowContext` provided by `EditableTableCollection`, replacing prop drilling of add-row concerns through `TableCollection -> Row -> NestedRow` and keeping the generic table API clean.
- Add unit tests for `Table` (add-row via context) and `NestedRow` (nested add-row via context).
- Add Storybook stories: flat editable table with functional row insertion, nested table with action logging, and editable table combining summary row with add-row.

https://github.com/user-attachments/assets/dc577ca2-5b4f-4b52-8b42-7dbe2f401214

### DIFF
--- a/packages/react/src/experimental/OneDataCollection/visualizations/collection/EditableTable/EditableTable.tsx
+++ b/packages/react/src/experimental/OneDataCollection/visualizations/collection/EditableTable/EditableTable.tsx
@@ -16,6 +16,7 @@ import { SummariesDefinition } from "../../../summary"
 import { CollectionProps } from "../../../types"
 import { TableCollection } from "../Table/Table"
 import { EditableCellRenderer } from "./components/EditableCellRenderer"
+import { AddRowProvider } from "./context/AddRowContext"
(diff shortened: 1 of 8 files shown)


sauldom102 (commented):
Some visual feedback, maybe we could better align cells to be centered

Image

sauldom102 (approved):

PR #3591, opened 2026-03-04
feat(F0Select): add group-header item 

miriam-mr90:
## Description

Add `group-header` as a new item type for the `F0Select` component when using static `options`. This allows consumers to visually separate options into labeled sections (e.g. "Assigned to me" / "All vacancies") without resorting to disabled items or manual separators.

A `SelectSeparator` is automatically rendered before each group header (except the first), so consumers only need to insert `{ type: "group-header", label: "..." }` entries in their options array.

Fully backward compatible — existing `item` and `separator` types continue to work unchanged.

## Screenshots

image

The dropdown shows two group headers ("Assigned to me" and "All vacancies") separating selectable options, with an automatic separator line between groups.

## Implementation details

- **`types.ts`**: Extended `F0SelectItemProps` union with a new `{ type: "group-header"; label: string }` variant alongside the existing `item` and `separator` types.
- **`F0Select.tsx`** (5 touch points):
  - `defaultSearchFn` — skips `group-header` items from search filtering (same as `separator`)
  - `selectable` callback — excludes `group-header` from selectable items
  - `itemsByValue` mapping — excludes `group-header` from the value map
  - `onChange` handler — excludes `group-header` from value extraction
  - `getItems` callback — renders a styled `div` matching the Figma "Dropdown header" spec (secondary text, `font-medium`, `text-sm`, padding 12/8/12px) with auto `SelectSeparator` before non-first group headers
- **`F0Select.stories.tsx`**: Added a `WithGroupHeaders` story demonstrating group headers with static options. Updated the `WithSearchBox` story's custom `searchFn` to handle the new type.

### Tests
- Existing F0Select tests pass (12 passed, 1 skipped — pre-existing)
- TypeScript compilation verified — no new type errors introduced
- Lint passes with 0 warnings

### Manual Verification
- Verified in Storybook via the new `WithGroupHeaders` story
- Group headers render with correct Figma styling (secondary color, font-weight 500, text-sm)
- Headers are non-selectable and excluded from search/selection logic
- Auto separator renders between groups but not before the first group

### DIFF
--- a/packages/react/src/components/F0Select/F0Select.tsx
+++ b/packages/react/src/components/F0Select/F0Select.tsx
@@ -626,8 +626,10 @@ const F0SelectComponent = forwardRef(function Select<
             item: <SelectSeparator key={`separator-${index}`} />,
           })
         } else if (mappedOption.type === "group-header") {
-          // Auto-insert separator before non-first group headers
-          if (hasSeenGroupHeader) {
+          // Auto-insert separator before non-first group headers,
+          // but skip if the previous item is already a separator
+          const lastItem = result[result.length - 1]
+          if (hasSeenGroupHeader && lastItem?.height !== 1) {
             result.push({
               height: 1,
               item: <SelectSeparator key={`group-header-separator-${index}`} />,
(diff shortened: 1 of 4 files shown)


eliseo-juan:

PR #3607, opened 2026-03-09
feat: add SelectCell for EditableTable with F0Select integration

laurafg:
## 🚪 Why?

The EditableTable component only supported text inputs for editable cells. We need a select cell type so columns can offer dropdown selection — either from static options or from an async data source — enabling richer inline editing workflows.

## 🔑 What?

- **New `SelectCell` component** wrapping `F0Select`, supporting both static `options` and async `source` + `mapOptions` data modes. Styled to match `TextCell` (transparent, borderless, full-height within the cell).
- **New `SelectCellConfig` type** added to `EditableTable/types.ts` as a discriminated union mirroring F0Select's two data modes: static options (array or per-row function) and async `DataSourceDefinition`.
- **Extended `EditableTableColumnDefinition`** with an optional `selectConfig` field, required when `editType` returns `"select"`.
- **Registered `SelectCell`** in the `editableCellMap` (in `consts.ts`) so the `"select"` edit type renders the new component instead of falling back to `TextCell`.
- **Fixed event propagation** in `EditableCellRenderer` — changed `onClickCapture`/`onMouseDownCapture` to `onClick`/`onMouseDown` so interactive components like `F0Select` can handle their own click events (the dropdown wouldn't open otherwise).
- **Duplicate onChange guard** — prevents `onCellChange` from firing twice per selection by comparing the incoming value against the current one before propagating.
- **Updated mock data** — the "Department" column now uses a static-options select in the default editable table stories.
- **Added `EditableTableWithDataSourceSelect` story** demonstrating the `source` + `mapOptions` pattern with a non-paginated data source on the "Role" column.

## 📸 Demo

https://github.com/user-attachments/assets/ca24efc8-566c-4864-b40c-0b880000396f

Screenshot 2026-03-13 at 10 19 42

### DIFF
--- a/packages/react/src/experimental/OneDataCollection/__stories__/mockData.tsx
+++ b/packages/react/src/experimental/OneDataCollection/__stories__/mockData.tsx
@@ -460,7 +460,14 @@ export const getMockVisualizations = (options?: {
           {
             id: "department",
             label: "Department",
-            editType: () => "text" as const,
+            editType: () => "select" as const,
+            selectConfig: {
+              options: DEPARTMENTS_MOCK.map((dept) => ({
+                value: dept,
+                label: dept,
+              })),
+              placeholder: "Select department",
+            },
             render: (item) => item.department,
             sorting: options?.table?.noSorting ? undefined : "department",
             order: options?.table?.allowColumnReordering ? 4 : undefined,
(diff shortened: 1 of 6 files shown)


laurafg:
> > When searching in the select, something weird happened in the name, maybe the arrow should be at the right, and then the loader wouldn't be in the text
> 
> you mean this loader, right? in our words like it should be outside the "input", no?
> 
> Screenshot 2026-03-13 at 09 41 28

but @eliseo-juan thats the default behaviour for selects, check it here https://ds.factorial.dev/?path=/docs/components-select--documentation#with-data-source-paginated

PR #3628, opened 2026-03-10
fix(OneDataCollection): align group header checkboxes with header and row checkboxes

juancruz-f7l:
## What

Fixes three visual alignment issues in `OneDataCollection`'s Table visualization when grouping is enabled:

1. **Group header checkbox alignment** — group header checkboxes were offset 6px to the right of header/row checkboxes (`px-3` padding vs `ml-1.5` margin).
2. **Group header label alignment** — group label text was misaligned with the first data column. Fixed by splitting the group header row into a dedicated checkbox cell (46px, matching `checkColumnWidth`) and a separate label cell, so each element lands at exactly the same horizontal position as its counterpart in the header and data rows.
3. **Data row alignment in selectable tables** — `TableCell` primitive used `p-2` (8px horizontal padding) while `TableHead` used `px-3` (12px), causing a 4px misalignment on all data cells in selectable tables. Fixed by changing `p-2` → `px-3 py-2` in `ui/table.tsx`.

## Changes

- `Table.tsx` — split group header row into separate checkbox cell + label cell; use `i18n.actions.selectAll` for checkbox title; fix filler cell `colSpan` (was double-counting the checkbox column after the split)
- `ui/table.tsx` — align `TableCell` horizontal padding with `TableHead` (`p-2` → `px-3 py-2`)

## Behaviour note

When `collapsible=false` and `selectable=true`, clicking the group label text previously toggled group selection (via `GroupHeader`'s `handleGroupClick`). With the structural split, the label cell renders `GroupHeader` with `selectable={false}`, so selection is only triggered by the checkbox. This is intentional — the checkbox is the unambiguous selection affordance. When `collapsible=true` (the common case), behaviour is unchanged since `showOpenChange` already takes priority over `selectable` in `handleGroupClick`.

### DIFF
--- a/packages/react/src/patterns/OneDataCollection/visualizations/collection/Table/Table.tsx
+++ b/packages/react/src/patterns/OneDataCollection/visualizations/collection/Table/Table.tsx
@@ -653,11 +653,13 @@ export const TableCollection = <
                           onOpenChange={(open) => setGroupOpen(group.key, open)}
                         />
                       </TableCell>
-                      <TableCell
-                        colSpan={columns.length - (frozenColumnsLeft || 1)}
-                      >
-                        &nbsp;
-                      </TableCell>
+                      {columns.length - (frozenColumnsLeft || 1) > 0 && (
+                        <TableCell
+                          colSpan={columns.length - (frozenColumnsLeft || 1)}
+                        >
+                          &nbsp;
+                        </TableCell>
+                      )}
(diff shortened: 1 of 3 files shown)


sauldom102 (dismissed):
@juancruz-f7l let's also align this too

Image

eliseo-juan (approved):
LGTM!

PR #3667, opened 2026-03-13
Fix: nested rows selectable

laurafg:
## 🚪 Why?

Selecting children rows in a DataCollection with nested records was broken in multiple ways: clicking a child checkbox selected the parent instead, children were excluded from bulk actions, "select all" only selected top-level parents (ignoring nested children), and Storybook stories had duplicate IDs causing all items to be checked at once.

## 🔑 What?

**Fix children row selection:**

- Replace `onCheckedChange` with `onSelectItem` prop in `Row`/`NestedRow` so each row calls `handleSelectItemChange` with its own item instead of inheriting the parent's closure
- Update `useSelectable` to accept and store the item reference directly in `handleSelectItemChangeInternal`, preventing failed lookups for children not present in `data.records`
- Relax the `currentPageItemIds` filter in `itemsStatus`/`selectedIds` to include items with a resolved reference (children are visible on the current page but not in top-level records)

**Cascade "select all" to nested children:**

- Add a `useEffect` in `NestedRow` that detects when a parent transitions from unselected → selected (e.g., via "select all") and cascades selection to all loaded children
- Track previous parent selection state with a ref to avoid infinite loops and double-triggers with the existing `handleSelectItemWithChildren` click handler

**Parent indeterminate/checked state from children:**

- Compute `parentIndeterminate` and `allChildrenSelected` in `NestedRow` based on loaded children's selection state
- When all children are individually selected, auto-select the parent via a `useEffect`
- Pass indeterminate/allChildrenSelected props to `Row` for correct checkbox rendering

**Hide checkbox on "load more" rows:**

- Override `selectable` to `() => undefined` on `LoadMore` rows so the column remains rendered (preserving table alignment) but no checkbox is displayed

**Fix Storybook mock data:**

- Generate unique IDs for child items in `mockData.tsx` (`${user.id}-child-0`, etc.)
- Fix `selectable` prop in nested records stories to use `item.id` instead of a hardcoded empty string

### Edge cases handled

| Scenario | Behavior |
|---|---|
| **Select all** (page or all-pages) | Parents are selected by `useSelectable`, then the `useEffect` in `NestedRow` cascades to all loaded children |
| **Deselect all** | `handleSelectAll(false)` resets the entire items Map to empty, clearing both parents and children in one operation |
| **Click parent checkbox directly** | `handleSelectItemWithChildren` cascades immediately on click; the `useEffect` does not double-trigger because the ref already tracks the parent as selected |
| **Expand a row after select-all** | When children load, the `children` dependency changes, the `useEffect` re-runs, sees the parent is selected, and cascades to newly loaded children |
| **Manually deselect a child** | Parent becomes indeterminate. The cascade `useEffect` does not re-select the child because the ref already has `wasSelected = true` (no transition detected) |
| **Deselect all children individually** | Parent auto-deselects via the existing `allChildrenSelected` effect (which detects 0 selected children and does nothing), and the indeterminate logic resets |

### Known limitation: selection scope is tied to expanded rows

"Select all" can only cascade to children that have been loaded — i.e., the parent row was previously expanded. Children of collapsed/never-expanded rows are not selected because they don't exist in memory yet (they are lazily loaded via `NestedDataProvider`).

This means **both selection and count only reflect what has been expanded**:

- If you click "select all" with 5 parents and 2 of them have been expanded (showing 3 children each), the selection will include 5 parents + 6 children = 11 items. The 3 unexpanded parents' children are not selected.
- If you then expand one of those 3 remaining parents, the cascade `useEffect` fires (parent is already selected → new children loaded → auto-selected), and the count increases.

**Counter accuracy per mode:**

- **Page-only mode**: the counter uses `checkedCount` (actual checked items in state). It accurately reflects what is selected — parents + expanded children. As you expand more rows, children get cascaded and the count goes up incrementally.
- **All-pages mode**: the counter uses `selectAllTotal - uncheckedCount`, where `selectAllTotal` comes from `paginationInfo.total` (top-level items only). This formula does not account for children at all, so even after children are loaded and selected, the displayed count can be lower than the actual number of selected items. This is a pre-existing architectural limitation of how `useSelectable` tracks counts separately from `NestedDataProvider`.

### DIFF
--- a/packages/react/src/experimental/OneDataCollection/__stories__/mockData.tsx
+++ b/packages/react/src/experimental/OneDataCollection/__stories__/mockData.tsx
@@ -1044,25 +1044,44 @@ const createdMixedNestedRecords = (filteredData: MockUser[]) => {
         ? [
             {
               ...user,
+              id: `${user.id}-child-0`,
               children: [
-                { ...user },
+                { ...user, id: `${user.id}-child-0-0` },
                 {
                   ...user,
+                  id: `${user.id}-child-0-1`,
                   detailed: index === 0,
                   children: [
-                    { ...user, detailed: index === 0 },
-                    { ...user, detailed: index === 0 },
+                    {
+                      ...user,
+                      id: `${user.id}-child-0-1-0`,
+                      detailed: index === 0,
+                    },
+                    {
+                      ...user,
+                      id: `${user.id}-child-0-1-1`,
+                      detailed: index === 0,
+                    },
                   ],
                 },
-                { ...user },
+                { ...user, id: `${user.id}-child-0-2` },
               ],
             },
             {
               ...user,
+              id: `${user.id}-child-1`,
               detailed: index === 0,
               children: [
-                { ...user, detailed: index === 0 },
-                { ...user, detailed: index === 0 },
+                {
+                  ...user,
+                  id: `${user.id}-child-1-0`,
+                  detailed: index === 0,
+                },
+                {
+                  ...user,
+                  id: `${user.id}-child-1-1`,
+                  detailed: index === 0,
+                },
               ],
             },
           ]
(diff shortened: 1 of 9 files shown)


eliseo-juan:

PR #3695, opened 2026-03-18
fix(CommunityPost): remove grey hover background from post card

ekaterina-spiridonova:
## 🚪 Why?

The community post cards now have card-level hover shadows applied by the consuming app (factorial). The inner `hover:bg-f1-background-hover` class on the CommunityPost div causes a grey background on hover, which conflicts with the intended white card hover state.

## 🔑 What?

- Removed `hover:bg-f1-background-hover` from the outer `div` className in `BaseCommunityPost`
- The hover visual feedback is now delegated to the card wrapper level (shadow transition), keeping the card background white on hover

### Tests
- No logic change, only a Tailwind class removal
- Existing tests unaffected

### Manual Verification
- Community post cards stay white on hover instead of turning grey

### DIFF
--- a/packages/react/src/sds/Home/Communities/Post/CommunityPost/index.tsx
+++ b/packages/react/src/sds/Home/Communities/Post/CommunityPost/index.tsx
@@ -108,7 +108,7 @@ export const BaseCommunityPost = ({
 
   return (
     <div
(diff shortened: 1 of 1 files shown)


eliseo-juan:

PR #3708, opened 2026-03-19
feat: Storybook component diff notifications via GitHub Issues

AnaTBTB:
## Summary

- Adds a script (`scripts/storybook-diff.mjs`) that fetches the live Storybook index and diffs it against a committed snapshot to detect when components are added or removed
- Adds a baseline snapshot (`scripts/storybook-snapshot.json`) of 314 components
- Extends the existing CI workflow (`.github/workflows/build-and-deploy-storybook.yaml`) with a new `notify-storybook-changes` job that runs after each successful Storybook deploy, opens a GitHub Issue if components changed, and auto-commits the updated snapshot with `[skip ci]`

## Why

The design system team needs to stay informed of structural changes to the F0 component library without manually checking Storybook. This automates that signal with low noise — only component-level additions/removals trigger a notification, not story-level changes within existing components.

## How it works

1. After every deploy to `https://ds.factorial.dev`, the `notify-storybook-changes` job runs
2. It fetches `https://ds.factorial.dev/index.json` and extracts the list of unique component names
3. It compares against the committed `storybook-snapshot.json`
4. If components were added or removed, it opens a GitHub Issue listing the diff
5. It updates the snapshot and commits it back to `main` with `[skip ci]` to avoid a loop

## Notes

- Uses the built-in `GITHUB_TOKEN` — no additional secrets required
- No Slack or external webhooks needed

### DIFF
--- a/.github/workflows/build-and-deploy-storybook.yaml
+++ b/.github/workflows/build-and-deploy-storybook.yaml
@@ -8,7 +8,7 @@ on:
   workflow_dispatch:
 
 permissions:
-  contents: read
+  contents: write
   pages: write
   id-token: write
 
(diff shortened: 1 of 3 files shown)


eliseo-juan:

PR #3920, opened 2026-04-13
docs(AGENTS): add mandatory pre-commit format step

desiree-np:
## Summary

- Add a **Pre-Commit Checklist** section to `packages/react/AGENTS.md` that makes `pnpm format` mandatory before every commit
- Add `pnpm format` and `pnpm format:check` to the Available Commands reference

## Why

Every PR has been failing the `Format` CI check because `oxfmt` was not being run before committing. This documents the requirement explicitly so any AI agent (or human) working in this package knows to run it.


sauldom102 (approved):
Nice Desii!

PR #3921, opened 2026-04-13
fix(EntitySelect): render popover inside dialogs and fix scroll/height issues

ReddyWish:
Portal EntitySelect dropdown into the dialog container when rendered inside F0Dialog (center/fullscreen), matching the pattern used by F0Select. Switch from absolute positioning to flex layout in the Content component, conditionally render SecondaryContent only when expanded, add overscroll-contain to prevent scroll chaining, and change overflow-auto to overflow-hidden to eliminate double scrollbars.

### DIFF
--- a/packages/react/src/deprecated/EntitySelect/Content/index.tsx
+++ b/packages/react/src/deprecated/EntitySelect/Content/index.tsx
@@ -70,7 +70,7 @@ export const Content = ({
 
   return (
     <div
-      className="relative overflow-hidden"
+      className="flex overflow-hidden"
       style={{
         height:
           singleSelector && (!actions || actions.length === 0)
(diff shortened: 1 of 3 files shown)


eliseo-juan (commented):
Doesn't make more sense to use the F0Select for you use case instead of keep updating a deprecated component?

ReddyWish:
@eliseo-juan That's what I was considering todo in the beginning, but EntitySelect has slightly more extendend functionality with sub-items, that's why I decided to stick to the small fix to deprecated component.

I think we can improve F0Select to cover this gap, but to me it looks like as a separate chore :upside_down_face:

eliseo-juan (approved):

PR #3946, opened 2026-04-15
fix(F0AiChat): surface file upload errors to the user

YaraDebian:
## Summary
The `catch` block in `useFileAttachments` silently swallowed upload errors — files were marked as `"error"` internally but rendered identically to successful uploads, then quietly excluded from the message payload on send. Users had no indication anything went wrong.

## Changes
- **`useFileAttachments.ts`**: Capture error message from thrown `Error` in catch block; also set `errorMessage` on the non-throw failure path (when `uploaded[idx]` is missing); guard against empty `err.message` by falling back to i18n generic message
- **`AttachedFilesList.tsx`**: Render error files with critical styling (red border, `AlertCircle` icon, filename in red); tooltip shows error message only when present; dismiss button is a semantic `<button>` with `aria-label` and `focusRing()` for keyboard accessibility; relies on the container's `aria-live="polite"` instead of per-item `role="alert"`
- **`types.ts`**: Add optional `errorMessage` field to `AttachedFile`
- **`i18n-provider-defaults.ts`**: Add `fileUploadError` translation key as generic fallback

## How it works
1. When `onUploadFiles` throws (e.g. "Each file must be 10MB or smaller"), the error message is captured and stored on the file
2. When `onUploadFiles` resolves but a specific file has no matching upload result, a generic fallback message is used
3. Failed files render with red critical styling and can be dismissed by the user via a keyboard-accessible button
4. Error details appear in a tooltip when `errorMessage` is present

image

### DIFF
--- a/packages/react/src/sds/ai/F0AiChat/components/input/ChatTextarea/AttachedFilesList.tsx
+++ b/packages/react/src/sds/ai/F0AiChat/components/input/ChatTextarea/AttachedFilesList.tsx
@@ -3,6 +3,7 @@ import { F0Icon } from "@/components/F0Icon"
 import { AlertCircle, CrossedCircle } from "@/icons/app"
 import { Skeleton } from "@/ui/skeleton"
 import { Tooltip } from "@/experimental/Overlays/Tooltip"
+import { focusRing } from "@/lib/utils"
 
 import { type AttachedFile } from "./types"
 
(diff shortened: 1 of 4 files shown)


YaraDebian:
build

packages/react/src/sds/ai/F0AiChat/components/input/ChatTextarea/AttachedFilesList.tsx
```diff
@@ -46,3 +56,38 @@ export const AttachedFilesList = ({
     </div>
   )
 }
+
+function ErrorFilePill({
```
YaraDebian, line 60: Not sure if we need a global component to be created here 🤔

sauldom102 (approved):

PR #3947, opened 2026-04-15
fix(EditableTable): improve nested row add button alignment and number cell clamping

RafaMellado:
- Fix add row button padding in detailed nested rows to align with
  sibling cells at the same depth level
- Hide add row button while children are still loading on first expand
- Clamp NumberCell values to min/max and force re-sync when the clamped
  value matches the current state (resetKey)
- Add MoneyCell component and tests
- Fix tree connector height calculation and sticky parent row scroll
  handling
- Fix count and summary value-display renderers

### DIFF
--- a/packages/react/src/hooks/datasource/useSelectable/useSelectable.ts
+++ b/packages/react/src/hooks/datasource/useSelectable/useSelectable.ts
@@ -420,12 +420,18 @@ export function useSelectable<
     (
       itemId: SelectionId | readonly SelectionId[],
       checked: boolean,
-      onlyIfNotPreviousState: boolean = false
+      onlyIfNotPreviousState: boolean = false,
+      fallbackItem?: R | readonly R[]
     ) => {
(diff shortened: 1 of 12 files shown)


packages/react/src/hooks/datasource/useSelectable/useSelectable.ts
```diff
@@ -420,12 +420,18 @@ export function useSelectable<
     (
       itemId: SelectionId | readonly SelectionId[],
       checked: boolean,
-      onlyIfNotPreviousState: boolean = false
+      onlyIfNotPreviousState: boolean = false,
```
laurafg: which was the issue here?

RafaMellado: For Nested Tables using Apollo, when you fetch the children asynchronously, the selectable was putting as an item always the parent. We need to pass the item being displayed in those cases

sauldom102 (approved):

PR #3969, opened 2026-04-20
feat(F0Button): changes to F0Button and added new icons

developerdanx:
## Description

- feat(F0Button): changes to F0Button and added new icons

## PR Summary

1. Fix fullWidth layout bug
- Problem: fullWidth used flex-1 on the outer wrapper, which caused the button to grow vertically in column (flex-col) layouts instead of taking full horizontal width.
- Fix: Replaced flex-1 with w-full. Non-fullWidth buttons use items-start to self-size.
2. Add isDark prop (experimental, ghost-only)
- Problem: The ghost variant (transparent background + dark text) is invisible when placed on dark surfaces.
- Fix: Added isDark?: boolean prop that, only when variant="ghost", switches:
  - Text and icon colors to "inverse"
  - Pressed background to bg-f0-background-inverse-secondary
  - Loading spinner border to border-f0-foreground-inverse
- All other variants ignore isDark completely.
- Marked with @experimental JSDoc.
3. Add iconPosition prop
- Problem: Icons could only render on the left side of the label.
- Fix: Added iconPosition?: "left" | "right" (default "left"). When "right", the icon renders after the label/emoji.
4. Align padding with Figma/web specs
- Problem: Padding was a flat px-2 for all sizes, and old responsive breakpoint classes (sm:px-3 lg:px-4) were incorrectly applied (they were screen-width breakpoints, not button-size variants).
- Fix: Created getButtonPadding() with per-size defaults and asymmetric padding when an icon + visible label are both present:
  - sm: px-2 default; icon side gets pl-1.5/pr-1.5 (6px vs 8px)
  - md: px-3 default; icon side gets pl-2/pr-2 (8px vs 12px)
  - lg: px-4 default; icon side gets pl-3/pr-3 (12px vs 16px)
- Removed the old -ml-0.5 negative icon margin and the incorrect responsive breakpoint padding.


sauldom102 (commented):
Regarding the showcase files, can we iterate all the possible combinations of the button props so that we avoid hardcoding everything? If we added/removed variant, we could automatically reflect it on the showcase file without having to change it

developerdanx:
> Regarding the showcase files, can we iterate all the possible combinations of the button props so that we avoid hardcoding everything? If we added/removed variant, we could automatically reflect it on the showcase file without having to change it

Agreed it's a good pattern, but applying it here would set a precedent for all showcases (F0Tag, F0Avatar, F0Chip, etc.) and that's a broader convention change. I'd prefer to keep this PR scoped to the button fixes and address showcase iteration as a follow-up across all components. WDYT?

sauldom102:
> > Regarding the showcase files, can we iterate all the possible combinations of the button props so that we avoid hardcoding everything? If we added/removed variant, we could automatically reflect it on the showcase file without having to change it
> 
> Agreed it's a good pattern, but applying it here would set a precedent for all showcases (F0Tag, F0Avatar, F0Chip, etc.) and that's a broader convention change. I'd prefer to keep this PR scoped to the button fixes and address showcase iteration as a follow-up across all components. WDYT?

Sure, makes sense

sauldom102 (approved):

PR #3983, opened 2026-04-21
chore: add GitHub Actions workflow for publishing Expo prod updates

developerdanx:
## Description

- add GitHub Actions workflow for publishing Expo prod updates and notifying Slack FCT-52555

### DIFF
--- a/.github/workflows/publish-expo-production.yaml
+++ b/.github/workflows/publish-expo-production.yaml
@@ -1,0 +1,231 @@
+name: "Publish Expo Production & Notify"
+
+on:
+  workflow_run:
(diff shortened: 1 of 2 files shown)


sauldom102 (approved):
Nice one Dan!

.github/workflows/publish-expo-production.yaml
```diff
@@ -0,0 +1,231 @@
+name: "Publish Expo Production & Notify"
```
eliseo-juan, line 1: I would add this process in our docs! We will need a guide about what to do with a P0 an so on

PR #4015, opened 2026-04-23
feat(F0AiChat): enrich requisition entity ref card

ReddyWish:
Enriches the requisition entity ref hover card: adds a status tag (with variant), a location shown as the card description, and detail rows for line manager and reason. Introduces an `EntityRefDetails` subcomponent intended for reuse across entity ref cards; currently consumed only by `RequisitionEntityRef`.


sauldom102 (approved):
Nice one!

PR #4020, opened 2026-04-23
feat: add glow around inputs modified by One

sauldom102:
## F0Form: AI fill glow indicator

When the AI co-creation flow fills fields via `fillForm`, the changed inputs now show a rotating gradient border — similar to `AIButton` — to visually signal that values were set by the assistant.

Screenshot 2026-04-23 at 12 47 49

### What changed

**New component: `GlowingFieldWrapper`**
Thin wrapper around a form control that renders a `animate-rotate-gradient` conic-gradient ring (same palette as `AIButton`) when `isGlowing` is true. The input's own border is suppressed while the glow is active to avoid a double-border.

**`FieldRenderer`**
Wraps `<FormControl>` (the input only, not label/help text) in `GlowingFieldWrapper`. When the user edits a glowing field, `clearFieldGlow` is called immediately via a wrapped `onChange` — so the border disappears on first interaction.

**`F0AiFormRegistry`**
Added two new methods:
- `subscribeFilledFields(formName, cb)` — subscribe to fill events for a form, returns unsubscribe
- `notifyFieldsFilled(formName, fieldNames)` — called after `setValues` succeeds

**`useFormFillAction`**
Calls `registry.notifyFieldsFilled(formName, fieldNames)` after setting values, to trigger the glow on the affected fields.

**`F0Form` (`F0FormSingleSchema`)**
Subscribes to `subscribeFilledFields` and maintains `aiGlowingFields` state. Provides it via a new `F0FormAiGlowContext` that propagates through sections without being re-provided by `F0FormSection`.

### Scope

Only applies to forms rendered **outside the canvas** (i.e. `F0FormSingleSchema`). Canvas forms are not affected. The glow is cleared per-field as soon as the user interacts with it.


eliseo-juan:

PR #4019, opened 2026-04-23
docs(F0Button): add MDX documentation page

desiree-np:
## Description

Adds a complete MDX documentation page for `F0Button` following the canonical [`f0-docs` skill](packages/react/.skills/f0-docs/) conventions: top-level sections are `## Anatomy` / `## Guidelines` / `## Accessibility`, and the auto-generated docs are disabled in favour of curated content with do/don't visuals, copy rules, and runtime examples.

## Final structure

```
## Anatomy            (Canvas Default + Controls)
  ### Type            (Canvas Variants + Type/Description/When-to-use table)
  ### Icon            (Canvas IconVariants + Layout DoDonts)
  ### Sizes           (Canvas Sizes + Size/Use case/Notes table)
  ### States          (Canvas States + State/Description/When table)
## Guidelines
  ### Design best practices   (critical variant rules + DoDonts)
  ### Content best practices  (Correct/Incorrect bullets paired with DoDonts)
  ### Behavior                (Async / Navigation / Polymorphic ref)
## Accessibility
  ### Screen reader behavior
```

Type / Icon / Sizes / States are siblings directly under `## Anatomy` (no intermediate `Variants` container) — only Type is technically a variant prop; Sizes is `size` and States are runtime states (`disabled`, `loading`).

## Implementation details

- feat: add `F0Button.mdx` with the structure above
- feat: document the strict rules for the `critical` variant (irreversible only, requires confirmation, max one per context, never a primary CTA) paired with a Do/Don't visual
- chore: set `tags: [\"!autodocs\", \"stable\"]` on the meta to disable the global autodocs and `tags: [\"no-sidebar\"]` on the 7 doc-only stories so they don't pollute the sidebar
- fix: add `mb-4 last:mb-0` to the `DoDonts` root for vertical spacing between consecutive instances within an MDX section

## Verification

- Visit `Components / Button / Button / Documentation` in Storybook
- Right-side TOC shows the structure above with no empty containers
- Sidebar shows only the `Documentation` entry under `Button / Button` (the 7 doc-only stories are hidden)
- All `<Canvas>` blocks render their stories; the `Show code` button is the source of truth (no manual code blocks)

## Skill alignment

Built against the repo-local skill at `packages/react/.skills/f0-docs/`. Improvements to that skill discovered while writing this PR (e.g. \"refactor without losing value\", `### Behavior` inclusion filter, inline Correct/Incorrect copy rules) ship in PR #4021.

PR #4021, opened 2026-04-23
docs(f0-docs): add authoring principles learned from F0Button MDX

desiree-np:
## Description

Codifies three authoring principles surfaced while writing the F0Button MDX documentation page (#4019). All three produced real, observable problems during that work — they belong in the skill so future component documentation does not repeat them. Also ports the still-useful pieces of the soon-to-be-removed global \`factorial-f0/component-documentation/\` tree (factorialco/factorial-skills#158) so nothing is lost.

## Implementation details

- feat: add **Refactor without losing value** principle — when restructuring an existing MDX, audit the section's existing content first and combine valuable visuals with new patterns instead of deleting them
- feat: add explicit inclusion filter for **\`### Behavior\`** — skip when \`<Controls>\` / Variants / global behavior already covers it; include only non-obvious runtime behavior (async loading, polymorphic rendering, controlled/uncontrolled rules, ref typing); never rename to \"Code\", \"Patterns\", or \"Advanced\"; no \"Related components\" subsection
- feat: add **Inline Correct / Incorrect for copy rules** pattern under \`### Content best practices\` — scannable \`**Correct:**\` / \`**Incorrect:**\` bullets paired with a visual \`<DoDonts>\` rendering the actual component; bullets and DoDonts complement each other, never replace each other
- chore: extend the validation checklist in \`mdx-authoring.md\` with matching items so the new principles are enforced during review

## Companion PRs

- factorialco/f0#4019 — applies these principles to F0Button MDX
- factorialco/factorial-skills#158 — removes the duplicated global tree

### DIFF
--- a/packages/react/.skills/f0-docs/SKILL.md
+++ b/packages/react/.skills/f0-docs/SKILL.md
@@ -93,7 +93,7 @@ If a subsection would only contain redundant info, omit it. If the component has
 
 ### Inline Correct / Incorrect for copy rules — pair with a visual DoDonts
 
-For wording and copy rules in `### Content best practices`, use **inline `**Correct:** / **Incorrect:**` bullets** for the rules themselves — more scannable than prose for textual examples. Use plain bold; never emojis (✅ / ❌).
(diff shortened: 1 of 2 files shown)


eliseo-juan (approved):
LGTM!

PR #4026, opened 2026-04-23
fix: add min-w-0 to TwoColumnLayout to enable horizontal scroll

larizzatg:
## Description
When using a DataCollection with table visualization inside TwoColumnLayout, the table's horizontal scrollbar never appeared.
  
This made call-to-action columns unreachable when the viewport narrowed (e.g. opening OneChat).

image

## Implementation details
Flex items default to min-width: auto, which prevents them from shrinking below their content's intrinsic width. The main element stretched to match the table, so the table never detected overflow.

Adding min-w-0 allows main to shrink within the available space, letting the table's overflow-auto trigger correctly.

### DIFF
--- a/packages/react/src/layouts/TwoColumnLayout/index.tsx
+++ b/packages/react/src/layouts/TwoColumnLayout/index.tsx
@@ -33,7 +33,7 @@ const _TwoColumnLayout = forwardRef<HTMLDivElement, TwoColumnLayoutProps>(
         >
           <main
             className={cn(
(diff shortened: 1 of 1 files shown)


viniciusmeneses:
@larizzatg does this fix also work for the "editable table"? We are going to replace the goals data collection with the editable version

larizzatg:
> @larizzatg does this fix also work for the "editable table"? We are going to replace the goals data collection with the editable version

I think so but I received a feedback from Jona to not use the same responsive model as the two column layout so I'm changing it now ad we will not using the layout at the end 😅

eliseo-juan:

PR #4037, opened 2026-04-24
feat(OneDataCollection): add loading/success/error status for bulk actions

juancruz-f7l:
## 🚪 Why?

[why]: # (✍️ _What needs is this Pull Request solving or how is it improving the product?_)

Bulk actions in `OneDataCollection` gave users no feedback during async operations. Clicking "Archive" or "Delete" on a set of records would fire the handler and immediately clear the selection — no spinner, no confirmation, no error recovery. Users had no way to tell whether the operation succeeded, was still running, or failed silently.

This PR brings the same `idle | loading | success | error` state machine that `FormActionBar` already uses for form submission — now to bulk-action bars.

## 🔑 What?

[what]: # (✍️ _What changes is this Pull Request introducing in the code?_)

- **New `OneDataCollectionActionBar` component** wraps `F0ActionBar` and moves the selection chrome (count + unselect button + warning) into `F0ActionBar`'s `leftContent` slot. `ActionBar/index.tsx` becomes a re-export so all existing import paths are preserved without change.
- **`onBulkAction` return type widens** from `void` to `void | Promise<void>` — non-breaking, sync consumers are unaffected.
- **`autoManageBulkActionStatus` prop** (default `false`, opt-in) — when `true`, a `Promise` returned from `onBulkAction` drives the bar through `loading → success → idle` (resolve) or `loading → error` (reject). On error the bar wiggles with `errorHighlight`, preserves selection for retry, and re-enables actions. On success, selection clears (unless `keepSelection: true`) and the bar auto-dismisses after 1.5 s.
- **Opt-in is intentional**: handlers that open a modal typically resolve when the modal _opens_, not when the user confirms — auto-managing those would flash a premature success. Without the flag, async handlers keep today's fire-and-forget behavior.
- **`bulkActionStatus` prop** (controlled mode) — overrides internal management for multi-step flows (confirm dialogs, server polling, wizards) where a single promise can't model the full lifecycle.
- **Bar visibility extended**: the bar stays open during `loading` and `success` even after selection is cleared, so users see the outcome.
- **Error auto-clear**: any selection change after an `error` resets to `idle`, preventing stale failure states across subsequent actions.

## ✅ Verification

https://github.com/user-attachments/assets/37e6724c-09e9-4ffb-ad3f-724f5b200c13

### Tests

- New integration test file: `packages/react/src/patterns/OneDataCollection/__tests__/bulkActionStatus.spec.tsx` — 5 tests covering:
  1. Sync handler clears selection immediately, never enters loading
  2. Async handler **without** `autoManageBulkActionStatus` preserves fire-and-forget (regression guard)
  3. Async handler **with** `autoManageBulkActionStatus` disables buttons while pending, dismisses on resolve
  4. Async rejection preserves selection and re-enables buttons for retry
  5. Controlled `bulkActionStatus="loading"` disables actions regardless of handler shape
- All 338 existing `OneDataCollection` unit tests pass (`28 test files`)
- New `WithAsyncBulkActions` Storybook story demonstrates `loading → success` and `loading → error` flows interactively (Archive resolves, Delete rejects)

### Manual Verification

- Storybook story `WithAsyncBulkActions`: select rows → click Archive → bar shows spinner → success checkmark → dismisses. Click Delete → bar shows spinner → error wiggle → buttons re-enable for retry.


juancruz-f7l:
## Review comments triage (67f2a44)

All four Copilot comments were valid. Here's what was done:

---

**Comment 1 — `.spec.tsx` → `.test.tsx`** ✅ Fixed

`bulkActionStatus.spec.tsx` renamed to `bulkActionStatus.test.tsx`. AGENTS.md explicitly forbids `.spec.ts` for Vitest tests — the sibling `index.spec.tsx` in the same `__tests__/` directory was the wrong precedent to follow.

---

**Comment 2 — Unselect button stays clickable during loading** ✅ Fixed

`OneDataCollectionActionBar` now derives `isInteractionDisabled = status === "loading" || status === "success"` at component scope and passes `disabled={isInteractionDisabled}` to the Unselect `F0Button`. This is necessary because `F0ActionBar`'s `isInteractionDisabled` flag only gates its own primary/secondary action buttons — `leftContent` is rendered raw and receives no implicit disabled propagation. Without this fix, clicking Unselect during an in-flight action would wipe the selection before a rejected promise could preserve it for retry.

New test: _"Unselect button is disabled while a bulk action is loading"_

---

**Comment 3 — Selection changes during pending promise → data loss on resolve** ✅ Fixed

Added a `selectionVersionRef` (monotonic counter, incremented on every selection change). The version is captured at bulk-action click time (`versionAtClick`). On promise resolve, `clearSelectedItems()` is skipped if `selectionVersionRef.current !== versionAtClick` — meaning the user changed selection while the operation was in-flight. Status still transitions to `success → idle` (the server call did succeed), but the user's updated selection is preserved for them to act on next.

New test: _"selection change during a pending promise preserves the new selection on resolve"_

---

**Comment 4 — Mobile secondary actions stay clickable during loading** ⚠️ Deferred

This is a real bug: `F0ActionBar`'s mobile path renders `<MobileDropdown items={secondaryActions} />` without propagating `isInteractionDisabled`. The `MobileDropdown` component has no `disabled` prop and its inner `<button>` elements fire unconditionally. The correct fix is upstream in `F0ActionBar` (pass `isInteractionDisabled` into the mobile render path), which benefits `FormActionBar` consumers as well. Addressing it here would expand the scope beyond `OneDataCollection`. Tracked for a follow-up PR against `F0ActionBar`.

juancruz-f7l:
## Second round review comments triage (f4ee490)

---

**Comment 1 — Auto-manage fights controlled mode** ✅ Fixed

Added a guard immediately after the `isPromise` check: when `controlledBulkActionStatus` is defined, we return early from the promise-handling branch entirely. No internal timer starts, `clearSelectedItems()` is not called, and `setInternalBulkActionStatus` is never touched. The consumer's external state machine owns the full lifecycle.

New test: _"controlled bulkActionStatus prevents auto-manage from clearing selection on resolve"_

---

**Comment 2 — Warning message hides Unselect button (regression)** ✅ Fixed

This was a genuine regression introduced during the refactor. The old `ActionBar/index.tsx` rendered selection chrome (count + Unselect) independently of `warningMessage` — the warning only replaced the _actions area_, not the left chrome. The new `leftContent` useMemo short-circuited on `warningMessage`, hiding the Unselect button entirely.

Fixed by restructuring `leftContent`: warning and selection chrome are now independent — both render when both are present, wrapped in a column-flex container on mobile that switches to row on `sm:`.

---

**Comment 3 — MobileDropdown secondary actions not disabled** ⚠️ Duplicate — already triaged

This is the same issue from the first review round (previously comment #4). See the triage already posted: https://github.com/factorialco/f0/pull/4037#issuecomment-4314271546. The fix belongs upstream in `F0ActionBar`. Docs updated to match actual behavior (see comment 5 below).

---

**Comment 4 — Test uses real 1500ms timer** ✅ Fixed

The "dismisses on resolve" test now uses `vi.useFakeTimers({ shouldAdvanceTime: true })` with `userEvent.setup({ advanceTimers: vi.advanceTimersByTime.bind(vi) })`. After `resolveHandler?.()` + `await Promise.resolve()` (flush microtasks), `vi.advanceTimersByTime(1500)` skips the dismiss delay instantly. Added `afterEach(() => vi.useRealTimers())` to the describe block as a safety net.

Result: `bulkActionStatus.test.tsx` suite time drops from ~1850ms → ~510ms.

---

**Comment 5 — Docs claim "disables every action"** ✅ Fixed

Updated `selectable-bulk-actions.mdx` to accurately describe what is and isn't disabled: primary actions and visible secondary buttons are disabled; overflow secondary actions in the mobile dropdown are not yet disabled, with a note that it's tracked upstream in `F0ActionBar`.

eliseo-juan:
I like it, but I would change the approach, I would move the loading to the button itself, not the whole bar, what do you think?

juancruz-f7l:
Feedback applied @eliseo-juan:

Single primary action:

https://github.com/user-attachments/assets/543e1004-20b5-4c0a-a413-0ed97ac4aa80

Multiple primary actions:

https://github.com/user-attachments/assets/c41956f4-c156-4154-9033-b35ae3992a9f

eliseo-juan (approved):
LGTM!

PR #4045, opened 2026-04-27
feat(editable-table): add onCellBlur

laurafg:
## Summary

- Add `onBlur` optional prop to `EditableCellProps` (shared across all cell types) and wire it through `TextCell` to the underlying `Input` component
- Wire `onCellBlur` end-to-end through all layers: table API → `EditableTable` → `EditableRowContext` → `EditableCellRenderer` → cell components
- Allow `NumberCell` (and `MoneyCell`) to emit null values when the user clears the input, instead of coercing to `0`

### `cells/index.ts`
- Added `onBlur?: () => void` to `EditableCellProps<R>`

### `cells/TextCell.tsx`
- Destructure and pass `onBlur` to `<Input>`

### `types.ts`
- Added `onCellBlur?: (item: R, columnId: string) => void` to `EditableTableVisualizationOptions`

### `context/EditableRowContext.tsx`
- Added `onCellBlur` to provider props and `handleCellBlur` to context value
- `handleCellBlur` calls `onCellBlur` with the current local item and column id

### `EditableTable.tsx`
- Destructures `onCellBlur`, stores in a ref, and passes to `EditableRowProvider`

### `components/EditableCellRenderer.tsx`
- Creates per-cell `onBlur` closure binding column id to `handleCellBlur`
- Passes `onBlur` to all cell components

## Usage

```tsx
<EditableTable
  onCellChange={async (updatedItem) => { /* save */ }}
  onCellBlur={(item, columnId) => {
    // Triggered when any editable cell loses focus
    // item contains the current row data, columnId identifies which cell
  }}
/>
```

### DIFF
--- a/packages/react/src/patterns/OneDataCollection/visualizations/collection/EditableTable/EditableTable.tsx
+++ b/packages/react/src/patterns/OneDataCollection/visualizations/collection/EditableTable/EditableTable.tsx
@@ -30,6 +30,7 @@ export const EditableTableCollection = <
   Grouping extends GroupingDefinition<R>,
 >({
   onCellChange,
+  onCellBlur,
   addRowActions,
   addRowActionsLabel,
   addNestedRowActions,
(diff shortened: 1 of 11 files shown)


viniciusmeneses:
@laurafg I moved the logic to allow `""` in the number input to a separate PR https://github.com/factorialco/f0/pull/4050 because the `onCellBlur` callback here needs more work, currently it's not firing for date cells and it doesn't look super straightforward to add it 🙏

eliseo-juan:

PR #4054, opened 2026-04-28
feat(F0Tabs): make tab id type generic

oscarbyte:
## Summary

- Make `F0TabItem` and `F0TabsProps` generic over the tab `id` type (`T extends string`, defaults to `string`).
- Consumers can now use string-literal unions for tab ids and get fully typed `activeTabId` / `setActiveTabId` without casts.
- `React.memo` does not preserve generics, so the memoized component is internally typed as `F0TabsProps<string>` and the public export is cast to a generic call signature. Runtime behaviour is unchanged.

## Motivation

Today, mobile/web consumers with a typed tab id union (e.g. `"my_courses" | "catalogue"`) must cast `setActiveTabId={handleSetActiveTabId as (id: string) => void}` because `F0Tabs` only accepts `Dispatch<string>`. Making the component generic removes the cast and prevents typos in `activeTabId`.

## API

```tsx
type ScreenTab = "overview" | "settings"
const tabs: F0TabItem<ScreenTab>[] = [
  { id: "overview", label: "Overview" },
  { id: "settings", label: "Settings" },
]
const [activeTab, setActiveTab] = useState<ScreenTab>("overview")

<F0Tabs<ScreenTab>
  tabs={tabs}
  activeTabId={activeTab}
  setActiveTabId={setActiveTab}
/>
```

The type parameter is inferred from the `tabs` array when literal `id` values are used; explicit annotation (as above) is only needed when callers want to enforce a specific union.

## Backward compatibility

`T extends string = string`, so all existing call sites compile unchanged. The 21 existing F0Tabs tests pass without modification.

## Caveats

- Web `TabsProps` is not currently generic; this introduces a small API divergence (RN is stricter when consumers opt in). The web component can be made generic later with the same pattern if desired.
- Because `React.memo` does not preserve generics, the public export uses an `as unknown as <T>(...) => ...` cast. This is purely a type-level wrapper — there is no runtime change.

## Validation

- `pnpm tsc` — no new errors introduced (pre-existing `F0BlurView` / `expo-blur` errors unrelated to this change).
- `pnpm test -- --watchman=false --testPathPatterns=F0Tabs` — 21/21 passing.
- `pnpm lint` — 0 warnings, 0 errors.
- `pnpm format:check` — clean.

---

> [!NOTE]
> **Low Risk**
> Type-surface-only change with a small export type cast; runtime behavior and rendering logic are unchanged, so risk is limited to potential TypeScript typing edge cases for consumers.
> 
> **Overview**
> `F0TabItem` and `F0TabsProps` are updated to be generic over the tab `id` type (`T extends string = string`), so `activeTabId`/`setActiveTabId` can be fully typed when consumers use string-literal unions.
> 
> Because `React.memo` drops generic inference, the memoized component is now internalized as `F0TabsProps<string>` and the public `F0Tabs` export is cast back to a generic call signature; runtime behavior is unchanged. Docs were updated with a typed-id usage example and notes about the generic/memo typing caveat.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit cecf1a3a8315e95e4404001be13e2bc16e2d3a27. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/react-native/src/components/F0Tabs/F0Tabs.md
+++ b/packages/react-native/src/components/F0Tabs/F0Tabs.md
@@ -73,6 +73,19 @@ import { F0Tabs } from "@factorialco/f0-react-native"
 
 // Embedded — renders only the first tab as plain text, no interaction
 <F0Tabs tabs={tabs} embedded />
+
+// Typed tab ids — narrow `id` to a string-literal union
+type ScreenTab = "overview" | "settings"
+const tabs: F0TabItem<ScreenTab>[] = [
+  { id: "overview", label: "Overview" },
+  { id: "settings", label: "Settings" },
+]
+const [activeTab, setActiveTab] = useState<ScreenTab>("overview")
+<F0Tabs<ScreenTab>
+  tabs={tabs}
+  activeTabId={activeTab}
+  setActiveTabId={setActiveTab}
+/>
 ```
 
 ## Props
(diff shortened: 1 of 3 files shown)


developerdanx (changes requested):
Thank you for the PR :) 

Here are some comments:

1. **Rename `AnyF0TabsProps`** → `F0TabsBaseProps`. "Any" reads like an `any` escape hatch; it's just `F0TabsProps<string>`.

2. **Simplify the export cast.** The generic signature is defined twice and uses `as unknown as`. Single assertion is enough:


```tsx
export const F0Tabs = F0TabsBase as <T extends string = string>(
  props: F0TabsProps<T>
) => React.ReactElement | null
```

3. **Use `React.ReactElement | null`** instead of `ReturnType<typeof F0TabsBase>` — standard return type, no coupling to internals.

Generic approach, type changes, and backward compat are all correct. Memo is preserved at runtime — the indirection is only needed because `React.memo` erases generics.

developerdanx (approved):

PR #4055, opened 2026-04-28
fix: editable table QA

RafaMellado:
## Why
Fix nested row connector rendering when add-row actions are the only visible child, and clean up money unit positioning logic in editable table cells.

## What
- Update nested row connector height handling to support add-row actions without relying on loaded child rows
- Mark add-row actions as child content so the vertical connector is rendered correctly
- Assign connector refs when `AddRowRow` is the first and only visible child
- Simplify `MoneyCell` unit position resolution while preserving locale-based fallback behavior

## Verification
- Manually verified the connector is shown when `AddRowRow` is the only child row
- Ran formatting and lint checks


eliseo-juan (approved):
LGTM

PR #4061, opened 2026-04-29
feat(F0AiChat): self-heal dashboard recipes via /compute repair

wili3:
## Description

Wires the agent-side dashboard self-heal contract (`/dashboard/compute` may now return `repairedSpecs` and `repairFailures`) into the f0 chat dashboard. When the agent heals a stale recipe, the canvas swaps `config.fetchSpecs` and persists the repaired recipe via the existing dashboard save callback, so future opens skip the repair round-trip. Per-item `{ error }` slices are no longer silently coerced to empty defaults — the existing `DashboardItem` error UI now renders for `/compute` failures.

## Behavior change

Item-level errors returned by `/dashboard/compute` (`{ error }` on a single item) now render `DashboardItem`'s `OneEmptyState` "Error loading data" + retry card instead of an empty chart / zero metric / empty table.

This affects two pre-existing cases that have nothing to do with self-heal:

- The agent's per-item SQL failures, which already returned `{ error: message }` on `main`.
- `useDashboardCompute`'s synthesized `{ error: "No result for item" }` when an item id is missing from a successful response.

Both used to render silently as empty visuals; they now surface the error to the user. This is the same UI already shown by `useDashboardItemData` for non-chat `F0AnalyticsDashboard` consumers, so it's battle-tested.

No backend coupling: works against the current agent, the new fields (`repairedSpecs`, `repairFailures`) are gracefully ignored when absent.

## Implementation details

- feat: extend `ComputeResponse` with optional `repairedSpecs` and `repairFailures`, plus a structured `ItemResult.reason` classifier
- feat: add an `onRepair` callback to `useDashboardCompute`, fired (best-effort) when a repaired recipe is returned
- feat: wire `handleRepairedSpecs` in the canvas wrapper — updates canvas content, best-effort `saveConfigToHistory`, and calls `canvasActions.dashboard.save` for already-saved dashboards; idempotent via a ref keyed on `savedDashboardId:toolCallId` and reset when `fetchSpecs` reference changes
- fix: propagate per-item compute errors through chart/metric/collection adapters so `DashboardItem`'s `OneEmptyState` error branch renders instead of empty defaults

  <details>
  <summary>Why?</summary>

  The three adapters (`mapChartItem`, `mapMetricItem`, `mapCollectionItem`) previously read only `r.chart` / `r.metric` / `r.collection` and fell back to empty defaults with `??`, never inspecting `r.error`. This made `DashboardItem`'s error branch dead code for chat dashboards and hid backend failures behind empty charts / zero metrics / empty tables.
  </details>

## Notes for downstream

The agent contract is implemented in [factorial-agent](https://github.com/factorialco/factorial-agent) (PR pending). After this PR's CI publishes an alpha, the factorial monorepo's `@factorialco/f0-react` will be pinned to that alpha to validate the full round-trip end-to-end.

### DIFF
--- a/packages/react/src/lib/providers/i18n/i18n-provider-defaults.ts
+++ b/packages/react/src/lib/providers/i18n/i18n-provider-defaults.ts
@@ -386,9 +386,9 @@ export const defaultTranslations = {
       dataExplanation: "Where does this data come from?",
     },
     dashboardDatasetFailure: {
-      title: "Some data couldn’t be loaded",
+      title: "This dashboard is no longer valid",
       description:
-        "We couldn’t load this section of the dashboard. Details: {{reason}}",
+        "We're sorry. Please contact support or ask One to generate a dashboard from scratch.",
     },
     pong: {
       title: "Pong",
--- a/packages/react/src/lib/providers/i18n/i18n-provider-defaults.ts
+++ b/packages/react/src/lib/providers/i18n/i18n-provider-defaults.ts
@@ -385,6 +385,11 @@ export const defaultTranslations = {
       retry: "Retry",
       dataExplanation: "Where does this data come from?",
     },
+    dashboardDatasetFailure: {
+      title: "Some data couldn’t be loaded",
+      description:
+        "We couldn’t load this section of the dashboard. Details: {{reason}}",
+    },
     pong: {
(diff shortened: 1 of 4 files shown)


eliseo-juan:

PR #4100, opened 2026-05-08
feat: add option creation in select from dataset question type

mgrol:
## Description

Add support for creating new items directly from dropdown-single questions in surveys. When a dataset provides an `onCreate` callback, form builders can enable an "Allow creation" toggle so respondents can create new entries inline when no search results match.

### SurveyFormBuilder
image

### SurveyAnsweringForm
image

## Implementation details

The F0Select component now accepts an `onCreate` prop that renders a "Create" button in the empty state when a search yields no results. 

In the `SurveyFormBuilder`, a new `allowCreate` property is added to the dropdown-single question type, along with an "Allow creation" toggle in the question actions menu, visible only when the linked dataset has an `onCreate` handler. 

The `SurveyAnsweringForm` wires `allowCreate` and `dataset.onCreate` into the form schema so the create option appears during answering.


packages/react/src/sds/surveys/SurveyAnsweringForm/hooks/useSurveyFormSchema.tsx
```diff
@@ -410,20 +411,36 @@ function buildFieldForQuestion(
...
+          const fieldWithCreate: F0Field =
+            q.allowCreate && dataset.onCreate
+              ? {
+                  ...field,
+                  onCreate: (searchValue: string) => {
+                    dataset.onCreate!(searchValue).then((record) => {
+                      const option = dataset.mapOptions(
+                        record
+                      ) as F0SelectItemObject<string>
+                      ;(onChange as (value: unknown) => void)(option.value)
```
sauldom102: What are we doing here?

mgrol: I updated many of the changes, opening it for review now 🙇‍♀️

sauldom102 (approved):
Looks nice! Thank you Maria!

PR #4131, opened 2026-05-13
fix(kanban): optimize lane updates

xavikp:
## Description

Fixes a set of compounding performance issues in KanbanCollection that caused unnecessary re-renders and serialization work on every render cycle.

Captura de pantalla 2026-05-14 a les 10 26 45

## Implementation details

- Removed 8 JSON.stringify memos in useDataCollectionLanesData that serialized filters, sortings, grouping and dataAdapter every render. Replaced with a single useDeepCompareMemoize over a plain-data descriptor.
- Batched per-lane setLanesHooks calls via queueMicrotask — N lanes loading in parallel now produce 1 parent re-render instead of N.
- Fixed onHookUpdate update loop — changed its signature to (laneId, value) so the parent passes the stable handleHookUpdate directly instead of a new inline arrow every render.
- Removed O(N×M) lanesSignature in Kanban.tsx — JSON.stringify over all records across all lanes on every render. laneItems now depends directly on lanesHooks.
- Memoized kanbanLanes, getKey, renderCard, dnd, and kanbanProps — previously all rebuilt inline every render. Narrowed renderCard deps from the whole (unstable) source ref to the specific fields it uses.
- Hoisted two pure helpers to module scope so they aren't recreated per render.

### DIFF
--- a/packages/react/src/patterns/OneDataCollection/hooks/useDataCollectionData/useDataCollectionLanesData.tsx
+++ b/packages/react/src/patterns/OneDataCollection/hooks/useDataCollectionData/useDataCollectionLanesData.tsx
@@ -1,4 +1,5 @@
-import { useCallback, useEffect, useMemo, useState } from "react"
+import { useCallback, useEffect, useMemo, useRef, useState } from "react"
+import { useDeepCompareMemoize } from "use-deep-compare-effect"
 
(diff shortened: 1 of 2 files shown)


eliseo-juan (approved):
LGTM

PR #4139, opened 2026-05-14
feat(ai): add proposal card

pol-torrent:
## Summary
- Add a reusable AI proposal card to the React SDS AI package.
- Include truncation/expand behavior, optional module metadata, action callbacks, data attribute forwarding, tests, and Storybook stories.
- Export `F0AiProposalCard` from the public AI SDS barrel.

## Validation
- pnpm --filter @factorialco/f0-react vitest:ci src/sds/ai/F0AiProposalCard/__tests__/F0AiProposalCard.test.tsx
- pnpm --filter @factorialco/f0-react run format
- pnpm --filter @factorialco/f0-react run tsc
- git diff --check


sauldom102 (changes requested):
A design feedback they have given while discussing it internally with the team:
- Primary button should be on the right-most of the card and not on the left.
- Cancel should be secondary variant instead of ghost

pol-torrent:
@sauldom102 primary action has been moved to the right, secondary action has been removed because it has no use case at the moment

sauldom102 (approved):
LGTM!

PR #4156, opened 2026-05-18
fix(ai): render custom placeholder when placeholders array has single element

viniciusmeneses:
## Summary

Fixes a bug where `ChatTextarea` ignored the custom placeholder when the `placeholders` array from `useAiChat` contained exactly one element, always showing the default i18n placeholder instead.

## Problem

The placeholder rendering in `TextareaField` has two paths:

1. **Static `<p>` tag** — renders when `!multiplePlaceholders` (i.e. array length <= 1)
2. **`TypewriterPlaceholder`** — renders only when `multiplePlaceholders` (array length > 1)

When `placeholders` has exactly 1 item:
- `multiplePlaceholders` is `false` (since `1 > 1` is `false`)
- The static path fires but always renders `resolvedDefaultPlaceholder` (the i18n default text)
- The custom placeholder value is never used

## Fix

The static placeholder now checks if there's a single custom placeholder and displays it instead of the default:

```tsx
{placeholders.length === 1
  ? placeholders[0]
  : resolvedDefaultPlaceholder}
```

This keeps existing behavior intact:
- **0 placeholders** → default i18n text (static)
- **1 placeholder** → custom text (static) ← fixed
- **2+ placeholders** → typewriter animation cycling through all items

Screenshot 2026-05-18 at 08 34 59

[Task](https://factorialmakers.atlassian.net/browse/FCT-54127)

### DIFF
--- a/packages/react/src/sds/ai/F0AiChat/components/input/ChatTextarea/TextareaField.tsx
+++ b/packages/react/src/sds/ai/F0AiChat/components/input/ChatTextarea/TextareaField.tsx
@@ -96,7 +96,9 @@ export const TextareaField = ({
             "overflow-hidden text-ellipsis whitespace-nowrap"
           )}
         >
-          {resolvedDefaultPlaceholder}
+          {placeholders.length === 1
+            ? placeholders[0]
+            : resolvedDefaultPlaceholder}
         </p>
       )}
       <textarea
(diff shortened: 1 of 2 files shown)


sauldom102 (approved):
Let's add some unit test for this to ensure this never breaks later on

PR #4160, opened 2026-05-18
feat: fix number input min cap onblur

aaroncaballero-f:
## Description

Defer min/max clamping to blur instead of applying it on every keystroke. Previously, NumberInput enforced min/max constraints in `handleChange`, which made it impossible to type intermediate values (e.g., deleting digits from "2283" to type "3000" would instantly clamp "2" back to "2283"). Now users can type freely and constraints are enforced on commit (blur or Enter).

Fixes FCT-54384: goal progress value snaps back to original when startValue != 0.

N/A — behavioral fix, no visual changes.

## Implementation details

- Removed real-time clamping from `handleChange` — it now preserves raw input and emits the unclamped value via `onChange`.
- Added `commitValue()` that clamps + reformats on blur and Enter (also normalizes trailing decimal separators like "17." → "17").
- Wired `onPressEnter` handling via `onKeyDown` since the raw `Input` component doesn't accept `onPressEnter` directly.
- Step buttons still validate bounds before calling `handleChange` (no behavior change there).
- Added tests for: typing below min during editing, clamping on blur (min/max), trailing separator normalization on blur.

### DIFF
--- a/packages/react/src/experimental/Forms/Fields/NumberInput/__tests__/NumberInput.test.tsx
+++ b/packages/react/src/experimental/Forms/Fields/NumberInput/__tests__/NumberInput.test.tsx
@@ -181,5 +181,123 @@ describe("NumberInput", () => {
       expect(input).toHaveValue("17,")
       expect(onChange).toHaveBeenLastCalledWith(17)
     })
+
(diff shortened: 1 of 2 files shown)


eliseo-juan:

PR #4163, opened 2026-05-18
feat(F0Alert): make description prop optional

rubenmoya:
## Summary

- Makes the `description` prop on `F0Alert` optional so consumers can render a title-only alert.
- When `description` is omitted, the title vertically centers with the avatar icon to avoid an off-center single-line layout; the original top alignment is preserved when a description is present.

## Changes

- `packages/react/src/components/F0Alert/types.ts` — `description: string` → `description?: string`.
- `packages/react/src/components/F0Alert/F0Alert.tsx` — conditionally render the description paragraph and apply `items-center` to the avatar+text row only when there's no description.

## Notes

- Non-breaking: loosens a required prop to optional.
- Format, typecheck, lint and existing F0Alert tests all pass locally.


fleveque (approved):
LGTM 🌮

PR #4200, opened 2026-05-21
feat: editable table status pill select

FaithAdekunle:
## Description

This PR adds status tag support to F0Select and also displays selected item's avatar when present.

- Adds a new `status` variant to `F0SelectTagProp`, allowing options to carry a `{ type: "status", text, variant }` tag.
- `SelectItem` renders a status pill inline; status-only items render as a standalone pill.
- `SelectedItems` renders the status pill in the trigger and also gains avatar rendering for selected options.
- New stories: `WithStatusTags`, `WithPersonAvatar` and `EditableTableWithStatusPillSelect`
Screenshot 2026-05-18 at 14 20 27
Screenshot 2026-05-18 at 14 21 20
Screenshot 2026-05-18 at 14 24 33

[Link to Figma Design](https://www.figma.com/design/A4204Nlu1QulvPCR8nr6GM/Headcount-planning?node-id=10382-87109&p=f&t=B5lO21aEFsYpdR5u-0)

### F0Select

- **`types.ts`**: Extended `F0SelectTagProp` union with `{ type: "status"; text: string; variant: StatusVariant }`, imported from `F0TagStatus/types`.
- **`SelectItem.tsx`**: Added an `isStatusOnly` short-circuit — when an item has only a status tag (no avatar/icon/description), render just a centered `F0TagStatus`. In the standard layout, added a `tag.type === "status"` branch alongside the existing `dot` / `person` / raw string branches.
- **`SelectedItems.tsx`**: Added an early return when the selected item's tag is a status tag, so the trigger renders the pill directly. Separately, added avatar rendering (`F0Avatar` at `size="xs"`) at the start of the standard layout, mirroring the existing `icon` slot.

### Stories

- New F0Select stories: `WithStatusTags`, `WithPersonAvatar`.
- New EditableTable stories: `EditableTableWithDateCellMinMax`, `EditableTableWithStatusPillSelect`.


sauldom102 (changes requested):
@FaithAdekunle let's consider copilot comments from above, I think they're worth checking and being addressed

sauldom102 (commented):
What's the real use-case for this? Do you have a Figma link? For me it's a bit weird in terms of design to allow to have a select for example and multiple options having the same variant 🤔 maybe we simply need to restrict that, or maybe having this as a regular select shouldn't be the case

sauldom102:
> What's the real use-case for this? Do you have a Figma link? For me it's a bit weird in terms of design to allow to have a select for example and multiple options having the same variant 🤔 maybe we simply need to restrict that, or maybe having this as a regular select shouldn't be the case

Okay I was just taking a look at the design, I think it makes sense to allow selecting statuses, we just need to restrict to have more than one option with same variant I'd say, if we have more than one with the same variant we could throw an error in development and just filter out duplicated options in production

FaithAdekunle:
> > What's the real use-case for this? Do you have a Figma link? For me it's a bit weird in terms of design to allow to have a select for example and multiple options having the same variant 🤔 maybe we simply need to restrict that, or maybe having this as a regular select shouldn't be the case
> 
> Okay I was just taking a look at the design, I think it makes sense to allow selecting statuses, we just need to restrict to have more than one option with same variant I'd say, if we have more than one with the same variant we could throw an error in development and just filter out duplicated options in production

I worry about this..cos it will effectively caps status-tagged selects to at most 5 options. I already implemented it but I will revert. See Copilot comment [here](https://github.com/factorialco/f0/pull/4200#discussion_r3337416395)

sauldom102 (changes requested):
When having "status" items in a select we should just likely force the select to have a minimum size of "md" even if specified as small. Because the spacing when the item is selected is a bit weird

Image

sauldom102 (changes requested):
We shouldn't be allowed to pass multiple items of different type to a select, we should throw an error in that case and also cover that case with a unit test

sauldom102 (approved):

PR #4203, opened 2026-05-21
feat: one table min width

FaithAdekunle:
## Description

This PR introduces `minWidth` to `OneTable` columns

- `TableHead` and `TableCell` accept an optional `minWidth` prop that overrides the minWidth derived from `width`, letting columns grow past `width` without shrinking below a floor.
- Column `minWidth` plumbed through `TableColumnDefinition`, `Table.tsx`, and `Row.tsx`.
- New `MinWidth` story.
Screenshot 2026-05-18 at 14 21 57

[Link to Figma Design](https://www.figma.com/design/A4204Nlu1QulvPCR8nr6GM/Headcount-planning?node-id=10382-87109&p=f&t=B5lO21aEFsYpdR5u-0)

### OneTable

- **`TableCell/index.tsx` and `TableHead/index.tsx`**: Added optional `minWidth` prop (`number | "auto"` and `ColumnWidth` respectively). When set, it overrides the `minWidth` style that previously always equaled `width`. Implemented as `colMinWidth = minWidth !== undefined ? getColWidth(minWidth) : colWidth` so existing behavior is preserved when the prop is omitted.
- **Why**: Previously `minWidth === width === maxWidth`, which prevented columns from growing past their declared width. The new prop decouples the lower bound so a column can flex upward while keeping a floor.

### Stories

- New OneTable story: `MinWidth`.

### DIFF
--- a/packages/react/src/experimental/OneTable/TableCell/index.tsx
+++ b/packages/react/src/experimental/OneTable/TableCell/index.tsx
@@ -46,6 +46,13 @@ interface TableCellProps {
   width?: number | "auto"
 
   /**
+   * Optional minimum width for the cell. When provided, overrides the
+   * minWidth derived from `width`, allowing the cell to shrink no further
+   * than this value.
+   */
+  minWidth?: number | "auto"
+
+  /**
    * When true, the header cell will stick in the specified position when scrolling horizontally
    * @default undefined
    */
@@ -99,6 +106,7 @@ export function TableCell({
   href,
   onClick,
   width = "auto",
+  minWidth,
   firstCell = false,
   sticky,
   colSpan,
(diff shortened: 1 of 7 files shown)


sauldom102 (commented):
Why's this change needed in the first place? 🤔 could you send the real use-case / Figma link of this?

FaithAdekunle:
@sauldom102 

Here is a screenshot from the figma design (link in pr description). The real use case for us is in editable table where a column may not yet have a value so it looks really narrow until a value is entered. The idea here is to let it have a min width so it does not look so thin even when empty. See the `Role` column in the screenshot.
Screenshot 2026-06-02 at 07 01 05

sauldom102 (approved):

PR #4207, opened 2026-05-21
fix: onApply as independant function

davidFactorial:
## Description

Updates F0Select so onApply works as an independent confirmation callback in multi-select mode. When onApply is provided, staged selections are only confirmed through Apply, and the callback receives the confirmed selection payload directly without relying on onChange.

## Implementation Details
- Changed onApply to receive the confirmed selection payload:
  value, originalItems, and options
- Kept onApply independent from onChange
- Ensured clicking Apply calls only onApply, not onChange
- Preserved internal committed selection tracking so Cancel and outside-close restore the last applied state
- Updated Storybook docs to describe the new onApply contract
- Added tests covering:
  - onApply receiving the confirmed payload
  - onApply working without onChange
  - staged changes being cancelled on outside click
  - Cancel restoring the committed selection


sauldom102 (changes requested):
As a behavior I think it makes total sense to have an "Apply selection" being enforced in certain cases, I would keep `onChange` though working as it was doing already, instead of `onApply` we could have a `withApplySelection` optional boolean prop that introduces this behavior, but `onChange` should keep being called as the value changes after the user clicks apply

I have been discussing this with @siguenzaraul as well and we think it makes sense

sauldom102 (approved):

PR #4211, opened 2026-05-21
feat(F0Card): add optional action prop to CardAlert

idowuarifayan:
## Description

Add an optional `action` prop to `CardAlert` (and by extension `F0Card`'s `alert` prop), allowing consumers to render an action button in the alert header strip.

The `action` prop is mutually exclusive with `dismissible` — TypeScript enforces this via a discriminated union, and the render uses a ternary so only one is ever shown.

### Design
Screenshot 2026-05-21 at 14 48 16

N/A — uses existing `F0Button` with `variant="outline"` and `size="sm"`, matching the pattern in `F0Alert`.

[Link to Figma Design](https://www.figma.com/design/SqIAXKMbJVUbwRYBvTILqW/Sidepanel-revamp?node-id=6144-20409&t=2hQ6kOmiHO9XGmoT-4)

## Implementation details

- Added `CardAlertAction` interface (`label`, `onClick`, `disabled?`) to `types.ts`
- Extended `CardAlertProps` discriminated union with a new `CardAlertWithAction` branch (`action: CardAlertAction`, `dismissible?: never`, `onDismiss?: never`); existing `CardAlertDismissible` gains `action?: never` to close the other direction
- `CardAlertHeader` renders `action` button or dismiss button via a ternary — never both
- 4 new unit tests: renders action button, calls `onClick`, does not render dismiss button when action is set, respects `disabled`
- New `WithAlertAction` story with a stateful toggle and a reset button (mirrors `WithDismissibleAlert` pattern)
- `WithAlertAction` variant added to `Snapshot` story


packages/react/src/components/F0Card/types.ts
```diff
@@ -41,14 +41,39 @@ type CardAlertDismissible = CardAlertBase & {
...
+
+/**
+ * An optional action button rendered in the alert header.
+ * Mutually exclusive with `dismissible` — only one can be shown at a time.
+ */
+export interface CardAlertAction {
+  /** Label text for the action button. */
+  label: string
+  /** Called when the action button is clicked. */
+  onClick: () => void
```
sauldom102: Could we actually have `href | onClick`? I'm thinking that maybe in some cases we just need to redirect to some other page or even have some activity url

idowuarifayan: Good catch!!! Thanks @sauldom102

sauldom102 (approved):

PR #4218, opened 2026-05-21
fix(RichTextEditor): keep fullscreen inside dialogs

pol-torrent:
## Description

This PR fixes RichTextEditor fullscreen behavior when the editor is rendered inside an F0Dialog.

When fullscreen is toggled outside a dialog, RichTextEditor keeps the existing behavior and portals the fullscreen editor to document.body. When it is already inside an open F0Dialog, it now avoids that portal remount and renders the fixed fullscreen editor in place. This keeps the editor in the dialog subtree during the fullscreen transition, so non-modal side dialogs do not treat the interaction as outside focus and close unexpectedly.

Architecture note: this is intentionally scoped to RichTextEditor. We are not changing F0Dialog outside-interaction semantics, not adding a generic portal abstraction, and not constraining fullscreen to the dialog panel. The fullscreen editor remains viewport-fixed.

## Implementation details

- use F0DialogContext.open as the stable signal for whether RichTextEditor is inside an active dialog
- skip the document.body fullscreen portal only when the editor is inside an open F0Dialog
- keep real fullscreen styling with fixed inset-0
- add a regression test for fullscreen RichTextEditor inside a right-positioned F0Dialog without closing the dialog

### DIFF
--- a/packages/react/src/components/RichText/RichTextEditor/index.test.tsx
+++ b/packages/react/src/components/RichText/RichTextEditor/index.test.tsx
@@ -50,8 +50,13 @@ test("keeps fullscreen editor inside side dialog", async () => {
     await screen.findByRole("button", { name: "Toggle fullscreen mode" })
   )
 
-  expect(
-    screen.getByRole("dialog").querySelector(".rich-text-editor-container")
-  ).toBeTruthy()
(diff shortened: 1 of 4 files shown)


sauldom102 (approved):
Super, thanks!

PR #4229, opened 2026-05-22
feat(NumberInput): consolidate F0AmountCalculator — add inline context and popover modes

idowuarifayan:
## 🚪 Why?

A numeric input was needed that supports **inline contextual content** (e.g. "of 300,00 €") and **popover editing** with deferred-commit Apply (including inline/block layout choice). Rather than building a separate wrapper component, `NumberInput` now directly supports these modes via opt-in props.

### Figma Design
Popover@2x

## 🔑 What?

**`NumberInput` extended** (`packages/react/src/experimental/Forms/Fields/NumberInput/`) — three new opt-in props:

| Prop | Type | Purpose |
|------|------|---------|
| `extraContent` | `ReactNode` | Context text rendered to the right of the input (e.g. "of 300,00 €") |
| `inputWidth` | `string` | CSS width constraint when `extraContent` is present |
| `popover` | `NumberInputPopoverConfig` | Icon-button trigger that opens the input in a floating panel |

**`NumberInputPopoverConfig`** supports:
- Custom trigger icon, side, alignment — `icon`, `side`, `align`
- Controlled open state — `open` + `onOpenChange`
- Labeled/icon-only trigger — `triggerLabel`
- Commit modes — `commitMode: "immediate"` (default) or `"deferred"`
- Apply button — `apply.label`, `apply.icon`, `apply.closeOnApply`, `apply.layout: "block" | "inline"`

**Deleted**: `F0AmountCalculator` — the entire directory was removed. All tests (56), stories, and docs consolidated into `NumberInput`.

## 🏡 Context

- [💼 Jira](_jira_link_)
- [🎨 Figma](https://www.figma.com/design/SqIAXKMbJVUbwRYBvTILqW/Sidepanel-revamp?node-id=6001-16697&t=UOutyIusaMAZHumz-4)


idowuarifayan:
Quick update on the Copilot threads (4347133298 + 4357157557):

Implemented now (worthwhile / non-breaking):
- ✅ `resolveStatus` priority aligned with `InputField` (`error > hint > status`) + test coverage
- ✅ `amountCalculator` mapping now honors explicit `config.clearable` override
- ✅ removed dead `readonly` mapping in `useSchemaDefinition`
- ✅ `AmountCalculatorFieldRenderer` value normalization now matches `NumberFieldRenderer` behavior (`Number(...)` coercion + finite check)
- ✅ MDX docs updated to match current popover min width (`220px`, `min-w-[220px]`)

These are included in commit `0bf04b440`.

Not applied for now (intentional / larger API decision):
- ℹ️ `hideLabel` semantics threads: currently intentional for F0Form integration to avoid duplicate label/messages. Changing this cleanly likely needs a dedicated prop (e.g. `hideMessages`) to avoid regressions.
- ℹ️ `experimentalComponent(...)` wrapper suggestion: skipped for now to keep consistency with current implementation/export pattern in this area.
- ℹ️ `React.ReactNode` namespace warning: not reproducible (typecheck passes).
- ℹ️ broader ARIA/id typing on `NumberInput` internals: valid long-term cleanup, but larger than this PR and not currently breaking runtime behavior.

If preferred, I can open a focused follow-up PR for the two API-shape items (`hideLabel` semantics + ARIA/id typing hardening).

sauldom102 (changes requested):
We already have `f0FormField.percentage(config)`, I see useful to render the percentage icon at the end of the input to give more context to the user that that field is a percentage, but we should try to reuse what we already have. The "of 400.00€" at the end could just be added as a helpText of the field.

We should try to simplify and use what we already have instead of creating new field types 🙏🏼

sauldom102 (changes requested):
As suggestion from design, let's always have the inline version with the button on the same row and let's just drop this version

Image

sauldom102 (commented):
A couple visual things to polish here:
- We could make the input smaller in this case, users normally just write at max "200.54" (6 characters)
- The input placeholder is overlapping with the "%" icon on the right
- We could make the box around "%" fully rect (same width and height) in this case as we only have one character shown there

Image

idowuarifayan:
> As suggestion from design, let's always have the inline version with the button on the same row and let's just drop this version
> 
> Image

Makes totally sense, I was trying to cover some of the iterations presented in the figma design, but yeah, I agree with you, visually, looks better inlined

priscilaguillen (approved):
LGTM

sauldom102 (approved):

PR #4235, opened 2026-05-25
feat(PageHeader): support AI-kind actions

igorsafonov-gif:
## Description

Adds support for AI-flavored actions in `PageHeader`. Marking a `PageAction` with `kind: "ai"` renders it as the branded gradient `AIButton` instead of the default outline button — same render path the SDS uses elsewhere for AI launchers (`AskOneButton`, the One-switch entry, etc.). Tooltip auto-derives from the action `label`, matching the existing header buttons (Settings, Latest from…).

The motivating use case is launching One's analytics-dashboard flow from a module page header (e.g. Trainings → "Learn how your employees complete trainings" → opens the chat with `displayDashboard`). To make those actions visually first-class without leaving the AI icon set, this PR also adds a `ChartLine` AI icon, mirrored from the existing app-set chart-line icon used in the `F0AnalyticsDashboard` chart-type dropdown.

Storybook: **Navigation → PageHeader → With AI Dashboard Launch**

## Implementation details

- `experimental/Navigation/Header/PageHeader/index.tsx`
  - Added `kind?: "default" | "ai"` to `PageAction`.
  - Added an AI render branch at the top of the internal `PageAction` renderer: when `kind === "ai"` and the action carries an `onClick`, render `<AIButton hideLabel>`. The cast to `AIButtonProps["icon"]` is intentional — `IconType` is a superset of `AIIconType` and the type contract for AI actions is "use an icon from `@/icons/ai`".
- `experimental/Navigation/Header/PageHeader/index.stories.tsx`
  - New `WithAIDashboardLaunch` story showing the canonical pattern with the Trainings module.
- `experimental/Navigation/Header/index.mdx`
  - New short "With AI Dashboard Launch" section pointing at the story and explaining `ChartLine` / `One` icon guidance.
- New AI icon:
  - `packages/core/assets/icons/ai/chart-line.svg` — SVG source (copy of the existing app `chart-line.svg`).
  - `packages/react/src/icons/ai/ChartLine.tsx` — generated-style `.tsx` mirroring the other AI icons; barrel export added in `icons/ai/index.ts`.
  - Reviewers: feel free to re-run `pnpm generate-icons` to re-emit from the SVG source — the hand-written file should round-trip identically.

## Notes / open questions

- `PageHeader` already mounts an AI affordance via `F0OneSwitch` (the rotating One avatar). Adding a second AI button via `kind: "ai"` is intentional for the dashboard-launch use case, but worth a sanity check from the PageHeader owners.
- `PageHeader` lives under `experimental/`, so the API is still cheap to evolve if a different shape is preferred (e.g. `variant: "ai"` to match `ResourceHeader`).


packages/react/src/experimental/Navigation/Header/PageHeader/index.tsx
```diff
@@ -25,6 +26,7 @@ import { ProductUpdates, ProductUpdatesProp } from "../ProductUpdates"
 export type PageAction = {
   label: string
   icon: IconType
+  kind?: "default" | "ai"
```
siguenzaraul: Instead of introducing an ad-hoc `kind` prop, you can reuse `ActionButtonVariant` from `@/ui/Action`, which already includes `"ai"` among its variants. This keeps the type aligned with the design system:

```ts
import { ActionButtonVariant } from "@/ui/Action"

variant?: ActionButtonVariant
```

packages/react/src/experimental/Navigation/Header/PageHeader/index.tsx
```diff
@@ -296,6 +298,18 @@ function PageAction({ action }: { action: PageAction }): ReactElement {
...
+      <AIButton
+        size="md"
+        label={action.label}
+        icon={action.icon as AIButtonProps["icon"]}
+        hideLabel
+        onClick={action.onClick}
+      />
+    )
+  }
+
```
siguenzaraul: You don't need a special case for `AIButton` here. You can remove this entire block — see the comment below on `ButtonInternal`.

packages/react/src/experimental/Navigation/Header/PageHeader/index.tsx
```diff
@@ -296,6 +298,18 @@ function PageAction({ action }: { action: PageAction }): ReactElement {
...
+  if (action.kind === "ai" && "onClick" in action) {
+    return (
+      <AIButton
+        size="md"
+        label={action.label}
+        icon={action.icon as AIButtonProps["icon"]}
+        hideLabel
+        onClick={action.onClick}
+      />
+    )
```
siguenzaraul: `F0Button` doesn't expose the `"ai"` variant, but `ButtonInternal` does — and it's the exact same button under the hood. If you switch to `ButtonInternal` and pass the variant through, the AI styling works automatically without needing the `AIButton` special case above:

```tsx
<ButtonInternal
  size="md"
  variant={action.variant ?? "outline"}
  label={action.label}
  icon={action.icon}
  hideLabel
/>
```

This applies to all three places where `F0Button` is used in this component (dropdown trigger, onClick handler, and href link).

packages/react/src/experimental/Navigation/Header/PageHeader/index.stories.tsx
```diff
@@ -100,6 +101,36 @@ export const WithActions: Story = {
...
+    module: {
+      name: "Trainings",
+      href: "/trainings",
+      id: "company_trainings" as const,
+    },
+    actions: [
+      {
+        label: "Settings",
+        icon: Settings,
+        href: "/trainings/settings",
```
siguenzaraul, line 116: If you apply the `kind` → `variant` rename, update this too:

```ts
variant: "ai",
```

siguenzaraul (approved):

PR #4248, opened 2026-05-27
docs: add example FeatureGrid to CRUD pattern operation pages

warcos-fact:
## Why

In the public Storybook build, `docsMode` collapses each story file into a single Documentation page and hides individual story entries from the sidebar. That means the per-operation CRUD pattern pages (Create, Read, Update, Delete) no longer expose their individual scenario stories — users land on a single Documentation page with no visual entry point into each example.

## What

Adds an `## Examples` section near the top of each per-operation page with a `FeatureGrid` linking to every scenario story, mirroring the pattern already used in `overview.mdx`.

### Per-operation grids

- **Create** — 7 cards (Table / List default + right dialog, Wizard, Kanban Lane, Editable Table Add Row)
- **Read** — 6 cards (Table default / dialog-to-page / page, List to page, Card / Kanban to dialog)
- **Update** — 7 cards (Table / List default + right dialog, Bulk variants, Card actions, Editable Table inline)
- **Delete** — 5 cards (Table single / bulk, Card, Kanban Bulk, Editable Table)

### Other tweaks

- Wrap each `FeatureGrid` in `<Unstyled>` so the inherited docs-theme link underline doesn't bleed into card text (matches the overview page).
- Trim the operation verb from card titles (e.g. `Table Create — Default` → `Table — Default`) since the page heading already scopes them.
- Replace the trailing `Related stories` list with a slim `Related documentation` section containing only cross-doc links (CRUD overview + agnostic principles).

### Out of scope

- `overview.mdx` already uses `FeatureGrid` correctly — untouched.
- `by-view.stories.tsx` keeps `tags: ["!autodocs"]`; overview already covers it.
- No changes to `src/experimental/CrudPatterns/`.

### DIFF
--- a/packages/react/src/patterns/OneDataCollection/__stories__/crud-patterns/delete/delete.mdx
+++ b/packages/react/src/patterns/OneDataCollection/__stories__/crud-patterns/delete/delete.mdx
@@ -19,13 +19,13 @@ step must make the scope and consequence explicit before the user commits.
     features={[
       {
         icon: Table,
-        title: "Table Single",
+        title: "Table — Single",
         description: "Item delete action triggers a Confirmation Dialog.",
         href: "/?path=/story/patterns-data-collection-crud-patterns-delete--table-delete-single",
       },
       {
         icon: Table,
-        title: "Table Bulk",
+        title: "Table — Bulk",
         description:
           "Selection and bulk action trigger a scoped Confirmation Dialog.",
         href: "/?path=/story/patterns-data-collection-crud-patterns-delete--table-delete-bulk",
@@ -38,7 +38,7 @@ step must make the scope and consequence explicit before the user commits.
       },
       {
         icon: Kanban,
-        title: "Kanban Bulk",
+        title: "Kanban — Bulk",
         description:
           "Selected Kanban cards use a scoped destructive Confirmation Dialog.",
         href: "/?path=/story/patterns-data-collection-crud-patterns-delete--kanban-delete-bulk",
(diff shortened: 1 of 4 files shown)


packages/react/src/patterns/OneDataCollection/__stories__/crud-patterns/delete/delete.mdx
```diff
@@ -10,6 +12,48 @@ import * as Stories from "./delete.stories"
...
+      {
+        icon: Kanban,
+        title: "Kanban Bulk",
+        description:
+          "Selected Kanban cards use a scoped destructive Confirmation Dialog.",
+        href: "/?path=/story/patterns-data-collection-crud-patterns-delete--kanban-delete-bulk",
+      },
+      {
+        icon: Pencil,
+        title: "Editable Table",
```
albertpmz: Tiny consistency nit on the card titles here: the `delete` grid uses bare titles (`Table Single`, `Table Bulk`, `Kanban Bulk`), whereas the `create`/`read`/`update` grids use the em-dash scoping convention (`Table — Default`, `Table — Right Dialog`, etc.).

Could we align these with the other three pages for visual consistency across the operation pages? e.g. `Table — Single`, `Table — Bulk`, `Kanban — Bulk`. Purely cosmetic, not blocking. 🙂

warcos-fact: Good catch — aligned in 507667f. Applied the `View — Variant` em-dash convention across all three pages (Create, Update, Delete) where the title has a `<View> <Modifier>` shape: `Table — Single`, `Table — Bulk`, `Kanban — Bulk`, `Kanban — Lane`, `Card — Actions`, `Editable Table — Add Row`, `Editable Table — Inline`.

Kept `Card` and `Editable Table` bare on the delete page (no variant to scope), and left `read.mdx` alone since its titles (`Table to Page`, `Card to Dialog`, …) read as natural noun phrases that an em-dash would break.

albertpmz (approved):
LGTM 🚀

albertpmz (approved):
Title formatting now consistent across all four operation pages — thanks for the extra cleanup in create/update too. LGTM 🚀

rubenmoya (approved):

PR #4259, opened 2026-05-28
docs(crud-patterns): add Right Dialog read variant for Table and List

warcos-fact:
## What

Adds the missing stay-and-read Right Dialog variant to the Read pattern docs for Table and List. Until now the Read page only surfaced Right Dialog as the preview step of the Dialog to Page flow, even though the CRUD overview already listed Right Dialog as a valid Read option when the user benefits from keeping the collection visible.

## Why

The Read documentation was incomplete vs the agnostic CRUD principles. A user reading a row or list item often benefits from keeping the rest of the collection in view (comparison, scanning, context). The existing Default Dialog covers the focused case, and Dialog to Page covers the preview-then-promote case, but the in-between — Right Dialog as the final read destination — was missing.

## Changes

- New stories in `read.stories.tsx`:
  - `TableReadRightDialog` — Right Dialog with no `Open Details` action; dialog is the final destination.
  - `ListReadRightDialog` — same pattern with the List visualization.
  - Both backed by a standalone `RightDialogScenario` (no shared state-machine refactor with `RightDialogToPageScenario`).
- `read.mdx` updates:
  - Two new FeatureGrid cards (`Table — Right Dialog`, `List — Right Dialog`).
  - New When-to-use bullet covering keep-collection-visible reads.
  - New Pattern options row for Right Dialog, between Default Dialog and Dialog to Page.
  - New Decision criteria row ("Does reading benefit from keeping nearby rows or items visible?" → Right Dialog).
  - Guidance paragraph clarifying that Right Dialog can be a final read destination, not only a preview.
- `overview.mdx`: two new Read scenarios so the table mirrors the operation page.

## Scope

- Documentation + Storybook stories only. No production code changes.
- Card, Kanban, and Editable Table reads are unchanged.
- Follow-up to #4248.

## Verification

- `pnpm --filter @factorialco/f0-react run format` — clean.
- `pnpm --filter @factorialco/f0-react run tsc` — only pre-existing `F0Graph` errors (unrelated; missing `@xyflow/react` types).

### DIFF
--- a/packages/react/src/patterns/OneDataCollection/__stories__/crud-patterns/overview.mdx
+++ b/packages/react/src/patterns/OneDataCollection/__stories__/crud-patterns/overview.mdx
@@ -351,6 +351,20 @@ to a page when the resource needs a durable destination.
         <td>
           <a
             target="_parent"
+            href="/?path=/story/patterns-data-collection-crud-patterns-read--table-read-right-dialog"
+          >
+            Table Read — Right Dialog
+          </a>
+        </td>
+        <td>
+          Clicking a row opens a Right Dialog while the collection stays
+          visible.
+        </td>
+      </tr>
+      <tr>
+        <td>
+          <a
+            target="_parent"
             href="/?path=/story/patterns-data-collection-crud-patterns-read--table-read-right-dialog-to-page"
           >
             Table Read — Dialog to Page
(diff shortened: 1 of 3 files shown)


warcos-fact:
Closing to make further modifications before re-opening.

PR #4267, opened 2026-05-29
fix(f0compose): disable React.StrictMode in dev

amadeu-thomson:
## Summary

- Removes `<React.StrictMode>` from `packages/f0compose/src/main.tsx` so localhost no longer crashes with "Maximum update depth exceeded"
- Experimental f0-react components (selectors, OneDataCollection lanes, F0AiChat) re-enter setState during StrictMode's double-invocation in dev. Vercel prod is unaffected because production React doesn't double-invoke.
- Adds an `AGENTS.md` note so future agents don't re-add it as boilerplate
- Unblocks teammates running f0compose prototypes locally

## Scope

Only `packages/f0compose/src/main.tsx` + `packages/f0compose/AGENTS.md`. No prod / build / Vercel impact.

## Test plan

- `pnpm dev` from `packages/f0compose/` → catalog + every prototype route loads without the strict-mode crash
- Vercel deploys unchanged (strict mode is dev-only)


amadeu-thomson:
Closing — packages/f0compose doesn't exist on main yet, so this PR can't ship just the StrictMode fix in isolation. Will revisit once f0compose lands on main.

PR #4274, opened 2026-05-31
feat(ResourceHeader): leading icon on metadata items + showBottomBorder

alexander-j-lyakhov:
## Description

Two small, additive, backward-compatible capabilities so pages can show an icon-prefixed
key/value row in the resource header **through the structured `metadata` API** (no custom-content
props):

- **`MetadataItem.icon?: IconType`** — an optional leading icon rendered before the label/value.
  Combined with `hideLabel`, an item reads as **"icon + value"** (e.g. a category or delivery
  icon). It's always visible (independent of `hideLabel`) and inherits the secondary label tone.
- **`ResourceHeader.showBottomBorder?: boolean`** — a 1px bottom border at the very bottom of the
  header, as a separator from the page content.

## Context — what changed since the first revision

The first revision added a `belowTitle?: ReactNode` slot. Per @sauldom102's review ("avoid props
accepting custom content, use metadata"), that's been **dropped** in favour of the structured
`icon` field above — the actual gap was that `metadata` items couldn't render a leading icon, not
that we needed arbitrary content. Everything our consumer needs (icon+value, label+value) is now
expressible via typed `metadata`.

## Implementation details

- `Metadata` (`MetadataItem`): added optional `icon`; rendered as the first child of the item row
  inside a secondary-toned wrapper so `F0Icon`'s `currentColor` picks up the label tone. No change
  to existing items (icon is opt-in) or to `hideLabel` behaviour.
- `ResourceHeader` / `BaseHeader`: `showBottomBorder` applies `border-b border-solid border-f1-border`
  via `cn(...)`; `belowTitle` removed.
- Story: `WithIconMetadataAndBorder` (icon metadata + bottom border).
- `tsc`, `oxlint`, and the Metadata/ResourceHeader unit tests pass; `oxfmt` applied.

## Consumer

factorialco/factorial#100343 (IT device-catalog detail page). It currently renders the params row
at the monorepo wrapper level; once this lands and `@factorialco/f0-react` is bumped, it will pass
icon-bearing `metadata` items straight to `ResourceHeader`.

### DIFF
--- a/packages/react/src/patterns/ResourceHeader/index.stories.tsx
+++ b/packages/react/src/patterns/ResourceHeader/index.stories.tsx
@@ -529,6 +529,26 @@ export const TeamHeader: Story = {
   },
 }
 
+export const WithContentBelowTitle: Story = {
+  tags: ["!dev"],
+  args: {
+    title: 'MacBook Pro 14" M5',
+    showBottomBorder: true,
+    belowTitle: (
+      <div className="flex flex-wrap items-center gap-x-6 gap-y-2 text-f1-foreground-secondary">
+        <span>Laptops</span>
+        <span>
+          Brand <span className="font-semibold text-f1-foreground">Apple</span>
+        </span>
+        <span>24h delivery</span>
+        <span>
+          SKU <span className="font-semibold text-f1-foreground">000017</span>
+        </span>
+      </div>
+    ),
+  },
+}
+
 export const WithLongDescription: Story = {
   tags: ["!dev"],
   args: {
--- a/packages/react/src/patterns/ResourceHeader/index.stories.tsx
+++ b/packages/react/src/patterns/ResourceHeader/index.stories.tsx
@@ -529,23 +529,33 @@ export const TeamHeader: Story = {
   },
 }
 
-export const WithContentBelowTitle: Story = {
+export const WithIconMetadataAndBorder: Story = {
   tags: ["!dev"],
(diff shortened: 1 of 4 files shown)


packages/react/src/patterns/ResourceHeader/index.stories.tsx
```diff
@@ -529,6 +529,26 @@ export const TeamHeader: Story = {
...
+        <span>
+          Brand <span className="font-semibold text-f1-foreground">Apple</span>
+        </span>
+        <span>24h delivery</span>
+        <span>
+          SKU <span className="font-semibold text-f1-foreground">000017</span>
+        </span>
+      </div>
+    ),
+  },
```
sauldom102: Why do you need this? Why not using metadata? We should avoid having props accepting such custom content

alexander-j-lyakhov: Good call — dropped the `belowTitle` slot. The real gap was that `metadata` items couldn't render a leading icon, so I added an optional `icon` to `MetadataItem` instead and kept everything in the typed `metadata` API. The consumer row is now plain metadata: `icon + value` (via `hideLabel`) for category/delivery, `label + value` for brand/SKU. `showBottomBorder` stays as a structured boolean. Reworked in 00c1bc1.

sauldom102 (approved):

PR #4282, opened 2026-06-01
feat(F0Sidepanel): add experimental side-anchored panel component

idowuarifayan:
## Description

Introduce **F0Sidepanel** in `experimental/Overlays/`, backed by a new internal **Sheet** primitive in `ui/Sheet/`. This is the f0 port of the existing Factorial monorepo `Sidepanel` component (`modules/core/components/Sidepanel`) plus the new affordances from the Sidepanel revamp Figma spec.

**Why this component exists:**

- The current `Sidepanel` lives in the Factorial monorepo, depends on `@factorialco/deprecated-design-system`, and uses `framer-motion` for slide animations. As consumers move to f0 there is no equivalent — they have to either keep the legacy import (blocks the design-system migration) or fall back to `F0Dialog` with `position="right"`, which renders as a centered-modal-pinned-to-the-edge and doesn't match the Sidepanel UX.
- The Sidepanel revamp design adds a new header pattern (close on the left, page-stepping navigation on the right) that no existing f0 component supports. Wrapping it as a Radix-based primitive lets the AppHeader's `navigation` shape be reused directly.
- Building on the `ui/Sheet/` primitive (modelled on `ui/Dialog/`) keeps animations Tailwind-only — no `framer-motion` dependency in the f0 surface — and unblocks any future "slide-from-edge" need without bespoke code.

**API parity + extensions:**

| Legacy `Sidepanel` prop | F0Sidepanel | Notes |
|---|---|---|
| `open`, `onClose`, `onCloseClicked` | ✓ | Preserves the 200ms exit-animation delay before `onClose` fires (legacy callers depend on this for post-close cleanup) |
| `title`, `closeLabel` | ✓ | `title` accepts string or ReactNode; `null` skips the title slot per the Figma layout |
| `width` (narrow / wide / wideXL / extended / full) | ✓ | Same names; `wide = 600px` matches Figma spec |
| `overlay`, `locked`, `allowBackgroundInteraction`, `headerBorder` | ✓ | |
| `footer`, `boxPadding` | ✓ | |
| `options` (kebab actions menu) | ✓ | Backed by `experimental/Navigation/Dropdown` |
| — | `navigation` *(new)* | Page-stepping: `{ counter, previous, next }` with `url \| onClick` step variants. Same shape as `experimental/Navigation/Header/PageHeader.navigation` for consistency |
| — | `side` *(new)* | `"left" \| "right"`, default `"right"`. Constrained to side-anchored placements |

**A11y:**

- Always renders a Radix `DialogTitle` (wrapped in `VisuallyHidden` from `@radix-ui/react-visually-hidden` when the visual title slot is empty), to satisfy Radix's screen-reader contract.
- Sets `onOpenAutoFocus={(e) => e.preventDefault()}` (same pattern `F0Dialog` uses) — avoids jarring focus shifts on open and sidesteps a jsdom focus-recursion issue when testing with Radix's `FocusScope`.

## Screenshots
Screenshot 2026-06-01 at 23 17 51

https://github.com/user-attachments/assets/c0aa697e-0718-4964-81d9-90aabd3a4730

[Link to Figma Design](https://www.figma.com/design/SqIAXKMbJVUbwRYBvTILqW/Sidepanel-revamp?node-id=6216-44093&t=8g7MQoNDlNnx9Idl-4)

Storybook coverage: `Default`, `LeftSide`, `WithTitle`, `WithFooter`, `WithNavigation`, **`WithLiveNavigation`** (steps through a 5-employee in-memory list — buttons auto-disable at boundaries), `WithOptions`, `Locked`, `WideXL`, `Extended`, `NoOverlay`, `Floating`, `NoBoxPadding`.

## Implementation details

**New files** (under `packages/react/src/`):

- `ui/Sheet/` — shadcn-style Radix Dialog wrapper with `top|right|bottom|left` side variants, toggleable overlay, Tailwind data-state slide animations. Mirrors the per-file structure of `ui/Dialog/`. Internal-only per AGENTS.md (`ui/*` is not part of the public f0 surface).
- `experimental/Overlays/F0Sidepanel/` — pattern component:
  - `F0Sidepanel.tsx` + `F0SidepanelInternal.tsx` — public wrapper + implementation
  - `types.ts` — `F0SidepanelProps`, `SidepanelSide`, `SidepanelWidth`, `SidepanelNavigation`, `NavigationStep`, `SidepanelOption` with `as const` arrays
  - `components/F0SidepanelHeader.tsx` — header (close, optional title, navigation, options menu)
  - `components/F0SidepanelNavigation.tsx` — counter + chevrons using `F0Button` with `ChevronUp`/`ChevronDown` (mirrors `PageHeader` layout)
  - `components/F0SidepanelFooter.tsx` — right-aligned footer with top border
  - `__stories__/F0Sidepanel.stories.tsx` — 13 stories including the live-navigation walkthrough
  - `__tests__/F0Sidepanel.test.tsx` — 12 Vitest tests covering mount/unmount, close timing (sync `onCloseClicked` + delayed `onClose`), `locked` behavior on Esc / outside-click / X, navigation callbacks + boundary disable, options menu open + invoke, title rendering

**Export wiring:**

- `experimental/Overlays/F0Sidepanel/index.tsx` wraps with `experimentalComponent("F0Sidepanel", …)` + `withDataTestId(…)`, matching the `Tooltip` pattern.
- Re-exported from `experimental/Overlays/exports.tsx`.

**Width mapping:**

| Variant | Width |
|---|---|
| `narrow` | `max-w-[360px]` |
| `wide` | `max-w-[600px]` (Figma spec) |
| `wideXL` | `max-w-[800px]` |
| `extended` | `max-w-[95vw]` |
| `full` | `max-w-full` |

**Validation:**

- `pnpm format` — clean
- `pnpm tsc` — no errors
- `oxlint` on changed files — 0 warnings / 0 errors
- `pnpm vitest run` against the new spec — 12/12 pass

**Follow-ups (out of scope for this PR):**

- A future `useResourceNavigation<T>` hook could absorb the prev/next/counter wiring in callers — proposed in the integration notes for the compensations team once the cursor-pagination question is settled.
- Foundations team may want to revisit the elevation token: f0's `shadow-lg` is intentionally subtle compared to Figma's `s400` (3-layer drop shadow). Not blocking the experimental ship.

### DIFF
--- a/packages/react/src/experimental/Overlays/F0Sidepanel/F0Sidepanel.tsx
+++ b/packages/react/src/experimental/Overlays/F0Sidepanel/F0Sidepanel.tsx
@@ -1,0 +1,8 @@
+import { F0SidepanelInternal } from "./F0SidepanelInternal"
+import { F0SidepanelProps } from "./types"
+
+export const F0Sidepanel = (props: F0SidepanelProps) => (
+  <F0SidepanelInternal {...props} />
+)
+
+F0Sidepanel.displayName = "F0Sidepanel"
--- a/packages/react/src/experimental/Overlays/F0Sidepanel/F0SidepanelInternal.tsx
+++ b/packages/react/src/experimental/Overlays/F0Sidepanel/F0SidepanelInternal.tsx
@@ -31,6 +31,7 @@ export const F0SidepanelInternal = ({
   locked = false,
   allowBackgroundInteraction = false,
   headerBorder = false,
+  loading = false,
   options,
   navigation,
   children,
@@ -96,6 +97,7 @@ export const F0SidepanelInternal = ({
           navigation={navigation}
           options={options}
           headerBorder={headerBorder}
+          loading={loading}
         />
 
         <div
--- a/packages/react/src/experimental/Overlays/F0Sidepanel/F0SidepanelInternal.tsx
+++ b/packages/react/src/experimental/Overlays/F0Sidepanel/F0SidepanelInternal.tsx
@@ -39,6 +39,12 @@ export const F0SidepanelInternal = ({
   const closeTimeoutRef = useRef<ReturnType<typeof setTimeout> | null>(null)
 
   useEffect(() => {
+    // If the parent re-opens us before the exit-animation timeout has fired,
+    // cancel the pending onClose so it doesn't run while the panel is visible.
+    if (open && closeTimeoutRef.current) {
+      clearTimeout(closeTimeoutRef.current)
+      closeTimeoutRef.current = null
+    }
     setIsOpen(open)
   }, [open])
 
--- a/packages/react/src/experimental/Overlays/F0Sidepanel/F0SidepanelInternal.tsx
+++ b/packages/react/src/experimental/Overlays/F0Sidepanel/F0SidepanelInternal.tsx
@@ -31,7 +31,6 @@ export const F0SidepanelInternal = ({
   locked = false,
   allowBackgroundInteraction = false,
   headerBorder = false,
-  shadow = true,
   options,
   navigation,
   children,
@@ -75,7 +74,6 @@ export const F0SidepanelInternal = ({
       <SheetContent
         side={side}
         withOverlay={overlay && !allowBackgroundInteraction}
-        withShadow={shadow}
         className={cn(widthClasses[width], "rounded-none")}
         onOpenAutoFocus={(event) => event.preventDefault()}
         onInteractOutside={(event) => {
--- a/packages/react/src/experimental/Overlays/F0Sidepanel/F0SidepanelInternal.tsx
+++ b/packages/react/src/experimental/Overlays/F0Sidepanel/F0SidepanelInternal.tsx
@@ -1,0 +1,110 @@
+import { useCallback, useEffect, useRef, useState } from "react"
+
+import { cn } from "@/lib/utils"
+import { Sheet, SheetContent } from "@/ui/Sheet/sheet"
+
+import { F0SidepanelFooter } from "./components/F0SidepanelFooter"
+import { F0SidepanelHeader } from "./components/F0SidepanelHeader"
+import { F0SidepanelProps, SidepanelWidth } from "./types"
+
+const EXIT_DURATION_MS = 200
+
+const widthClasses: Record<SidepanelWidth, string> = {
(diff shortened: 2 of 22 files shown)


idowuarifayan:
Addressed all three Copilot review comments in `ca7110f9e`:

**1. Stale exit-animation timeout** — When the parent flips `open` back to `true` within the 200ms exit window (e.g. quick toggle after `onCloseClicked`), the pending `onClose()` would still fire while the panel is visible again. Now cleared in the `[open]` effect when `open` returns to `true` while a timeout is pending. Added a regression test that toggles `open` `true → false → true` inside the window and asserts `onClose` stays uncalled.

**2 & 3. `??` → `||` for navigation title fallback** — An empty string title (`""`) was passing through `??` as a valid label, producing an empty accessible name (and a runtime warning from F0Button's label enforcer). Switched to `||` to match `PageHeader`'s navigation behavior.

**Bonus — i18n the default labels** — While in the file, replaced the hardcoded `"Previous"` / `"Next"` strings with `useI18n().navigation.previous` / `.next` (the keys already exist in `defaultTranslations`, used by `OneCalendar`). No new i18n entries needed.

Tests: 13/13 pass (12 previous + the new re-open regression test).

idowuarifayan:
Closing this in favor of `F0Dialog` with `position="right"`, per the Foundations team's recommendation.

`F0Dialog` covers the core API surface (controlled open/close, title, content, options menu, mobile bottom-sheet adaptation) and is a more consolidated home for right-anchored panels in f0.

What we lose by closing:
- **Page navigation** (1/N counter + previous/next chevrons) — the Figma revamp's main new pattern. Not currently in `F0Dialog`. To be revisited as a follow-up against F0Dialog if/when the team decides to extend it.
- **`onCloseClicked` with 200ms exit-delay before `onClose`** — legacy timing contract from the monorepo `Sidepanel`. Consumers will need to migrate to `F0Dialog`'s synchronous `onClose`.
- Some minor knobs (`locked`, `headerBorder`, `overlay` toggle independent of modality, `loading` skeletons).

The branch (`feat/f0-sidepanel`) is staying around in case any of this gets ported to `F0Dialog` later — in particular the navigation component, the i18n choices, and the close-timing logic are reusable.

PR #4286, opened 2026-06-02
fix: granularityDefinitions type

rubenmoya:
## Description

While `granularityDefinitions` in Factorial there is a type error when trying to access to any of the properties since it says it can be `undefined`.

This happens because the variable is typed as `Record<string, GranularityDefinition>`, so the `as const` does nothing.

## Screenshots

Before:
image

After:
image

## Implementation details

I used `satisfies Record<string, GranularityDefinition>`, instead, which not only makes the `as const` work but also prevents users from using wrong types in the object.


eliseo-juan (approved):
LGTM

arnaub (approved):

PR #4309, opened 2026-06-03
fix(TableOfContent): use background-hover for active item and reduce padding without icon

oskar-hernandez:
## Summary

- Active item now uses `bg-f1-background-hover` instead of `bg-f1-background-selected` for a more subtle selection indicator
- Left padding when no icon is present changed from `pl-0` to `pl-0.5` (2px) for better text alignment

## Test plan

- [ ] Open Storybook → Experimental > Navigation > TableOfContent
- [ ] Verify active item shows `background-hover` color
- [ ] Verify items without icons have 2px left padding
- [ ] Verify items with icons still have `pl-7` left padding


siguenzaraul (changes requested):
to long, split the PR into two

siguenzaraul (approved):

PR #4313, opened 2026-06-03
feat(F0Dialog): resource-aware header variant + opt-in metadata row gap

ReddyWish:
## Description

Adds an opt-in, resource-aware variant to `F0Dialog` so side panels (e.g. an ATS
application preview) can show a resource identity band — avatar, title, description,
status, metadata — with the dialog owning its top chrome (expand / back / prev-next /
kebab / close), instead of composing that chrome inside the dialog body.

The header is modeled as **independent layers** rather than one coupled flag —
controls row ⟂ identity band ⟂ tabs ⟂ footer. Every addition is **optional and
defaults to current behavior**: existing `F0Dialog`, `ResourceHeader`, `BaseHeader`,
`Metadata`, and `PageHeader` consumers render unchanged.

## Implementation details

**F0Dialog — controls + resource-aware header (`patterns/F0Dialog`)**
- `DialogControls` discriminated union for the left control slot:
  - `{ kind: "resource", expand?, navigation? }` — top-level: "Open detail" + prev/next
  - `{ kind: "back", label, onClick }` — drilled-in sub-page: a single back affordance

  The two navigation modes are mutually exclusive, so illegal combinations (e.g. back
  *and* prev/next at once) are unrepresentable rather than something the component has
  to defensively reconcile.
- Decoupled the identity band from the controls: the resource layout is entered on
  `resourceHeader || controls`, and `BaseHeader` renders **only** when `resourceHeader`
  is set. A visually-hidden `DialogTitle` (from `resourceHeader.title`, falling back to
  `title`) keeps an accessible name. Identity band, controls, `tabs`, and footer
  (`primaryAction` / `secondaryAction`) each compose independently.
- `otherActions` (kebab) stays a **separate, orthogonal** prop. It answers a different
  question than `controls` ("what can I do to this resource?" vs. "what navigation mode
  am I in?"), is resource-scoped, and remains available in *any* control mode — main
  view and drilled-in sub-page alike — so it isn't folded into the union.

**Metadata row gap (`Metadata`, `BaseHeader`, `ResourceHeader`)**
- Opt-in `rowGap` / `metadataRowGap`: `none` (default, == prior `gap-y-0`) | `xs` |
  `sm` | `md` (12px). Threaded ResourceHeader → BaseHeader → Metadata.

**PageNavigation extraction (`Navigation/Header`)**
- Moved the inline prev/next/counter UI + `NavigationProps` type out of `PageHeader`
  into a shared `PageNavigation` component (exported from `Navigation/Header`, reused by
  `PageHeader` and the dialog header). Behavior identical.

## Why this shape (the flexibility it buys)

1. **Any combination is expressible** — identity band with or without controls, controls
   with or without an identity band, tabs with neither. The main view and a sub-page are
   just different selections over the same orthogonal slots.
2. **The drill-down "back" pattern is a first-class dialog affordance** — any `F0Dialog`
   consumer gets it for free, instead of hand-rolling a back button inside body content.
3. **"Open detail" survives as a persistent escape hatch** at the top level without
   entangling it with back.
4. **Adding a new navigation mode is additive** — a union arm + a switch case, with no
   change to consumers of other modes.
5. **The contract documents intent** — the prop shape says "resource mode, these tabs,
   this identity, these actions," instead of a pile of optionals a reader must mentally
   reconcile.

## Note on the API surface check

The `ResourceHeader` / `ResourceHeaderProps` "breaking change" flags are **false
positives**. The only change there is the new **optional** `metadataRowGap` prop
(default `"none"`). Existing consumers compile unchanged — nobody is forced to pass it.

The check trips because `ResourceHeaderProps` is a `Pick<ComponentProps<typeof
BaseHeader>, …>`, and `"metadataRowGap"` added to that key list; the structural
differ can't see inside the `Pick` to tell the added member is optional, so it
conservatively reports the parameter type as "changed." This is an **additive, minor**
change — not a major bump.

Screenshot 2026-06-03 at 22 06 22

https://github.com/user-attachments/assets/6fafee52-0b5b-40d8-8130-0c73887a05f6

### DIFF
--- a/packages/react/src/experimental/Information/Headers/BaseHeader/index.tsx
+++ b/packages/react/src/experimental/Information/Headers/BaseHeader/index.tsx
@@ -57,6 +57,7 @@ interface BaseHeaderProps {
     actions?: MetadataAction[]
   }
   metadata?: MetadataProps["items"]
+  metadataRowGap?: MetadataProps["rowGap"]
   /** Renders a 1px bottom border at the very bottom of the header. */
   showBottomBorder?: boolean
 }
@@ -74,6 +75,7 @@ export function BaseHeader({
   otherActions = [],
   status,
   metadata = [],
+  metadataRowGap = "none",
   showBottomBorder = false,
 }: BaseHeaderProps) {
   const allMetadata: BaseHeaderProps["metadata"] = [
@@ -165,7 +167,7 @@ export function BaseHeader({
 
         {allMetadata.length > 0 && (
           <div className="flex flex-wrap items-center gap-x-3 gap-y-1 md:hidden">
-            <Metadata items={allMetadata} />
+            <Metadata items={allMetadata} rowGap={metadataRowGap} />
           </div>
         )}
 
@@ -307,7 +309,7 @@ export function BaseHeader({
       </div>
       {allMetadata.length > 0 && (
         <div className="hidden flex-wrap items-center gap-x-3 gap-y-1 md:block">
-          <Metadata items={allMetadata} />
+          <Metadata items={allMetadata} rowGap={metadataRowGap} />
         </div>
       )}
     </div>
--- a/packages/react/src/experimental/Information/Headers/Metadata/index.tsx
+++ b/packages/react/src/experimental/Information/Headers/Metadata/index.tsx
@@ -107,6 +107,15 @@ interface MetadataItem {
   }
 }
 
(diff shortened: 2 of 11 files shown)


sauldom102 (approved):

sauldom102 (changes requested):
Oh wait, there are some breaking changes

sauldom102 (approved):

PR #4322, opened 2026-06-04
ci(react): warn on breaking public API changes vs main

sauldom102:
## Why

It's currently too easy to ship a breaking change to the public API of `@factorialco/f0-react` without noticing — e.g. #4199 renamed components (`Input` → `F0TextInput`), and a follow-up regression silently degraded prop types to `any`. Neither was caught in CI. This PR adds a **non-blocking** safeguard that detects breaking public-API changes on every PR, comments on the PR, and labels it.

## What counts as breaking (precise, member-level)

The check compares the PR's public API surface against `main` across all three shipped entries (`f0`, `experimental`, `ai`), classifying each export **structurally**:

| Change | Result |
| --- | --- |
| Export removed / renamed | ❌ breaking |
| Property/parameter removed or retyped | ❌ breaking |
| New **required** property/parameter | ❌ breaking |
| Optional → required | ❌ breaking |
| Kind or type-parameters changed | ❌ breaking |
| New export (component/type) | ✅ safe |
| New **optional** property/parameter | ✅ safe |
| Required → optional (widening) | ✅ safe |

So adding an optional prop or a whole new component does **not** warn; only genuinely consumer-breaking changes do.

## How it works

- **`.github/workflows/api-surface.yaml`** — on PRs touching the React/core package, two parallel jobs build the rolled-up `.d.ts` on the PR head and on `main` and upload them as artifacts (the base build uses the existing `build` script, so it works even though this workflow doesn't exist on `main` yet). A diff job runs the analyzer, then:
  - posts/updates a PR comment via the existing `add-or-update-pr-comment` action,
  - emits a `::warning::` annotation, and
  - adds/removes a **`breaking changes`** label on the PR.
  - It **never blocks merge.**
- **`packages/react/.scripts/check-api-surface.ts`** — snapshots each entry's public exports via the **TypeScript compiler API** (resolving structural types, so api-extractor rollup noise like `*_2` suffixes / dangling `./types` imports and primitive prototype methods are normalized away), then compares member-by-member. Both sides are analyzed by the same TS in one process → deterministic. Run locally: `pnpm --filter @factorialco/f0-react check:api-surface --base <dir> --head <dir>`.

## Tests

`packages/react/.scripts/__tests__/check-api-surface.test.ts` — 14 cases that build real `.d.ts` fixtures and run the analyzer end-to-end, asserting **both directions**:
- **Caught:** removed export, renamed export, removed prop, retyped prop, added-required prop, optional→required, changed param type, removed param.
- **Not flagged:** identical surface, brand-new export, new optional prop, required→optional, new optional param.
- Plus a cross-directory determinism check.

All 14 pass and are discovered by the `unit` vitest project (runs in the existing `test.yaml`).

### DIFF
--- a/.github/workflows/api-surface.yaml
+++ b/.github/workflows/api-surface.yaml
@@ -72,6 +72,7 @@ jobs:
     permissions:
       contents: read
       pull-requests: write
+      issues: write
     steps:
       - uses: actions/checkout@v4
       - uses: pnpm/action-setup@v4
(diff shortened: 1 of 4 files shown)


eliseo-juan (approved):
lgtm

PR #4337, opened 2026-06-05
fix(SurveyFormBuilder): stop title caret from jumping to end on every keystroke

rubenmoya:
## Summary

- Caret jumps to the end of question title/description textareas on every keystroke in `SurveyFormBuilder`, making mid-string editing impossible. Reproduces in every consumer (e.g. Factorial's `newFormEditor` and `procurement/FormEditor`).
- Root cause is inside f0: `flattenElements` produces new `FlatFormItem` literals every call, so `Reorder.Item value={item}` gets a fresh identity every render. framer-motion v12's `Reorder` keys on `value` reference equality, so it tears down each Item's DOM — recreating the `<textarea>` and re-firing `BaseQuestion`'s mount focus effect, which `.focus()`es the new node and snaps the caret to the end.
- Fix in two parts:
  - New `stabilizeFlatItems(prev, next)` helper in `Form/utils.ts` reuses the previous render's `FlatFormItem` reference whenever the same-id item is deep-equal. Wired through `Form/index.tsx`'s flatItems memo so `Reorder.Item` now gets a stable `value` for unchanged items → no DOM teardown → caret stays put.
  - `BaseQuestion`'s `shouldFocusTitleOnMountRef` is replaced with a one-shot `pendingFocusRef` that flips to `false` after the first focus — belt-and-suspenders against StrictMode double-invocations and any future upstream identity churn.

## What changed

| File | Change |
|---|---|
| `Form/utils.ts` | new `stabilizeFlatItems(prev, next): { items, map }` helper |
| `Form/index.tsx` | thread a `useRef<Map>` through the existing flatItems memo via `stabilizeFlatItems` |
| `QuestionTypes/BaseQuestion/index.tsx` | one-shot guard on the mount-focus effect |
| `Form/__tests__/utils.test.ts` | 5 new vitest cases for `stabilizeFlatItems` |

Drag-reorder behavior is unchanged — `Reorder.Item` still receives the same `value` shape, just with stable identity when content hasn't changed.

## Test plan

- [x] `pnpm vitest run src/sds/surveys/SurveyFormBuilder/` — 78/78 pass (35 in `utils.test.ts` incl. 5 new cases, 18 in `index.test.tsx`, others).
- [x] `pnpm tsc` — no new errors (pre-existing `F0Graph`/`@xyflow/react` failures are unrelated).
- [x] `pnpm lint src/sds/surveys/SurveyFormBuilder/` — clean.
- [ ] Storybook smoke: open a question, place caret mid-string in the title and description textareas, type a character, confirm caret stays put.
- [ ] Storybook regression: drag-reorder a question between sections, confirm motion animation and final order are unchanged.
- [ ] End-to-end smoke in a consumer (publish a prerelease or `pnpm link` into Factorial's `newFormEditor/form.tsx` and `procurement/components/FormEditor`): same mid-string typing check plus confirm "+ Add question" still auto-focuses the new question's title (intended UX — caret at end is correct for a freshly added empty question).

### DIFF
--- a/packages/react/src/sds/surveys/SurveyFormBuilder/Form/__tests__/index.test.tsx
+++ b/packages/react/src/sds/surveys/SurveyFormBuilder/Form/__tests__/index.test.tsx
@@ -1,7 +1,13 @@
 import { userEvent } from "@testing-library/user-event"
+import { useState } from "react"
 import { describe, expect, it, vi } from "vitest"
 
-import { zeroRender as render, screen, waitFor } from "@/testing/test-utils"
+import {
+  act,
+  zeroRender as render,
+  screen,
+  waitFor,
+} from "@/testing/test-utils"
 
 import { SurveyDatasets, SurveyFormBuilderElement } from "../../types"
 import { SurveyFormBuilder } from "../index"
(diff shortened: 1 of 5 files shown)


rubenmoya:
Third commit (c84870ecd) — this one has a **failing-test-first** regression guard, so I'm confident it's the actual fix.

**Why the previous attempt failed:** the local mirror's sync effect was self-defeating. It pulled the prop down whenever it differed from the last-emitted value — but with an eventually-consistent parent (CopilotKit's `useCoAgent`), the re-render right after a keystroke still carries the **stale** prop, which differs from what was just emitted. So the effect reverted the textarea to the stale value and threw the caret to the end anyway. A lagging value is indistinguishable from a genuine external edit when you only compare "prop vs last-emitted".

**Correct fix:** while the field is being actively edited (focused), the local value is authoritative and incoming props are ignored; props are accepted only when the field is idle (not focused) — covering AI edits and programmatic resets. Guarded the sync effects on `document.activeElement !== titleRef/descriptionRef.current`.

**Test discipline this time:** added a regression test that backs the title with a lagging parent, focuses the textarea, pushes a stale prop, and asserts the value is NOT overwritten while focused but DOES update once blurred. Confirmed it **fails on the prior code** (`expected 'STALE' to be 'Hello'`) and passes after the fix. Full suite: 79/79.

rubenmoya:
Superseded by the real fix in the factorial monorepo: factorialco/factorial#101127.

Root cause turned out to be CopilotKit's `useCoAgent` eventual consistency (the controlled `value` goes stale for a render between the keystroke and the agent echo), not anything in f0. The fix belongs at that async boundary (`useOne`), which keeps `SurveyFormBuilder` a plain controlled component — no shadow state in the design system. Closing this in favor of that.

The Reorder.Item identity-stabilization in here was a minor perf idea but isn't needed for the bug; happy to reopen it as a standalone perf PR if it's worth it.

PR #4339, opened 2026-06-05
feat(ui): add ButtonGroup — a data-driven action bar

warcos-fact:
## What

Adds **`ButtonGroup`**, a **data-driven action bar** for arranging a row of action buttons. Actions are passed as **props, not children** — which keeps the surface constrained and structurally guarantees invariants (e.g. at most one primary). This is the outcome of review feedback on the original children-based draft.

- **`primaryAction`** — a single solid button, pinned at the trailing edge (or a split button via `type: "split"`).
- **`secondaryActions`** — outline buttons, split buttons, or inline `{ type: "separator" }` hairlines (or a single link). Plain secondaries that don't fit shed into a **"⋯" menu by width measurement**; split buttons and the primary are pinned and never collapse.
- **`otherActions`** — `DropdownItem[]`, always reachable through the "⋯" menu (supports separators and `critical`).
- **Responsive** — `stack="sm" | "md" | "container-md"` collapses into a full-width column (`fullWidthOnStack`, `reverseOnStack` to promote the primary to the top), with a responsive `size={{ base, md }}` and an automatic `Dropdown` ↔ `MobileDropdown` swap. The active branch is chosen in JS so only one mounts at a time (and the overflow machinery initializes with its DOM present).
- **Destructive actions** — `critical` renders red. Guard it in `otherActions` (behind the ⋯) by default; mark an inline secondary `critical` for a one-click delete when the resource is cheap to recreate.

## Notes

- Evolved from the original "layout primitive" (children-based) after review (@sauldom102, @rubenmoya). The prop shape mirrors `F0CardRow` and is consistent with the `primaryAction` / `secondaryActions` / `otherActions` triplet already used by `BaseHeader` and `TimeLine`.
- Folds the earlier `ButtonGroupOverflow` into `ButtonGroup` (deleted) — width-driven overflow is now native via `useOverflowCalculation`. No change to the public export barrel; there were no external consumers.
- Docs: autodocs + an MDX page documenting the row/stacked behavior, the ⋯ auto-swap, split-pinning, and the guard-by-default rule for destructive actions; behaviors stories cover the Trailing / Split / Reflowing patterns.

## Verification

Driven through Storybook across every pattern (Trailing, Split, Reflowing, width-overflow, critical) — which surfaced and fixed two bugs invisible to type-check/lint: an infinite render loop (unmemoized array into `useOverflowCalculation`) and secondaries failing to measure when the group started stacked (observer attached before the row DOM existed). Type-check, lint, Chromatic, and the type build are green.


packages/react/src/ui/ButtonGroup/ButtonGroupOverflow.tsx
```diff
@@ -0,0 +1,109 @@
...
+      className={cn("relative flex w-full min-w-0 items-center", className)}
+      style={{ gap }}
+    >
+      {/* Hidden full-width measurement copy, used to compute the visible split. */}
+      <div
+        ref={measurementContainerRef}
+        aria-hidden="true"
+        className="pointer-events-none invisible absolute left-0 top-0 flex items-center whitespace-nowrap"
+        style={{ gap }}
+      >
```
rubenmoya: I'd add an `inert` here to make sure it's not interactive at all

rubenmoya (approved):
LGTM, just a minor comment

packages/react/src/ui/ButtonGroup/__stories__/ButtonGroup.stories.tsx
```diff
@@ -0,0 +1,89 @@
...
+}
+
+/** A `ButtonGroupSeparator` divides logical groups (hidden when stacked). */
+export const WithSeparator: Story = {
+  render: (args) => (
+    <ButtonGroup {...args}>
+      <F0Button variant="outline" label="Discard" onClick={() => {}} />
+      <ButtonGroupSeparator />
+      <F0Button variant="default" label="Save" onClick={() => {}} />
+    </ButtonGroup>
```
sauldom102: Instead of passing buttons of the group like this I think we should pass them via `actions` prop accepting the props of F0Button, otherwise devs are going to pass whatever they want inside of here

warcos-fact: Sounds good, I'll review this.

PR #4307, opened 2026-06-03
feat(cocreation): co-creation pattern definition

warcos-fact:
## Description

Defines the **co-creation pattern** in Storybook — the docs, the interactive `creation-with-ai` and `standard-flow` stories with their mock data, and the shared mock clarifying runtime that backs the flow.

The flow exercises a couple of capabilities the SDS didn't expose yet, so this PR also ships a small amount of **component code** (below). Rebased on `main` now that the stack it was built on (#4550, #4533, #4535) has merged.

## Component changes (public API)

- **`refactor(headers)`** — rename `BaseHeader` / `ResourceHeader` `closeAction` → `onClose`. Drops the per-action `tooltip`; the close button is now labelled via `i18n.actions.close` and also renders on mobile (previously desktop-only).
- **`feat(ai)`** — add `revealChatOnCanvasToggle` to `F0AiChat`. Lets a host suppress the chat content's re-fade when the canvas opens/closes while the chat is docked as a side panel. Defaults to `true` (unchanged behavior); fullscreen transitions still reveal.

## Docs & stories

- **docs** — co-creation pattern MDX: `index`, `guidelines`, `phases`, `building-blocks`, `standard-flow`, `creation-with-ai`, `accessibility`, `adoption`.
- **stories** — `creation-with-ai` and `standard-flow` with mock data (`mockData`, `mockToast`, `survey-mocks`, `tab-configs`).
- **mocks** — extend the shared `F0AiChat` mock runtime with a clarifying-question flow; add a welcome-suggestions demo to the `F0AiChatTextArea` story.

## Internal refactors (from review)

- Centralize the story's story-local `CanvasContent` / `CanvasEntityDefinition` casts behind typed helpers.
- Extract repeated survey toast titles into constants.

PR #4340, opened 2026-06-05
feat(ui): add F0CardRow component

warcos-fact:
## What

Adds **`F0CardRow`** — the horizontal, row-shaped counterpart to `F0Card`:

- Optional **avatar** on the left (any avatar type: person, company, team, file, flag, icon, emoji, module, alert, date, pulse), rendered at a fixed `lg` size.
- **Title** + optional **description** (wraps across lines when long).
- Trailing **actions**: `primaryAction`, `secondaryActions`, `otherActions` (⋯ overflow menu), or an icon-only **confirm/reject** variant.
- Optional **alert** banner.
- **`stackAt`** container breakpoint (`sm`/`md`/`lg`/`never`, default `never`): below it the actions drop onto their own full-width line (with a footer hairline) and secondary buttons fold into the ⋯ menu; the primary stays pinned.

Includes a unit test suite and autodocs (description + use cases + stories).

## Notes

- **Built on `ButtonGroup` (#4339, now merged).** `CardRowActions` is a thin adapter over `ButtonGroup`: the `primaryAction` / `secondaryActions` / `otherActions` triplet maps straight through, and `ButtonGroup` owns the row layout, the width-driven ⋯ overflow, and pinning the primary at the trailing edge. The card layers on only its own concerns:
  - stops click propagation so an action never triggers the row's `onClick` / overlay-link navigation;
  - the stacked own-line **footer hairline** at the container breakpoint (`stackAt`, a container query — reacts to the card's own width, not the viewport);
  - the icon-only **confirm/reject** (✗ / ✓) variant, mapped to two pinned secondaries.
- The actions wrapper takes the remaining row space (`flex-1`) so `ButtonGroup` has a bound wider than its content to measure against — a shrink-to-fit container would shed its tail into the ⋯ menu.
- The `F0HILActionConfirmation` migration to `F0CardRow` and the removal of the experimental `F0Card` `oneLiner` variant stay with the cocreation PR (#4307), since both originate there. Once this lands, that PR (or a follow-up) can point F0HIL at `F0CardRow` and drop `oneLiner`.

## Verification

Type-check and the F0Card unit suite (45 tests) pass against the merged `ButtonGroup`. Driven through Storybook across the row patterns — inline (primary + ⋯ + secondary), confirm/reject, stacked own-line + footer, and genuine width-overflow (a secondary sheds into ⋯ while the primary stays pinned).

### DIFF
--- a/packages/react/src/components/F0Card/F0CardRow.tsx
+++ b/packages/react/src/components/F0Card/F0CardRow.tsx
@@ -1,7 +1,6 @@
 import { forwardRef } from "react"
 
 import { F0Link } from "@/components/F0Link"
-import { type TagStatusProps } from "@/components/tags/F0TagStatus"
 import { DropdownItem } from "@/experimental/Navigation/Dropdown"
 import { withDataTestId } from "@/lib/data-testid"
 import { withSkeleton } from "@/lib/skeleton"
@@ -21,6 +20,7 @@ import {
   CardRowActions,
   type CardRowConfirmAction,
   type CardRowStackAt,
+  type CardRowStatus,
   cardRowClassName,
 } from "./components/CardRowActions"
 import { type CardAlertProps } from "./types"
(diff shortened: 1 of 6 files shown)


packages/react/src/components/F0Card/F0CardRow.tsx
```diff
@@ -0,0 +1,244 @@
...
+          </F0Link>
+        )}
+
+        <div className={cardRowClassName[stackAt]}>
+          <div className="flex min-w-0 flex-row items-center gap-3">
+            {avatar && <CardAvatar avatar={avatar} size="lg" />}
+            <div className="flex min-w-0 flex-col gap-0">
+              {title && (
+                <Text variant="body" content={title} className="font-medium" />
+              )}
```
rubenmoya: Title is not optional, do we need this check?

rubenmoya (approved):
LGTM

packages/react/src/components/F0Card/F0CardRow.tsx
warcos-fact: Good catch — `title` is required, so the guard never short-circuits. Removed it in 27cb23bae, rendering the title unconditionally.

siguenzaraul (approved):

PR #4341, opened 2026-06-05
feat(sds): render F0HILActionConfirmation with F0CardRow

warcos-fact:
## What

Updates **`F0HILActionConfirmation`** to render with **`F0CardRow`** instead of its bespoke `div` + buttons layout: the confirm action becomes the primary, cancel becomes a secondary.

## Sequencing

Third step of the ship sequence:
1. ButtonGroup (#4339)
2. F0CardRow (#4340)
3. **F0HIL → F0CardRow** (this PR)

Stacked on `feat/f0-card-row` so it compiles against `F0CardRow`. **Retarget to `main` once #4340 merges.** The dead `oneLiner` variant has been removed from the cocreation branch, so nothing competes with this.

### DIFF
--- a/packages/react/src/sds/ai/F0HILActionConfirmation/F0HILActionConfirmation.tsx
+++ b/packages/react/src/sds/ai/F0HILActionConfirmation/F0HILActionConfirmation.tsx
@@ -1,4 +1,4 @@
-import { F0Button } from "@/components/F0Button"
+import { F0CardRow } from "@/components/F0Card"
 import Check from "@/icons/app/Check"
 
 import { F0HILActionConfirmationProps } from "./types"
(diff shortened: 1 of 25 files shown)


packages/react/src/sds/ai/F0HILActionConfirmation/F0HILActionConfirmation.tsx
```diff
@@ -1,35 +1,31 @@
...
-          type="button"
-          variant="outline"
-          size="md"
-          onClick={onCancel}
-          label={cancelText}
-        />
-      </div>
-    </div>
+    <F0CardRow
+      title={text ?? ""}
```
rubenmoya: `title` is mandatory in `F0CardRow`, that's why we're using `?? ""`, but if that happens we will show a card without any title... do we want that? Sounds like a broken UI to me

rubenmoya (approved):

packages/react/src/sds/ai/F0HILActionConfirmation/F0HILActionConfirmation.tsx
warcos-fact: Good catch. We had the same issue in the previous PR that builds the CardRow. I'll review and fix.

warcos-fact: Reviewed — this is **not** addressed yet. `text` is still optional (`text?: string` in `types.ts`) and we still coerce with `title={text ?? ""}`.

`F0CardRow` guards the title with `{title && …}`, so an empty `text` won't crash, but it renders a confirmation row with **no prompt** — just the description + ✓/✗ buttons — which is the broken UI you flagged.

Fix I'm going with: make `text` required (`text: string`) and drop the `?? ""`. A HIL confirmation without a prompt has no meaning, and there are no external consumers yet (only the stories, which all pass `text`), so tightening the type is safe and removes the coercion entirely. Pushing that now.

siguenzaraul (approved):

PR #4343, opened 2026-06-05
feat(F0PdfViewer): add experimental PDF viewer component

miriam-mr90:
## Summary

Adds a new `F0PdfViewer` component under `components/` (with experimental flag) — a read-only PDF viewer built on `react-pdf` (`pdf.js`) that renders a document as a continuous, scrollable list of pages with a toolbar for page navigation, zoom, rotation, printing and downloading.

This single F0 component covers the current ATS viewer **and** the Expenses viewer, so both modules could drop their own bespoke PDF viewers.

### Component
- New `F0PdfViewer` at `packages/react/src/components/F0PdfViewer/`, composed as `withDataTestId(experimentalComponent("F0PdfViewer", withSkeleton(F0PdfViewerBase, F0PdfViewerSkeleton)))`
- API: `url`, `filename?` (defaults to `"document.pdf"`), `page?`, `pagesToDisplay?`, `initialScale?` (`page-width` default / `page-fit` / `0.5`…`4`), `withCredentials?`, `rotatable?`, `initialRotation?`, `onRotationChange?`, `onPdfLoaded?`, `onPageChange?`
- Continuous scroll, visible-page tracking, page-fit/width that recompute from container size (accounting for page padding + scrollbar gutter so pages stay centered), links open in a new tab

### Rotation (opt-in)
- `rotatable` adds a toolbar control that rotates the document in 90° steps
- Rotation is **uncontrolled**: the viewer owns the state but exposes `initialRotation` + `onRotationChange`, so persistence stays with the consumer (e.g. Expenses keeps its per-file `localStorage`). Keeps F0 free of storage/domain concerns
- `page-width`/`page-fit` recompute with swapped dimensions at 90°/270°

### `@/ui/` layer + worker
- `src/ui/pdf.tsx` re-exports react-pdf `Document`/`Page` (+ pdf.js types) so the component never imports the third-party primitive directly
- `pdfWorker.ts` auto-configures the pdf.js Web Worker (version-matched to the bundled `pdfjs-dist`) and exposes `configurePdfWorker(src)` to override for CSP/offline

### Toolbar
- Page `X / Y` + prev/next, zoom −/+, scale `F0Select`, rotate (when `rotatable`), print, download — icon-only `F0Button variant="outline"` (prev/next at `sm`/24px, the rest `md`/32px per Figma)

image

### Styling
- Tailwind + `f1-` tokens; a minimal co-located `.css` only for what must target react-pdf's internal DOM (canvas/text-layer z-index, opaque viewer surface, text-layer `::selection`)

### Packaging
- `react-pdf` + `pdfjs-dist` added as **peerDependencies** (and devDependencies for F0's own dev/build); both are externalized from the bundle, so consumers that don't use the viewer don't pay for pdf.js

### Stories / Docs / Tests
- Stories `Default`, `PageFit`, `Rotatable`, `Skeleton`, `Snapshot`; manual MDX docs page (`!autodocs`) with **Sizing** and **Rotation** sections; the Docs preview renders in a bounded iframe (`docs.story.inline: false` + `height`) so it doesn't overflow the docs column
- Unit tests: component (mocked `@/ui/pdf`), plus pure `scales`, `visiblePage`, and `pdfActions`

## Accessibility & API
- Toolbar has `role="toolbar"`; icon-only buttons carry accessible names; the `page X / Y` indicator is announced via a polite `aria-live` region
- The selectable text layer gets a legible `::selection` (foreground + selection background tokens) so selected text stays readable
- Loading and skeleton states expose `role="status"` + `aria-busy` / `aria-live`
- Root accepts arbitrary `data-*` attributes and `data-testid`

## Notes 👀 
- Lives under `components/` (the `experimental/` folder is deprecated) and opts into experimental status via `experimentalComponent()` + the Storybook `experimental` tag
- **New dependencies**: this is the first F0 component to pull in `react-pdf` + `pdfjs-dist`. Would love Foundations' take on adding them as peer deps and on the worker auto-config strategy (`import.meta.url`/`?url` + `configurePdfWorker()` override, version-matched, no CDN)
- F0 has no global `::selection`; the viewer scopes one to its text layer (the monolith relies on a global `::selection`). Worth considering a global one in F0 separately
- Covers both the ATS and Expenses use cases; Documents (fillable forms + save) would need a follow-up
- Inline review comments below capture the "why" behind the non-obvious bits (kept out of the code per our convention)

## ⚠️ About the "Breaking public API changes" check

The API surface check flags ~19 "breaking changes", but they all stem from a **single additive change**: this PR adds a new `pdfViewer` block (with the `rotate` key) to `defaultTranslations`. Since translation keys are required in `TranslationsType`, that new key ripples through everything that consumes the i18n type — `TranslationsType`, `buildTranslations`, `I18nProvider`/`I18nProviderProps`, `useI18n`, the granularity definitions and `NavigationFilter`, plus the `experimental`/`ai` re-exports of those.

Nothing is broken at runtime (F0 ships `defaultTranslations` with `pdfViewer` included). The only impact is type-level: consumers that pass a **complete** `translations` object to `I18nProvider` will need to add the new `pdfViewer` keys when upgrading — the standard consequence of adding i18n keys for any new F0 component. The check is non-blocking.

### Screenshot
image


packages/react/src/ui/pdf.tsx
```diff
@@ -0,0 +1,10 @@
+"use client"
+
+import { Document, Page, pdfjs } from "react-pdf"
+
+import "react-pdf/dist/esm/Page/AnnotationLayer.css"
+import "react-pdf/dist/esm/Page/TextLayer.css"
+
+export { Document, Page, pdfjs }
```
miriam-mr90, line 8: AGENTS rule: components must not import third-party primitives directly. This `@/ui/pdf` wrapper re-exports react-pdf's `Document`/`Page` (and the pdf.js types) so the component depends on the UI abstraction layer, not react-pdf directly.

packages/react/src/components/F0PdfViewer/F0PdfViewer.styles.css
```diff
@@ -0,0 +1,24 @@
+.F0PdfViewer__surface {
```
miriam-mr90, line 1: These rules target react-pdf's internally-rendered DOM (`.react-pdf__*`, `.textLayer`), which Tailwind utilities can't reach. `__surface` is an OPAQUE equivalent of `f1-background-tertiary`: that token is translucent (~4% alpha), so the sticky translucent toolbar tint-stacked over the (also translucent) viewer surface and read darker than over the white document. Compositing the tint over an opaque base (`--neutral-0` + `--neutral-5`) keeps toolbar and viewer identical everywhere and stops the document showing through while scrolling, while staying theme/dark-mode aware.

packages/react/src/components/F0PdfViewer/F0PdfViewer.styles.css
```diff
@@ -0,0 +1,24 @@
...
+  z-index: 0;
+}
+
+.F0PdfViewer__page .react-pdf__Page__textContent {
+  z-index: 1;
+  pointer-events: auto;
+  user-select: text;
+}
+
+.F0PdfViewer__page .textLayer ::selection {
```
miriam-mr90, line 21: pdf.js renders the selectable text layer with `color: transparent` (visible glyphs come from the canvas beneath). Without an explicit `::selection` color, selecting that transparent text is unreadable. We force a legible foreground + selection background (tokens) so selected text stays readable. Note: the ATS monolith gets this from a GLOBAL `::selection`; F0 has none, so we scope it to the viewer's text layer.

packages/react/src/components/F0PdfViewer/pdfActions.ts
```diff
@@ -0,0 +1,42 @@
...
+  if (!pdf) return
+
+  const blobUrl = URL.createObjectURL(await getPdfBlob(pdf))
+  const iframe = document.createElement("iframe")
+  iframe.style.display = "none"
+  iframe.src = blobUrl
+  iframe.onload = () => {
+    iframe.focus()
+    iframe.contentWindow?.print()
+    const cleanup = () => {
```
miriam-mr90, line 18: Print uses a hidden iframe + blob URL. We revoke the object URL and remove the iframe on `afterprint` (with a timeout fallback) so repeated prints don't leak iframes / object URLs.

packages/react/src/components/F0PdfViewer/F0PdfViewer.tsx
```diff
@@ -0,0 +1,293 @@
...
+    }, [currentPage, onPageChange])
+
+    useEffect(() => {
+      if (
+        pages.length > 0 &&
+        (initialScale === "page-width" || initialScale === "page-fit")
+      ) {
+        applyDynamicScale(initialScale)
+      }
+      // eslint-disable-next-line react-hooks/exhaustive-deps
```
miriam-mr90, line 221: exhaustive-deps is intentionally disabled here: this effect should only re-run when the pages are first measured (`pages.length`) or `initialScale` changes, to apply the dynamic page-width/page-fit scale once dimensions are known — not on every `applyDynamicScale` identity change.

miriam-mr90:
@factorialco/foundations One thing I'd like your input on 👀

This component uses **`react-pdf` + `pdfjs-dist`** (pdf.js), added as peer deps and externalized, so they're **not bundled into F0**; only apps that use the viewer pull them in.

`pdf.js` renders PDFs in a background **Web Worker**. Instead of loading that worker from a **CDN** (needs internet, blocked by strict security policies), the component **ships it with the app**, matched to the bundled pdf.js version, with `configurePdfWorker()` as an escape hatch.

When this was first proposed, **Sergio Carracedo OK'd this approach**, but since he's no longer around to review it, I'd like to confirm you're still good with it 🙏

sauldom102 (commented):
Great contribution Miriam!! ❤️ Some nitpick:
- Text selection is not working fully fine at the moment

https://github.com/user-attachments/assets/ce520d17-bd15-4780-9aec-b564ecd61898

sauldom102 (commented):
Could we also add some story/tests for when we have a pdf with fillable fields? Maybe it's out of your scope/use-case for recruitment but I think it would be great to validate that that works fine

rubenmoya:
When the PDF is loading, there is a time where we show multiple spinners: 

image

I'm guessing it's one per page. Could we use some kind of skeleton?

miriam-mr90:
> Great contribution Miriam!! ❤️ Some nitpick:
> 
> * Text selection is not working fully fine at the moment
> 
>  Screen.Recording.2026-06-10.at.16.15.42.mov

Fixed!!
image

packages/react/src/components/F0PdfViewer/types.ts
```diff
@@ -0,0 +1,45 @@
...
+  /** File name used when downloading the document. Defaults to "document.pdf". */
+  filename?: string
+  /** Zero-based page index to scroll to initially (and on change). */
+  page?: number
+  /** Restrict rendering to a subset of pages (zero-based indexes). */
+  pagesToDisplay?: number[]
+  /** Initial zoom level. Defaults to "page-width". */
+  initialScale?: F0PdfScale
+  /** Send credentials with the document request. Defaults to true. */
+  withCredentials?: boolean
```
sauldom102, line 34: What is this withCredentials used for? Is it something F0 needs to care about? 🤔

miriam-mr90:
> When the PDF is loading, there is a time where we show multiple spinners:
> 
> image
> I'm guessing it's one per page. Could we use some kind of skeleton?

@rubenmoya You were right! it was react-pdf's default spinner, one per page.

Now, each page shows a page-shaped **skeleton** placeholder while it renders instead of a spinner, and I also removed a duplicated document-level loading state

packages/react/src/components/F0PdfViewer/types.ts
miriam-mr90: `withCredentials` is forwarded to react-pdf's `Document` (the pdf.js fetch). It controls whether the PDF request sends credentials (cookies/auth) on **cross-origin** requests. It's a no-op for same-origin.

All three current monolith viewers (ATS, Documents and Expenses) fetch the PDFs from authenticated backend endpoints and already hardcode `withCredentials: true` today. 
I just surfaced it as a prop (defaulting to `true` to match that behavior) instead of hardcoding it...

**What `false` would do:**
  - Same-origin or public PDFs → no difference, still loads fine.
  - Cross-origin + auth-gated PDF → the request goes **without** the session cookie → backend responds 401/403 → the PDF fails to load (react-pdf shows its default "Failed to load PDF file." message, since we don't render a custom error state).

There's also the flip side: `withCredentials: true` on a cross-origin request requires the server to send `Access-Control-Allow-Credentials: true` (and a concrete origin, not `*`), so a consumer with public cross-origin PDFs behind a CDN that isn't set up for credentialed CORS would want to set it to `false`.

That's why it's a configurable prop: `true` by default (covers the current authenticated use cases), with the option to turn it off. But happy to flip the default or drop it if you'd rather keep it implicit 🙃

sauldom102 (approved):

PR #4349, opened 2026-06-07
feat(F0AudioPlayer): add audio player for recordings

ReddyWish:
Adds an experimental `F0AudioPlayer` / `F0AudioPlayerCard` for playing back
AI call recordings (e.g. the candidate side panel in the ATS). No existing
f0 component covered audio playback.

## Changes
- **`F0AudioPlayer`** — compact controls bar: play/pause, draggable seek
  bar, and `current / total` time.
- **`F0AudioPlayerCard`** — the recording card from the design: 40×40 play
  avatar + title/date header, kebab top-right, full-width scrubber below.
- **Kebab menu** — playback speed (`1x / 1.5x / 2x`) styled like the f0
  single-select rows (tinted selected row + right-aligned `CheckCircle`,
  no radio dot). Consumers can append items via `actions: AudioPlayerMenuAction[]`.
- **Seek bar** — built on the `ui/slider` primitive (isolated in
  `AudioScrubber`); the fill mirrors Radix's thumb in-bounds offset so it
  always meets the thumb (fixes the gap near the start). Viridian fill,
  buffered indicator, thumb on hover/focus.
- **Internals** — `useAudioPlayer` wraps the `HTMLAudioElement` and mirrors
  its events into React; `usePlayerController` owns state + controlled
  `playing` sync and is shared by both layouts.
- Exported via `experimentalComponent()` + `withDataTestId`; added to
  `components/exports.ts`.

## Out of scope
- Waveform / transcript sync, volume.

Screenshot 2026-06-07 at 20 34 58

Screenshot 2026-06-08 at 10 10 45

Screenshot 2026-06-08 at 14 02 03


packages/react/src/components/F0AudioPlayer/F0AudioPlayer.tsx
```diff
@@ -0,0 +1,89 @@
...
+      playbackRates: _playbackRates,
+      playing: _playing,
+      defaultPlaying: _defaultPlaying,
+      onPlayingChange: _onPlayingChange,
+      onPlay: _onPlay,
+      onPause: _onPause,
+      onSeek: _onSeek,
+      onTimeUpdate: _onTimeUpdate,
+      onEnded: _onEnded,
+      onError: _onError,
```
rubenmoya: It feels weird to do all this destructuring and then not using any of it

packages/react/src/components/F0AudioPlayer/components/AudioScrubber.tsx
```diff
@@ -0,0 +1,115 @@
...
+  return (
+    <Slider
+      value={[value]}
+      onValueChange={handleValueChange}
+      onValueCommit={handleValueCommit}
+      min={0}
+      max={max || 1}
+      step={1}
+      disabled={isDisabled}
+      aria-label="Seek"
```
rubenmoya: This label is hardcoded

packages/react/src/components/F0AudioPlayer/components/AudioScrubber.tsx
```diff
@@ -0,0 +1,115 @@
...
+          className={cn(
+            "pointer-events-none absolute top-1/2 size-3 -translate-x-1/2 -translate-y-1/2 rounded-full bg-f1-background-selected-bold opacity-0 transition-opacity",
+            "group-hover:opacity-100 group-has-[:focus-visible]:opacity-100",
+            "group-has-[:focus-visible]:ring-1 group-has-[:focus-visible]:ring-f1-special-ring group-has-[:focus-visible]:ring-offset-1"
+          )}
+        />
+      )}
+
+      <SliderThumb
+        aria-label="Seek"
```
rubenmoya: This label is hardcoded, and it's the same as the previous element

packages/react/src/components/F0AudioPlayer/components/AudioScrubber.tsx
```diff
@@ -0,0 +1,115 @@
...
+    <Slider
+      value={[value]}
+      onValueChange={handleValueChange}
+      onValueCommit={handleValueCommit}
+      min={0}
+      max={max || 1}
+      step={1}
+      disabled={isDisabled}
+      aria-label="Seek"
+      aria-valuetext={`${formatPlaybackTime(value)} of ${formatPlaybackTime(max)}`}
```
rubenmoya: This is hardcoded

packages/react/src/components/F0AudioPlayer/components/PlaybackMenu.tsx
```diff
@@ -0,0 +1,109 @@
...
+              )
+            })}
+          </>
+        )}
+        {extraItems.length > 0 && (
+          <>
+            {showSpeed && <DropdownMenuSeparator />}
+            {extraItems.map((action, index) => (
+              <DropdownMenuItem
+                key={index}
```
rubenmoya: `label` would be better for the key, same index doesn't guaranty same item

packages/react/src/components/F0AudioPlayer/components/PlaybackMenu.tsx
```diff
@@ -0,0 +1,109 @@
...
+                onSelect={() => action.onClick?.()}
```
rubenmoya: Why is `onClick` optional? What's the use case of an item without a click action?

packages/react/src/components/F0AudioPlayer/components/PlaybackMenu.tsx
```diff
@@ -0,0 +1,109 @@
...
+import {
+  DropdownMenu,
+  DropdownMenuContent,
+  DropdownMenuItem,
+  DropdownMenuLabel,
+  DropdownMenuSeparator,
+  DropdownMenuTrigger,
+} from "@/ui/dropdown-menu"
+
+import type { AudioPlayerMenuAction } from "../types"
```
rubenmoya, line 15: I'd rather define this here and import it in types, since this component is the only user of that type.

packages/react/src/components/F0AudioPlayer/components/PlaybackMenu.tsx
```diff
@@ -0,0 +1,109 @@
...
+  const showSpeed = playbackRates.length > 0
+
+  return (
+    <DropdownMenu>
+      <DropdownMenuTrigger asChild>
+        <F0Button
+          variant="ghost"
+          size="sm"
+          icon={Ellipsis}
+          label="Recording options"
```
rubenmoya: This is hardcoded

packages/react/src/components/F0AudioPlayer/components/PlayPauseButton.tsx
```diff
@@ -0,0 +1,37 @@
...
+  disabled,
+  size = "md",
+  onToggle,
+}: PlayPauseButtonProps) => {
+  const Icon = isPlaying ? SolidPause : SolidPlay
+
+  return (
+    <button
+      type="button"
+      aria-label={isPlaying ? "Pause" : "Play"}
```
rubenmoya: This is hardcoded

packages/react/src/components/F0AudioPlayer/F0AudioPlayer.tsx
```diff
@@ -0,0 +1,89 @@
...
+      ...rest
+    } = props
+
+    const controller = usePlayerController(props)
+
+    return (
+      <div
+        ref={ref}
+        role="group"
+        aria-label={ariaLabel ?? "Audio player"}
```
rubenmoya: Hardcoded string

packages/react/src/components/F0AudioPlayer/F0AudioPlayerCard.tsx
```diff
@@ -0,0 +1,113 @@
...
+    ...rest
+  } = props
+
+  const controller = usePlayerController(props)
+
+  return (
+    <div
+      ref={ref}
+      role="group"
+      aria-label={ariaLabel ?? "Audio player"}
```
rubenmoya: Hardcoded string

packages/react/src/components/F0AudioPlayer/components/PlaybackMenu.tsx
```diff
@@ -0,0 +1,109 @@
...
+  playbackRate,
+  playbackRates,
+  onRateChange,
+  disabled,
+  extraItems = [],
+}: PlaybackMenuProps) => {
+  const showSpeed = playbackRates.length > 0
+
+  return (
+    <DropdownMenu>
```
rubenmoya, line 36: Can't we reuse the existing [Dropdown](https://f0.factorial.dev/?path=/docs/experimental-dropdown--documentation) component?

packages/react/src/components/F0AudioPlayer/usePlayerController.ts
```diff
@@ -0,0 +1,77 @@
...
+    onTimeUpdate,
+    onEnded,
+    onError,
+  } = props
+
+  const audioRef = useRef<HTMLAudioElement>(null)
+
+  const [, setIsPlaying] = useControllableState<boolean>({
+    prop: playing,
+    defaultProp: defaultPlaying,
```
rubenmoya: Is this even used? We pass `playing` and `defaultPlaying` to `useControllableState`, but then we ignore the hook getter and use the `playing` from props, and `player.isPlaying`. We have multiple sources of truths.

packages/react/src/components/F0AudioPlayer/variants.ts
```diff
@@ -0,0 +1,11 @@
+import { cva } from "cva"
+
+export const timeVariants = cva({
```
rubenmoya: Instead of putting this in the root and importing it in two places using `../` repeating the same HTML, why don't we create a component? It would also cleanup the `formatPlaybackTime`, we can put those two things in the component itself. One file instead of 2 and duplicated HTML.

rubenmoya (changes requested):

packages/react/src/components/F0AudioPlayer/components/AudioScrubber.tsx
ReddyWish: Hey @rubenmoya had some doubts here, so yes, I would prefer dynamic value, but this static strings ('Play', 'Seek', 'Playback speed') have no data to derive from, so I just decided to keep hardcoded strings in this cases. So you mean better to wire them through useI18n() ?

packages/react/src/components/F0AudioPlayer/components/AudioScrubber.tsx
ReddyWish: https://github.com/factorialco/f0/pull/4349#discussion_r3425865777

packages/react/src/components/F0AudioPlayer/components/AudioScrubber.tsx
ReddyWish: https://github.com/factorialco/f0/pull/4349#discussion_r3425865777

packages/react/src/components/F0AudioPlayer/components/PlaybackMenu.tsx
ReddyWish: agree

packages/react/src/components/F0AudioPlayer/F0AudioPlayerCard.tsx
ReddyWish: @rubenmoya this is a little bit different, it's overridable, not fixed as above, but question is the same - better to wire it through useI18n() ?

packages/react/src/components/F0AudioPlayer/components/PlaybackMenu.tsx
ReddyWish: @rubenmoya  I wanted to use it in a first place, but it would not satisfy our ui/ux here, we need a selected item, right alignment and couple more tiny nuances, so I would need to make some changes to the Dropdown, but i considered it not optimal for now in terms of time

packages/react/src/components/F0AudioPlayer/usePlayerController.ts
ReddyWish: valid. `player.isPlaying` is a real source of truth, I will drop `useControllableState`

packages/react/src/components/F0AudioPlayer/variants.ts
ReddyWish: yes, it is valid as well, will be fixed

packages/react/src/components/F0AudioPlayer/F0AudioPlayer.tsx
ReddyWish: @rubenmoya mmm it noisy, yes, but it was pulled out just to avoid getting into `...rest` and get spread onto the `<div>` as invalid DOM attrs. I will consider a shared getDataAttributes(props) that picks only data-* 👍

packages/react/src/components/F0AudioPlayer/components/PlaybackMenu.tsx
ReddyWish: Did not notice it, good catch, I think agent did it because items could be href links. But anyway it should't be optional thx 👍

packages/react/src/components/F0AudioPlayer/components/PlaybackMenu.tsx
ReddyWish: @rubenmoya I would keep it as is. `AudioPlayerMenuAction` is a public api, not just playbackMenu's .. it's re-exported from index.tsx, and it's documented in the MDX, so card consumers use it too. Also moving it into PlaybackMenu and importing it back into types.ts would also invert the dependency direction 🤔 wdyt?

packages/react/src/components/F0AudioPlayer/components/AudioScrubber.tsx
rubenmoya: Yeah, it would be better to have them through `useI18n`, otherwise clients with the language set to Portuguese, Spanish, Catala, etc, would still have the labels in Enlish

packages/react/src/components/F0AudioPlayer/components/PlaybackMenu.tsx
rubenmoya: I'd try to avoid as much duplication as possible, I'd say it's better to add that new item type to the dropdown and reuse it here. What are those "tiny nuances"?

What do you think @sauldom102?

ReddyWish: @rubenmoya @sauldom102 I think i can add new item type yes, we just need to decide, by "tiny nuances" i mean style alignment with the design, which can be covered if we would apply new item type

rubenmoya (approved):

packages/react/src/components/F0AudioPlayer/components/PlaybackMenu.tsx
ReddyWish: Ok, I will proceed with considering the dropDown in next iterations of the audio player, for will keep it as is, for saving time

PR #4351, opened 2026-06-07
feat(OneTable): add onHeaderClick to table column headers

FaithAdekunle:
## What

Adds an optional `onHeaderClick` callback to `TableHead` and to the OneDataCollection **table column definition**. When provided, the column header label becomes an interactive button.

Key points:
- **Independent of sorting** — `onHeaderClick` and `onSortClick` can coexist on the same header (the sort button remains separate).
- The column-level prop is exposed via the existing `Pick<ComponentProps<typeof TableHead>, …>` in the table column type, so it flows through the existing `{...column}` spread in `Table.tsx` with no render changes.

## Why

Consumers need to trigger an action scoped to a clicked column header — e.g. the Financial Workspace **employee-cost** table opens a workforce/headcount side panel when a month header is clicked. The legacy table supported this via a per-header `onClick`; the OneDataCollection table had no equivalent, so the behavior was lost in the migration. This restores that capability at the design-system level.

## Changes

- `experimental/OneTable/TableHead/index.tsx` — new `onHeaderClick?` prop; label renders as a focusable button when set.
- `patterns/OneDataCollection/visualizations/collection/Table/types.ts` — expose `onHeaderClick` on the column definition.
- Stories:
  - `OneTable` → `ClickableHeader` (incl. coexistence with sorting).
  - OneDataCollection Table → `TableWithClickableHeaders` (column-level API).

## Follow-up (downstream)

Once released, bump `@factorialco/f0-react` in the `factorial` monolith and wire `onHeaderClick` into the Financial Workspace `DataCollectionTable` (separate PR).

### DIFF
--- a/packages/react/src/experimental/OneTable/TableHead/index.tsx
+++ b/packages/react/src/experimental/OneTable/TableHead/index.tsx
@@ -47,6 +47,13 @@ interface TableHeadProps {
   onSortClick?: () => void
 
   /**
+   * Callback fired when the header label is clicked. Independent of sorting —
+   * when provided, the header label becomes an interactive button. Use this to
+   * trigger actions such as opening a side panel for the column.
+   */
+  onHeaderClick?: () => void
+
+  /**
    * Optional tooltip text. When provided, displays an info icon next to the header content
    * that shows this text in a tooltip when hovered.
    */
@@ -87,6 +94,7 @@ export function TableHead({
   minWidth,
   sortState = "none",
   onSortClick,
+  onHeaderClick,
   info,
   infoIcon = InfoCircleLine,
   sticky,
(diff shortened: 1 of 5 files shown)


albertpmz:
Could you please add a video demonstrating this new behavior? It'd be interesting to see how both the onClick trigger and sorting button coexist 🙏🏼

FaithAdekunle:
> Could you please add a video demonstrating this new behavior? It'd be interesting to see how both the onClick trigger and sorting button coexist 🙏🏼

The header text itself serves as the onClick trigger while the sorting button remains as is.

https://github.com/user-attachments/assets/c4ebd4fa-6a4b-4bbb-a404-234fe5fd23d3

PR #4382, opened 2026-06-09
feat(hourDistribution): per-datapoint neutralLabel for tooltip

neil-krichi:
## Summary
- Adds `neutralLabel?: string` to `HourDistributionDataPoint` and `BarSeriesDataPoint`
- When set, the per-point label overrides the chart-level `justifiedAbsenceLabel` in the tooltip
- Backward-compatible: existing consumers without `neutralLabel` get the same behavior

## Why
The Factorial Time Tracking distribution column uses `hourDistribution` to show worked hours per day. Currently, weekends and actual absences both show "Justified absence" in the tooltip. With this change, consumers can pass a per-point label like "Non-workable", "Sick leave", or "Christmas Day" to differentiate them.

## Test plan
- [x] Existing tests still pass (10/10)
- [x] New tests: per-point label override for partial absences, full-day bars, and fallback behavior
- [x] New story: `HourDistributionWithMixedLabels` showing weekends vs leave types

### DIFF
--- a/packages/react/src/ui/value-display/types/barSeries/barSeries.tsx
+++ b/packages/react/src/ui/value-display/types/barSeries/barSeries.tsx
@@ -35,6 +35,8 @@ export interface BarSeriesDataPoint {
   secondaryValue?: number
   neutralValue?: number
   neutralFullHeight?: boolean
+  /** Per-point label for the neutral segment. Overrides any chart-level label when set. */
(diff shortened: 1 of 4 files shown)


andrewwekesa (approved):

albertpmz (approved):
LGTM!

PR #4384, opened 2026-06-09
feat(EditableTable): cellHint position + apply disabled state to hint

RafaMellado:
## What

Two improvements to the Editable Table `cellHint`:

1. **`position` option** — `cellHint` can now return `position: "left" | "right"` to place the hint icon at the start (default) or end of the cell content.

2. **Disabled cell fix** — when a cell is disabled, the hint icon used to render on a white background while the content was greyed out, making it look like the hint lived *outside* the cell. The disabled background now covers the whole cell (icon included), so the hint reads as part of the disabled cell.

## How

- `types.ts` — added optional `position?: "left" | "right"` to the `cellHint` return type.
- `cells/index.ts` — added `position` to the shared `hint` prop.
- `BaseCell.tsx` — render the hint icon before or after the content based on `position`; added a `disabled` prop that applies `bg-f1-background-disabled` to the whole cell container.
- `DisabledCell.tsx` — pass `disabled` to `BaseCell` and drop the inner-div background so the disabled background spans the full cell.

All cell components already forward `hint` to `BaseCell`, so both behaviors work across every editable cell type.

## Usage

```ts
cellHint: (item) => ({
  icon: AlertCircle,
  message: "Inferred from catalog",
  position: "right",
})
```

### DIFF
--- a/packages/react/src/patterns/OneDataCollection/visualizations/collection/EditableTable/components/cells/BaseCell.tsx
+++ b/packages/react/src/patterns/OneDataCollection/visualizations/collection/EditableTable/components/cells/BaseCell.tsx
@@ -1,7 +1,8 @@
 import { ReactNode } from "react"
 
 import { F0Icon, type IconType, type F0IconProps } from "@/components/F0Icon"
-import { cn, focusRing } from "@/lib/utils"
+import { F0Box } from "@/lib/F0Box"
+import { cn } from "@/lib/utils"
 import {
   Tooltip,
   TooltipContent,
@@ -48,16 +49,16 @@ export function BaseCell({
     <TooltipProvider delayDuration={100}>
       <Tooltip>
         <TooltipTrigger asChild>
-          <button
-            type="button"
+          <F0Box
+            role="img"
             aria-label={hint.message}
-            className={cn(
-              "pointer-events-auto flex shrink-0 cursor-pointer items-center rounded px-1",
-              focusRing()
-            )}
+            display="flex"
+            alignItems="center"
+            shrink={false}
+            paddingX="xs"
           >
             <F0Icon icon={hint.icon} size="md" color={hint.iconColor} />
-          </button>
+          </F0Box>
         </TooltipTrigger>
         <TooltipContent
           side="top"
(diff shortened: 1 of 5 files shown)


packages/react/src/patterns/OneDataCollection/visualizations/collection/EditableTable/components/cells/BaseCell.tsx
```diff
@@ -29,14 +30,47 @@ export function BaseCell({
...
+          <button
+            type="button"
+            aria-label={hint.message}
+            className={cn(
+              "pointer-events-auto flex shrink-0 cursor-pointer items-center rounded px-1",
+              focusRing()
+            )}
+          >
+            <F0Icon icon={hint.icon} size="md" color={hint.iconColor} />
+          </button>
```
sauldom102: Why an ad-hoc button? We have F0Button

packages/react/src/patterns/OneDataCollection/visualizations/collection/EditableTable/components/cells/index.ts
```diff
@@ -24,7 +24,12 @@ export type EditableCellProps<R extends RecordType> = {
   ) => void
   item: R
   isLastColumn?: boolean
-  hint?: { icon: IconType; message: string; iconColor?: F0IconProps["color"] }
+  hint?: {
+    icon: IconType
+    message: string
+    iconColor?: F0IconProps["color"]
+    position?: "left" | "right"
```
sauldom102, line 31: It's going to be confusing for users if you start placing the icons on the cells in different places every time

packages/react/src/patterns/OneDataCollection/visualizations/collection/EditableTable/components/cells/BaseCell.tsx
RafaMellado: It's a button wrapping the icon, not a real button like that. Maybe it can be a div. This is the example 

Image

packages/react/src/patterns/OneDataCollection/visualizations/collection/EditableTable/components/cells/index.ts
RafaMellado: Captura de pantalla 2026-06-09 a las 16 57 33

we have this problem

In the first column, you can see the 'rellenar' tag having the icon on the left. In other side, we have a correct position in the salary column, that's why we need to put this in different positions

WDTY?

PR #4393, opened 2026-06-09
feat(NotesTextEditor): add PDF/file attachment block

adomingc:
## 🚪 Why?

NotesTextEditor can't attach documents (PDFs, Word, spreadsheets) — users have to host files elsewhere and paste a raw URL. This adds a first-class attachment block.

## 🔑 What?

A `/file` slash command (in *Blocks*) uploads via the consumer's pipeline and renders a Notion-style chip; click to open in a new tab, delete hidden in readonly.

**Feature**
- `FileAttachment/index.tsx` (new) — `fileAttachment` Tiptap atom + React NodeView reusing `FileItem`; exposes `FileUploadConfig`, `insertFileFromFile`, `isSafeAttachmentUrl`.
- Wiring: `fileUploadConfig` prop on `NotesTextEditor`, `/file` command, `richTextEditor.file` i18n key, 2 Storybook stories.

**Hardening + shared infra** — shared with the image uploader via a new `shared/optimisticUpload.ts` helper (both `FileAttachment` and `Image` migrated, dropping the duplicated upload state machine):
- Guard `editor.isDestroyed` after `await` (no throw on unmount mid-upload).
- Reconcile by position instead of `setNodeSelection` (no caret hijack while typing).
- Sanitize `src` in `parseHTML` (unsafe schemes never enter the model).
- Accept by MIME *or* extension (empty/alternate `file.type` no longer silently rejected).
- All failures routed through `onError("upload-failed")`.

## ✅ Verification

- `FileAttachment/index.test.ts` — 15 tests passing; `tsc`/oxfmt/oxlint green.
- Storybook `NotesTextEditor`: Default (`/file` → upload), WithFileAttachment, WithFileAttachmentReadonly.

> **Reviewers:** stored-XSS safety ultimately depends on the backend serving uploads with `Content-Disposition: attachment` / non-HTML `Content-Type`.

https://github.com/user-attachments/assets/95ca302d-c596-43f1-b1e5-539d51d7527a

## 📌 Follow-up

- The *uploading* state renders a small custom chip (`div` + `Spinner` + filename) because `F0FileItem` has no loading state. Consider adding a `loading?: boolean` prop to `F0FileItem` and reusing it here, so the chip is fully consistent and the custom markup disappears.


NabeelahY (approved):
LGTM

PR #4415, opened 2026-06-11
feat(avatars): align F0AvatarModule sizes to canonical scale + stabilize

desiree-np:
## What

Aligns `F0AvatarModule`'s size scale to the canonical avatar names (`xs`/`sm`/`md`/`lg` = 20/24/32/40px) so it is consistent with the rest of the avatar family, and promotes it to **stable** (stable tags, `satisfies Meta`, Snapshot story, MDX docs, unit tests; removes `@experimental`; deterministic gradient id via `useId`).

The two badge-only sub-sizes (`2xs`/`3xs` = 12/10px) are kept for the corner-badge use and documented as not for standalone use.

## Migration (intentional `size` rename — not a blocker)

The `size` prop values were renamed (and shifted in pixels). **All internal consumers are already migrated in this PR.** If you use `F0AvatarModule` directly, remap:

| Before | After |
|---|---|
| `lg` | `md` |
| `md` | `sm` |
| `sm` | `xs` |

Plus a new larger `lg` (40px) and the badge-only `2xs`/`3xs`. The Public API check will flag this as a breaking change — expected, since the API changed on purpose.

## Verification
- Unit tests green; `F0Card` (CardAvatar consumer) 48/48 green.
- Typecheck clean for Module + all 8 consumers; lint/format clean.

### DIFF
--- a/packages/react/src/components/avatars/F0AvatarModule/F0AvatarModule.tsx
+++ b/packages/react/src/components/avatars/F0AvatarModule/F0AvatarModule.tsx
@@ -1,4 +1,5 @@
 import { cva, type VariantProps } from "cva"
+import { useId } from "react"
 
 import { BaseAvatarProps } from "../internal/BaseAvatar"
 import { ModuleId, modules } from "./modules"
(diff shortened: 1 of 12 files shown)


eliseo-juan (approved):
Ask claude about breaking changes, because this could have a lot of them in the monorepo

siguenzaraul (approved):

PR #4421, opened 2026-06-11
fix(dialog-alike): portal center modals to a top-level overlay root

warcos-fact:
## Problem

Center/modal `dialog-alike` dialogs (`F0Dialog`) portal into `#content`, which sits at `z-10` inside `ApplicationFrame`'s `isolate` stacking context — **below** the AI canvas (`z-[21]`) and chat (`z-20`). A dialog's local `z-50` is confined to `#content`'s own stacking context (globally the `z-10` layer), so the AI panels paint over it: **the dialog renders behind the chat/canvas**.

This is a **latent, general bug** (reproducible on `main`, not feature-specific): any center modal opened while the AI canvas/chat is active is occluded. It surfaced in the cocreation flow ([#4307](https://github.com/factorialco/f0/pull/4307)).

## Fix

- **Top-level overlay root**: `LayoutProvider` now renders `<div id="f0-overlay-root">` inside `#f0-layout` — above app content (escapes the ApplicationFrame `isolate` context) but inside the themed layout (f1 tokens + `.dark` still apply). `display:contents` keeps it out of flow; dialogs position with `fixed`.
- **Variant-aware default portal** in `DialogWrapper`: center modals → `#f0-overlay-root`, side drawers → `#content` (unchanged). `DialogContent` gains `defaultContainerId` (default `"content"`), falling back `#f0-overlay-root → #content → document.body`.
- **`container` override prop** threaded through `F0Dialog`/`F0Drawer` → internals → `DialogWrapper` → `DialogContent`, for callers needing a specific portal target.

## Verification

- `tsc --noEmit` clean; pre-commit lint passes.
- `#f0-overlay-root` exists under `#f0-layout` (`display:contents`).
- A center `F0Dialog` portals into `#f0-overlay-root` and renders/themes correctly.
- `F0Drawer` (left/right) does **not** use the overlay root — unchanged (`#content`).

## Related

Unblocks the proper fix for [#4307](https://github.com/factorialco/f0/pull/4307) (cocreation dialogs render behind the AI chat/canvas), which currently uses a `container={null}` workaround. **Merge this first**; that branch then drops the workaround and relies on this default.


developerdanx (changes requested):
Can we add some test and a demo of how this is working please?

developerdanx (approved):

PR #4424, opened 2026-06-11
feat(F0Form): add showOnlySelectedSection styling option

sauldom102:
## Summary

Adds a new `styling.showOnlySelectedSection` option to `F0Form`. When the sections sidepanel is enabled, the form renders only the section selected in the sidepanel instead of stacking all sections — useful for large forms that would otherwise overwhelm the user.

```tsx
<F0Form
  formDefinition={formDefinition}
  styling={{ showSectionsSidepanel: true, showOnlySelectedSection: true }}
/>
```

Also includes three related fixes:
- Sections hidden by `renderIf` no longer appear as sidepanel items (previously, selecting one showed an empty content area).
- The sidepanel is hidden entirely on small viewports (max-width 560px), where sections stack as in the regular layout.
- Row layout now responds to the space actually available to the row via a container query: fields sharing a `row` stack vertically and take the full width when their container is narrower than 480px — regardless of viewport size (previously a viewport breakpoint kept them side by side and cramped next to the sidepanel or inside panels).

## How it works

- Works in both **single-schema** and **per-section** modes. The option is a no-op unless `showSectionsSidepanel` is enabled (the sidepanel is the only way to switch sections).
- Inactive sections are **hidden with CSS rather than unmounted**, so field registration, values, dirty state, and validation are fully preserved while switching — and submit still validates the whole form.
- Clicking a sidepanel item swaps the content in place and resets the container scroll to the top (no anchor scrolling).
- **Error navigation reveals hidden sections**: submitting with an error in a non-visible section (or using the prev/next error arrows in the action bar) switches to the section containing the field before focusing it. Implemented via a new optional `onBeforeFocusField` hook point in `useErrorNavigation`; regular forms keep the exact same synchronous focus behavior as before.
- **`renderIf` filtering**: sidepanel items are derived from the sections whose `renderIf` currently evaluates to true (mirroring what `SectionRenderer` renders), re-evaluated as form values change. The form only subscribes to value changes when conditional sections actually exist, so non-conditional forms keep their current render behavior. If the active section becomes hidden while `showOnlySelectedSection` is on, the form falls back to the first visible section. Note: `renderIf` on sections is a single-schema-mode feature — per-section mode ignores it today (unchanged by this PR).
- **Small screens**: on viewports up to 560px (the same small-screen convention used by dialogs, via `useMediaQuery` from usehooks-ts) the sidepanel is not rendered and `showOnlySelectedSection` is disabled along with it, so all sections stack and remain reachable.
- **Row layout**: each `RowRenderer` declares its own `@container` and switches to horizontal layout at `@[480px]` of container width (same threshold as the previous `xs` viewport breakpoint, now measured where it matters). `w-full` on row children makes stacked fields span the container; in row mode it is inert because `flex-1` sets flex-basis 0, which takes precedence over width for main-axis sizing. Verified in Storybook: at a 600px viewport with the sidepanel the fields stack full-width; on desktop they stay side by side with equal widths.

## Test plan

- [x] 10 new tests covering: section switching in both modes, value preservation across switches, error-driven section reveal on submit, the option being a no-op without the sidepanel, `renderIf`-driven sidepanel filtering, active-section fallback when the selected section becomes hidden, small-screen behavior in both modes, and row layout container-query classes
- [x] Full F0Form suite passes (437 tests, 13 files)
- [x] `tsc --noEmit` and oxlint clean
- [x] New `WithOnlySelectedSection` Storybook story with a 4-section form
- [x] Row fix verified visually in Storybook: 600px viewport + sidepanel (stacks full-width), 400px viewport (stacks full-width), desktop (side by side, equal widths)

### DIFF
--- a/packages/react/src/patterns/F0Form/F0Form.tsx
+++ b/packages/react/src/patterns/F0Form/F0Form.tsx
@@ -136,15 +136,26 @@ function F0FormPerSection<T extends F0PerSectionSchema>(
 
   const sectionIds = useMemo(() => Object.keys(schema), [schema])
 
+  // Only effective when the sidepanel is actually rendered (it provides the
+  // only way to switch between sections).
+  const showOnlySelectedSection =
+    showSectionsSidepanel &&
+    (styling?.showOnlySelectedSection ?? false) &&
+    !!sections &&
+    sectionIds.length > 0
+
   const handleSectionClick = useCallback(
     (sectionId: string) => {
+      // When only the selected section is shown, the content swaps in place —
+      // there is no anchor to scroll to.
+      if (showOnlySelectedSection) return
       const anchorId = generateAnchorId(name, sectionId)
       const element = document.getElementById(anchorId)
       if (element) {
         element.scrollIntoView({ behavior: "smooth", block: "start" })
       }
     },
-    [name]
+    [name, showOnlySelectedSection]
   )
 
   const [activeSection, setActiveSection] = useState<string | undefined>(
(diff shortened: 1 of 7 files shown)


albertpmz (approved):
LGTM! 🚀

PR #4429, opened 2026-06-12
fix(F0CardRow): keep a lone CTA out of the ⋯ menu

warcos-fact:
## Description

A card row whose only action was a single secondary button could collapse entirely into the overflow "⋯" menu — `ButtonGroup`'s overflow hook always reserves the ellipsis-button width, so in a tight row the lone outline button didn't fit and was shed, leaving just an ellipsis. The root cause was forcing a card's only CTA into the *secondary* slot purely for outline styling (secondaries shed; the primary is pinned). This PR lets a lone CTA live in the pinned primary slot while still rendering as an outline button, and takes the chance to move `F0CardRow` to the Experimental section.

See **Storybook → Experimental/Card Row → Secondary Primary Action**: the lone "Download" action renders as a pinned outline button and stays visible at every width (verified down to a ~180px row — no "⋯").

## Implementation details

- fix: add `variant: "primary" | "secondary"` to `CardPrimaryAction` so a lone CTA can render as an outline button while staying pinned (never sheds into the ⋯ menu)

  <details>
  <summary>Why a new prop instead of fixing overflow?</summary>

  Secondary actions are the ones that shed under width pressure; the primary is always pinned. The bug came from putting the card's only action in the secondary slot just to get outline styling. Letting the *primary* carry an outline variant fixes it at the source — the action lives in the pinned slot and is never measured for overflow.
  </details>

- fix: plumb a `primaryVariant` prop through `ButtonGroup` (applied in both the row and stacked branches); the full `F0Card` honours the variant on its primary `F0Button` for consistency
- refactor: move `F0CardRow` (and its exclusive `CardRowActions`) into `src/experimental/F0CardRow`, so it surfaces under the Experimental Storybook section with the 🚧 maturity tag; shared `F0Card` primitives (`CardActions`, `CardAlert`, `CardAvatar`, `types`) stay put and are imported via `@/` aliases
- chore: keep a `@deprecated` re-export at `components/F0Card` for backward compatibility, add the canonical export to `experimental/exports.ts`, and point `F0HILActionConfirmation` at the new path
- test: existing F0CardRow suite (17 tests) moves with the component and passes; added a `SecondaryPrimaryAction` story


developerdanx (approved):

warcos-fact:
Superseded by #4441, which contains these exact commits (the F0CardRow → Experimental move + the lone-CTA outline fix) plus the opt-in whole-row click-target restore. Consolidating all F0CardRow work into a single PR to avoid duplicate/conflicting merges. No work lost — it all ships via #4441.

PR #4438, opened 2026-06-12
fix(ButtonGroup): never collapse or squeeze a confirm/reject pair

warcos-fact:
## Description

The accept/reject CardRow (`F0HILActionConfirmation`) routed its pair through `ButtonGroup` as `primaryAction={confirm}` + `secondaryActions={[reject]}`, which inherited the row's width-driven overflow machinery. As a result, in the inline icon-only layout the reject (✗) button collapsed into the "⋯" menu and the confirm (✓) button squeezed to nothing in the tight band just above the stack breakpoint — leaving an ellipsis with no visible primary. This standardizes a "never collapse / never squeeze" behaviour in `ButtonGroup` itself (rather than hand-rendering the pair) so the confirm/reject row always shows both buttons inline.

## Implementation details

- feat: add a `canOverflow` prop to `ButtonGroup` (default `true`); when `false`, secondary buttons never shed into the "⋯" menu — they always render inline and the hidden measurement copy is skipped (`otherActions` still populate the menu when present)

  <details>
  <summary>Why?</summary>

  `useOverflowCalculation` always reserves the ellipsis-button width upfront, so a lone icon-only secondary never fits and collapses entirely. A confirm/reject pair must never collapse, so it opts out of overflow rather than relying on the width heuristic.
  </details>

- fix: pin row item widths with `shrink-0` (cluster children plus the trailing primary/split buttons) so a `ButtonGroup` overflows by shedding into the menu, never by squeezing a button to invisibility — benefits every call site; the first child (the `flex-1` cluster) keeps its shrink so the title truncates against it

- fix: opt the `CardRowActions` confirm/reject branch into `canOverflow={false}` so the reject (✗) button can never shed into the "⋯" menu

## Verification

- Inline band (~470px): both ✗ and ✓ render at full size, no ellipsis
- Below the stack breakpoint: labelled stacked layout unchanged
- Regression: default `canOverflow` still sheds secondaries into "⋯", with no squeezing
- `tsc` clean, lint clean, `F0Card` unit tests 48/48 pass

### DIFF
--- a/packages/react/src/components/F0Card/components/CardRowActions.tsx
+++ b/packages/react/src/components/F0Card/components/CardRowActions.tsx
@@ -233,6 +233,9 @@ export function CardRowActions({
           secondaryActions={reject ? [reject] : undefined}
           size={size}
           gap={GAP}
+          // The confirm/reject pair must always stay inline — the reject button
+          // must never shed into the "⋯" menu.
+          canOverflow={false}
         />
       )
     }
--- a/packages/react/src/ui/ButtonGroup/ButtonGroup.tsx
+++ b/packages/react/src/ui/ButtonGroup/ButtonGroup.tsx
@@ -103,6 +103,12 @@ export interface ButtonGroupProps {
   fullWidthOnStack?: boolean
   /** Reverse the stacked column so the primary lands on top. */
   reverseOnStack?: boolean
+  /**
+   * When `false`, secondary buttons never shed into the "⋯" menu — they always
+   * render inline (e.g. a confirm/reject pair that must never collapse).
+   * `otherActions` still populate the menu when present. @default true
+   */
+  canOverflow?: boolean
   className?: string
 }
 
@@ -190,6 +196,7 @@ interface ButtonGroupBranchProps {
   size: ButtonSize
   gap: number
   align: "end" | "between"
+  canOverflow: boolean
 }
 
 /**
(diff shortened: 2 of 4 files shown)


developerdanx (changes requested):
Can we add some tests here please?

warcos-fact:
Added tests + a visual demo (per the review). 🟢

**Tests** — `src/ui/ButtonGroup/__tests__/ButtonGroup.test.tsx` (ButtonGroup had no tests; the `CardRowActions` change is already covered by `F0CardRow.test`, so tests target ButtonGroup). jsdom has no layout, so the width-driven overflow can't run — the tests cover the `canOverflow` branch, which is pure render logic:
- `canOverflow={false}` keeps every secondary inline, renders no `⋯` trigger, and skips the hidden measurement copy.
- by default the group can shed into the `⋯` menu.
- `otherActions` still surface in the menu even when `canOverflow={false}`.

**Demo** — new behaviors story **ButtonGroup/Patterns → “Confirm/reject never collapses”**, an icon-only pair in a 96px box:
- **Issue (default):** the reject sheds into the `⋯` menu, leaving `⋯ + ✓`.
- **Fix (`canOverflow={false}`):** `✗ + ✓` stay inline.

See it in the **Chromatic** visual review linked above (story “Confirm/reject never collapses”). Screenshot of the two states is attached on the PR description.
image-1781522129882

developerdanx (approved):

PR #4443, opened 2026-06-13
fix(datacollection): keep kanban lanes from sticking on loading skeletons

ReddyWish:
## Changes
- **Kanban visualization** — lane skeleton now keys on `isInitialLoading` instead of `isLoading`. A lane that has resolved once never re-shows a full skeleton on a background refetch or when `isLoading` is stranded.
- **`ui/Kanban`** — read-only display state (`loading`, `total`, `hasMore`, `loadingMore`, `fetchMore`) is read live from the `lanes` prop at render; `localLanes` now owns only drag item-order. Fixes empty lanes never clearing their skeleton (their item-signature never changes, so the old snapshot never re-synced) and keeps optimistic drag intact.

## Why
On any `OneDataCollection` kanban backed by `cache-and-network`, a lane could stay on a loading skeleton after its data loaded — counter correct, cards stuck — until the user toggled views. Affects all kanban consumers.

## Testing
- Added regression tests: a lane whose source emits `{loading:true, data}` and never settles still renders its cards; an empty lane that settles `loading → false` clears its skeleton.
- OneDataCollection + datasource suites green (635 tests); tsc / oxlint / oxfmt clean.

### DIFF
--- a/packages/react/src/patterns/OneDataCollection/visualizations/collection/Kanban/Kanban.tsx
+++ b/packages/react/src/patterns/OneDataCollection/visualizations/collection/Kanban/Kanban.tsx
@@ -215,7 +215,7 @@ export const KanbanCollection = <
           variant: l.variant,
           total: totalItems,
           hasMore,
-          loading: laneData?.isLoading ?? true,
+          loading: laneData ? laneData.isInitialLoading : true,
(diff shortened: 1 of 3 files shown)


eliseo-juan (approved):
LGTM

PR #4457, opened 2026-06-15
fix(F0CardRow): restore opt-in whole-row click target

warcos-fact:
## Description

#4376 removed the whole-row click target from `F0CardRow` entirely (along with
the `link` / `onClick` / `disableOverlayLink` props), which broke consumers whose
entire card surface is the action (e.g. `F0AiChatWelcomeCards`).

This restores it, but **opt-in**: the pointer cursor, hover/focus affordance, and
overlay link only apply when `link` or `onClick` is set. Cards that pass neither
stay exactly as #4376 left them — static rows whose only interactive parts are
their actions.

N/A — behavioural change. Verify under **Components / F0CardRow**: a row with
`onClick`/`link` shows a pointer cursor + hover border and is clickable; a row
with neither stays static.

## Implementation details

- Re-add `link`, `onClick`, `disableOverlayLink` to `F0CardRowProps`.
- Gate the affordance on `const clickable = !!link || !!onClick` — pointer cursor +
  `hover`/`focus-within` border/shadow + the absolute overlay `F0Link` only render
  when clickable.
- Restore `CardRowActions`' click-stops (`stopPropagation`) so action clicks don't
  bubble to a now-clickable row.
- Restore the `link` / `onClick` argTypes in the story.

> Note: branches off `main`, where `F0CardRow` still lives in `components/F0Card`.
> #4441 moves it to `experimental/F0CardRow`; whichever lands second needs a
> trivial rebase to the new path.

### DIFF
--- a/packages/react/src/components/F0Card/F0CardRow.tsx
+++ b/packages/react/src/components/F0Card/F0CardRow.tsx
@@ -1,9 +1,10 @@
 import { forwardRef } from "react"
 
+import { F0Link } from "@/components/F0Link"
 import { DropdownItem } from "@/experimental/Navigation/Dropdown"
 import { withDataTestId } from "@/lib/data-testid"
 import { withSkeleton } from "@/lib/skeleton"
-import { cn } from "@/lib/utils"
+import { cn, focusRing } from "@/lib/utils"
 import { Card } from "@/ui/Card"
 import { Skeleton } from "@/ui/skeleton"
 import { Text } from "@/ui/Text"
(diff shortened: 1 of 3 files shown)


warcos-fact:
Closing as redundant: the opt-in whole-row click target this PR restores already landed on `main` via #4441 (`refactor(F0CardRow): mark experimental via wrapper + lone-CTA outline + opt-in row click`). This branch was cut before #4441 merged, so it adds nothing new and would revert #4441's `experimentalComponent` marking. Superseded by #4441.

PR #4470, opened 2026-06-16
feat(F0Form): add native period field type

aaroncaballero-f:
## Description

Adds a native `period` field type to F0Form for selecting a period by granularity (Year / Half year / Quarter / Month / Range). It renders the existing `F0DatePicker` but keeps the full `DatePickerValue` (`{ value: { from, to }, granularity }`) as the form value instead of collapsing it to a single `Date` like the `date` field, so the chosen granularity and range survive. This removes the need for the `fieldType: "custom"` + `renderCustomField` escape hatch previously used to embed a period selector — the form now wires label, required asterisk, size, error and validation natively.

## Storybook

No static screenshots attached — the field renders via Storybook (Chromatic captures the visual diffs):

- `Forms/F0Form` → **PeriodField** (also included in the `AllFieldTypes` canvas under the f0FormField API docs)
- `Forms/F0WizardForm` → **WithPeriodField**

## Implementation details

- feat: add `period` field type (`F0PeriodField` / `F0PeriodConfig` / renderer) that preserves the full `DatePickerValue` rather than collapsing it to a `Date`
- feat: add `f0FormField` overload + `F0PeriodFieldConfig` for `z.custom<DatePickerValue>()` with `fieldType: "period"`
- feat: expose `granularities`, `presets`, `minDate`/`maxDate` (static or dynamic) and `displayFormat`
- feat: add an opt-in `selectOnCellOnly` prop (default `false`) threaded through OneCalendar → DatePickerPopup → F0DatePicker; when `true`, switching granularity only changes the view and the value is committed/closed only on a cell click. The `period` field enables it.

  <details>
  <summary>⚠️ Touches Foundations-owned primitives</summary>

  `selectOnCellOnly` is added to `OneCalendar`, `DatePickerPopup` and `F0DatePicker`. The default (`false`) preserves the current behavior for every existing consumer (date/daterange fields, date navigators, reporting), so it is backward-safe — but it does modify shared calendar primitives and needs a Foundations look.
  </details>

- chore: wire `period` through `renderFieldInput`, `useSchemaDefinition` and the public exports
- test: unit tests for the field (inference, config propagation, clearable, required asterisk, full-value round-trip) and for `selectOnCellOnly` (OneCalendar + DatePickerPopup)
- docs: F0Form / F0WizardForm stories + a "Period field" section in the F0Form MDX

  <details>
  <summary>Why a dedicated type instead of a flag on \`date\`?</summary>

  The `date` field's contract is a `Date` value. Preserving the period would force its value to become a `DatePickerValue` behind a runtime flag (polymorphic value), break its `Date`-based renderIf/constraints, and misreport the type in schema introspection. A dedicated `period` type keeps a clean value contract; `F0DatePicker` is not modified.
  </details>


packages/react/src/patterns/F0Form/__stories__/F0Form.mdx
```diff
@@ -286,6 +286,28 @@ const formSchema = z.object({
...
 The base call accepts any Zod schema (`z.string()`, `z.number()`, `z.date()`, `z.enum()`, `z.object()`, `z.array()`, etc.) and any supported field config.
 
+## Period field
+
+`period` is a granularity-aware period selector (Year / Half year / Quarter / Month / Range). Unlike `date` — which stores a single `Date` and re-displays it with the first granularity — a `period` field keeps the **full value** (`{ value: { from, to }, granularity }`), so the chosen granularity and range are preserved.
+
+It has no shortcut: declare it with the base call and `z.custom<DatePickerValue>()`. Label, the required asterisk, size, and error handling are wired by the form automatically — no manual re-wiring.
+
+```tsx
+goalPeriod: f0FormField(z.custom<DatePickerValue>().optional(), {
```
sauldom102: Why not declaring it with `z.object({ value: { from, to }, granularity })` instead? Could we also declare its "shortcut" like we are doing with other fields? (https://f0.factorial.dev/?path=/docs/patterns-forms-f0form--documentation#using-shortcuts-recommended) So that we could do something like `f0FormField.datePeriod({ ... })` and we avoid to import `DatePickerValue`

aaroncaballero-f: done, switched to a z.object({ value, granularity }) with a f0FormField.datePeriod(...) shortcut, so no DatePickerValue import needed and required now validates properly
thanks!

sauldom102 (approved):

PR #4471, opened 2026-06-16
fix(SurveyFormBuilder): foreground input text in dark mode

warcos-fact:
## Description

In dark mode, the typed-in value of the question title, section title, and select-option inputs rendered as black instead of the theme foreground (white). These form controls set a text color only for the disabled state (or none at all), and because Tailwind Preflight is disabled in this repo, form controls don't inherit the page's `text-f1-foreground` — they fall back to the browser's default (black) text color regardless of theme. This sets the foreground color explicitly so the inputs read correctly in both light and dark mode.

## Implementation details

- fix: set `text-f1-foreground` on the question title, section title, and select-option inputs so their value text uses the theme foreground in dark mode
- fix: tokenize the placeholder with `placeholder:text-f1-foreground-tertiary`, matching the description and checkbox inputs that already rendered correctly


warcos-fact:
image (2)
image (3)

eliseo-juan (approved):
LGTM

PR #4472, opened 2026-06-17
feat(value-display): add categoryBarChart cell renderer

AndresRojoFactorial:
## Why

OneDataCollection tables had no way to render proportional distribution bars (e.g. gender breakdown, work-location split). The existing `barSeries` cell renders vertical bars — useful for time-series, not for showing parts-of-a-whole. The standalone `CategoryBarChart` kit component exists but can't be used inside table cells because `valueDisplayRenderers` only accepts registered cell types.

## What

New `categoryBarChart` value-display cell type that renders a horizontal stacked proportional bar inside OneDataCollection table columns.

### Added
- **`categoryBarChart.tsx`** — Cell renderer: horizontal proportional bar with per-segment tooltips, categorical/custom color support, empty-state fallback (`–`). Tooltip interaction adapted for table cells (`pointer-events-auto`, `delayDuration`, `disableHoverableContent`, `z-[100]`) following the `barSeries` pattern.
- **`renderers.tsx`** — Registered as `categoryBarChart` in `valueDisplayRenderers`.
- **`resolveColor()`** — Helper that accepts both raw CSS values (`#hex`, `rgb()`, `hsl()`) and f0 theme token names (e.g. `categorical-1`). `getColor()` only handles tokens; passing hex directly produced `hsl(var(--chart-#F5A51C))` → invisible bars.
- **9 unit tests** — empty, undefined, segments, zero total, skip-zero, percentage format, custom colors, hideTooltip, duplicate names.
- **5 stories** — GenderDistribution, WorkLocation, WithCustomColors, SingleCategory, Empty.
- **MDX docs** — Value type reference with Canvas examples.

### Design decisions
- **Standalone implementation, not a kit wrapper** — Same visual style as `CategoryBarChart` kit (same colors, gap, tooltip layout) but adapted to the cell renderer signature `(args, meta) => JSX`. Keeps the value-display layer self-contained like all other cell types.
- **`gap-1`** — Matches the kit's segment spacing for visual consistency.
- **Individual `TooltipProvider` per segment** — Required for hover to work inside OneDataCollection table rows (table captures pointer events). Same pattern as `barSeries`.
- **Raw CSS color passthrough** — Consumers pass domain-specific hex colors (e.g. gender: `#F5A51C`, location: `#07A2AD`); the renderer uses them directly instead of forcing theme tokens. Falls back to `getCategoricalColor(index)` when no color is provided.

### Usage

```tsx
{
  type: 'categoryBarChart',
  value: {
    dataPoints: [
      { name: 'Female', value: 12, color: '#6B9CA8' },
      { name: 'Male', value: 8, color: '#F5A51C' },
    ],
  },
}
```

### Gender distribution in table cell
Screenshot 2026-06-17 at 12 09 14

### Empty state
Screenshot 2026-06-17 at 12 08 11

### Storybook
Screenshot 2026-06-17 at 12 11 45

## Test plan

- [x] 9 unit tests passing (empty, undefined, segments, zero-total, skip-zero, percentage, custom-color, hideTooltip, duplicate-names)
- [x] All 60 value-display tests passing (no regressions on barSeries, hourDistribution, compound, icon)
- [x] Linter: 0 warnings, 0 errors (oxlint)
- [x] Format: oxfmt clean
- [x] Cycle dependencies: no new cycles
- [x] Tested in Factorial monorepo (slot-1) with patched `node_modules` — renders correctly in TeamsListPage Gender column with hex colors
- [ ] Storybook visual review

### DIFF
--- a/packages/react/src/ui/value-display/types/categoryBarChart/categoryBarChart.tsx
+++ b/packages/react/src/ui/value-display/types/categoryBarChart/categoryBarChart.tsx
@@ -23,6 +23,17 @@ export interface CategoryBarChartCellValue {
 
 const CELL_MIN_HEIGHT_PX = 40
 
+function resolveColor(color: string | undefined, index: number): string {
+  if (!color) return getCategoricalColor(index)
+  if (
+    color.startsWith("#") ||
+    color.startsWith("rgb") ||
+    color.startsWith("hsl")
+  )
+    return color
+  return getColor(color)
+}
+
 function formatPercentage(value: number, total: number): string {
   const pct = (value / total) * 100
   return pct % 1 === 0 ? pct.toFixed(0) : pct.toFixed(1)
@@ -72,9 +83,7 @@ export const CategoryBarChartCell = (
       <div className="flex h-2 w-full gap-1 overflow-hidden">
         {dataPoints.map((point, index) => {
           const percentage = (point.value / total) * 100
-          const color = point.color
-            ? getColor(point.color)
-            : getCategoricalColor(index)
+          const color = resolveColor(point.color, index)
 
           if (percentage === 0) return null
 
--- a/packages/react/src/ui/value-display/types/categoryBarChart/categoryBarChart.tsx
+++ b/packages/react/src/ui/value-display/types/categoryBarChart/categoryBarChart.tsx
@@ -23,17 +23,6 @@ export interface CategoryBarChartCellValue {
 
 const CELL_MIN_HEIGHT_PX = 40
 
-function resolveColor(color: string | undefined, index: number): string {
-  if (!color) return getCategoricalColor(index)
-  if (
-    color.startsWith("#") ||
-    color.startsWith("rgb") ||
-    color.startsWith("hsl")
-  )
-    return color
-  return getColor(color)
-}
-
 function formatPercentage(value: number, total: number): string {
   const pct = (value / total) * 100
   return pct % 1 === 0 ? pct.toFixed(0) : pct.toFixed(1)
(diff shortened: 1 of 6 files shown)


rferrando (approved):

packages/react/src/ui/value-display/types/categoryBarChart/categoryBarChart.tsx
```diff
@@ -0,0 +1,135 @@
...
+}
+
+export interface CategoryBarChartCellValue {
+  dataPoints: CategoryBarDataPoint[]
+  hideTooltip?: boolean
+}
+
+const CELL_MIN_HEIGHT_PX = 40
+
+function resolveColor(color: string | undefined, index: number): string {
```
siguenzaraul: Do we need this amount of flexibility? or we can just use f0 tokens?

siguenzaraul:
Probably we need to use another color for the title, to dark

Captura de pantalla 2026-06-18 a las 12 00 46

packages/react/src/ui/value-display/types/categoryBarChart/categoryBarChart.tsx
AndresRojoFactorial: It is not need, I'm changing it to use f0 tokens.

AndresRojoFactorial:
@siguenzaraul I agree the tooltip title color doesn't look great — it's too dark against the tooltip background. However, this is not specific to the categoryBarChart: it comes from the shared `TooltipContent` component in `src/ui/tooltip.tsx`, which applies the `dark` class to force dark-mode styling on all tooltips globally. Fixing the contrast for the label text here would mean overriding the tooltip's own color scheme in this one component, while the underlying issue affects every tooltip that renders secondary text.

I'd prefer to address this in a separate PR that improves the `TooltipContent` defaults (or introduces a dedicated secondary-text token for dark tooltips) so all consumers benefit. For now I've simplified the tooltip setup in this component (single `TooltipProvider`, relying on default positioning/z-index) to keep it aligned with the shared component behavior.

Let me know if you'd rather I apply a local fix here in the meantime.

siguenzaraul (approved):

PR #4488, opened 2026-06-17
feat(OneTable): support structured rich header info

zygisS22:
## Why

`OneTable` / `OneDataCollection` column headers expose `info` as a plain string, rendered as a short dark text tooltip. Catalog-backed columns need a richer header affordance with a title, description, and optional "Learn more" link, while keeping the design-system surface and content structure controlled by F0.

## What

`TableHead.info` and `PropertyDefinition.info` now accept `string | TableHeaderInfo`:

- `string` -> unchanged short text tooltip.
- `TableHeaderInfo` -> structured hoverable card rendered by F0.

```ts
type TableHeaderInfo = {
  title: string
  description: string
  link?: {
    label: string
    onClick: () => void
  }
  // Accessible name for the info icon. Defaults to the column label.
  label?: string
}
```

F0 owns the hover surface, content layout, inverse text styling, focus ring, and dismiss behavior. Consumers provide only the data/action fields.

## Why HoverCard

The card is hover-revealed but can contain an interactive link action, so it uses F0's `HoverCard`, not `Tooltip`. The legacy string `info` path still uses the existing non-interactive tooltip and remains backward compatible.

## Accessibility

- `label` controls the info icon accessible name; when omitted, string column labels are used.
- The link action is a real button with F0 focus-ring styling.
- Clicking the link closes the hover card before invoking the consumer action.

## Testing

- `pnpm --filter @factorialco/f0-react run format`
- `pnpm --filter @factorialco/f0-react run tsc`
- `pnpm --filter @factorialco/f0-react run lint`
- `pnpm run vitest --project=unit --reporter=dot src/experimental/OneTable/TableHead/__tests__/TableHead.test.tsx` from `packages/react`

## Consumer

Unblocks the Factorial AI Reports feature by allowing semantic dashboard columns to pass structured catalog metadata and a `Learn more` action. The Factorial PR should consume this structured object rather than a custom render callback.

### DIFF
--- a/packages/react/src/experimental/OneTable/index.stories.tsx
+++ b/packages/react/src/experimental/OneTable/index.stories.tsx
@@ -246,6 +246,9 @@ export const InfoHeader: Story = {
   ),
 }
 
+// `info` only provides the hover surface — the consumer composes the card body
+// (here: title, muted meta line, description, and a "Learn more" action that
+// uses `close` to dismiss the card before opening another surface).
 export const RichInfoHeader: Story = {
   render: () => (
     <OneTable>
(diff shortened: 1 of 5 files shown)


sauldom102 (changes requested):
The "Learn more" link is not visible at all

Image

zygisS22:
> The "Learn more" link is not visible at all
> 
> Image

Whoops, sorry 😓 

fixing it ASAP

zygisS22:
@sauldom102 just fixed it, was rendering with the same colour as the background so it was invisible 😞 
image

Could you review it again plss!

packages/react/src/experimental/OneTable/index.stories.tsx
```diff
@@ -246,6 +246,65 @@ export const InfoHeader: Story = {
...
+                      close()
+                      alert("Open data catalog on this field")
+                    }}
+                    className="mt-1 w-fit rounded-xs font-medium text-f1-foreground-inverse underline underline-offset-2 transition-colors hover:text-f1-foreground-inverse-secondary focus-visible:outline-none focus-visible:ring-1 focus-visible:ring-f1-special-ring"
+                  >
+                    Learn more
+                  </button>
+                </div>
+              ),
+            }}
```
sauldom102: Let's avoid custom children/render, let's just better specify title, description and link as separate props here, we don't want any custom thing to be here

zygisS22: Makes sense, just fixed and test it locally 👍🏿

sauldom102 (approved):

PR #4491, opened 2026-06-18
feat(F0SegmentedBar): add segmented score/rating bar

ReddyWish:
A focused, standalone primitive for displaying a value as a discrete,
segmented bar. First consumer is the ATS competency side panel
(red/orange/green per level).

[FoSegmentedBar documentation](https://66a7a8d7d124220c363457cc-vicyelppmb.chromatic.com/?path=/docs/experimental-f0segmentedbar--documentation)

## Changes
- `experimental/F0SegmentedBar/` — `F0SegmentedBar.tsx`, `types.ts`, `index.tsx`
  - `value`/`max` segmented bar, `color?: SegmentColorToken`, optional `label`
  - `value` floored and clamped to `[0, max]`; non-positive/non-finite
    `max` renders an empty bar (no NaN, no Infinity crash)
  - `role="progressbar"` with i18n'd `aria-valuetext`; `aria-label` from
    `label` with a translated fallback; `motion-reduce:transition-none`
  - track + pill styling aligned with the existing `Progress` primitive
- `lib/providers/i18n` — `segmentedBar.label`, `segmentedBar.valueText` defaults
- `experimental/exports.ts` — export `F0SegmentedBar`
- Stories (`tags: ["experimental"]`) + MDX docs (Overview/Guidelines/Code/Examples)

Screenshot 2026-06-19 at 09 28 12

Screenshot 2026-06-19 at 09 28 54

### DIFF
--- a/packages/react/src/lib/providers/i18n/i18n-provider-defaults.ts
+++ b/packages/react/src/lib/providers/i18n/i18n-provider-defaults.ts
@@ -707,6 +707,10 @@ export const defaultTranslations = {
     submit: "Submit",
     stepOf: "Step {{current}} of {{total}}",
   },
+  segmentedBar: {
+    label: "Segmented bar",
+    valueText: "{{value}} of {{max}}",
+  },
 } as const
 
 type TranslationShape<T> = {
--- a/packages/react/src/lib/providers/i18n/i18n-provider-defaults.ts
+++ b/packages/react/src/lib/providers/i18n/i18n-provider-defaults.ts
@@ -707,10 +707,6 @@ export const defaultTranslations = {
     submit: "Submit",
     stepOf: "Step {{current}} of {{total}}",
   },
-  segmentedBar: {
-    label: "Segmented bar",
-    valueText: "{{value}} of {{max}}",
-  },
 } as const
 
 type TranslationShape<T> = {
--- a/packages/react/src/experimental/F0SegmentedBar/F0SegmentedBar.tsx
+++ b/packages/react/src/experimental/F0SegmentedBar/F0SegmentedBar.tsx
@@ -26,13 +26,13 @@ const F0SegmentedBarBase = forwardRef<HTMLDivElement, F0SegmentedBarProps>(
       <div
         ref={ref}
         role="progressbar"
-        aria-label={label ?? i18n.t("segmentedBar.label")}
+        aria-label={label}
         aria-valuemin={0}
         aria-valuemax={segmentCount}
         aria-valuenow={filled}
-        aria-valuetext={i18n.t("segmentedBar.valueText", {
-          value: filled,
-          max: segmentCount,
+        aria-valuetext={i18n.t("audioPlayer.position", {
+          current: filled,
+          total: segmentCount,
         })}
         className={cn("flex h-2 w-full gap-1")}
       >
--- a/packages/react/src/experimental/F0SegmentedBar/F0SegmentedBar.tsx
+++ b/packages/react/src/experimental/F0SegmentedBar/F0SegmentedBar.tsx
@@ -1,0 +1,61 @@
+import { forwardRef } from "react"
+
+import { getColor } from "@/kits/Charts/utils/colors"
+import { withDataTestId } from "@/lib/data-testid"
+import { experimentalComponent } from "@/lib/experimental"
+import { useI18n } from "@/lib/providers/i18n"
+import { cn } from "@/lib/utils"
+
+import { F0SegmentedBarProps } from "./types"
+
+export type { F0SegmentedBarProps, SegmentColorToken } from "./types"
+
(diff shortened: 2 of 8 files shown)


sauldom102 (commented):
We already have this component https://f0.factorial.dev/?path=/docs/kits-charts-progresschart--documentation that is also meant to display a progress, maybe we should better adapt that one to be segmented instead of creating a new one

ReddyWish:
> We already have this component https://f0.factorial.dev/?path=/docs/kits-charts-progresschart--documentation that is also meant to display a progress, maybe we should better adapt that one to be segmented instead of creating a new one

@sauldom102 Yeah, I actually looked at ProgressChart before going with a separate component. The thing is they're not really the same — ProgressChart (and the other bars in f0) are continuous, just one fill at value/max%. F0SegmentedBar is discrete: N equal on/off segments for stuff like "competency level 3 of 5", so max is the segment count, not a percentage denominator. That difference also changes the a11y story. ProgressChart just sets aria-label="60.0%", so a screen reader announces a percentage, which is what you want for a continuous bar, but for a discrete rating that's kinda misleading; "60%" isn't what the bar means. So F0SegmentedBar uses the proper progressbar pattern: aria-valuemin/aria-valuemax/aria-valuenow for the raw numbers, plus aria-valuetext to override the default numeric read-out so it announces "3 of 5" instead of a bare number or a percentage. Bolting that onto ProgressChart would mean branching its a11y behavior by variant. so i thought that maybe it is better to keep it simple and separate then complicate it.

I asked my agent to generate comparison table and it looks smth like this

| | `ProgressChart` | `F0SegmentedBar` |
|---|---|---|
| **Semantics** | continuous proportion — "60% of the way there" | discrete ordinal — "level 3 of 5" |
| **`max` means** | denominator for a percentage (default 100) | the number of segments |
| **Render** | one `Progress` fill grown to `value/max%` | N equal pills, each on/off |
| **Layer** | Charts kit — carries `ChartConfig`/`ChartPropsBase` generics | plain UI primitive |
| **a11y** | `aria-label="60.0%"` | `role=progressbar` + `aria-valuetext` "3 of 5" |

Let me know wdyt please

packages/react/src/lib/providers/i18n/i18n-provider-defaults.ts
```diff
@@ -702,6 +702,10 @@ export const defaultTranslations = {
     submit: "Submit",
     stepOf: "Step {{current}} of {{total}}",
   },
+  segmentedBar: {
+    label: "Segmented bar",
+    valueText: "{{value}} of {{max}}",
+  },
```
sauldom102: I would get rid of these two:
1. For the valueText: we already have audioPlayer.position which already is `"{{current}} of {{total}}"`
2. For the label: I would always require the label in the component, not allowing it to be optional, if we want to hide it we could include a hideLabel optional boolean prop there

sauldom102:
> > We already have this component https://f0.factorial.dev/?path=/docs/kits-charts-progresschart--documentation that is also meant to display a progress, maybe we should better adapt that one to be segmented instead of creating a new one
> 
> @sauldom102 Yeah, I actually looked at ProgressChart before going with a separate component. The thing is they're not really the same — ProgressChart (and the other bars in f0) are continuous, just one fill at value/max%. F0SegmentedBar is discrete: N equal on/off segments for stuff like "competency level 3 of 5", so max is the segment count, not a percentage denominator. That difference also changes the a11y story. ProgressChart just sets aria-label="60.0%", so a screen reader announces a percentage, which is what you want for a continuous bar, but for a discrete rating that's kinda misleading; "60%" isn't what the bar means. So F0SegmentedBar uses the proper progressbar pattern: aria-valuemin/aria-valuemax/aria-valuenow for the raw numbers, plus aria-valuetext to override the default numeric read-out so it announces "3 of 5" instead of a bare number or a percentage. Bolting that onto ProgressChart would mean branching its a11y behavior by variant. so i thought that maybe it is better to keep it simple and separate then complicate it.
> 
> I asked my agent to generate comparison table and it looks smth like this
> 
> `ProgressChart`	`F0SegmentedBar`
> **Semantics**	continuous proportion — "60% of the way there"	discrete ordinal — "level 3 of 5"
> **`max` means**	denominator for a percentage (default 100)	the number of segments
> **Render**	one `Progress` fill grown to `value/max%`	N equal pills, each on/off
> **Layer**	Charts kit — carries `ChartConfig`/`ChartPropsBase` generics	plain UI primitive
> **a11y**	`aria-label="60.0%"`	`role=progressbar` + `aria-valuetext` "3 of 5"
> Let me know wdyt please

Fair enough, let's keep them separate, I'm just a bit more concerned about having a similar chart-like component outside of the chart kit, but I would prefer having it in experimental first so it's fine!

sauldom102 (approved):
LGTM! 🚀 🚀 🚀

PR #4502, opened 2026-06-18
feat(SurveyFormBuilder): redesign the blocked section

warcos-fact:
## Description

Redesigns the **blocked (locked) section** in the `SurveyFormBuilder`. Extracted from #4307 (co-creation patterns) so the SurveyFormBuilder/`F0TagRaw` changes can be reviewed and merged on their own; #4307 is stacked on top of this branch.

A locked section now reads as a single muted grey rounded panel with a white "Locked" tag in its title, replacing the previous explanatory alert.

| Light | Dark |
| --- | --- |
| Grey `rounded-2xl` panel wrapping the section header + its question cards, white "Locked" tag at the right of the title, white default cards inside. | Same treatment using semantic tokens. |

_(Storybook: **Surveys → SurveyFormBuilder → Default**.)_

## Implementation details

- **Whole-section panel**: a locked section's header + its questions are grouped into one `rounded-2xl bg-f1-background-secondary` wrapper (`Form/index.tsx`). Verified safe inside `Reorder.Group` (framer-motion orders by measured position + context, not DOM nesting; locked items aren't draggable).
- **Alert → "Locked" tag**: the section notice/alert is removed; a white `F0TagRaw` (lock icon + "Locked") sits at the right of the title. The section description is no longer shown inline — it's surfaced as an **instant tooltip** on the tag.
- **Cards**: locked question cards roll back to the default white card styling, drop the hover-border effect, and force the `not-allowed` cursor on every descendant (`[&_*]:!cursor-not-allowed`). Right padding mirrors the drag-and-drop gutter so all cards share the same width.
- **`F0TagRaw`**: adds an optional `className` passthrough (backward compatible) so the tag can render white.

Builder-only: the answering/preview renderer (`SurveyAnsweringForm`) is untouched and shows locked sections as regular questions.

### DIFF
--- a/packages/react/src/components/tags/F0TagRaw/F0TagRaw.tsx
+++ b/packages/react/src/components/tags/F0TagRaw/F0TagRaw.tsx
@@ -2,13 +2,17 @@ import { forwardRef } from "react"
 
 import { F0Icon } from "@/components/F0Icon"
 import { useTextFormatEnforcer } from "@/lib/text"
+import { cn } from "@/lib/utils"
 
 import type { F0TagRawProps } from "./types"
 
 import { BaseTag } from "../internal/BaseTag"
 
 export const F0TagRaw = forwardRef<HTMLDivElement, F0TagRawProps>(
-  ({ text, additionalAccessibleText, icon, onlyIcon, info }, ref) => {
+  (
+    { text, additionalAccessibleText, icon, onlyIcon, info, className },
+    ref
+  ) => {
     useTextFormatEnforcer(
       text,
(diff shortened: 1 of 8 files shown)


warcos-fact:
Closing: the blocked-section redesign can't be cleanly extracted onto main — it depends on cocreation-branch additions (the question-level `notice` prop and the `surveyFormBuilder.labels.locked` i18n key) that aren't on main, and pulling those in cascades into broader cocreation work. Keeping this work in #4307 instead.

PR #4514, opened 2026-06-19
fix(F0NumberInput): show thousand separators when not focused

FaithAdekunle:
## Description

Editable number and money cells (`F0NumberInput`, used by the EditableTable Number/Money cells) rendered values without thousand separators (`50000` instead of `50,000`), while read-only/display cells formatted correctly. This surfaced in Headcount Planning salary inputs for new and approved headcount rows (FCT-55894). The input now shows locale-aware grouping separators while blurred and reverts to the raw ungrouped value on focus so typing and parsing are unaffected.

## Implementation details

- fix: add locale-aware thousand separators to `F0NumberInput` display when the input is not focused (`50000` → `50,000` en-US, `5000.54` → `5.000,54` es-ES); raw value returns on focus

  <details>
  <summary>Why blur-only?</summary>

  Grouping separators inside the editable string would break `extractNumber` parsing and caret/selection handling mid-edit. Keeping the raw value while focused means `handleChange`, `handleBeforeInput`, clamping, and the deferred-popover path are all untouched — grouping is purely a display concern applied on blur.
  </details>

- fix: account for grouping separators in EditableTable number-cell width measurement (`useNumberCellLayout`) so wider grouped values aren't clipped
- test: cover grouped-on-blur / raw-on-focus behavior and `onBlur` passthrough in `F0NumberInput` and `NumberCell` specs

## Notes

- No consumer (factorial) code change required — the fix is entirely within `F0NumberInput`, so all money/number input cells benefit automatically.
- Visual change: best verified in the `F0NumberInput` and EditableTable money-cell Storybook stories.

https://github.com/user-attachments/assets/0b12c419-9895-415b-adcc-82b392eb2778

### DIFF
--- a/packages/react/src/components/F0NumberInput/__tests__/F0NumberInput.test.tsx
+++ b/packages/react/src/components/F0NumberInput/__tests__/F0NumberInput.test.tsx
@@ -24,6 +24,71 @@ describe("F0NumberInput", () => {
     expect(input).toHaveValue("123,46")
   })
 
+  describe("thousand separators", () => {
+    test("groups the value when not focused (en-US)", () => {
+      render(
+        <F0NumberInput
+          locale="en-US"
+          value={5000}
+          maxDecimals={0}
+          label="Number Input"
+        />
+      )
+      expect(screen.getByRole("textbox")).toHaveValue("5,000")
+    })
+
+    test("groups the value when not focused (es-ES)", () => {
+      render(
+        <F0NumberInput
+          locale="es-ES"
+          value={5000.54}
+          maxDecimals={2}
+          label="Number Input"
+        />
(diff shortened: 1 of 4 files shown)


sauldom102 (approved):

sauldom102 (changes requested):
I discovered something that is not working fully right

sauldom102 (commented):
A couple things not working fully right:
- We are automatically displaying less decimals that the ones added manually by the user.
- When writing dots/commas it's not having a consistent behavior compared to the formatting as you can see in the video attached

https://github.com/user-attachments/assets/61174612-9099-4f33-9d15-8f8963b4ba1f

PR #4528, opened 2026-06-20
feat(F0AudioPlayer): add expandable detail panel with Summary/Transcript tabs

ReddyWish:
## Changes
- `F0AudioPlayerCard` gains an optional `details` prop — an array of
  `{ value, label, content }` tabs (labels passed already translated,
  mirroring the existing `actions` prop). When provided, a "View detail" /
  "Hide detail" outline button appears in the header and toggles an
  expandable panel below the progress bar.
- The panel reuses the `TabNavigation` / `TabNavigationLink` primitive
  (same underline tabs as page/side-panel nav) and an `@/ui/ScrollArea`
  for the content, capped at `detailsMaxHeight` (default `200`, overridable)
  and scrollable beyond it.
- Expanded state is controlled or uncontrolled: `expanded`,
  `defaultExpanded`, `onExpandedChange` (follows the card's existing
  `playing` pattern).
- Animated expand/collapse via `motion/react`, respecting reduced motion.
  The panel is a labelled `role="region"` and the toggle exposes
  `aria-expanded`.
- New i18n defaults: `audioPlayer.viewDetail` / `audioPlayer.hideDetail`.
  Exposes the `AudioPlayerDetailTab` type.
- Backward compatible: without `details`, the card and the bare
  `F0AudioPlayer` are unchanged. The detail toggle remains usable even
  when the player is `disabled` (that only governs audio controls).

https://github.com/user-attachments/assets/3f69f417-9910-4137-ac9f-9c1d3126e1e6

### DIFF
--- a/packages/react/src/components/F0AudioPlayer/F0AudioPlayerCard.tsx
+++ b/packages/react/src/components/F0AudioPlayer/F0AudioPlayerCard.tsx
@@ -51,10 +51,14 @@ const F0AudioPlayerCardBase = forwardRef<
     defaultProp: defaultExpanded,
     onChange: onExpandedChange,
   })
-  const [activeTab, setActiveTab] = useState(details?.[0]?.value)
-  const activeContent =
-    details?.find((tab) => tab.value === activeTab)?.content ??
-    details?.[0]?.content
+  const [selectedTab, setSelectedTab] = useState(details?.[0]?.value)
+  // Guard against a stale selection if `details` changes (e.g. a recycled
+  // card in a list): fall back to the first tab so the highlight and content
+  // stay in sync.
+  const activeTab = details?.some((tab) => tab.value === selectedTab)
+    ? selectedTab
+    : details?.[0]?.value
+  const activeContent = details?.find((tab) => tab.value === activeTab)?.content
 
   return (
     <div
@@ -106,6 +110,7 @@ const F0AudioPlayerCardBase = forwardRef<
                 }
                 onClick={() => setExpanded(!isExpanded)}
                 aria-expanded={isExpanded}
+                aria-controls={detailsId}
               />
             )}
             {(controller.playbackRates.length > 0 || actions) && (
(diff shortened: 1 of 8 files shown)


sauldom102 (changes requested):
Instead of Tabs component I would use in this card the [SegmentedControl component](https://f0.factorial.dev/?path=/docs/experimental-components-f0segmentedcontrol--documentation), it could look cleaner

Besides that, I would avoid having `text-sm` in the entire card, it worsens the readability

With `text-sm`:

Image

With `text-base`:

Image

sauldom102 (changes requested):
When having expandable details and they are hidden, it's causing the card to slightly grow at the bottom side, you can see that visually if you compare it with the card above without expandable details 👀 

Image

ReddyWish:
> Instead of Tabs component I would use in this card the [SegmentedControl component](https://f0.factorial.dev/?path=/docs/experimental-components-f0segmentedcontrol--documentation), it could look cleaner
> 
> Besides that, I would avoid having `text-sm` in the entire card, it worsens the readability
> 
> With `text-sm`:
> 
> Image
> With `text-base`:
> 
> Image

@sauldom102 can you pls advice a better solution here in this case, because Segmented cont
Screenshot 2026-06-23 at 10 58 12
rol is not fully aligned with our design

ReddyWish:
> When having expandable details and they are hidden, it's causing the card to slightly grow at the bottom side, you can see that visually if you compare it with the card above without expandable details 👀
> 
> Image

@sauldom102 yes, that's what I did not notice, thank you, i will fix that

sauldom102:
> > Instead of Tabs component I would use in this card the [SegmentedControl component](https://f0.factorial.dev/?path=/docs/experimental-components-f0segmentedcontrol--documentation), it could look cleaner
> > Besides that, I would avoid having `text-sm` in the entire card, it worsens the readability
> > With `text-sm`:
> > Image
> > With `text-base`:
> > Image
> 
> @sauldom102 can you pls advice a better solution here in this case, because Segmented cont Screenshot 2026-06-23 at 10 58 12 rol is not fully aligned with our design

I would ask @isabellatangari 👀 because the segmented control likely didn't exist when this design was done, I think it could look a bit more compact and clean, also it wouldn't visually "clash" with the tabs that you already have above in the dialog

sauldom102 (approved):
Great one Ilya!

PR #4533, opened 2026-06-22
feat(ai): add F0AiProcessingOverlay

warcos-fact:
## Description

Adds `F0AiProcessingOverlay`, a floating "applying changes" affordance for the F0AiChat family: while `active`, it blurs and locks (`pointer-events-none`) the wrapped panel/canvas and floats a centered pill (the One icon + a status label) over it, signalling that the surface is being regenerated and shouldn't be edited. It lifts the existing survey form-builder pattern into a reusable component so any surface paired with the chat can use it.

Extracted from the co-creation patterns PR (#4307) so it can be reviewed on its own.

## Implementation details

- feat: add `F0AiProcessingOverlay` (`active`, `label`, `className`, `children`) under `sds/ai`, exported from `sds/ai/exports.ts`
- feat: add `ai.applyingChanges` ("Applying changes") default translation used as the pill's fallback label

Visual behaviour is covered by the component's usage in the co-creation stories.

ui

## Breaking changes

None in practice. The `api_breaking` bot flags the new `ai.applyingChanges` translation key because `TranslationsType` requires every key, so a consumer who hand-authors a **complete** translation object from scratch would need to add it. This is intentional and the standard cost of any translatable string:

- The component, its props, and exports are purely additive.
- The idiomatic pattern (`{ ...defaultTranslations, ...overrides }`) auto-inherits the new English default, so the vast majority of consumers are unaffected.
- Affected consumers fix it with a single line; the check is non-blocking, hence `feat(ai):` rather than `feat!:`.


developerdanx (approved):
💪

siguenzaraul (approved):

PR #4547, opened 2026-06-22
fix(EditableTable): fade disabled cell text (FCT-56243)

omedale:
Screenshot 2026-06-23 at 11 10 56 AM

### Problem
Disabled cells in the editable table rendered their text in the same full-strength color as editable cells, so a disabled cell was visually indistinguishable from an editable one.

## 🔑 What?
- Disabled cells now render their value in the secondary foreground (`text-f1-foreground-secondary`) alongside the existing disabled background, so they read as inactive.
- Added a Storybook story (`EditableTableWithDisabledCells`) showcasing the disabled-cell appearance.

## 🏡 Context
- [💼 Jira](https://factorialmakers.atlassian.net/browse/FCT-56243)

## 🏷️ Skill tags
`skill:frontend`

### DIFF
--- a/packages/react/src/patterns/OneDataCollection/__stories__/visualizations/editable-table/editable-table.stories.tsx
+++ b/packages/react/src/patterns/OneDataCollection/__stories__/visualizations/editable-table/editable-table.stories.tsx
@@ -336,6 +336,53 @@ export const EditableTableWithEditableCallback: Story = {
   },
 }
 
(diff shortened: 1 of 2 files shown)


albertpmz:
hey @warcos-fact can you weigh in here? 👀

omedale:
We agreed to go with the current approach @albertpmz 

cc: @warcos-fact

albertpmz (approved):

PR #4553, opened 2026-06-23
feat(tags): add size prop to F0TagRaw (md/sm)

smalonso:
## Description

Adds a `size` prop to the `F0TagRaw` component so it can be rendered in two sizes: `md` (default) and `sm`.

## Why

Until now the tag only had a single, fixed size with `text-base` (14px) typography. In denser UIs (compact lists, inline metadata, table cells) a smaller tag is needed, but there was no supported way to get one without ad-hoc overrides. This adds a first-class, token-based small variant.

## Implementation details

- **`internal/BaseTag`** — added a `size?: "md" | "sm"` prop (default `"md"`), the shared core that all tag variants render through. `md` keeps the current `text-base` (14px); `sm` uses `text-sm`, the design-system typography token for **12px** (`fontSize.sm` = `0.75rem`). Only the typography token changes — padding/shape are untouched.
- **`F0TagRaw`** — exposed the `size` prop in its types and forwarded it to `BaseTag`.
- **Storybook** — added a `size` control (inline radio `md`/`sm`), a new `Sizes` story showing both side by side, and `sm` examples in the snapshot story for visual regression coverage.

Defaults to `"md"`, so existing usages are unaffected.

## Screenshots

Available in Storybook under **Tags/TagRaw → Sizes**.

image

image

### DIFF
--- a/packages/react/src/components/tags/F0TagRaw/F0TagRaw.tsx
+++ b/packages/react/src/components/tags/F0TagRaw/F0TagRaw.tsx
@@ -10,7 +10,7 @@ import { BaseTag } from "../internal/BaseTag"
 
 export const F0TagRaw = forwardRef<HTMLDivElement, F0TagRawProps>(
   (
-    { text, additionalAccessibleText, icon, onlyIcon, info, className },
+    { text, additionalAccessibleText, icon, onlyIcon, info, className, size },
     ref
   ) => {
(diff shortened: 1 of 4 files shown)


sauldom102 (approved):

smalonso:
Ty!

PR #4575, opened 2026-06-26
fix(ai): keep chat steady on canvas toggle

warcos-fact:
## Description

`F0AiChat` re-faded its content on every visualization-mode change (via `useRevealOnChange`) to mask layout reflow. With the chat docked as a side panel, opening or closing the canvas flips `sidepanel`↔`canvas`, so the chat re-faded on every toggle even though its content barely changes. This adds an opt-in `revealChatOnCanvasToggle` flag (default `true`, so behaviour is unchanged for existing consumers); when disabled it collapses `sidepanel` and `canvas` into a single reveal state, keeping the chat steady while toggling the canvas. Fullscreen transitions still reveal, and the canvas panel's own open/close animation is untouched.

Extracted as a standalone fix from the Creation-with-AI work — the cocreation PR #4307 cherry-picks this commit and sets the flag to `false` in its story.

## Implementation details

- fix: add an opt-in `revealChatOnCanvasToggle` prop to `F0AiChat` and the `ApplicationFrame` `ai` config (defaults to `true`)
- fix: when disabled, treat `sidepanel` and `canvas` as one mode-change reveal state so toggling the canvas no longer re-fades the chat content; fullscreen transitions still reveal

### DIFF
--- a/packages/react/src/patterns/ApplicationFrame/index.tsx
+++ b/packages/react/src/patterns/ApplicationFrame/index.tsx
@@ -387,7 +387,11 @@ function ApplicationFrameContent({
                     }
                   }}
                 >
-                  <F0AiChat />
+                  <F0AiChat
+                    revealChatOnCanvasToggle={
+                      ai?.revealChatOnCanvasToggle ?? true
+                    }
+                  />
                 </motion.div>
               )}
             </motion.div>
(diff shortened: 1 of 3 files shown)


sauldom102 (approved):

warcos-fact:
@siguenzaraul fixed this in a different commit.

PR #4584, opened 2026-06-29
feat(react-native): support Expo SDK 56 (React Native 0.85)

developerdanx:
## Summary

Upgrades `@factorialco/f0-react-native` from **Expo SDK 54** to **Expo SDK 56** (React Native 0.85, React 19.2.3, Hermes v1, New Architecture).

This is the design-system half of the SDK 56 upgrade; the Factorial mobile app will consume the published version next.

## Version matrix

| | Before (SDK 54) | After (SDK 56) |
|---|---|---|
| expo | ~54.0.32 | ~56.0.12 |
| react-native | 0.81.5 | 0.85.3 |
| react / react-dom / react-test-renderer | 19.1.0 | **19.2.3** |
| reanimated / worklets | 4.1.7 / 0.5.1 | 4.3.1 / 0.8.3 |
| screens / safe-area / gesture-handler / svg | 4.16 / 5.6 / 2.28 / 15.15.1 | 4.25.2 / 5.7.0 / 2.31.1 / 15.15.4 |
| expo-blur / expo-clipboard / expo-image | 15.0 / 8.0 / 3.0 | ~56.x (SDK-aligned) |
| uniwind | ^1.2.7 | **^1.9.0** |

## Notable changes

- **F0BlurView** migrated to the SDK 55+ expo-blur API: `experimentalBlurMethod` → `blurMethod`.
- **babel.config.cjs** is now a function that adds `react-native-worklets/plugin` **only for app bundles** (Metro caller). The builder-bob library build keeps it off, so consumers workletize at bundle time (unchanged published behaviour); the bundled playground app now workletizes its own Reanimated worklets.
- **Jest**: preset switched to `@react-native/jest-preset` (the RN 0.85 `react-native/jest-preset` shim throws), `react-native-worklets` mocked, `@types/react-test-renderer` added, and the `react-test-renderer` 19.2 type tightening handled in `F0Button` / `F0Step` / `PressableFeedback` specs.
- **Peer deps** realigned: `expo-image` / `expo-blur` loosened, `expo-clipboard` moved from `dependencies` to `peerDependencies`. Root pnpm React-island override → 19.2.3.
- Snapshots updated for the React 19.2 `useId()` id format change (`«rN»` → `_rN_`) in svg gradient/clipPath ids.

## Why uniwind 1.9 / react 19.2.3 (not the latest patch)

Both surfaced during on-device testing and were then verified against upstream sources.

### `uniwind` → ^1.9.0
[...]

PR #4585, opened 2026-06-29
chore(react-native): use native Liquid Glass tabs in the playground tab bar

developerdanx:
## Summary

**Stacked on #4584 (Expo SDK 56).** Converts the playground bottom tab bar from the JS expo-router `<Tabs>` to **Native Tabs** (`expo-router/unstable-native-tabs`), so it renders the platform's real system tab bar via react-native-screens.

- **iOS 26** — native `UITabBar` with the **Liquid Glass** material. It's the default appearance; we deliberately set no `backgroundColor`/`blurEffect`, so the system glass material and automatic light/dark adaptation are preserved.
- **Android** — native Material 3 bottom navigation (Liquid Glass is iOS-only).
- Icons via SF Symbols (`sf=`) on iOS and Material symbols (`md=`) on Android (`paintpalette.fill`/`palette`, `square.grid.2x2.fill`/`widgets`).

Playground-only — not part of the published package surface (`files: src, lib`), so there is **no library API change** (hence `chore`).

## Before → after (iOS 26)

Flat, full-width JS tab bar with emoji icons → floating **Liquid Glass** capsule with native SF Symbols and system theming.

## Verification

- Verified on **iPhone 17 Pro (Expo Go, SDK 56.0.0)**: renders as the floating Liquid Glass tab bar and navigates between Design Tokens ↔ Components (the bottom bar reports as a native `Tab Bar` with `paintpalette.fill` / `grid 2x2` SF Symbols in the a11y tree).
- `pnpm tsc` clean (playground is type-checked).

## Base / merge order

Stacked on `chore/expo-sdk-56-upgrade` — requires SDK 56 / RN 0.85 (Native Tabs + Liquid Glass). Rebase/retarget onto `main` after #4584 merges.

https://github.com/user-attachments/assets/ecb60848-9e27-4e52-93b3-db9945910325


eliseo-juan (approved):
LGTM!

PR #4587, opened 2026-06-29
feat(OneDataCollection): add column add/remove affordances to the table settings popover

zygisS22:
## Description

Adds two **opt-in** affordances to the table column-settings popover (the sliders icon on a `OneDataCollection` / `F0AnalyticsDashboard` table):

1. **"Add column"** — an entry at the top of the popover that calls a consumer-provided `onAddColumn`.
2. **Remove a column** — a trash button revealed on hover over a column row that calls `onRemoveColumn(columnId)`.

Both are opt-in: nothing renders unless the matching callback is provided, so every existing table is unchanged. **Removing is intentionally distinct from the existing hide toggle** — hide only flips visibility (persisted in settings), whereas remove asks the consumer to drop the column from its `columns` set. Frozen/leading columns (and columns flagged `noRemoving`) are never removable.

## Design reference

[Figma — AI Analytics › column settings (node 13028-9147)](https://www.figma.com/design/1smmEIugmR0CNeu7NvK33y/AI-Analytics?node-id=13028-9147&p=f&t=uLcCKhWKF5jQoSBg-0)

The placement of **"Add column" at the top of the settings popover** and the **hover-to-remove (trash)** affordance both come straight from this frame. It also shows the wider AI-Analytics context these column controls are being built for (driving AI-authored analytics tables/dashboards), which is why the controls live inside the existing table-settings popover rather than as a separate toolbar.

## Screenshots & recording

**"Add column" entry at the top of the popover** (Name is locked → not removable):

Add column entry in the table settings popover

**Hover a column row → a trash button appears** (distinct from the visibility toggle):

Hover reveals the remove/trash button on a column row

**Full flow (screen recording — add + remove):** ▶️ 

https://github.com/user-attachments/assets/917ee9ec-304a-463d-9942-d92ac1707db8

## How to try it

Storybook (`pnpm --filter @factorialco/f0-react dev`):
- **Full public-API demo:** `Patterns/Data Collection/Visualizations/Table → With Column Add Remove`
- **Component-level:** `Patterns/Data Collection/Internal/SortAndHideList → With Removable Items`

Open the table story, click the sliders icon → the popover shows "Add column" on top and a trash button on hover over Email/Role. (The demo's "Add column" just re-adds the next hidden column since the story has no picker; in real use it opens the consumer's column picker.)

## Implementation details

The popover content flows `Settings → VisualizationSettingsRenderer → (table) SettingsRenderer → TableSettings → SortAndHideSettings → SortAndHideList`. The two callbacks are threaded down that chain as optional props:

- **`TableVisualizationOptions`** (public) gains `onAddColumn?: () => void` and `onRemoveColumn?: (columnId: ColId) => void`. `TableColumnDefinition` gains `noRemoving?: boolean` (parity with `noHiding`) to lock specific columns.
- **`SettingsRenderer`** now also renders the popover when only add/remove (and not hide/reorder) is enabled, and forwards the callbacks.
- **`TableSettings`** computes a per-row `removable` flag: removable when `onRemoveColumn` is set, the column is not one of the frozen/leading non-editable columns, and it is not flagged `noRemoving`.
- **`SortAndHideSettings`** renders the "Add column" entry at the top of the section and forwards `onRemove` to the list.
- **`SortAndHideList`** renders a hover-revealed (`group-hover` / `focus-within`) ghost trash `ButtonInternal` per `removable` row, beside the existing visibility switch.
- i18n: added `collections.table.settings.addColumn` and `…removeColumn`.

The hide/reorder persistence (`settings.visualization.table`) is untouched; add/remove mutate the caller's column set instead.

### Tests / stories
- Unit: `SortAndHideList.test.tsx` — remove button appears only for `removable` rows when `onRemove` is set, and fires with the right item.
- Stories: `SortAndHideList → WithRemovableItems`; `Table → WithColumnAddRemove` (exercises the full public API).

### Verification
- `oxfmt --check` ✅ · `oxlint --max-warnings 0` ✅ · `build:types` (tsc) ✅ · `pnpm tsc` incl. the new story ✅ · `vitest` (3 new + existing settings/column tests) ✅


albertpmz (approved):
LGTM! 👍

PR #4596, opened 2026-06-30
feat: expose full base color palette for F0TagDot

ReddyWish:
## Changes

`F0TagDot` (and RN `F0Tag.Dot`) only allow-listed 11 of the 15 chromatic
base-color families. The omitted four — `radical`, `orange`, `red`,
`grass` — are exactly the hues consumers reach for, so the gap pushed
them onto the `customColor` escape hatch with raw `hsl(var(--*))` vars.

- **react** — add `radical`, `orange`, `red`, `grass` to `tagDotColors`.
  `NewColor` and the value-display `dotTag` cell derive from this list,
  so the colors flow through automatically.
- **react-native** — same four added to `dotTagColors` for parity.
- **docs** — TagDot Storybook description now notes the full palette is
  available and steers actual *status* use cases to `TagStatus`.

> Note: `red`/`orange`/`grass` sit close to the `critical`/`warning`/
> `positive` hues. `F0TagDot` is for categories, not statuses — the doc
> note reinforces using `TagStatus` when a status is meant.

### DIFF
--- a/packages/react-native/src/components/F0Tag/F0Tag.types.ts
+++ b/packages/react-native/src/components/F0Tag/F0Tag.types.ts
@@ -18,6 +18,10 @@ export const dotTagColors = [
   "flubber",
   "indigo",
   "camel",
+  "radical",
+  "orange",
+  "red",
+  "grass",
 ] as const
 
 /**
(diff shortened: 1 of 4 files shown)


developerdanx (changes requested):
Hi @ReddyWish 

Please attach demo on Android and iOS of the result

PR #4612, opened 2026-07-03
feat(F0Graph): viewport node-array windowing (opt-in) + perf harness [FCT-57915]

factmarc10010:
## Description

Adds **opt-in, non-breaking** viewport virtualization to `F0Graph` for [FCT-57915](https://factorialmakers.atlassian.net/browse/FCT-57915), targeting the "broken orgchart" case (thousands of expand-visible root nodes with no manager). Covers **A0 (perf harness)**, **A1 (node-array windowing)**, **A2 (viewport-driven data loading)** and **A3 (exposure through OneDataCollection, incl. the two-phase hydration seam)**. A4 (release) is a post-merge step.

Draft — opening for review. Reviewer tip: the commits are ordered by capability (windowing → A2 → OneDataCollection pass-through → Seam A → review fixes), so reviewing commit-by-commit follows the design.

### A0 · Perf harness
- **`ViewportWindowing`** story: 3,000 root nodes with a live *visible vs rendered* counter, a windowing toggle and a padding slider.
- **`onRenderedNodesChange(count)`** callback exposing the post-windowing node count.

### A1 · Node-array windowing
- Props **`enableNodeWindowing`** / **`nodeWindowPadding`** (default `600`px). Only the React Flow nodes whose layout box intersects the padded viewport are materialized, instead of the full expand-visible array (`onlyRenderVisibleElements` only culls the DOM).
- Layout, tree structure and bounds still computed in full → positions, `fitView` and navigation stay correct.
- **`useViewportGeometry`** derives the padded viewport rect from the RF store, **quantized to a grid** so the O(N) intersection is throttled to camera cell-crossings, not every pan frame.
- Windowing **activates only after the first settled viewport** so the mount-time `fitView` frames the whole graph first.
- **Navigation fallbacks:** fit-view / find-me / focused-node fall back to full-layout `fitBounds` / `setCenter` (off-screen targets aren't in the RF store).
- **aria-owns** trimmed to children still in the window (no dangling references).

### A2 · Viewport-driven data loading
- Props **`loadVisibleNodeData(ids)`** / **`visibleDataDebounceMs`** (default `200`ms). **`useViewportDataLoader`** watches the on-screen node ids and, once the camera settles, calls the consumer with the ids that newly entered the viewport — **debounced + batched, each id requested at most once**. Build the tree from a lightweight structure; fetch heavy per-node data only for what's on screen.
[...]

PR #4621, opened 2026-07-04
feat(F0Graph): open centered on a node (first-frame focus)

factmarc10010:
## Description

Adds an instant **"open already looking at this node"** for the graph — the org-chart case of landing on the current user (or their branch root) without a fit-to-all-then-pan — and the correctness fixes needed to make it hold as you interact (collapse/expand no longer move the view).

Rebased on `main` (includes #4612 windowing + #4625 `focusOnEntry`).

## Preview / Demo

- **`Graph/F0Graph` → InitialFocus** — passes `initialFocusNodeId` (a deep, off-center node); the graph opens already framed on it (no fit-to-all + pan). Compare with **LargeTree** (same data) which opens fit-to-all.
- **`Data Collection/Visualizations/Graph` → FocusOnRoot** — the OneDataCollection path with `focusOnEntry: "ceo-a"` (controlled expand + lazy loading): opens framed on the root, and collapsing/expanding keeps that frame.

### Before (performance issues already resolved in another PR) 
Dropped in the middle of the orgchart without any pattern 

https://github.com/user-attachments/assets/d6111c95-ead0-4ade-ae7a-6e280350058c

### After 
Now centered on your "root" the manager of the manager of the manager so you have a familiar starting point

https://github.com/user-attachments/assets/fdd9c23f-0f94-49ce-b085-1c7f8bc276e7

## Behavior

- With `focusOnEntry` (OneDataCollection) / `initialFocusNodeId` (F0Graph): the graph **opens already framed on the target + its direct children, no animation** — no fit-to-all first, no second pan.
- **Collapse / expand keeps the view put**: the toggled node stays visually fixed; the camera is only moved by explicit actions (pan, "Find me", search).
- Default (no target): unchanged — fit-to-all.
- Fallback: target not resolvable → fit-to-all (never a blank frame).
- Search / "Find me" reveals: unchanged.

## What's in this PR

**1. Feature — first-frame focus (two layers)**
- **`initialFocusNodeId`** (F0Graph, view layer): frame on a node **that is present** + its direct children on the mount-time fit.
- **`focusOnEntry`** (OneDataCollection, data layer): pre-resolves the node's ancestor path and expands it **before first paint** (the tree is lazy, so the node may not be loaded yet), then feeds it to `initialFocusNodeId`. This is the pre-paint, no-pan version of a reveal.

**2. Correctness fixes (so the frame survives interaction)**
- **fly-to no-refire**: the `focusedNode` fly-to only fires when the focus actually changes, not on every layout re-render.
- **one-shot initial fit**: the initial frame is applied once, imperatively — React Flow's queued `fitView` prop is no longer used, so a later layout change can't re-fire it as a fit-all.
- **anchor via viewport pan**: on collapse/expand the layout engine repositions the toggled node (dagre re-centers a parent over its children). Instead of offsetting node positions (which held for one commit then snapped back on the next node-windowing render — the "root jumps left on the first collapse" bug), we translate the **viewport** by the anchor's `(dx, dy)` delta before paint. Node positions stay raw, so `getNodePosition` / `contentBounds` (reveal, fit-view, windowing) stay consistent.

### F0Graph
- New **`initialFocusNodeId`** prop. Pure `resolveInitialFitViewNodes` picks the target + its present direct children (or `undefined` → fit-all fallback). Zoom capped by `INITIAL_FOCUS_MAX_ZOOM`. The fit is applied once, imperatively (`didInitialFitRef`), not via the `fitView` prop.
- `useGraphRenderModel` keeps node positions raw and reports the anchor's reflow delta via **`onAnchorReflow`**; `F0GraphView` applies it with `reactFlow.setViewport` (before paint, so no flicker).

### OneDataCollection graph viz
- `useDataCollectionTreeData` **pre-resolves the `focusOnEntry` ancestor path before first paint** (extracted `resolvePath`, shared with `revealNode`). No focus/highlight on entry — the initial viewport centers it. Resolution failure → silent fallback.
- `GraphCollection` passes `focusOnEntry` to the hook and `initialFocusNodeId` to F0Graph. `resolveGraphReveal` no longer reveals on entry; search / Find-me unchanged.

The monorepo needs nothing new — it already passes `focusOnEntry = root of the tree`; F0 frames the first paint and keeps it stable on collapse/expand.

## Acceptance criteria

- [x] First frame centered on the target + direct children, no jump
- [x] **Collapse / expand does not move the viewport** (only explicit actions do)
- [x] Default (no target) unchanged — fit-to-all
- [x] Fallback to fit-to-all when target can't be resolved
- [x] Search / "Find me" unchanged
- [x] Windowing + two-phase hydration: no regression

## Testing

- `utils.test.ts` — `resolveInitialFitViewNodes` (frame target / fallback).
- `F0Graph.focusNoRefire.test.tsx` — fly-to fires only on focus change; initial fit is one-shot (no re-fit on collapse).
- `useGraphRenderModel.test.ts` — anchor reports the `(dx, dy)` delta on reflow, and stays silent when the position is unchanged.
- `useDataCollectionTreeData.test.tsx` — `focusOnEntry` pre-resolution.
- Full F0Graph + OneDataCollection Graph suites green; `tsc` and `oxlint` clean.

### DIFF
--- a/packages/react/src/patterns/F0Graph/F0Graph.tsx
+++ b/packages/react/src/patterns/F0Graph/F0Graph.tsx
@@ -112,6 +112,16 @@ export interface F0GraphProps<T = unknown> {
   // ---- Navigation ----
   focusedNode?: string
   highlightedNodes?: Set<string>
+  /**
+   * Id of a node to frame on the **first paint**, instead of the default
+   * fit-to-all. The initial viewport opens already centered on this node with
+   * no animation — use it to open "already looking at" a node (e.g. the current
+   * user). The node must exist in the initial `nodes` (expand its ancestors
+   * before mount); if it's absent, this falls back to the normal fit-to-all.
+   * Only affects the initial frame — later `focusedNode` / "Find me" reveals
+   * still animate with the smooth pan.
+   */
+  initialFocusNodeId?: string
 
   // ---- Layout ----
   /** Layout sizing hint passed to the built-in layout engine. Defaults to 256. Override for compact nodes (icons, file rows). */
--- a/packages/react/src/patterns/F0Graph/__stories__/F0Graph.stories.tsx
+++ b/packages/react/src/patterns/F0Graph/__stories__/F0Graph.stories.tsx
@@ -718,6 +718,30 @@ export const LargeTree: Story = {
   },
 }
 
+/**
+ * `initialFocusNodeId`: open already centered on a specific node instead of
+ * fitting the whole tree. The graph mounts framed on a deep, off-center member
+ * (no fit-to-all, no pan) — the "open looking at me" behaviour used by the
+ * org chart. Compare with `LargeTree` (same data) which opens fit-to-all.
+ */
+export const InitialFocus: Story = {
+  parameters: {
+    docs: {
+      description: {
+        story:
+          "Opens framed on `initialFocusNodeId` (a deep, off-center node) on the first paint — no fit-to-all then pan. The node must be present in the initial `nodes` (here `defaultExpandDepth: 2` makes members visible); if it's absent, F0Graph falls back to fit-to-all.",
+      },
+    },
+  },
+  args: {
+    nodes: makeLargeTree(600),
+    renderNode: renderEmployee,
+    showControls: true,
+    defaultExpandDepth: 2,
+    initialFocusNodeId: "dept-4-member-100",
+  },
+}
+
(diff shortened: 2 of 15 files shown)


sauldom102 (commented):
Can you add a story with this prop being passed so that we can see the behavior? Also, please add some video in the PR

sauldom102 (approved):

PR #4642, opened 2026-07-08
feat(F0AudioPlayer): support lazy src resolution and known duration

ReddyWish:
Adds an on-demand source to `F0AudioPlayer` / `F0AudioPlayerCard`. Until now the
component required the URL up front and loaded it eagerly on mount. `src` now
also accepts a resolver function that produces the URL only when playback is
actually requested. Backward-compatible for the string case.

Why the lazy form is useful as a general capability:
- **No work until intent.** If the URL is expensive or side-effectful to produce
  (a network call, a signed/credentialed URL, a token mint), it isn't produced
  for a player the user never presses play on — which matters most in lists,
  where eager loading would resolve every item.
- **Resolution is tied to the moment of use.** For time-bounded URLs, the
  validity window starts at play rather than at render, so it isn't half-spent
  before the user interacts. Presigned recording URLs (e.g. ATS call recordings)
  are one example, but the pattern applies to any deferred source.
- **Metadata without a fetch.** `duration` lets the card show total time and an
  active seek bar with `preload="none"`, so nothing hits the network until play.

## Changes
- **`src: string | (() => Promise<string>)`** — a single required prop. Pass a
  URL string for the eager path, or a function that resolves the URL on the
  first play for the lazy path. Enforces exactly one source form at the type
  level (no separate optional props that could both be omitted or both set).
- **`duration?: number`** — seeds the total time and an active seek bar before
  any audio loads; superseded by real metadata once it arrives.
- **`preload` defaults to `"none"`** when `src` is a function (was always
  `"metadata"`), so lazy mode makes zero network requests before intent.
- **Expiry handling** — on a media error in lazy mode, the URL is re-resolved
  once and playback resumes, so an expired presigned URL self-heals.
- All lazy logic lives in the shared `usePlayerController` (which derives the
  eager string vs lazy resolver from `src`); both the player and the card pick
  it up. `usePlayerController` now exposes `currentSrc`.

Consumer shape:

```tsx
<F0AudioPlayerCard
  src={() => generatePlaybackUrl(callId)}  // presign, fires on play
  duration={recordingDurationSeconds}       // total time shown before load
  title="AI Call with Alex Williams"
  subtitle="May 9, 2025 – 10:00am"
/>
```

## Testing
- `pnpm vitest src/components/F0AudioPlayer` — 34 pass (resolve-on-play,
  no-fetch-until-play, `preload=none` default, duration display, expiry
  re-resolve, getSrc-rejection recovery).
- `pnpm lint` / `oxfmt` — clean.
- `tsc --noEmit` — no new errors.
- Added a `LazySource` Storybook story.

## Notes
- Experimental component; API is not yet stable.
- Seek-before-load: with `duration` known but audio not yet loaded, the scrubber
  is draggable before first play (sets `currentTime` on a src-less element —
  harmless). Left as-is to keep the change tight; can gate it in a follow-up.

### DIFF
--- a/packages/react/src/components/F0AudioPlayer/types.ts
+++ b/packages/react/src/components/F0AudioPlayer/types.ts
@@ -26,13 +26,27 @@ export interface AudioPlayerDetailTab {
 export interface F0AudioPlayerProps
   extends WithDataTestIdProps, DataAttributes {
   /**
-   * The audio source URL.
+   * The audio source URL. Provide this or `getSrc`.
    */
-  src: string
+  src?: string
+
+  /**
+   * Lazily resolves the audio source URL the first time playback is requested.
+   * Use for on-demand credentials (e.g. presigned URLs) so the URL is only
+   * fetched on user intent. Ignored when `src` is set.
+   */
+  getSrc?: () => Promise<string>
+
+  /**
+   * Known total duration in seconds. Lets the player show the total time and an
+   * active seek bar before the audio loads (e.g. with `preload="none"`).
+   * Superseded by the real duration once metadata loads.
+   */
+  duration?: number
 
   /**
    * How much of the audio to preload.
-   * @default "metadata"
+   * @default "metadata" ("none" when `getSrc` is set)
    */
   preload?: "none" | "metadata" | "auto"
 
--- a/packages/react/src/components/F0AudioPlayer/types.ts
+++ b/packages/react/src/components/F0AudioPlayer/types.ts
@@ -26,16 +26,12 @@ export interface AudioPlayerDetailTab {
 export interface F0AudioPlayerProps
   extends WithDataTestIdProps, DataAttributes {
   /**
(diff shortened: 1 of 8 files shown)


packages/react/src/components/F0AudioPlayer/types.ts
```diff
@@ -26,13 +26,27 @@ export interface AudioPlayerDetailTab {
...
    */
-  src: string
+  src?: string
+
+  /**
+   * Lazily resolves the audio source URL the first time playback is requested.
+   * Use for on-demand credentials (e.g. presigned URLs) so the URL is only
+   * fetched on user intent. Ignored when `src` is set.
+   */
+  getSrc?: () => Promise<string>
```
sauldom102: I would better express it like this:

```suggestion
  src: string | () => Promise<string>
```

So that we always enforce at type-level that we provide either one or the other

sauldom102 (approved):

PR #4646, opened 2026-07-09
fix(F0Graph): don't re-center focusedNode on layout changes

factmarc10010:
## Problem

The graph re-centers on the current `focusedNode` on the **first layout change** after focusing (e.g. collapsing the focused node's children) — the viewport jerks and re-centers even though `focusedNode` never changed.

## Root cause

The "fly to `focusedNode`" effect in `F0GraphView` depended on `centerOnNode`:

```ts
useEffect(() => { … fly to focusedNode … }, [focusedNode, reactFlow, enableNodeWindowing, centerOnNode])
```

`centerOnNode`'s identity changes on every layout-affecting render:
`collapse/expand → new layout.nodes → new positionMap → new getNodePosition → new centerOnNode`.
So the effect re-runs and re-applies the fly-to with the **same** `focusedNode` — the entry fly-to firing again. (It re-fires on *every* such layout change while a `focusedNode` is set, not only the first.)

Note it isn't triggered by `focusOnEntry` per se: entry uses `initialFocusNodeId` and leaves `focusedNode` undefined. It bites once something sets `focusedNode` — search-select, "Find me", or a consumer that pins it — and then the tree relayouts.

## Fix

The effect now depends **only on `focusedNode`**, reading the fly-to logic through a ref (`flyToFocusedRef`, refreshed each render). It fires only when `focusedNode` transitions to a new value (entry, search-select, "Find me"), never on layout re-renders while it's unchanged.

## Non-goals / unchanged
- Initial open centering, search-select centering, "Find me" centering — all still work (each sets a new `focusedNode`).
- Node windowing (#4612) and the fly-to's windowing/`fitView` paths — untouched.

## Testing
- New `F0Graph.focusNoRefire.test.tsx`: `focusedNode` unchanged across a collapse ⇒ no extra `fitView`; a genuine `focusedNode` change ⇒ flies to the new node. Full F0Graph suite green (229); typecheck / oxlint / oxfmt clean.

### DIFF
--- a/packages/react/src/patterns/F0Graph/__tests__/F0Graph.focusNoRefire.test.tsx
+++ b/packages/react/src/patterns/F0Graph/__tests__/F0Graph.focusNoRefire.test.tsx
@@ -1,0 +1,125 @@
+import React, { act } from "react"
+import {
+  afterEach,
+  beforeAll,
+  beforeEach,
+  describe,
+  expect,
+  it,
+  vi,
+} from "vitest"
+
+import { zeroRender } from "@/testing/test-utils"
+
+import type { GraphNode } from "../types"
+import { F0Graph } from "../F0Graph"
+
+// Spy React Flow instance so we can count fly-to (fitView) calls. Only the
+// public `useReactFlow` is mocked — the ReactFlow component renders normally.
+// (Name must start with `mock` for the hoisted vi.mock factory.)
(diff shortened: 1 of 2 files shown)


factmarc10010:
Folded into #4621 (cherry-picked the focusedNode re-center fix + test). Closing in favor of that PR.

PR #4653, opened 2026-07-09
fix(ai): welcome suggestions popover closes when switching to the last group

zygisS22:
## Bug

On any AI chat welcome screen, with a suggestion-group popover open, clicking the **last** group tag closes the popover instead of switching to that group. Clicking the open group's own tag also re-opened it instead of closing for every tag except the last.

Reproduce before this fix: open the One chat welcome → click "Analyze" → click "Create" (the third and last tag) → the panel closes instead of showing Create's items.

## Root cause

`WelcomeScreenSuggestionsRow` renders one shared `Popover`, but wrapped each group button in its own `PopoverTrigger`. Radix registers a single trigger per popover—the last one mounted—so its internal dismissal and toggle behavior raced the row's `activeIdx` state.

## Fix

- Use plain group buttons inside the existing `PopoverAnchor`; the row now owns open, switch, and close behavior consistently for every group.
- Ignore pointer-down-outside dismissal for clicks inside the group row, preventing it from racing a group-button click.
- Preserve the popup semantics Radix previously supplied with `aria-haspopup`, `aria-expanded`, `aria-controls`, and an accessible dialog label.
- Restore focus after selection or Escape while preserving focus when the host or an outside click deliberately moves it elsewhere.

## Storybook

`WithWelcomeSuggestions` now has an executable interaction that opens **Analyze**, switches to the third **Create** group, verifies Create's items are visible, and verifies Analyze's items are gone.

Before
https://github.com/user-attachments/assets/9f9444f5-6e8a-4ad1-814f-fa470e762bee

After

https://github.com/user-attachments/assets/07c467af-c357-4829-939a-6e87ebd5329f

## Tests

- 13 focused `WelcomeScreenSuggestionsRow` tests pass, covering switching to and from the last group, same-tag close, ARIA state, selection focus, host-directed focus, Escape, and outside-click dismissal.
- Formatting, TypeScript, focused lint, cycle-dependency checks, and independent code/accessibility/Storybook/test-coverage reviews pass.
- The Storybook interaction was verified locally and finishes with the third Create group open.

### DIFF
--- a/packages/react/src/components/F0Button/internal-types.ts
+++ b/packages/react/src/components/F0Button/internal-types.ts
@@ -31,6 +31,14 @@ export type ButtonInternalProps = Pick<
      */
     "aria-expanded"?: boolean
     /**
+     * Identifies the expandable region controlled by the button.
+     */
+    "aria-controls"?: string
+    /**
+     * Describes the type of popup opened by the button.
+     */
+    "aria-haspopup"?: React.AriaAttributes["aria-haspopup"]
+    /**
      * Forwarded to the underlying button. Use `-1` to take the button out of the
      * tab order (e.g. when a parent manages focus via roving tabindex).
      */
--- a/packages/react/src/kits/ai/F0AiChatTextArea/__stories__/F0AiChatTextArea.stories.tsx
+++ b/packages/react/src/kits/ai/F0AiChatTextArea/__stories__/F0AiChatTextArea.stories.tsx
@@ -1,5 +1,6 @@
-import { Meta, StoryObj } from "@storybook/react-vite"
+import type { Meta, StoryObj } from "@storybook/react-vite"
 import { useRef, useState } from "react"
+import { expect, userEvent, within } from "storybook/test"
 
 import { F0AiChatTextArea } from "../F0AiChatTextArea"
 import type { F0AiChatTextAreaSubmitPayload } from "../types"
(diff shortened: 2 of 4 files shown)


eliseo-juan (approved):
LGTM!

siguenzaraul (approved):

zygisS22:
Updated this branch with `origin/main` (merge commit, no rebase / no force-push) so it's no longer conflicting.

The conflict was purely structural: main moved the AI tree from `packages/react/src/sds/ai/**` to `packages/react/src/kits/ai/**` (pure renames, no content change). Git's rename detection carried this PR's edits onto the new `kits/ai/` paths with no manual resolution needed, so the toggle-behavior fix is unchanged. Note the Storybook ID is now `kits-ai-f0aichattextarea--with-welcome-suggestions`.

Verified locally on the merged branch: `tsc --noEmit` clean, `oxfmt --check` clean, and 49 unit tests pass across `kits/ai/F0AiChatTextArea` (including all 13 in `WelcomeScreenSuggestionsRow.test.tsx`). I also ran the story side-by-side against unmodified `main`: main closes the popover when you click the last group, this branch switches to it correctly and still toggles closed on a second click. Popover placement measures identical to main at both tall and short viewports, so nothing shifted there.

All 21 substantive checks are green. The four remaining `🤖 Agentic Checks` jobs are stuck QUEUED on runners; that workflow isn't a required status check and has ended `skipped`/`cancelled` on every recent PR, so it isn't gating.

Conflict-free with main and CI green — ready to merge whenever you are.

PR #4664, opened 2026-07-10
feat(OneDataCollection): add liveUpdate to graph visualization

AndresRojoFactorial:
## Description

Adds a declarative `liveUpdate` option to the Data Collection graph visualization so consumers can reflect real-time / collaborative changes (rename, re-parent, remove) on the already-loaded tree **in place** — preserving the user's expansion and viewport — instead of forcing a full reset that collapses everything back to `defaultExpandDepth`. The batch reconciles the structure of the affected parents locally, so moving a node's only child away no longer leaves the old parent with an eternal loading expander, and a leaf parent that gains a child immediately shows its expand affordance without refetching.

## Implementation details

- feat: add `liveUpdate?: { version, upsert?, remove? }` to `GraphVisualizationOptions` and thread it through `GraphCollection` into `useDataCollectionTreeData`

  <details>
  <summary>How it works</summary>

  Bumping `version` applies the batch exactly once (the number dedups against React re-renders; the initial value is adopted on mount so no spurious batch fires). `upsert` records are matched by node id: an existing node has its `data`, `childrenCount` and parent refreshed (re-parenting via `getParentId`); an unknown record is inserted when attachable. `remove` ids are dropped with their descendants and pruned from the expanded set. No re-fetch, no collapse.

  ```tsx
  const [live, setLive] = useState({ version: 0, upsert: [], remove: [] })

  // On a websocket event: employee 42 moved under manager 7
  setLive((prev) => ({
    version: prev.version + 1,
    upsert: [updatedEmployee42], // getParentId(updatedEmployee42) === "7"
    remove: [],
  }))

  <OneDataCollection
    visualizations={[{ type: "graph", options: { ..., liveUpdate: live } }]}
  />
  ```

  </details>

- fix: reconcile `childrenCount`/`childrenLoaded` of the parents touched by a batch from the in-memory tree

  <details>
  <summary>Why?</summary>

  Previously the structure came only from the record accessors, so callers had to also upsert the old and new parents with fresh counts — and nobody does. Two visible bugs followed:

  1. **Eternal spinner**: moving the only child of an expanded parent left the parent with a stale `childrenCount > 0` and no children in memory, which the render model interprets as "expanded, waiting for children" → permanent loading expander until a full reload.
  2. **Missing expander**: moving a node under a leaf parent (`childrenCount: 0`) left the new parent without the expand affordance, so the moved child was unreachable.

  Now the hook tracks the parents whose child set the batch touches (old/new parent of a move, parent of a removal) and recomputes their counts from the in-memory children. A believed-leaf parent that gains children this way is also marked fully loaded, so expanding it shows the moved child immediately instead of fetching data the server may not reflect yet. Callers only send the records that changed.

  </details>

- fix: prune subtrees re-parented under a not-yet-loaded parent instead of promoting them to floating roots

  <details>
  <summary>Why?</summary>

  A dangling `parentId` would be promoted to a root by the tree builder, drawing the moved subtree as a disconnected tree at the top level. It now leaves the loaded tree (with its descendants) and reappears naturally when its new parent is expanded.

  </details>

- refactor: split `applyLiveUpdate` into pure module-level helpers (`collectSubtreeIds`, `findUnattachedIds`, `applyUpsertRecords`, `reconcileTouchedParents`)

  <details>
  <summary>Why?</summary>

  Review feedback: the function had grown past comfortable reading size, and its subtree removal used a `shift()`-based `while` loop that both mutated the condition it observed and cost O(n) per step. `applyLiveUpdate` is now a five-step orchestrator; the subtree walk uses a read cursor over an append-only frontier — O(n) overall, and the collected-ids guard bounds it by the number of loaded nodes, so it terminates even against cyclic parent links. Parent reconciliation also counts children in a single pass instead of one scan per touched parent.

  </details>

- test: cover in-place upsert/rename, insert under a loaded parent, cascade removal, re-parenting reconciliation (old parent stops advertising children, new leaf parent gains its expander with no refetch), last-child removal, and moves under a not-loaded parent
- docs: add a `LiveUpdate` Storybook story with interactive rename / move / remove controls over the simulated org-chart backend

#### Video Demo

https://github.com/user-attachments/assets/3c705c27-3e6d-45bc-b3fc-a3e070c6e297


sauldom102 (commented):
Would this work as well if we wanted to change some item name or other attributes?

AndresRojoFactorial:
@sauldom102 
> Would this work as well if we wanted to change some item name or other attributes?

For Renaming works, because it is related with the node data... maybe for inner values or representations of F0Graph (such as dim, focuses especial tags) are out of the scope or should be adress by a different function, these one is focused more on the tree representation.

https://github.com/user-attachments/assets/fd187f57-95fb-46a1-a34b-2571ceafe88f

siguenzaraul (approved):

PR #4689, opened 2026-07-14
feat(DetailsItem): add fullWidth prop to stretch to container width

AngelMsaa:
## Description

Adds an optional `fullWidth` boolean prop (default `false`) to the experimental `DetailsItem` so it can drop its ~320px width cap and stretch to 100% of its container. Purely additive and backwards-compatible: existing consumers render exactly as before.

**Motivation.** `DetailsItem` caps its width at ~320px (`max-w-80`, on `DataList`'s inner row). There was no way to make it fill its container, so consumers reached for unrelated props to bypass the cap. The Factorial monolith org-chart side panel sets `isHorizontal` + `verticalLayout` purely to drop the cap and fill the drawer, which couples "fill width" to "layout direction" and is confusing. `fullWidth` is the explicit, intent-revealing replacement, kept orthogonal to `isHorizontal` / `verticalLayout` (those control layout direction; this controls width).

New Storybook story: **Experimental / List / DetailsItem → `FullWidth`** (and **Experimental / List / DetailsItemsList → `FullWidth`**). Each renders inside a 600px container to contrast the two states. See the Chromatic build on this PR for the rendered story.

Verified in Storybook at desktop width (the `md:max-w-80` cap only applies at `md`+):

| State | inner `max-width` (computed) | measured width in 600px box |
| --- | --- | --- |
| default (unset) | `320px` | 320px (capped) |
| `fullWidth` | `none` | 566px (fills container) |

## Implementation details

- `DetailsItemType`: added `fullWidth?: boolean` with a JSDoc note (flows through `DetailsItemsList`'s `details: DetailsItemType[]` and surfaces in autodocs prop tables).
- `DetailsItem`: destructure `fullWidth = false`; when set, add `[&>div]:md:max-w-none` to the wrapper, dropping the `md:max-w-80` cap that lives on the child `DataList` row. This reuses the same arbitrary-variant idiom the component already uses to reach that child (e.g. `xs:[&>div]:flex-col`), so no change to `DataList`'s public API.
- `DetailsItemsList`: plumb `fullWidth={item.fullWidth}` through the per-item render (that layer picks props explicitly rather than spreading).
- Stories: added `FullWidth` stories to `DetailsItem` and `DetailsItemsList` with a 600px wrapper and a short doc note on when to use it.
- Tests: added coverage asserting the cap-override class is present only when `fullWidth` is set.

Local checks: `oxfmt --check`, `oxlint` (0 warnings), `tsc` typecheck, `vitest` (DetailsItem), and `storybook build` all pass.

> Draft for human review; do not merge.

### DIFF
--- a/packages/react/src/experimental/Lists/DetailsItem/__tests__/DetailsItem.test.tsx
+++ b/packages/react/src/experimental/Lists/DetailsItem/__tests__/DetailsItem.test.tsx
@@ -14,6 +14,33 @@ describe("DetailsItem", () => {
     expect(screen.getByText("Hello")).toBeInTheDocument()
   })
 
+  describe("fullWidth", () => {
+    // The ~320px cap (`md:max-w-80`) lives on DataList's inner div; DetailsItem
+    // drops it from the outside via the `[&>div]:md:max-w-none` variant only
+    // when `fullWidth` is set.
+    const capOverride = "[&>div]:md:max-w-none"
+
+    it("keeps the width cap by default", () => {
+      const { container } = render(
+        <DetailsItem title="Email" content={{ type: "item", text: "a@b.co" }} />
+      )
+
+      expect(container.firstChild).not.toHaveClass(capOverride)
+    })
+
+    it("drops the width cap when fullWidth is set", () => {
+      const { container } = render(
+        <DetailsItem
+          title="Email"
+          content={{ type: "item", text: "a@b.co" }}
+          fullWidth
+        />
+      )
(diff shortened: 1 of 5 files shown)


AngelMsaa:
Closing in favor of #4696, which removes the ~320px max-width cap outright instead of adding a `fullWidth` opt-in prop. See #4696 for the alternative approach.

_🤖 Managed by [Claude Code](https://claude.com/claude-code)_

PR #4695, opened 2026-07-15
fix(F0Form): autosubmit drops the first change on a pristine form

albertpmz:
## Description

Fixes a bug where **F0Form autosubmit silently drops the first change on a pristine form**. Toggling a single switch (or any single discrete change) with `submitConfig: { type: "autosubmit" }` did nothing — no submit fired — until a *second* change was made. Present in every released version (3.8.2, 4.34.0, 4.35.3) and `main`.

### Root cause

The debounce read `form.formState.isDirty` **synchronously inside the `form.watch` callback**. `watch` fires on the change before react-hook-form has recomputed `isDirty`, so the first change saw a stale `false`, scheduled no timer, and was lost. A second change worked because `isDirty` was already `true` by then.

### User impact

Single-control autosubmit forms are the ones that break (no second change to rescue the first). Example: Performance → review process → Visibility settings — three autosubmit switches with `hideActionBar: true`. An admin flips one toggle, the switch visibly moves, there's no Save button and no action bar, so it looks saved — but the mutation never fired and it reverts on reload.

N/A — no visual change; behavior-only fix.

## Implementation details

Moved the `isDirty` / `isSubmitting` guards **inside** the debounced `setTimeout`, where RHF state has settled (the timer fires ~800ms after the change):

```js
form.watch(() => {
  if (timer) clearTimeout(timer)
  timer = setTimeout(() => {
    if (!form.formState.isDirty) return      // now reads settled state
    if (form.formState.isSubmitting) return
    submit()
  }, delay)
})
```

Safe:
- `form.watch` never fires on mount → no spurious submit on load.
- Not-dirty / no-interaction path still no-ops (the guard just runs later).
- `form.reset()` after a successful submit triggers watch → schedules a timer → guard sees `isDirty === false` → no-op.
- Debounce, rapid-change, and validation behavior unchanged.

**Tests**
- New regression test: toggle a single switch on a pristine form, advance the timer, assert exactly one submit. It **fails on the old code, passes on the new** (verified by reverting the fix with the test in place).
- Full `F0Form` suite green: **442 passed**.

https://claude.ai/code/session_01Bb3Ln6Rk4Tn6mi1X2MBom4


albertpmz:
Closing: the core fix (reading `formState.isDirty` inside the debounce timer instead of synchronously in the `watch` callback) already landed on `main` via #4674 and has been shipped since `@factorialco/f0-react` v4.39.0. The remaining extras here (unmount guard, `isSubmitting` re-check, extra tests) can be resubmitted separately if wanted.

PR #4717, opened 2026-07-16
feat(TimelineRow): animate disclosure + motion polish

jespinosa19:
## What & why

`TimelineRow` was structurally solid but **static**: the multitask/nestedtask body appeared and disappeared instantly, and the nestedtask chevron hard-swapped between two icons (while the multitask header already animated its chevron via `GroupHeader → ChevronToggle`). This PR improves those animations, **reusing existing F0 primitives/patterns only** — no new tokens, no bespoke components.

### Animated expand/collapse
- New `TimelineCollapse` wraps the collapsible body of both `MultitaskRow` and `NestedtaskRow` and animates `height: 0 ↔ auto` + `opacity` with `overflow-hidden`, `initial={false}` (no animation on first paint) and `useReducedMotion` — **200ms, `easeOut`**. Same pattern F0 already ships in `F0ClarifyingPanel` (Motion's native `height:auto`, no ResizeObserver).
- Before: `{expanded && <body/>}` (instant show/hide). After: smooth, interruptible open/close.

### Animated chevron
- `NestedtaskHeader` now uses the **`ChevronToggle`** primitive (smooth 200ms rotation, respects reduced motion) instead of swapping `ChevronUp`/`ChevronDown`. Now consistent with the multitask `GroupHeader`, which already used it.

### Progress counter no longer jitters
- Added `tabular-nums` to the nestedtask progress counter (`completedCount/taskCount`) so the digits keep a fixed width as progress updates.

### Larger hit target
- Extended the nestedtask disclosure `<button>` to a 40px-tall target via an `::after` pseudo-element (visible control is ~20px; no layout shift, no overlap with adjacent rows).

## Scope notes
- **Multitask progress tag** (`F0TagStatus` `x/y`) left untouched — it's a primitive; tabular-nums would belong there, not here.
- **Per-item stagger on expand**: considered and intentionally *not* added — F0's collapsibles fade the body as a unit; child stagger fought the `height:auto` animation for no clear gain.
- **`text-wrap`**: N/A — `F0Text` doesn't take `className` for descriptions, and titles are single-line by layout.

## Verification
- Storybook (`SDS / TimelineRow`): exercised `NestedtaskInProgress` and `MultitaskWithNestedtask` (incl. collapse-within-collapse) — smooth open/close, chevron rotates, body unmounts after exit with no leftover gap, no console errors.
- `tsc --noEmit` (tsconfig-build): 0 errors in `sds/TimeLine`.
- `oxlint` + `oxfmt`: clean (also enforced by pre-commit hooks).

https://github.com/user-attachments/assets/fde7f693-e541-4252-8721-77cb81673445


jespinosa19:
Heads-up @danilojpferreira — this adds some motion to the nestedtask variant + content slot you built: animated expand/collapse (shared `TimelineCollapse`, same `height:auto` pattern as `F0ClarifyingPanel`) and a rotating chevron via `ChevronToggle` (so it now matches the multitask `GroupHeader`), plus `tabular-nums` on the progress counter and a 40px hit target. No API changes. Would love your eyes on it whenever you get a sec 🙏

danilojpferreira:
> Heads-up @danilojpferreira — this adds some motion to the nestedtask variant + content slot you built: animated expand/collapse (shared `TimelineCollapse`, same `height:auto` pattern as `F0ClarifyingPanel`) and a rotating chevron via `ChevronToggle` (so it now matches the multitask `GroupHeader`), plus `tabular-nums` on the progress counter and a 40px hit target. No API changes. Would love your eyes on it whenever you get a sec 🙏

@jespinosa19 as an improvement, add to the PRs the snapshot or videos. Some feedbacks can be also based on that.

jespinosa19:
> > Heads-up @danilojpferreira — this adds some motion to the nestedtask variant + content slot you built: animated expand/collapse (shared `TimelineCollapse`, same `height:auto` pattern as `F0ClarifyingPanel`) and a rotating chevron via `ChevronToggle` (so it now matches the multitask `GroupHeader`), plus `tabular-nums` on the progress counter and a 40px hit target. No API changes. Would love your eyes on it whenever you get a sec 🙏
> 
> @jespinosa19 as an improvement, add to the PRs the snapshot or videos. Some feedbacks can be also based on that.

Video attached! Thanks!!

eliseo-juan (changes requested):
The animation itself looks right (and I checked motion-dom 12.17 — `height: "auto"`
*is* restored as `auto` on completion via `DOMKeyframesResolver`/`getFinalKeyframe`,
so no stale pixel height). Four things I'd fix before merge:

**1. `overflow-hidden` is permanent and clips the focus ring and the new hit target**

It's a class, not an animated style, so it outlives the animation. Two concrete
consequences:

- `focusRing()` is `ring-1 ring-offset-1` (`lib/utils.ts:8`) → 2px **outside** the
  element box. The first/last focusable inside the collapsible body (nested headers,
  `Actions`) gets its focus indicator clipped — a WCAG 2.4.11 / 2.4.13 regression on
  keyboard navigation.
- The new `after:h-10` extends ~10px above/below a ~20px control. Inside a multitask
  collapse, `overflow: hidden` clips **hit-testing** too, so the larger hit target
  this PR adds doesn't work in exactly the nested case.

Suggestion: keep `overflow-hidden` only while animating (clear it in
`onAnimationComplete`, re-apply on exit start).

**2. New component with no tests**

`TimelineCollapse.tsx` ships without a test file — blocking per our review
checklist. The existing suite only covers static `expanded` true/false, so nothing
covers the mount-already-expanded path (`initial={false}`), the exit path, or
reduced motion.

**3. Adopting `ChevronToggle` pulls in an a11y violation**

`ChevronToggle` renders `<F0Icon role="button">` with no accessible name
(`ui/ChevronToggle/ChevronToggle.tsx:38`) → axe `button-name`, now nested inside the
real `<button>`. With axe coverage just extended to WCAG 2.2 A/AA and CI failing on
new `skipCi`, this either breaks the a11y job or propagates the debt. The icon
should be `aria-hidden` — fixing it in `ChevronToggle` also fixes the multitask
`GroupHeader`. (Colour is fine: `F0Icon` defaults to `currentColor`, so the
`text-f1-icon-secondary` wrapper works.)

**4. Content stays reachable during the 200ms exit**

`aria-expanded` is already `false` while the `role="region"` body (still referenced
by `aria-controls`) is in the DOM, so a screen reader can reach "hidden" content;
and if focus was inside it, it's dropped to `<body>` on unmount. Add
`aria-hidden`/`inert` on exit (or move focus first).

**One note on the justification**: the description says this is "the same pattern F0
already ships in `F0ClarifyingPanel`", but there that pattern lives in the *story* —
the shipped component explicitly delegates the animation to its parent *because
nested height animations conflict* (`F0ClarifyingPanel.tsx:30-35`). That's directly
relevant here: `MultitaskRow`'s collapse contains `NestedtaskRow` collapses. Could
you verify toggling an inner row *while the outer one is mid-animation*?

The real precedent is `F0Accordion/components/AccordionItem.tsx:67-83`, which is
identical — that makes three copies of this block. Worth extracting a shared
collapse primitive into `ui/` instead of a fourth local one (your version is
actually the best of the three: it's the only one handling reduced motion).

jespinosa19:
Thanks! Addressed all four:

- **`overflow-hidden`**: now clips only while animating, so the focus ring and
  the 40px hit target aren't cropped once open (driven off `useIsPresent` + a
  `settled` flag, which also covers already-expanded mounts and re-clips on exit).
- **Tests**: added open/closed, mount-already-expanded (`initial={false}`), the
  exit path, and reduced motion.
- **`ChevronToggle`**: icon is `aria-hidden` with no `role` now, which also clears
  it on the multitask `GroupHeader`.
- **Exit a11y**: the collapsing body is marked `inert` + `aria-hidden` while it
  leaves (via `useIsPresent`).

Also verified the nested toggle inside the multitask collapse: the outer wrapper
stays at `height: auto` and reflows, no stale pixel or stuck state. And pulled the
block into a shared `ui/Collapse` so it's not a fourth local copy; left the
`AccordionItem` and `F0ClarifyingPanel` migration for a follow-up.

PR #4720, opened 2026-07-16
feat(F0FileItem): add optional subtitle prop with doubled padding

albertpmz:
## Description

Adds an optional `subtitle` prop to the experimental `F0FileItem`. When provided, the component renders a pre-formatted secondary line under the file name and doubles its padding (both axes) so the two-line block stays visually balanced.

This unblocks the Factorial finance/expenses F0Form migration, where the signed-invoice file row (`DropItem`) shows an upload date under the file name — something `F0FileItem` couldn't express until now. The prop is kept general (a plain string) rather than date-specific so it also covers file size or any other secondary text.

## Screenshots

image

## Implementation details

- **New prop** `subtitle?: string` on `F0FileItemProps`. It's a plain, pre-formatted string so the component stays presentational and locale-agnostic — the consumer owns any formatting.
- **Layout:** the name is now wrapped in a `flex-col` alongside the subtitle. The subtitle renders muted (`text-f1-foreground-secondary`, `text-xs`) under the name. The no-actions `pr-3` moved from the ellipsis onto the wrapper.
- **Padding:** `cva` moved all padding into `compoundVariants` keyed on a new `withSubtitle` variant, doubling both axes when a subtitle is present:
  - `md`: `py-0.5 pl-0.5 pr-1.5` → `py-1 pl-1 pr-3`
  - `lg`: `p-1` → `p-2`
- **Story:** added `WithSubtitle`.
- **Tests:** added `__tests__/F0FileItem.test.tsx` (none existed) covering name/subtitle render, absence of the subtitle node when omitted, and padding for both sizes. 5/5 pass. `oxlint` + `oxfmt` clean.

No release note needed here beyond the auto-generated changelog (release-please).

### DIFF
--- a/packages/react/src/components/F0FileItem/F0FileItem.tsx
+++ b/packages/react/src/components/F0FileItem/F0FileItem.tsx
@@ -34,6 +34,12 @@ export interface F0FileItemProps extends HTMLAttributes<HTMLDivElement> {
   actions?: F0FileAction[]
   disabled?: boolean
   size?: F0FileItemSize
+  /**
+   * Pre-formatted date rendered under the file name (e.g. an upload date).
+   * The consumer owns locale and formatting. When set, the item grows to two
+   * lines and its vertical padding is doubled.
+   */
+  date?: string
 }
 
 /** @deprecated Use F0FileAction */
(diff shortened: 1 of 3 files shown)


sauldom102 (changes requested):
Why not better naming it `subtitle`? I assume we could pass any string here, being it a date or not

albertpmz:
Change the date prop for subtitle – not tighten it

albertpmz:
> Why not better naming it `subtitle`? I assume we could pass any string here, being it a date or not

Done! Thanks for the heads up 🫡

PR #4728, opened 2026-07-16
feat(F0Form): add delete persistence + confirmation to entitiesList

albertpmz:
## What

Adds **delete-action persistence + confirmation** to the F0Form `entitiesList` field, making removal symmetric with how add/edit already persist (`createFormDefinition` / `updateFormDefinition`).

New config on `F0EntitiesListOptions`:

```ts
onRemove?:      (item: T) => Promise<{ success: boolean } | void>
confirmRemove?: (item: T) => ConfirmDialogOptions
removableIds?:  Array<string | number>
```

Removal flow (`performRemove`): **confirm → persist → splice**
1. Shows a confirmation via `dialogs.confirmation(...)` — `confirmRemove(item)` copy if provided, else a sensible default. Cancel/dismiss aborts (nothing changes).
2. If confirmed, `await onRemove(item)`. A throw or `{ success: false }` **keeps the row** and surfaces a critical alert.
3. Only splices the row from the field value on success.

All four removal paths route through it — list-view overflow, editable-table, and both `rowActions` `remove()` helpers. A per-row pending state disables the action while `onRemove` is in flight.

Also: removal gets its own gating axis, **`removableIds`** — the remove counterpart to `editableIds`, and independent of it. A row can be editable but not removable (e.g. a pinned resource owner) or vice versa. Omitted → every row removable, mirroring `editableIds`. Wired through both visualizations: list-view and the editable-table (`F0FormEditableTable` gains a `canRemoveRow` prop).

## Why

`entitiesList` could self-persist add and edit (each dialog has its own `onSubmit`), but **remove only spliced the local array** — there was no hook to persist a deletion, and no confirmation. Per the "CRUD patterns / Delete & destructive" doc, every destructive action must confirm. Consumers with server-backed lists had to bolt on a custom `rowAction` to actually delete. This closes that gap natively.

## Behavior changes (flag for reviewers)

- The built-in remove now **always shows a confirmation** (previously immediate).
- Removal is gated by the new **`removableIds`**, independent of `editableIds`. Omitted → all rows removable (matches the prior default), so this is additive: existing lists behave unchanged unless they set `removableIds`.
- `onRemove` is optional; without it, removal stays value-only, just confirmed.

## Testing

- New list-view gating suite (`removableIds`): row outside it has no remove but stays editable; `removableIds` omitted → all removable (back-compat); edit/remove gate **independently** (removable-not-editable and editable-not-removable); removable row → Remove fires `dialogs.confirmation` + `onRemove(item)` and the row disappears.
- New editable-table test: the remove button shows only for rows inside `removableIds`.
- `entitiesList` renderer suite: **19 passed** (+5). Full `src/patterns/F0Form` (+ `F0FormEditableTable`): **494 passed**. `tsc` 0 errors. oxlint + oxfmt clean.
- Storybook: extended `Patterns/Forms/F0Form → Entities List Field Form Definitions` to demonstrate `onRemove`/`confirmRemove` and pin the first member via `removableIds` (logs CREATE/UPDATE/DELETE).

### DIFF
--- a/packages/react/src/patterns/F0Form/fields/entitiesList/types.ts
+++ b/packages/react/src/patterns/F0Form/fields/entitiesList/types.ts
@@ -253,6 +253,16 @@ export interface F0EntitiesListOptions<T = EntitiesListItem> {
    * (pencil) action that opens the edit dialog.
    */
   editableIds?: Array<string | number>
+  /**
+   * Restricts which items can be removed, matched against each item's `id`
+   * property. The remove counterpart to {@link editableIds}, and independent
+   * of it — a row can be editable but not removable, or vice versa. When
+   * omitted, every item is removable. Items without an `id` (e.g. rows just
+   * added by the user and not yet persisted) stay removable. A row hidden from
+   * this list shows no remove action (`list-view`) / no remove button
+   * (`editable-table`).
+   */
+  removableIds?: Array<string | number>
   /** Minimum number of rows required (defaults to 1 unless the field is optional) */
   minItems?: number
   /** Maximum number of rows allowed. When reached the add button is hidden. */
@@ -356,6 +366,8 @@ export type F0EntitiesListField = F0BaseField & {
   labels?: F0EntitiesListLabels
   /** Ids of the items that can be edited (matched against `item.id`) */
   editableIds?: Array<string | number>
+  /** Ids of the items that can be removed (matched against `item.id`) */
+  removableIds?: Array<string | number>
   /** Maximum number of rows allowed */
   maxItems?: number
   /** Per-column presentation options, keyed by item-schema property name */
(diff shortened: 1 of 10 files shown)


packages/react/src/patterns/F0Form/fields/entitiesList/types.ts
sauldom102, line 255: Do we have an equivalent `removableIds` property here? 🤔 If not we should add it I think

albertpmz: Great point! I've added a symmetrical prop to manage deletion permissions separately. If not provided, all rows will be deletable by default. 🙌🏼

sauldom102 (approved):

PR #4732, opened 2026-07-17
feat(upselling-kit): add dismiss option and confirmation opt-out to UpsellingAlert

daviz-fct:
## Description

Two related improvements to `UpsellingAlert`:

1. **Dismiss affordance** — adds an optional dismiss button, following the same pattern already used by `F0Alert`. When an `onDismiss` callback is provided, a close button is shown so users can dismiss the upsell; the consumer owns the dismiss behaviour (e.g. persist it to hide the alert for N days and let it reappear afterwards).
2. **Confirmation opt-out (bug fix)** — `UpsellingAlert` rendered its action via `UpsellingButton` without forwarding `showConfirmation` (which defaults to `true`), so there was no way to turn it off. When a consumer's `onRequest` only opens a modal or navigates instead of creating an upselling request, the button still showed the success dialog once `onRequest` resolved — surfacing a bogus **"Meeting requested! / One of our experts will contact you..."** confirmation for a request that was never made (reported from production). `showConfirmation` can now be set on the alert's `action`; it defaults to `true`, so existing behaviour is unchanged.

## Type of change

- [x] New feature (non-breaking change which adds functionality)
- [x] Bug fix (non-breaking change which fixes an issue)

image

## Implementation details

**Dismiss**
- feat: add optional `onDismiss` prop to `UpsellingAlert`; when provided, renders an `sm` outline close button with the `Cross` icon, consistent with `F0Alert`
- feat: position the dismiss button to the right of the upselling action and pin it to the top-right in the stacked (responsive) layout, aligning it with the action button
- feat: wrap the close button in a leaf component so `useI18n()` is only read when `onDismiss` is set, keeping the alert renderable without an `I18nProvider`
- docs: add a `Dismissible` story and an `onDismiss` boolean control to toggle the dismiss button, with an example handler that hides the alert for a number of days

**Confirmation opt-out**
- feat: add optional `showConfirmation` to the alert's `action` (typed from `UpsellingButtonProps["showConfirmation"]`) and forward it to `UpsellingButton`; defaults to `true` so current consumers are unaffected
- docs: add a `WithoutConfirmation` story documenting the modal-only / navigation use case

<details>
<summary>Verification</summary>

- `pnpm lint` (oxlint) — 0 errors on the changed files
- `pnpm format:check` — clean on the changed files
- `pnpm build:types` — only pre-existing, unrelated errors (missing optional peer deps: `axe-core`, `docx-preview`, `react-virtuoso`, `maplibre-gl`); nothing in `UpsellingKit`

</details>

## Follow-up (consumer side)

Once this releases and the version is bumped in the `factorial` monorepo, `frontend/src/modules/upselling_kit/components/UpsellAlert` should pass `showConfirmation: false` in its `action` (its `handleRequest` only opens the `ProductModal`).

### DIFF
--- a/packages/react/src/sds/UpsellingKit/UpsellingAlert/index.stories.tsx
+++ b/packages/react/src/sds/UpsellingKit/UpsellingAlert/index.stories.tsx
@@ -105,6 +105,26 @@ export const Dismissible: Story = {
   },
 }
 
+/**
+ * When the action only opens a modal or navigates instead of creating an
+ * upselling request, set `showConfirmation: false` so the success dialog
+ * ("request sent") isn't shown for a request that was never made.
+ */
+export const WithoutConfirmation: Story = {
+  args: {
+    ...Default.args,
+    title: "Open the product details",
+    description:
+      "Use showConfirmation: false when the action only opens a modal or navigates, so no success dialog is shown for a request that was never sent.",
+    action: {
+      ...Default.args!.action!,
+      label: "Learn more",
+      showConfirmation: false,
+      onRequest: () => alert("Open product modal"),
+    },
+  },
+}
+
 export const Snapshot: Story = {
   parameters: withSnapshot({}),
   args: Default.args,
--- a/packages/react/src/sds/UpsellingKit/UpsellingAlert/index.stories.tsx
+++ b/packages/react/src/sds/UpsellingKit/UpsellingAlert/index.stories.tsx
@@ -6,6 +6,17 @@ import { withSnapshot } from "@/lib/storybook-utils/parameters"
 
 import { UpsellingAlert } from "."
 
+/**
+ * Example dismiss handler: persist the dismissal so the alert stays hidden for
+ * a number of days and reappears afterwards.
+ */
+const dismissForDays = () => {
+  const HIDE_FOR_DAYS = 7
+  const hideUntil = Date.now() + HIDE_FOR_DAYS * 24 * 60 * 60 * 1000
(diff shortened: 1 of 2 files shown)


packages/react/src/sds/UpsellingKit/UpsellingAlert/index.stories.tsx
```diff
@@ -6,6 +6,17 @@ import { withSnapshot } from "@/lib/storybook-utils/parameters"
...
 import { UpsellingAlert } from "."
 
+/**
+ * Example dismiss handler: persist the dismissal so the alert stays hidden for
+ * a number of days and reappears afterwards.
+ */
+const dismissForDays = () => {
+  const HIDE_FOR_DAYS = 7
+  const hideUntil = Date.now() + HIDE_FOR_DAYS * 24 * 60 * 60 * 1000
+  window.localStorage.setItem("upselling-alert-hidden-until", String(hideUntil))
```
sauldom102: We might need a custom key for each upselling alert, otherwise each product upselling alert might overwrite each other times

daviz-fct: Good catch — fixed in 672e8b76d. `dismissForDays` now takes a per-alert `storageKey` (namespaced by module/feature), and each story passes its own key, so multiple upselling alerts no longer overwrite each other's dismissal state. This is the example handler in the story; the component itself just calls `onDismiss`, leaving persistence and key choice to the consumer.

sauldom102 (approved):
Great David! Thank you!

9Shuck (approved):

PR #4751, opened 2026-07-20
feat(toast)!: stabilize F0Toast + docs

desiree-np:
## What

Stabilizes the real `F0Toast` / `ToastProvider` to the behaviour spec we agreed on. Two commits:

1. **Position / width / duration** — bottom-centre of the content area, fixed 384px (full-width below `sm`), and action-aware duration (5s, 10s with an action).
2. **Loading, actions, close + a real bug fix + docs** —
   - `loading` variant (spinner) that resolves in place on the same `id`
   - inline actions on the trailing edge (link left, primary button right); **no close (✕) when there's an action** (the action is the primary interaction and the toast auto-dismisses, so a second control would only compete with it). A toast that does **not** auto-dismiss (`persistent`) keeps the ✕ even with an action, so it's never a dead-end.
   - top-align icon + controls when there's a description; `layout="position"` so in-place updates don't distort the text
   - **fix:** a toast auto-closed when resolving in place. The countdown reset now keys off the toast's content (not only `duration`), so a same-duration update (e.g. `success → error`) restarts the timer instead of inheriting the previous one.
   - docs consolidated into a single hand-written guide (`toast.mdx`) + one stories file; old `toasts.mdx` + playground retired.

## BREAKING CHANGE (intentional, major bump)

Two behaviour/API breaks, flagged so the release is a major bump:

- **`loading` variant** widens the public variant union from **4 → 5** (`F0ToastVariant`, and the `variant` of `ToastOptions` / `ToastProviderItem` / `toasts.open`). Additive at runtime — existing callers are unaffected — but an *exhaustive* `switch` over the union would need the new case.
- **Default duration for action-less toasts drops from 10s to 5s.** Every existing toast without an action now dismisses in half the time. This is the change most likely to generate "toasts disappear too fast" reports after the bump — consuming teams should pass an explicit `duration` if they relied on 10s.


packages/react/src/ui/Toast/F0Toast.tsx
```diff
@@ -69,8 +72,19 @@ const F0Toast = forwardRef<HTMLDivElement, F0ToastProps>(
...
     const [remainingTime, setRemainingTime] = useState(duration || 0)
+    const [prevDuration, setPrevDuration] = useState(duration)
     const [isPaused, setIsPaused] = useState(false)
 
+    // Reset the countdown when the toast is updated in place (its `duration`
+    // changes) — e.g. error → loading → success on the same id. Done DURING render
+    // (not in an effect) so `remainingTime` is never momentarily 0 while
+    // `duration > 0`: otherwise the auto-close effect below sees 0 in the commit
+    // between resolving and the reset, and dismisses the toast mid-resolution.
+    if (duration !== prevDuration) {
```
albertpmz: The reset infers "new toast" from a duration change, but `store.addItem` (store.ts:54) replaces in place and most variant pairs recompute to the same 5000. `success` → `error` on the same id at t=4.8s keeps `remainingTime ≈ 200ms`, so the error toast is dismissed a few frames after it paints — same bug, narrower.

Suggest a real identity signal: bump a `revision` in `addItem` when replacing, pass it through, compare that here instead of `duration`.

packages/react/src/ui/Toast/F0Toast.tsx
```diff
@@ -202,38 +233,38 @@ const F0Toast = forwardRef<HTMLDivElement, F0ToastProps>(
...
+                    handleActionClick(buttonAction, buttonAction.onClick)
+                  }
+                />
+              ))}
+            </div>
+          )}
+
+          {/* Close — only when there's no action (so it's never mis-clicked for
+              the action) and not while loading */}
+          {onClose && !isLoading && !hasActions && (
```
albertpmz: Hiding ✕ is right when the toast auto-dismisses, but this hides it when there's no timer too, giving two undismissable states:

- `persistent: true` + an action with `keepOpen` — no timer, no ✕, and the action doesn't close it.
- a `loading` toast that never resolves (navigate away mid-request, abort on unmount) — no timer, no ✕, actions suppressed at line 240. It survives for the rest of the session since the store is module-level.

```suggestion
          {onClose && !isLoading && (!hasActions || !duration) && (
```

albertpmz (commented):
Nice cleanup overall — the position/width/duration spec, the inline action layout and the `layout="position"` reasoning all read well, and the render-phase reset is the correct pattern for that bug. Three things I'd want resolved before this goes out as a major, all left as line comments:

1. The countdown-reset fix only covers duration *changes*, so the same early-dismiss survives for same-duration in-place updates.
2. `persistent` + an action, and a `loading` toast that never resolves, are both undismissable now that ✕ is suppressed.
3. No test covers any of the new behaviour — including the bug this PR fixes and the 10s → 5s default it changes.

One for the description rather than a line: the **default duration halving (10s → 5s)** is a behaviour break for every existing action-less toast in the app and isn't listed under BREAKING CHANGE. It's the change most likely to generate "toasts disappear too fast" reports after the bump, so worth calling out explicitly for consuming teams.

packages/react/src/hooks/toast/__tests__/toasts.test.tsx
```diff
@@ -242,7 +242,7 @@ describe("toasts API", () => {
       })
       expect(screen.getByText("Temporary")).toBeInTheDocument()
 
-      // Default lifetime is 10s; advance past it.
+      // Default lifetime is 5s (10s with an action); advance past it.
```
albertpmz: Two gaps here:

- The test still advances `10_100` (line 247), so it passes for a 5s *or* a 10s default — the constant this PR changes isn't actually asserted. Tighten to `5_100`, plus a sibling asserting a toast with `actions` lives past `5_100` and is gone at `10_100`.
- No regression test for the fix itself: `open({id, variant:"loading"})` → `open({id, variant:"success"})`, assert still mounted after a tick and gone at 5s. The render-phase `setState` in F0Toast.tsx:83 reads as a lint smell, so the next reader will plausibly tidy it back into `useEffect(…, [duration])` — the exact bug fixed here — and nothing goes red.

desiree-np:
Thanks @albertpmz — all four addressed (c0efb0a4c + 00340a767):

**1. Same-duration in-place reset** — good catch. The countdown reset now keys off the toast's content (`duration` + `variant` + `title` + `description`), not just `duration`, so a same-id `success → error` (both 5s) restarts instead of inheriting the old `remainingTime`. Added a regression test.

**2. Undismissable states** — two different cases:
- **`persistent` + a keepOpen action**: fixed. A toast that does *not* auto-dismiss now keeps the ✕ even with an action — a persistent toast must always be dismissable.
- **`loading` that never resolves**: kept ✕-less on purpose. A `loading` toast is a wait state — the user doesn't act on it or dismiss it; it's resolved by the caller when the operation settles. A ✕ would let someone hide a still-running operation, and an auto-timeout could kill a legitimately long one (a big export/sync). The only real risk is an *orphaned* loading (caller never resolves it), which is a caller contract — now documented: settle it (success/error) or close it by id in every path, including on failure or navigation.

**3. Tests** — tightened the default-duration test to 5s (it was passing for 5s *or* 10s), added a sibling asserting the 10s action window, and the resolve-in-place regression from #1.

**4. 10s → 5s callout** — added to the description under BREAKING CHANGE (the default halving for action-less toasts is the change most likely to surface as "toasts disappear too fast").

packages/react/src/ui/Toast/F0Toast.tsx
desiree-np: Fixed in c0efb0a4c — the countdown reset now keys off the toast's content (`duration` + `variant` + `title` + `description`), not just `duration`, so a same-id `success → error` (both 5s) restarts instead of inheriting the old `remainingTime`. Added a regression test (`restarts the countdown when a toast resolves in place on the same id`).

packages/react/src/ui/Toast/F0Toast.tsx
desiree-np: Two cases here:
- `persistent` + a keepOpen action: fixed in c0efb0a4c — a toast that does *not* auto-dismiss now keeps the ✕ even with an action (`!hasActions || !willAutoDismiss`), so persistent is never a dead-end.
- `loading` that never resolves: kept ✕-less by design. `loading` is a wait state the user doesn't dismiss — it's resolved by the caller when the op settles; a ✕ would let someone hide a still-running op, and a timeout could kill a legitimately long one (big export/sync). The orphaned case (caller never resolves) is a caller contract, now documented in 00340a767.

packages/react/src/hooks/toast/__tests__/toasts.test.tsx
desiree-np: Done in c0efb0a4c — tightened the default test to 5s (still present at `4_800`, gone by `5_100`), added a sibling asserting the 10s window with an action (alive past `5_100`, gone by `10_100`), and the resolve-in-place regression from the F0Toast comment above.

packages/react/src/ui/Toast/F0Toast.tsx
```diff
@@ -69,8 +72,23 @@ const F0Toast = forwardRef<HTMLDivElement, F0ToastProps>(
...
+    const contentSig = `${duration ?? ""}|${variant}|${title ?? ""}|${description ?? ""}`
+    const [prevSig, setPrevSig] = useState(contentSig)
     const [isPaused, setIsPaused] = useState(false)
 
+    // Reset DURING render (not in an effect) so `remainingTime` is never
+    // momentarily 0 while `duration > 0`, which would trip the auto-close effect
+    // below and dismiss the toast mid-resolution.
+    if (contentSig !== prevSig) {
+      setPrevSig(contentSig)
+      setRemainingTime(duration || 0)
```
albertpmz, line 92: `c0efb0a4c` reset `remainingTime` on in-place replacement but not `isPaused`, so the same bug survives through the other timer variable (that diff touches no `isPaused` or mouse-handler line).

Both handlers (lines 158-168) are guarded on `duration && duration > 0`, and `loading` has `duration === undefined`, so neither fires while loading. Hover an error toast, click Retry (goes `loading`), move away: `mouseLeave` can't clear `isPaused`, so the **settled** toast never ticks. Measured: still mounted after 60s with a live progress bar. Mirror case: hovering a `loading` toast never pauses it, so it dismisses under the cursor at 6s.

Fix, since the effect on line 140 already bails on `!duration`:

```tsx
const handleMouseEnter = () => setIsPaused(true)
const handleMouseLeave = () => setIsPaused(false)
```

Note this is the settled toast, not the loading one, so it's separate from the caller-owned-loading decision.

packages/react/src/ui/Toast/F0Toast.tsx
```diff
@@ -179,19 +205,32 @@ const F0Toast = forwardRef<HTMLDivElement, F0ToastProps>(
...
+            "pointer-events-auto flex flex-row gap-3",
+            // top-align icon + controls with the title when there's a
+            // description (multi-line); centre for a single-line toast
+            description ? "items-start" : "items-center"
+          )}
+        >
+          {isLoading ? (
             <div className="flex-shrink-0">
-              <F0AvatarAlert type={avatarType} size="sm" />
+              <Spinner size="small" className="text-f1-foreground-inverse" />
```
albertpmz, line 215: `Spinner` sets its own `aria-live`/`aria-busy` (`ui/Spinner/index.tsx:30-32`), so a loading toast nests two live regions inside the root's `role="status" aria-live="polite"`. Counted: 2 for loading, 1 for every other variant. New here, `main` doesn't import `Spinner`. Duplicate or dropped announcements depending on the screen reader.

Title already carries the message, so `aria-hidden="true"` on the spinner wrapper.

packages/react/src/ui/Toast/F0Toast.tsx
desiree-np: Fixed in 4efeaeac9 — the in-place reset now clears `isPaused` alongside `remainingTime`, so a pause set while hovering an error toast can't carry over through `loading` (where the guarded mouse handlers can't clear it) and freeze the resolved toast. Added a regression test: hover an error toast → Retry (→ loading) → resolve to success, and it still counts down and dismisses.

packages/react/src/ui/Toast/F0Toast.tsx
desiree-np: Fixed in 4efeaeac9 — the spinner wrapper is now `aria-hidden="true"`, so the toast keeps a single live region (the root `role="status"`). The title carries the message, and the `Spinner`'s own aria-live/aria-busy no longer nests a second one.

packages/react/src/ui/Toast/F0Toast.tsx
albertpmz: Hey! `setIsPaused(false)` in the reset clears the leak, but the handlers are still guarded on `duration > 0`, so hovering a toast while `loading` registers no pause, and the settled toast then counts down under a cursor hovering over the toast. And with force-clear, a new leak is introduced: any toast updating in place discards a legitimate pause, so a hovered toast that updates also dismisses under the cursor.

For example, let's say a user bulk-archives 12 items and the toast resolves from `loading` to "12 people archived" + an 'Undo' button **while the user has its cursor hovered over the toast**. The 10s will run while they read the text, and the button will be gone; hover is inferred from `duration` and changes rather than tracked. Tracking it unconditionally fixes all three, and the effect on line 140 already bails on `!duration`:

```
const handleMouseEnter = () => setIsPaused(true)
const handleMouseLeave = () => setIsPaused(false)
```

with `setIsPaused(false)` dropped from the reset.

desiree-np: You're right on both — the force-clear was a band-aid. Refactored in e8c8eaa85: pause is now derived from live hover (`isHovered`, set by the mouse handlers with **no** `duration` guard). So a hover during `loading` registers and holds the resolved toast under the cursor, and an in-place update no longer discards a legitimate hover pause. Removed the reset-on-update entirely. Updated the regression test: hovered → resolve in place → stays paused past its duration → resumes and dismisses on mouse leave.

albertpmz (approved):
LGTM! Great work 👏🏼

PR #4752, opened 2026-07-20
feat(F0Map): add map pattern built on MapLibre

AngelMsaa:
## Description

Adds **F0Map**, a data-driven map pattern for the f0 design system and the spatial sibling of F0Graph: pass `F0MapPoint[]` and the map owns marker rendering, clustering, label collision, selection and accessibility. It renders f0-themed light/dark MapLibre styles over keyless OpenFreeMap vector tiles, replacing the need for per-product Google Maps implementations. RFC: [F0Map - a map pattern for the f0 design system](https://app.notion.com/p/3a35e6e051ee816cb9e0f7a9446f43d3).

_Storybook: `Map/F0Map` (WithMarkers, WithClusters, WorldClusters, CurrentLocation). Screenshots to be added._

## Implementation details

- feat: `F0Map` shell - controlled/uncontrolled selection, separate `highlightedId` reveal channel, imperative handle (`getMap`, `focusMarker`, `fitToMarkers`, `clearSelection`), cooperative gestures, theme via `.dark` ancestor detection, tile-error retry banner and WebGL list fallback
- feat: semantic `F0MapMarker` variants (`default` / `workplace` / `employee` / `company`) over an internal `BaseMapMarker` engine; avatar-derived accent colors, selected pin indicator, zoom-based size step
- feat: screen-space clustering (`useClusters`) rendered as an avatar-pile `F0MapCluster` with zoom-to-bounds expansion

  <details>
  <summary>Scale disclaimer</summary>

  Markers are DOM elements with JS screen-space clustering, designed for
  workplace-scale datasets. Soft limit 200 markers (warns above it).
  Thousands of points (e.g. employee distribution) need a GL-native
  clustering path - planned follow-up, out of scope here. Consumers should
  pass a referentially stable `markers` array; `interactive`,
  `gestureHandling`, `minZoom` and `maxZoom` are read on mount.
  </details>

- feat: collision-aware marker labels (`useLabelCollision`) with pocket connectivity and canvas text measurement
- feat: accessibility - `F0MapList` operable text alternative (also the WebGL fallback), skip-to-list link, live marker-count announcements, keyboard-focusable clusters, i18n `map.*` namespace
- feat: current-location dot with persisted opt-in (`useCurrentLocation`), silent-only auto-show via the Permissions API, locate control
- feat: f0-themed MapLibre styles (light + dark) generated from OpenFreeMap Bright via `buildStyles.mjs` - colors derived from f0 tokens by perceptual matching, muted over the neutral ramp
- feat: export the pattern from `experimental/exports.ts`
- perf: zoom state stored as a threshold boolean and cluster/label/collapse recomputes bail out when output is unchanged, so camera animations don't re-render the marker layer per frame
- test: unit suite with a mocked `maplibre-gl` (markers, selection, imperative handle, WebGL fallback, gestures)

**Still pending (follow-up commits on this PR):** hook + sub-component unit tests, sub-component MDX docs, `F0MapControls`/`F0MapList` stories, and a Chromatic snapshot story.


factmarc10010 (commented):
Really strong, carefully-built pattern — clean separation (`F0Map` shell → semantic `F0MapMarker` → `BaseMapMarker` engine → GL layers → hooks), comments that explain the *why*, solid controlled/uncontrolled selection, reduced-motion, theming, imperative handle, and a genuine a11y scaffold. The perf instincts (zoom as a threshold boolean, identity-preserving bail-outs so camera frames don't re-render the marker layer) are the right ones. Nothing here needs rewriting — my notes are mostly product/scope decisions rather than code defects. Inline comments cover the specifics; the three I'd want to align on before merge are below.

### 1. Keyless OpenFreeMap dependency (biggest strategic risk)
Tiles, sprites *and* glyphs all resolve to OpenFreeMap hosts — a free, community-run, best-effort service with no SLA, no API key, and here **no fallback host, no self-host switch, no env/prop override**. For a design-system pattern that products across the org will embed, that's a single point of failure entirely outside our control. The PR frames this as replacing per-product Google Maps, but Google Maps ships an SLA. I'd want infra/product sign-off, plus at minimum a tile-host override and ideally a self-hosting story, before this becomes the sanctioned map.

### 2. Current-location silently geolocates on any map where permission was already granted
Documented as intentional, which is exactly why I'm flagging it: `useCurrentLocation` is always called, and on mount — if the Permissions API reports `granted` — it fetches and renders the dot **even when the consumer never set `showCurrentLocation`**. The prop only gates the *prompt*, not whether location is used at all. For a shared component that violates least-surprise (a workplace map that never asked for location still tracks the user). I'd gate the whole location path behind the prop. See the inline notes on `useCurrentLocation`.

### 3. Keyboard/AT: markers are focusable buttons *and* there's a parallel operable list
The shell says "the map has no focusable pins" and adds a skip-to-list link + sr-only list as the operable path — but `BaseMapMarker` renders real `<button>`s (always, since `onClick` is always passed) with no `tabIndex={-1}`. So keyboard users tab through up to ~200 canvas buttons in DOM (not visual) order, and AT announces every marker *and* every list item — the same data twice. Pick one operable path: either make the pins `tabIndex={-1}`/`aria-hidden` and rely on the list, or drop the list.

### Testing gaps (the PR already calls these out)
The suite is genuinely behavioral for a11y / WebGL-fallback / style-swap re-add. But clustering (`useClusters`), label collision (`useLabelCollision`) and arc math (`utils/arc`) are effectively untested because the maplibre mock returns fixed `project`/`cameraForBounds`, and there's nothing for `useCurrentLocation`, `useIsDarkContext`, `pointLabel` branches, `highlightedId`, or tile-error/retry. The trickiest logic (geolocation + the pure geometry functions) is the least covered — those pure functions don't need a real map, just per-case projection stubs. I'd want at least `useCurrentLocation` + the pure functions covered before we treat this as a stable pattern.

### Minor
- Barrel/PR description list variants as default/workplace/employee/company, but the code has 5 (includes `stop`).
- "locate" control uses the `SearchPerson` icon (a crosshair/target usually reads better for "locate me").
- The `styledata` listeners in the GL layers call `setData` on every `styledata` event (fires often) — minor redundant churn; `load`+`style.load` already cover the re-add.
- `useIsDarkContext` only observes the ancestor chain captured at attach — a `.dark` wrapper inserted as a *new* ancestor later isn't detected.
- `F0MapListProps.className` is marked `@private` but is part of the exported public props type.

packages/react/src/patterns/F0Map/hooks/useCurrentLocation.ts
```diff
@@ -0,0 +1,93 @@
...
+    // silently" contract. So always confirm with the Permissions API first.
+    const query = navigator.permissions?.query({
+      name: "geolocation" as PermissionName,
+    })
+    if (query) {
+      query
+        .then((status) => {
+          // Granted (here or elsewhere) → show silently. Otherwise a prompt is
+          // only acceptable when explicitly asked for (`promptOnMount`).
+          if (status.state === "granted" || promptOnMount) request()
```
factmarc10010: **Location is fetched regardless of `showCurrentLocation`.** This hook is always called by `F0Map`, and here — when the Permissions API reports `granted` — it calls `request()` and the dot renders on any map, even ones that never opted into the feature. The `showCurrentLocation` prop only maps to `promptOnMount`, i.e. it gates the *prompt*, not whether we geolocate at all. For a shared design-system component I'd expect the prop to gate the whole path: no `showCurrentLocation` → no geolocation, no dot. As-is the prop name is misleading and it's a least-surprise/privacy concern.

packages/react/src/patterns/F0Map/hooks/useCurrentLocation.ts
```diff
@@ -0,0 +1,93 @@
+import { useCallback, useEffect, useState } from "react"
+
+// App-wide opt-in flag: once the user grants location for any f0 map, we
+// remember it so every map auto-shows their location without re-prompting.
+const OPT_IN_KEY = "f0:map:current-location-opt-in"
```
factmarc10010: This opt-in flag is a single global `localStorage` key with no tenant/company scoping and it's never cleared (no logout reset, no expiry). On a shared/kiosk browser, or after switching companies, a prior grant silently re-enables silent location. It mostly only *drives* behavior in the `.catch` and no-Permissions-API paths (old Safari) — which is precisely where it silently fetches with no fresh consent. Worth scoping it and clearing it on logout, or dropping it in favor of the Permissions API alone.

packages/react/src/patterns/F0Map/F0Map.tsx
```diff
@@ -0,0 +1,616 @@
...
+        ...FLY_OPTS,
+        center: c,
+        zoom: Math.max(mapRef.current.getZoom(), 13),
+        animate: !reduceMotion,
+      })
+    )
+  }, [requestLocation, reduceMotion])
+
+  // Activating a list item selects the marker and flies to it (how keyboard /
+  // screen-reader users navigate; the map has no focusable pins).
```
factmarc10010, line 333: This says "the map has no focusable pins", but `BaseMapMarker` renders real `<button>`s (always — `F0MapMarkersLayer` always passes `onClick`) with no `tabIndex={-1}`. So the pins *are* in the tab order: keyboard users tab through up to ~200 canvas buttons in DOM (not visual) order, and AT announces every marker plus every list item — duplicate representations of the same points, which is what the skip-to-list link + sr-only list were meant to avoid. Recommend one operable path: either `tabIndex={-1}`/`aria-hidden` on the pins and rely on the list, or drop the list.

packages/react/src/patterns/F0Map/styles/assets/build-map-assets.sh
```diff
@@ -0,0 +1,41 @@
+#!/usr/bin/env bash
+# ---------------------------------------------------------------------------
+# Builds the one hosted asset F0Map needs beyond OpenFreeMap's free tiles:
```
factmarc10010: This whole Inter-glyph pipeline looks orphaned. The script says to generate/upload Inter SDF PBFs and "set the glyphs URL in `styles/index.ts`", but nothing reads a glyphs URL and `buildStyles.mjs` never sets `text-font` — the generated styles still point `glyphs` at OpenFreeMap (Noto Sans). So the "labels use f0's Inter font" premise is currently unfulfilled. Either wire it up or remove the script and the comments that reference it, so it doesn't read as done.

packages/react/src/patterns/F0Map/components/F0MapControls/types.ts
```diff
@@ -0,0 +1,28 @@
...
+  onZoomOut?: () => void
+  /**
+   * Fit all markers in view. When omitted (e.g. no markers), the button is
+   * hidden.
+   */
+  onFit?: () => void
+  /**
+   * Recenter on the user's location. When provided, a "locate" button renders
+   * as the first control; omit it (e.g. no geolocation) for a clean toolbar.
+   * Returning a promise drives the button's loading spinner while it resolves.
```
factmarc10010: This "returning a promise drives the button's loading spinner" contract isn't implemented — `F0MapControls` just does `onClick={onLocate}` with no loading state, and `F0Map`'s `handleLocate` returns void. Either implement the spinner or drop the claim (and the `Promise<void>` return) so the type doesn't promise behavior that doesn't exist.

packages/react/src/patterns/F0Map/styles/index.ts
```diff
@@ -0,0 +1,44 @@
+import maplibregl, { type StyleSpecification } from "maplibre-gl"
+// Bundled with the package as a same-origin URL asset (not a runtime CDN
+// fetch), so it works offline and adds no third-party dependency.
+import rtlTextPluginUrl from "@mapbox/mapbox-gl-rtl-text/mapbox-gl-rtl-text.js?url"
+
+import darkStyle from "./f0-dark.json"
```
factmarc10010, line 6: `f0-light.json` / `f0-dark.json` are ~84KB each (168KB) generated artifacts imported directly, so they bundle into the JS for every `@factorialco/react` consumer even when F0Map is unused — worth confirming tree-shaking actually drops them (a JSON import reached through a used barrel usually won't shake out). And `buildStyles.mjs` is a manual one-off (not in CI or any package script) that hard-`fetch`es OpenFreeMap at generation time with no retry, so the committed output can silently drift from upstream. Either add a CI regen+diff, or accept them explicitly as vendored assets with a documented regen step.

packages/react/src/patterns/F0Map/F0Map.tsx
```diff
@@ -0,0 +1,616 @@
...
+          <a
+            href={`#${listId}`}
+            className="sr-only rounded-md focus:not-sr-only focus:absolute focus:left-2 focus:top-2 focus:z-30 focus:bg-f1-background focus:px-3 focus:py-2 focus:text-sm focus:text-f1-foreground focus:shadow-md focus:outline-none focus:ring-1 focus:ring-f1-special-ring"
+          >
+            {i18n.map.skipToList}
+          </a>
+          {/* Announces the marker count to screen readers when it changes. */}
+          <div role="status" aria-live="polite" className="sr-only">
+            {`${markers.length} ${
+              markers.length === 1 ? i18n.map.location : i18n.map.locations
```
factmarc10010, line 526: This live-region count uses a manual `location`/`locations` JS ternary, while `map.cluster` uses proper i18n `{{count}}` pluralization. This breaks for languages with more than two plural forms — use the i18n plural mechanism here too for consistency.

siguenzaraul (approved):

PR #4755, opened 2026-07-21
feat(chat): separate voice notes from file uploads

zygisS22:
## What changed

`F0ChatCapabilities` exposes `canSendVoiceNotes`, so hosts can allow file uploads without also authorizing voice-note recording. Existing consumers keep the current behavior unless they explicitly opt out; dictation remains independently controlled by `transcribe`.

## Why

Supplying `uploadFiles` currently auto-enables voice notes in the chat composer. Factorial's Communications chat wants uploads without voice notes, and hiding the recorder with CSS or dropping uploads are both worse. The explicit capability keeps the choice with the host.

- defaults compatibly for existing hosts
- can disable voice notes while leaving attachments enabled
- does not affect optional voice dictation

## Verification

Focused ChatComposer tests cover upload preservation and dictation independence (4 tests). TypeScript, lint, and formatting pass.

Split out of #4713 so that PR stays a single concern (the composer `toolbarStart` slot).

### DIFF
--- a/packages/react/src/sds/chat/F0Chat/__tests__/ChatComposer.voiceNote.test.tsx
+++ b/packages/react/src/sds/chat/F0Chat/__tests__/ChatComposer.voiceNote.test.tsx
@@ -120,6 +120,42 @@ const recordVoiceNote = async (user: ReturnType<typeof userEvent.setup>) => {
 }
 
 describe("ChatComposer voice notes", () => {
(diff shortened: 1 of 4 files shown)


zygisS22:
Closing — not needed for the Analytics switcher work. If the Communications upload/voice-note coupling becomes a real need during the f0 version bump, this can be reopened or redone then.

PR #4757, opened 2026-07-21
feat(F0Form): forward onOpenChange/loading/searchFn/searchEmptyMessage on select fields

albertpmz:
## Description

Extend F0Form's native `fieldType: 'select'` config to forward four more props to the underlying `F0Select` — `onOpenChange`, `loading`, `searchFn`, `searchEmptyMessage` — plus fix a pre-existing gap where `icon`/`onCreate` were declared but never threaded.

Today a schema-declared select can't lazy-fetch options on first open, show a loading spinner while they fetch, customize search matching, or set a custom empty-state message. Consumers have to drop to a `fieldType: 'custom'` wrapper around `F0Select` for any of those (e.g. a contacts/vendor picker that lazy-loads on open). This makes those cases expressible with a plain native select field.

Additive/optional props only — non-breaking.

New Storybook story `F0Form/SelectLazyFetchOnOpen` demonstrates the lazy-fetch-on-first-open flow (options start empty, `onOpenChange` triggers a fake fetch, `loading` spinner shows meanwhile, searchable).

## Implementation details

- **`fields/select/types.ts`** — `onOpenChange`, `loading`, `searchEmptyMessage` added to `F0SelectConfigBase`; `searchFn` added to the options variant only, with `searchFn?: never` on the `source` and `customFieldName` variants so the "options-only" constraint (mirroring `F0Select`, where `searchFn` is invalid alongside `source`) is type-enforced.
- **`useSchemaDefinition.ts`** — threads the new props onto the resolved field; `searchFn` only in the non-`source` branch; `icon`/`onCreate` now threaded (the renderer already read them but the config never set them — silent drop).
- **`SelectFieldRenderer.tsx`** — forwards the props to `F0Select`. Field-level `loading` is OR-merged with the form's async-defaultValues loading (`loading: loading || field.loading`) so neither clobbers the other. No `searchFn` in source mode.
- **Tests/story** — 5 unit tests (prop forwarding, the loading merge, `searchFn` source-mode exclusion) + the lazy-fetch story.

Verified locally: `tsc --noEmit` 0 errors, `vitest --project=unit` 5/5, oxlint/oxfmt clean.

Note: this is the design-system enabler; the form field still stores a string value, so object-valued selects still store an id and re-map at submit.

### DIFF
--- a/packages/react/src/patterns/F0Form/__stories__/F0Form.stories.tsx
+++ b/packages/react/src/patterns/F0Form/__stories__/F0Form.stories.tsx
@@ -7,6 +7,7 @@ import { F0Button } from "@/components/F0Button"
 import { createDataSourceDefinition } from "@/hooks/datasource"
 import { Archive, ArchiveOpen, ExternalLink, Plus, Settings } from "@/icons/app"
 import { useF0FormDefinition } from "@/patterns/F0WizardForm"
+import { forms } from "@/patterns/forms"
 
 import type {
   FileUploadHookReturn,
@@ -14,8 +15,6 @@ import type {
   FileUploadStatus,
 } from "../fields/types"
 import type { RenderCustomFieldSelectConfig } from "../types"
-
-import { forms } from "@/patterns/forms"
 
 import {
   f0FormField,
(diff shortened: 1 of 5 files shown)


albertpmz:
Closed – current config is already enough for the use case we're trying to apply it to.

PR #4759, opened 2026-07-21
perf(OneDataCollection): debounce the collection→URL sync (fixes per-keystroke Sentry session churn)

factmarc10010:
## Why

`useDataCollectionUrlSync` reflects every collection state change (filters / search / sortings / visualization) back into the URL through `history.replaceState`, **with no debounce**. The search box updates `currentSearch` on **every keystroke**, so typing writes the URL once per letter.

The browser Sentry SDK (`@sentry/browser` v8, `browserSessionIntegration` — on by default) treats each `history` change as a **navigation** and **rotates the release-health session** on it. So each keystroke emitted **two session envelopes** — one `exited` (previous session) and one `init` (new session):

```
{"type":"session"} {"init":false,"status":"exited","errors":0,...}
{"type":"session"} {"init":true,"status":"ok","errors":0,...}
```

Effect: session counts are **inflated** by typing, crash-free-session rate is **diluted** with trivially-`ok` sessions, plus needless `replaceState`/network churn — across **every** OneDataCollection in the app, not just the one where it was noticed (the People org chart).

## What

Debounce the collection→URL write by **300ms**, so a burst of changes collapses into a **single** `replaceState` once the user pauses.

- The write must carry the **full** state: `setDataCollectionUrlParams` clears and rewrites all `dc_` params, so a search-only write would drop the filter/sort params. Hence the whole sync is debounced, not just search. Filters/sortings/visualization change on discrete clicks (no burst), so the ~300ms delay on them is imperceptible.
- The debounced function is stabilized with `useCallback([])` — `useDebounceCallback` re-creates its debounced instance whenever `func` changes, so an inline arrow would yield a new debounced identity every render, re-running the effect and cancelling the pending write on each render (which under frequent re-renders would never let it fire). `syncDataCollectionUrlParams` reads `window.location` itself and takes the full state as its argument, so it needs no deps.
- Pending writes are **cancelled** on deps-change / unmount, so a stale snapshot can't land after a newer one — nor stamp this collection's `dc_` params onto the page the user navigated to.

The URL → collection apply (on mount, after storage hydration) is unchanged and still immediate.

## ⚠️ Monitoring concern (please read)

This changes **Sentry Release Health**, which is monitoring-critical — flagging it explicitly:

- **Error/exception reporting is untouched.** Those are `type: event` envelopes on a separate path; only `type: session` (release health) envelopes are affected. Factorial's `Sentry.init` only adds `extraErrorDataIntegration` (no BrowserTracing), so no performance/navigation transactions are involved either.
- **Session counts will drop** after this ships (fewer navigation-triggered rotations). This is **more accurate**, not less — today's counts are inflated by per-keystroke rotations. But anyone watching Release Health dashboards/alerts keyed on **session volume** should be aware so a lower baseline isn't mistaken for an outage or a data-loss regression.
- The change is **app-wide** (core ODC hook), so it deserves review from an ODC / observability owner rather than riding along a feature PR — which is why it's split out here.

If we'd rather minimize blast radius, an alternative is to debounce **only search-triggered** writes and keep filter/sort/visualization writes immediate (they don't burst). Happy to switch to that if preferred.

## Testing

- `useDataCollectionUrlSync.test.tsx` updated to fake timers with a `flushUrlSync()` helper; added a coalescing test (typing `a`→`ad`→`ada` yields a single write with `ada`). 11/11 pass locally.
- `oxfmt` + `oxlint` clean.

### DIFF
--- a/packages/react/src/patterns/OneDataCollection/hooks/__tests__/useDataCollectionUrlSync.test.tsx
+++ b/packages/react/src/patterns/OneDataCollection/hooks/__tests__/useDataCollectionUrlSync.test.tsx
@@ -1,10 +1,15 @@
-import { afterEach, describe, expect, it, vi } from "vitest"
+import { act } from "@testing-library/react"
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest"
 
 import { DATA_COLLECTION_URL_PARAM_PREFIX } from "@/lib/providers/datacollection/dataCollectionUrlParams"
 import { FiltersDefinition } from "@/patterns/OneFilterPicker/types"
 import { zeroRenderHook as renderHook } from "@/testing/test-utils"
 
 import { useDataCollectionUrlSync } from "../useDataCollectionUrlSync"
+
+// The collection → URL write is debounced (URL_SYNC_DEBOUNCE_MS = 300ms), so
+// advance past it before asserting the URL. Comfortably above the debounce.
+const flushUrlSync = () => act(() => void vi.advanceTimersByTime(400))
 
 const FILTERS = {
   department: { type: "in", label: "Department", options: { options: [] } },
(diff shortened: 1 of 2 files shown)


eliseo-juan (approved):
LGTM

PR #4764, opened 2026-07-21
fix(SurveyFormBuilder): [FCT-59912] fix Add-option contrast + enforce a11y

fleveque:
## Description

https://factorialmakers.atlassian.net/browse/FCT-59912

The SurveyFormBuilder `SelectQuestion`, `RatingQuestion`, `DropdownSingleQuestion`, and `Form` stories were opted out of the axe check via `withSkipA11y()` with `// TODO: Fix a11y issues` notes. Auditing them found **one real violation and three stale opt-outs**: `RatingQuestion` (5 stories), `DropdownSingleQuestion` (4 stories) and the `Form` composite (9 stories) were already axe-clean, while `SelectQuestion` had a genuine **WCAG 1.4.3 (contrast)** failure. This fixes it and turns the check on for all of them.

This clears **every** SurveyFormBuilder entry from `.storybook/a11y-skip-allowlist.json`, completing the surveys group of the skipCi burndown: **36 → 32 files, 54 → 39 skip call-sites**.

## Implementation details

- fix: the **"Add option"** `F0Button` was dimmed via `opacity-50`, dropping its text contrast to **3.42:1** (below the 4.5:1 AA minimum). Raised it to **`opacity-70` (6.73:1)** — keeping the intended muted/secondary look while meeting AA (rather than removing the dimming, which would make it too prominent).
- fix: the SurveyFormBuilder "Applying changes" loading label (`ApplyingChangesTag`) had no explicit text color, inheriting a near-white color on its light pill (invisible). Set `text-f1-foreground` — resolves it in light theme, but CI surfaced a residual contrast reading (~1.3:1) that couldn't be reproduced locally, so the **`ApplyingChanges` story is scoped to `test: "todo"`** (reported, non-blocking) with a tracked follow-up.
- chore: replace the per-story `withSkipA11y()` opt-outs with meta-level `a11y: { test: "error" }` across `SelectQuestion` (5), `RatingQuestion` (5), `DropdownSingleQuestion` (4), and the `Form` composite (9). These now **block CI** on a11y violations (except the one `ApplyingChanges` story on `test: "todo"`). Removed the now-unused `withSkipA11y` imports and stale TODO comments.
- chore: drop all four SurveyFormBuilder entries from `.storybook/a11y-skip-allowlist.json` (the burndown list only shrinks), verified by `a11ySkipAllowlist.test.ts`.

## Verification

CI green: **22 of 23 SurveyFormBuilder stories enforced** at `test:"error"` (the 5 SelectQuestion stories flipped FAIL→PASS after the contrast fix; RatingQuestion, DropdownSingleQuestion + Form were already clean), with `ApplyingChanges` on `test:"todo"` pending the follow-up. Lint + typecheck + allowlist enforcement test clean.

The `DropdownSingleQuestion` opt-outs were confirmed stale by running axe over all 2050 stories twice (WCAG 2.0/2.1/2.2 A+AA, scoped to `#storybook-root`, 1280×720): all 4 of its stories reported **0 violations** in both runs, so enforcing them needed no component changes.

## Screenshots — "Add option" before / after

Small visual change: the "Add option" button's dimming is raised from **50% (3.43:1, fails AA)** to **70% (6.73:1, passes)** — still visibly muted/secondary, just legible. **Design: this keeps the muted intent while meeting AA — please confirm this treatment, or tell us the muted color you'd prefer** (a compliant color token would be more robust than opacity across themes/backgrounds).

| Before (dimmed, fails contrast) | After (fixed) |
| --- | --- |
|addoption-before|addoption-after|

## Note for review

⚠️ These SurveyFormBuilder stories are **not snapshotted by Chromatic** (f0 defaults `disableSnapshot: true`; they don't opt in via `withSnapshot`), so this visual change is **not** covered by visual regression — hence the manual screenshots above. Please eyeball it for design sign-off. (Adding a snapshotted `Snapshot` story is part of the component's Definition-of-Done and is out of scope here.)

ℹ️ A commit was added after the initial approval (`615c1d71d`): it removes the last four `DropdownSingleQuestion` opt-outs and its allowlist entry. No component code changed — stories only, plus the allowlist. Re-approval may be needed depending on branch protection.

### DIFF
--- a/packages/react/.storybook/a11y-skip-allowlist.json
+++ b/packages/react/.storybook/a11y-skip-allowlist.json
@@ -30,10 +30,7 @@
     "src/patterns/OneDataCollection/components/ActionBar/index.stories.tsx": 1,
     "src/patterns/OneDataCollection/hooks/useDataCollectionItemNavigation/__stories__/useDataCollectionItemNavigation.stories.tsx": 1,
     "src/kits/ai/F0AiChat/__stories__/F0AiFormTools.stories.tsx": 1,
-    "src/kits/surveys/SurveyFormBuilder/Form/index.stories.tsx": 1,
     "src/kits/surveys/SurveyFormBuilder/QuestionTypes/DropdownSingleQuestion/index.stories.tsx": 4,
-    "src/kits/surveys/SurveyFormBuilder/QuestionTypes/RatingQuestion/index.stories.tsx": 5,
-    "src/kits/surveys/SurveyFormBuilder/QuestionTypes/SelectQuestion/index.stories.tsx": 5,
     "src/ui/OnePagination/index.stories.tsx": 1,
     "src/ui/Select/__stories__/select.stories.tsx": 1,
     "src/ui/value-display/types/country/__stories__/country.stories.tsx": 1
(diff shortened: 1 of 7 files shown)


eliseo-juan (approved):
LGTM!

PR #4770, opened 2026-07-21
feat(Card): default the header link to a ghost appearance

warcos-fact:
## What

Changes the default appearance of `CardLink` (the widget-header link, top-right of every `Widget`) from a boxed control to a ghost icon button.

| | Before | After |
|---|---|---|
| background | `bg-f1-background-inverse-secondary` (dark: `bg-f1-background-tertiary`) | `bg-transparent` |
| border | `border-f1-border` | `border-transparent` |
| hover | `hover:border-f1-border-hover` | `hover:bg-f1-background-secondary-hover` |

Size, layout, typography, focus ring and the icon are unchanged, so the button does not move or resize.

## Why

The boxed style reads as a chip; a ghost affordance is the intended treatment for the subtle top-right action on widget headers.

## Scope

`CardLink` is shared, so this updates the default for **every** widget header link across f0-react (intentional).

## Tests

Adds a `CardLink` unit test asserting the ghost defaults (`bg-transparent`, `border-transparent`, `hover:bg-f1-background-secondary-hover`, and absence of the old fill). Full `Card` suite passes (17/17). Chromatic will cover the visual diff.

### DIFF
--- a/packages/react/src/ui/Card/Card.tsx
+++ b/packages/react/src/ui/Card/Card.tsx
@@ -147,9 +147,9 @@ const CardLink = React.forwardRef<
 >(({ className, title, icon = ChevronRight, href, ...rest }, ref) => {
   const sharedClassName = cn(
     "group inline-flex aspect-square h-6 items-center justify-center gap-1", //layout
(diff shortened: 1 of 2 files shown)


eliseo-juan (approved):
LGTM!

PR #4776, opened 2026-07-22
feat(F0DatePicker): add defaultMonth to control initial view

FaithAdekunle:
## Description

Adds an opt-in `defaultMonth` prop to `F0DatePicker` and the `F0Form` date field that controls which month the calendar opens on when the field has no value yet, without preselecting a value. This lets consumers open, for example, an end-date picker on the already-selected start date instead of today.

## Implementation details

- feat: add `defaultMonth` prop to `F0DatePicker`, forwarded through `DatePickerPopup` to `OneCalendar`
- feat: add `defaultMonth` to the `F0Form` date field config (`F0DateConfig`), accepting a static `Date` or a dynamic `({ values }) => Date` function, resolved alongside `minDate`/`maxDate`
- fix: `OneCalendar` no longer resets the view to today when nothing is selected

  <details>
  <summary>Why?</summary>

  On mount the calendar ran `setSelected(defaultSelected)`, and with no selection the view date fell back to a hardcoded `new Date()` — clobbering any provided initial month. The empty-selection fallback now uses `defaultMonth` (which itself still defaults to today), so a passed `defaultMonth` is respected.
  </details>
- docs: add `WithDefaultMonth` story and `defaultMonth` argType to the DatePicker Storybook docs


sauldom102 (changes requested):
Could you explain me a little bit your use case? Is it because you're setting a min/max date and the month opened by default is useless? Just wondering if we could be smarter here and instead of accepting a default month we should be opening the month depending on this min/max date passed, so we don't have to be setting this up every time

FaithAdekunle:
@sauldom102 Yes we set min and max values and yes that could be a solution for our use case. Ours is a case of start and end date where the start date becomes the min for end date. So making the calendar month for the end date field default to whatever the min is could work for us as well.

Thoughts??

sauldom102:
> @sauldom102 Yes we set min and max values and yes that could be a solution for our use case. Ours is a case of start and end date where the start date becomes the min for end date. So making the calendar month for the end date field default to whatever the min is could work for us as well.
> 
> Thoughts??

That would be great yes! I would then drop that default initial month we set here and would make it depending on the min/max date

sauldom102 (approved):

PR #4808, opened 2026-07-23
feat(categoryBarChart): center cell and add loading skeleton

AndresRojoFactorial:
## Description

Fixes the Teams-hierarchy table where the in-cell category bar chart was pushed to the top of the row (FCT-59059). The bar now sits in a text-line-height box and is vertically centered, so it lines up with the row's text without changing the row height and without touching the shared table primitive (the `<td>` stays `align-top` by design). 
The cell also gains a `loading` flag that renders a skeleton — the same size as the loaded bar — instead of flashing the empty dash before the data arrives.

## Implementation details

- fix: center the categoryBarChart bar within a text-line-height box (`flex h-5 items-center`) so it aligns with the row's text — no `h-full` (it doesn't resolve against the cell's `min-height`) and no min-height/padding that would make the row taller than a normal one
- feat: add a `loading` flag to the cell value that renders a `Skeleton` (`h-2 w-full`, matching the loaded bar) in place of the `–` dash while the row's data is still loading; `loading` takes precedence over the data points
- test: cover the loading state (skeleton instead of dash, matching size, precedence over data)
- docs: document the `loading` flag, add a `Loading` story and a `Loading` toggle on `DataCollectionExample` (skeleton inside a table), and update the MDX

### Before
image

### After
image

### DIFF
--- a/packages/react/src/ui/value-display/types/categoryBarChart/__stories__/categoryBarChart.mdx
+++ b/packages/react/src/ui/value-display/types/categoryBarChart/__stories__/categoryBarChart.mdx
@@ -17,6 +17,7 @@ type value = {
     color?: string        // optional color name or hex; falls back to categorical palette
   }>
   hideTooltip?: boolean   // when true, suppresses hover tooltips
+  loading?: boolean       // when true, renders a skeleton (same size as the bar) until data arrives
 }
 ```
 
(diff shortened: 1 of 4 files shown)


eliseo-juan (approved):
Lgtm!

PR #4815, opened 2026-07-23
feat(F0AnalyticsDashboard): polish metric (KPI) widget

igorsafonov-gif:
## What

Visual polish for the metric (big-number KPI) widget in `F0AnalyticsDashboard`:

- **Smaller value** — the number drops one type step (`text-4xl` → `text-3xl`).
- **Height-reactive alignment** — once the widget body is taller than 220px, the value centers both vertically and horizontally instead of hugging the bottom-left of a tall tile. Below that it stays bottom-left as before.
- **Header offset** — the centered value is nudged up 16px so it reads as optically centered against the whole card (which includes the title/description header), not just the body.

Height is tracked with a `ResizeObserver` (reusing the existing `useContainerSize` hook), so alignment reacts to grid resizes and fullscreen toggles.

## Scope

Single file: `MetricItem.tsx`. Applies to every metric widget (analytics dashboards and the in-chat dashboard canvas).

## Notes

- The 220px threshold is measured on the widget **body** (area below the header). Happy to retune if the intended reference is the full card height.
- Existing `MetricItem` unit tests pass unchanged (they assert formatted text, not layout).


zygisS22 (approved):
Approved after code, accessibility, Storybook, and local Factorial review.

Verified:
- Focused MetricItem suite: 11/11 passing, including fit-to-overflow-to-fit resizing and trend updates on the same mounted node.
- Formatting, TypeScript, lint, unit tests, Storybook tests, Chromatic, alpha publish, and public API checks pass.
- Published height variants cover the 144px compact bottom-left layout, 336px centered layout, positive and negative trends, and long overflowing values.
- Tall KPI content is centered with the intended 16px optical upward offset.
- Overflowing values receive a visible focus ring and keyboard focus only when scrolling is needed; ArrowRight scrolls the value, while fitting values stay out of the tab order.
- In local Factorial, the tall Active employees KPI rendered 34 centered in the card with the expected optical offset.

The accessibility gap found during review is fixed in this exact head. No remaining blockers.

siguenzaraul (approved):

PR #4821, opened 2026-07-23
feat(OneDataCollection): add grouping to the Kanban visualization

victorquiros98:
## Description

Adds the collapsible grouping mode to the Kanban visualization — the last built-in visualization missing it (table, card and list already support it). Motivated by onboarding, which groups tracked employees by policy **version**, where each version has its own set of columns (phases).

> **Draft for alignment (#f0-support).** The generic grouping follows the existing table/card/list pattern. The per-group columns API and the stacked-board layout are the two points I'd like Foundations to steer before finalizing — see "Open questions" below.

## Type of change

- Linked #f0-support thread: _pending — will link the alignment thread_

[Link to Figma Design](https://www.figma.com/design/Kb82lpJ0O95VatfTkMizsu/Onb-Policies-Co-creation?node-id=16518-30987)

Storybook: **Data Collection / Visualizations / Kanban → `KanbanWithGrouping`** (generic) and **Kanban Versions → `PerVersionColumns`** (per-version columns preview).

image

image

## Implementation details

- feat: add grouping to the Kanban visualization — pivots the per-lane grouped data into one board per group (reusing `GroupHeader` + `useGroups`); per-group counter, filter/toolbar above all groups, sort by column (not by group), collapsible header. Empty groups are omitted (standard data-driven grouping).
- refactor: extract a `KanbanBoard` component (one board's render + DnD/index maps) so the collection can render one board per group without changing flat behaviour.
- feat: add optional per-group columns via `getLanesForGroup(groupKey) => lanes` on the Kanban visualization options — each group renders its own columns while the fetch stays over the global union of lanes. Enables the onboarding case where each policy version has its own phases.

  <details>
  <summary>Why / status</summary>

  No visualization has per-group columns today. This is contained to the Kanban visualization (no changes to the fetch, source, or other visualizations) and the API shape is pending Foundations review.
  </details>
- test: cover the pivot, collapse/expand, and per-version columns (empty versions omitted).
- docs: add a "Kanban grouping" section to the Grouping docs.

### Open questions for Foundations

1. **API shape** for per-group columns — is `getLanesForGroup(groupKey) => lanes` on the Kanban options the right form?
2. **Stacked-board layout/height** — the Kanban UI is `h-full` (built for one board), so N stacked boards need a bounded height or the layout breaks. There's no existing F0 pattern; a provisional fixed height is used to preview. Affects all grouped Kanban, not just versions.


packages/react/src/patterns/OneDataCollection/visualizations/collection/Kanban/Kanban.tsx
```diff
@@ -289,40 +236,202 @@ export const KanbanCollection = <
...
+          variant: lane.variant,
+          color: lane.color,
+          total: items.length,
+          hasMore: false,
+          loading: laneData ? laneData.isInitialLoading : true,
+          loadingMore: false,
+          fetchMore: undefined,
+        }
+      })
+      const itemCount = boardLanes.reduce(
```
eliseo-juan: This computes the group counter from the records currently fetched, but the datasource already provides an authoritative one: group.itemCount, produced by grouping.groupBy[field].itemCount(groupId, filters) in packages/react/src/hooks/datasource/useData.ts:506. List and Table both use it directly (List/index.tsx:194, Table.tsx:620).

User-visible consequence: switching from Kanban to Table on the same dataset shows a different number for the same group. And under pagination the Kanban number is just "how much we've loaded so far" presented as a total.

Please pass group.itemCount through the pivot (it can be a Promise, which GroupHeader already handles via Await) and drop the reduce.

packages/react/src/patterns/OneDataCollection/visualizations/collection/Kanban/Kanban.tsx
```diff
@@ -289,40 +236,202 @@ export const KanbanCollection = <
...
+                data-testid={`kanban-group-${board.key}`}
+              >
+                <GroupHeader
+                  className="cursor-pointer select-none rounded-md px-3.5 py-3 transition-colors hover:bg-f1-background-hover"
+                  showOpenChange={collapsible}
+                  label={board.label}
+                  itemCount={board.itemCount}
+                  open={openGroups[board.key]}
+                  onOpenChange={(open) => setGroupOpen(board.key, open)}
+                />
```
eliseo-juan: Group-level selection is missing. useSelectableLanes already exposes groupAllSelectedStatus and handleSelectGroupChange (packages/react/src/hooks/datasource/useSelectable/useSelectable.ts:174, :538), and both Card (Card/index.tsx:458-469) and List (List/index.tsx:202-212) wire selectable / select / onSelectChange into GroupHeader.

So in a selectable collection, grouped Kanban is the only visualization where you can't select a whole group — with no indication that the capability exists elsewhere. Please wire the same three props.

packages/react/src/patterns/OneDataCollection/visualizations/collection/Kanban/Kanban.tsx
```diff
@@ -289,40 +236,202 @@ export const KanbanCollection = <
...
+          id: lane.id,
+          title: lane.title,
+          items,
+          variant: lane.variant,
+          color: lane.color,
+          total: items.length,
+          hasMore: false,
+          loading: laneData ? laneData.isInitialLoading : true,
+          loadingMore: false,
+          fetchMore: undefined,
```
eliseo-juan, line 361: hasMore: false, fetchMore: undefined and total: items.length hard-code away pagination. With infinite-scroll — which is exactly what KanbanWithGrouping uses (60 items, perPage: 20) — a grouped board silently shows only the first page, and each lane's counter reports the loaded count as the total.

This is documented as a limitation in grouping.mdx:195-196, but documenting a regression doesn't make it a design decision: a user looking at a grouped board has no way to tell that 40 of 60 records aren't there. Either keep total from paginationInfo.total so the counter stays truthful, or decide grouping requires a non-paginated source and fail loudly. Silently truncating is the one option I'd rule out.

packages/react/src/patterns/OneDataCollection/visualizations/collection/Kanban/Kanban.tsx
```diff
@@ -289,40 +236,202 @@ export const KanbanCollection = <
...
+    for (const lane of lanes) {
+      const data = lanesHooks[lane.id]?.data
+      if (data?.type !== "grouped") continue
+      for (const group of data.groups) {
+        if (!labelByKey.has(group.key)) {
+          labelByKey.set(group.key, group.label)
+          orderedKeys.push(group.key)
+        }
+      }
+    }
```
eliseo-juan: orderedKeys is built by first appearance while walking lanes, so currentGrouping.order is never honoured. useData groups with groupBy(data, GROUP_ID_SYMBOL) — a Map, i.e. insertion order within that one lane's fetched records (packages/react/src/hooks/datasource/useData.ts:501).

Concrete failure: lane todo contains only group B, lane doing contains A and B. Board order comes out B, A with order: "asc".

Note that the test at index.spec.tsx:375 asserts exactly this accidental order rather than the grouping order, so it stays green with the bug in place. Group order needs to derive from the grouping sort, not from lane traversal.

packages/react/src/patterns/OneDataCollection/visualizations/collection/Kanban/Kanban.tsx
```diff
@@ -289,40 +236,202 @@ export const KanbanCollection = <
...
+    // Group axis is data-driven: a group appears only if it has items (standard
+    // grouping behaviour), so empty versions are simply not shown.
+    const orderedKeys: string[] = []
+    const labelByKey = new Map<string, string | Promise<string>>()
+    for (const lane of lanes) {
+      const data = lanesHooks[lane.id]?.data
+      if (data?.type !== "grouped") continue
+      for (const group of data.groups) {
+        if (!labelByKey.has(group.key)) {
+          labelByKey.set(group.key, group.label)
```
eliseo-juan: The label is taken from whichever lane produced the group first. But useData computes it as groupConfig.label(groupKey, mergedFilters) — and mergedFilters is that lane's merged filters, since each lane fetches with its own filters (packages/react/src/hooks/datasource/useData.ts:503).

The signature receives filters precisely because the label may depend on them, so a consumer whose label varies with filters gets the one computed under an arbitrary lane's filter set. Same applies to itemCount once #1 is fixed. The pivot needs a lane-independent source for group metadata.

packages/react/src/patterns/OneDataCollection/visualizations/collection/Kanban/Kanban.tsx
```diff
@@ -289,40 +236,202 @@ export const KanbanCollection = <
...
+      const boardLanes: KanbanProps<R>["lanes"] = groupLaneDefs.map((lane) => {
+        const laneData = lanesHooks[lane.id]
+        const group =
+          laneData?.data?.type === "grouped"
+            ? laneData.data.groups.find((g) => g.key === key)
+            : undefined
+        const items = group?.records ?? []
+        return {
+          id: lane.id,
+          title: lane.title,
```
eliseo-juan, line 353: If getLanesForGroup returns a lane id that isn't in source.lanes, lanesHooks[lane.id] is undefined, so loading: laneData ? laneData.isInitialLoading : true pins that lane in a skeleton forever — no error, no warning, just a column that never loads.

Right now the only thing preventing this is a JSDoc sentence in types.ts saying the ids must exist. Please validate and console.warn in dev (or filter unknown ids out) rather than relying on the comment.

packages/react/src/patterns/OneDataCollection/visualizations/collection/Kanban/Kanban.tsx
```diff
@@ -289,40 +236,202 @@ export const KanbanCollection = <
...
+            <KanbanBoard<R>
+              lanes={flatLanes}
+              renderCard={renderCard}
+              getKey={getKey}
+              onCreate={onCreate}
+              idProvider={idProvider}
+              allowReorder={false}
+              loading={kanbanLoading}
+            />
+          ) : (
```
eliseo-juan, line 462: Two problems with this fallback path:

a) It omits onMove, so KanbanBoard renders without DndProvider — drag & drop silently disappears while it's showing.

b) It's not only a transient loading state. If currentGrouping.field isn't a key of grouping.groupBy, useData returns type: "flat" (packages/react/src/hooks/datasource/useData.ts:515), the data?.type !== "grouped" guard continues for every lane, and groupedBoards stays empty permanently. The collection then renders a flat board, with no group headers and no DnD, for what the consumer configured as a grouped Kanban.

That's the case the removed throw new Error("Grouping is not supported in Kanban yet") used to catch. Worth a dev-mode error for "grouping field not found in groupBy", and passing onMove through the fallback.

packages/react/src/patterns/OneDataCollection/visualizations/collection/Kanban/Kanban.tsx
```diff
@@ -289,40 +236,202 @@ export const KanbanCollection = <
...
       {lanesProvider}
       {lanesSelectProvider}
-      {!onMove ? (
-        <Kanban<R> {...kanbanProps} />
+      {isGrouped ? (
+        <div
+          className="flex flex-col gap-6"
+          aria-busy={kanbanLoading}
+          aria-live="polite"
+        >
```
eliseo-juan, line 448: aria-live="polite" is set unconditionally on the container that wraps all boards, so every change inside it — initial load, lane load, drag, collapse/expand — re-announces the entire board tree to screen reader users.

List scopes it to loading only: aria-live={showFullscreenLoading ? "polite" : undefined} (packages/react/src/patterns/OneDataCollection/visualizations/collection/List/index.tsx:186). Please match that, or move the announcement into a small dedicated status region.

packages/react/src/patterns/OneDataCollection/__stories__/visualizations/kanban-versions.stories.tsx
```diff
@@ -0,0 +1,215 @@
...
+          defaultOpenGroups: true,
+          groupBy: {
+            version: {
+              name: "Version",
+              label: (id) => `Version ${String(id).replace("v", "")}`,
+            },
+          },
+        },
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any -- preview adapter
+        dataAdapter: dataAdapter as any,
```
eliseo-juan: as any with an eslint-disable in committed code. If the adapter genuinely doesn't typecheck against useDataCollectionSource, that's a signal about the source API worth surfacing, not casting away. createDataAdapter from the existing story helpers should cover this — KanbanWithGrouping uses it without a cast.

eliseo-juan (approved):
LGTM!

PR #4834, opened 2026-07-24
fix(OneDataCollection): use column IDs for summary keys

PauloPhagula:
## Description

Use stable column IDs for summary-row React keys so tables can safely render multiple columns with the same visible label.

## Type of change

- [x] Bug fix

## Implementation details

- fix: derive summary-cell keys from the column ID, preserving the label fallback
- test: cover duplicate labels with distinct column IDs and assert no React console error

## Verification

- Focused TableCollection unit suite passes: 42 tests
- Formatting, targeted linting, and TypeScript checks pass

### DIFF
--- a/packages/react/src/patterns/OneDataCollection/visualizations/collection/Table/Table.tsx
+++ b/packages/react/src/patterns/OneDataCollection/visualizations/collection/Table/Table.tsx
@@ -873,7 +873,7 @@ export const TableCollection = <
                       )}
                       {columns.map((column, cellIndex) => (
                         <TableCell
-                          key={`summary-${String(column.label)}`}
+                          key={`summary-${column.id ?? String(column.label)}`}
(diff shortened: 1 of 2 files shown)


sauldom102 (commented):
Why would we want to have a table with columns with same label on the first place?

PauloPhagula:
> Why would we want to have a table with columns with same label on the first place?

In compensations one of the main use-cases is to compare inputs and outputs, i.e. that the values the HR admin sent to the bookkeeper are indeed equal (at least very close) to what they calculated.

On the sample image below, I highlighted some columns. Notice how the title is repeated, but the icon is different, representing that one is the inputs (HR admin) and the other output (Bookkeeper provided)

Screenshot 2026-07-28 at 11 51 11

It is exactly in these cases where the warning is shown

PauloPhagula:
CC @sauldom102

PR #4840, opened 2026-07-27
feat: smooth zoom transition for F0PdfViewer

jespinosa19:
## What & why

Zooming a PDF was **abrupt**: react-pdf re-rasterizes the page at the new scale on every step, tearing the canvas down and rebuilding it, so each step flashed a blank frame and read as a hard, flickery jump (most noticeable zooming out). This PR makes the zoom **ease smoothly**, reusing only existing primitives (a CSS transform plus a debounce), no new tokens or components.

### Decoupled render scale
- New `renderScale` is the scale react-pdf actually rasterizes at; it lags the target `scale` during a manual zoom.
- During the gesture only a cheap GPU `transform: scale(scale / renderScale)` on the already-rendered page animates (180ms, `cubic-bezier(0.2, 0, 0, 1)`), so there is no canvas rebuild and no blank frame.
- Once the user settles, `renderScale` catches up (debounced 200ms) and react-pdf re-rasterizes **once**, at the final size, so the page stays crisp. The visual size is identical across that swap, so there is no jump.
- Page-fit / page-width (and the initial fit) set `renderScale` immediately, so nothing animates on load.

## Scope notes
- **Only the zoom transition.** Toolbar, scroll, paging, rotate and visible-page tracking are untouched.
- `transform-origin: top center` so the page grows / shrinks from where it sits in the scroll, not the viewport centre.
- Respecting `prefers-reduced-motion` could be a nice follow-up, but the transition is short and transform-only.

## Verification
- Storybook (`Components / F0PdfViewer`): zoom in / out across scales, the page eases smoothly with no blank flash between steps and no animation on first load; the page is crisp again after settling.
- Page navigation, page-fit / page-width, rotate and visible-page tracking on scroll all still work.

### DIFF
--- a/packages/react/src/components/F0PdfViewer/F0PdfViewer.tsx
+++ b/packages/react/src/components/F0PdfViewer/F0PdfViewer.tsx
@@ -136,6 +136,10 @@ const PdfViewerBase = forwardRef<
   const [pdf, setPdf] = useState<PDFDocumentProxy | null>(null)
   const [pages, setPages] = useState<PageMetrics[]>([])
   const [scale, setScale] = useState(1)
+  // The scale react-pdf actually rasterizes at. It lags `scale` during manual
+  // zoom so the gesture animates with a CSS transform instead of rebuilding the
+  // canvas every step; it catches up (debounced) once the user settles.
+  const [renderScale, setRenderScale] = useState(1)
   const [currentPage, setCurrentPage] = useState(0)
   const [selectedScale, setSelectedScale] = useState<F0PdfScale>(initialScale)
   const [rotation, setRotation] = useState<F0PdfRotation>(initialRotation)
@@ -200,7 +204,10 @@ const PdfViewerBase = forwardRef<
           : (container.clientHeight - toolbarHeight - PAGE_VIEWPORT_PADDING) /
             pageHeight
 
+      // Fit changes rasterize immediately (no transform gap), so they never
+      // flash or animate on load.
       setScale(computed)
+      setRenderScale(computed)
       setSelectedScale(value)
     },
     [pages, currentPage, rotation]
@@ -320,6 +327,16 @@ const PdfViewerBase = forwardRef<
     return () => container.removeEventListener("click", handleClick, true)
   }, [])
 
+  // Manual zoom animates via a cheap CSS transform on each page (see the render
+  // below) so the gesture stays smooth. Re-rasterizing at the new scale is what
+  // flashes (react-pdf tears the canvas down and rebuilds it), so we defer it:
+  // once the user settles, catch renderScale up to scale in one debounced step.
+  useEffect(() => {
+    if (renderScale === scale) return
+    const id = setTimeout(() => setRenderScale(scale), 200)
+    return () => clearTimeout(id)
+  }, [scale, renderScale])
+
   return (
     <div
       ref={ref}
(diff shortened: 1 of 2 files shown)


eliseo-juan (changes requested):
Nice idea to decouple the render scale, but I think the mechanism can't hold the
guarantees the description claims. A few blocking things:

**1. `transform` doesn't change layout → pages overlap and get clipped during the whole gesture**

The `transform: scale()` lives on the page wrapper (`F0PdfViewer.tsx:371-400`), so
its layout box stays at `renderScale` for the gesture **plus** the 200ms debounce:

- With `transform-origin: top center` the painted page escapes its box and covers
  the next page in a multi-page document.
- `scrollHeight` / `scrollWidth` don't grow, so zooming past the viewport width
  clips the page against the surface's `overflow-hidden` (`:327`) instead of
  producing a horizontal scrollbar.
- When `renderScale` finally catches up, layout reflows and everything shifts.

So the hard jump isn't removed, it's deferred — and now happens twice (visual
scale, then layout shift). Repro: multi-page PDF, scroll to page 2, zoom to 200%.

Could we reserve the final size on the wrapper (explicit `width`/`height` at the
target scale) instead of only transforming, so layout and paint stay in sync?

**2. Visible-page tracking reads layout, not the transform**

`calculateVisiblePage` uses `offsetTop`/`offsetHeight` (`visiblePage.ts`) and
`goToPage` uses `element.offsetTop` (`:173-178`) — both blind to the transform.
During that window the toolbar page indicator and next/prev page target offsets
that don't match what's on screen.

**3. The skeleton is sized off the wrong scale**

`pageSkeletonWidth/Height` still derive from `scale` (`:151-152`) while `Page`
rasterizes at `renderScale`, so the `loading` skeleton shown while re-rasterizing
is sized at the target scale, not the rendered one.

Related: "the visual size is identical across that swap, so there is no jump" —
when react-pdf recreates the canvas, assigning `canvas.width` clears it, and the
transform resets to 1 in the same commit. Could you attach a capture at 300% on a
throttled machine? I'd expect one blank frame at the end of the gesture.

**4. Division by zero / NaN**

`applyDynamicScale` computes `(clientHeight - toolbarHeight - padding) / pageHeight`.
In a zero-height or hidden container (tab, drawer, collapsed panel — realistic for
a document preview) that's 0 or negative, and it's now written into `renderScale`
too → `transform: scale(Infinity)` / `scale(NaN)`. Needs a guard:

    const ratio = renderScale > 0 ? scale / renderScale : 1

**5. `prefers-reduced-motion`**

An animated zoom in a document viewer is exactly the vestibular-trigger case, and
we already have `useReducedMotion()`. It's one line here rather than a follow-up:

    transition: shouldReduceMotion || renderScale === scale ? "none" : "transform 180ms ..."

**6. No tests for a new state machine on a stable component**

This adds a second scale state plus a timer to a component in `components/`
(stable), and `F0PdfViewer.test.tsx` already exists. Minimum: fake timers, assert
which `scale` reaches `Page` before/after the debounce, and that fit modes never
open a transform gap.

jespinosa19:
Thanks! Addressed in `46b6aab`:

**1. Layout vs paint (overlap / clip / deferred reflow).** Fixed by reserving the page box at the target size and transforming only the canvas inside it. A "sizer" div now gets `width/height = intrinsic * scale` (the final size) and owns the frame + `overflow-hidden`; the canvas is scaled within it from `renderScale` up or down to the target. So the layout box is always the target size: scroll grows immediately (a real horizontal scrollbar past ~100%, no clipping), neighbouring pages never get covered, and there is no reflow when `renderScale` catches up. Your repro (page 2, 200%) now scrolls correctly.

**2. Visible-page tracking blind to the transform.** Falls out of #1: the page ref now lives on the target-sized sizer, so `offsetTop`/`offsetHeight` in `goToPage` and `calculateVisiblePage` match what is on screen.

**3. Skeleton sized off the wrong scale.** Fixed, the per-page loading skeleton is now `intrinsic * renderScale` (what is actually rasterizing). And you are right about the blank frame: react-pdf hides the canvas during re-raster, so there is a brief blank at the end of the gesture. It now happens inside the stable, target-sized frame, so it reads as an in-place content refresh rather than a size jump.

**4. Division by zero / NaN.** Fixed. `applyDynamicScale` bails when the computed scale is non-finite or `<= 0` (hidden or zero-height container), and the zoom ratio is clamped to a finite positive value, so `scale(Infinity)` / `scale(NaN)` can no longer happen.

**5. `prefers-reduced-motion`.** Added via `useReducedMotion()`; the transition is skipped entirely when it is set.

**6. Tests.** Added two: the box is reserved at the target size the moment you zoom (layout leads), and the crisp re-raster is debounced (`Page` keeps its old `renderScale` right after the click, then catches up). One note: I kept fit-to-view animating (product request) rather than instant, which is now safe because the sizer keeps layout correct regardless of the transform gap. Happy to make fit instant if you would prefer.

PR #4841, opened 2026-07-27
feat(EditableTable): support per-row dateConfig function for date cells

FaithAdekunle:
## Description

Lets an editable-table date column's `dateConfig` be a **function of the row item** (`(item) => DateCellConfig`) as well as a static object. This makes a date cell's `minDate`/`maxDate` able to depend on other fields in the same row — e.g. bounding an end date's lower bound by that row's start date — which the current static, per-column config can't express.

https://github.com/user-attachments/assets/bd96ae32-310c-4817-83c9-b11c1c10ccce

## Type of change

## Implementation details

- `EditableTable/types.ts` — `dateConfig?: DateCellConfig | ((item: R) => DateCellConfig)`, mirroring the existing `numberConfig.units` static-or-function convention.
- `EditableTable/components/cells/DateCell.tsx` — resolves the config against the current row item before reading `minDate`/`maxDate`.
- The read-only cell only uses `dateConfig` as a truthy "is this a date column?" flag, so a function value works there unchanged.
- Added an interactive story (`EditableTableWithPerRowDateConfig`) with rows carrying different start/end dates, demonstrating per-row bounds.

Backward compatible — the static-object form is unchanged.

### DIFF
--- a/packages/react/src/patterns/OneDataCollection/__stories__/visualizations/editable-table/editable-table.stories.tsx
+++ b/packages/react/src/patterns/OneDataCollection/__stories__/visualizations/editable-table/editable-table.stories.tsx
@@ -1,5 +1,5 @@
 import { Meta, StoryObj } from "@storybook/react-vite"
-import { format } from "date-fns"
+import { format, parseISO } from "date-fns"
 import { useMemo, useRef, useState } from "react"
 import { action } from "storybook/actions"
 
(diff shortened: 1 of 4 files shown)


sauldom102 (changes requested):
Can we include some unit tests?

sauldom102 (approved):

PR #4843, opened 2026-07-27
feat(F0AiMessagesContainer): welcome screen caption and subtitle

igorsafonov-gif:
## Description

Adds two optional props to the AI chat welcome (start) screen so hosts can label mode-specific chats:

- `initialMessageCaption` — a static line rendered **above** the gradient typewriter phrase, same size (`text-2xl font-semibold`) in `text-f1-foreground-secondary`. E.g. "Analytics mode:".
- `initialMessageSubtitle` — a smaller (`text-lg`) secondary line rendered **below** the phrase. E.g. "Ask about employees, contracts, absences, and presence. More data soon."

Both are pass-throughs to the internal `WelcomeScreen` (`caption`/`subtitle`); nothing changes when omitted. First consumer: One's Analytics mode start screen in the factorial monorepo (wiring lands there after this releases).

## Type of change

## Implementation details

- `WelcomeScreen` wraps the phrase in a column: optional caption (`mb-1`), phrase, optional subtitle (`mt-3`). Inline `style={{ minHeight: 28 }}` replaced with `min-h-[28px]`.
- Two adjacent a11y fixes surfaced by review of these lines:
  - The typewriter now respects `prefers-reduced-motion` (`useReducedMotion()` from `@/lib/a11y`): renders the first phrase statically with no rotation (WCAG 2.2.2).
  - `aria-label` was set on a plain `<p>` (prohibited on the paragraph role, and it fought the caption by exposing a partially-typed slice). It now applies only in the interactive (`role="button"`) case; a `sr-only` span carries the full stable phrase, with the typed slice `aria-hidden`.
- New `Snapshot` story (`withSnapshot({})`) whose `play` waits for the typed span to match the sr-only span, pinning Chromatic's capture to the settled state.

## Testing

- `WelcomeScreen.test.tsx`: caption/subtitle presence, styling, and DOM order; both omitted; reduced-motion static render; no `aria-label` on the non-interactive paragraph. Existing interactive/fullscreen tests updated to query via the sr-only phrase.
- New `F0AiMessagesContainer.test.tsx`: integration test asserting the caption/subtitle props reach the welcome screen.
- `pnpm vitest run src/kits/ai`: 268 tests green; `tsc`, `oxlint`, `format:check` clean.

## Follow-up commits (design review in the consuming app)

- **Rotation order + immediacy**: swapping the phrase list restarts the typewriter at the first phrase of the new list and begins typing immediately (the 400ms settle delay stays first-mount-only). Regression-tested.
- **Visual polish**: caption sits flush above the phrase, subtitle drops to `text-base`, both fade in from transparent over 500ms.
- **Layout stability**: the welcome suggestions row reserves two chip rows (`min-h-[72px]`, bottom-packed) so suggestion-set swaps can't rewrap 1↔2 rows and shift the greeting.
- **History markers**: `ChatThread` gains optional `icon`/`trailingLabel`, rendered by `ThreadItem` (icon before title, hover-revealed label) — first consumer marks Analytics chats. Companion PRs: factorialco/factorial-agent#2296, factorialco/factorial#107629.

https://github.com/user-attachments/assets/75ad024f-f0e8-41a7-88f5-b1afff34602b


warcos-fact:
Is that Chat/Analytics toggle part of the UI or is it just for the demo?

igorsafonov-gif:
@warcos-fact it's part of the UI, but it's already shipped.

siguenzaraul (approved):

PR #4853, opened 2026-07-27
feat(react): stamp data-f0-component-name via Component factory

sauldom102:
## Description

Stamps a static `data-f0-component-name` attribute onto the root DOM node of every component built with the `Component(meta, …)` factory, using `meta.name`. This marks *which* F0 component rendered a node — a per-component-type identity that is the same for every instance — so tooling can map a rendered DOM node back to its F0 component. It is always present (independent of the XRay debug overlay) and adds no extra DOM nodes.

This is deliberately separate from `data-testid`, which is a per-instance, consumer-supplied test hook (each `<F0Button>` can carry a different id) — a different concern, left untouched.

### Why this matters

This is groundwork for **static analysis of the rendered DOM in the monorepo and the composer**. Once every f0 component identifies itself in the DOM, we can run analysis over a rendered tree to **guarantee composition rules** (e.g. which components may/may not nest inside which) and to **assert that certain components include certain things** (required children, expected structure, etc.). Without a reliable, component-type-level marker on the DOM there's no dependable way to attribute a node to an f0 component — `data-testid` can't serve this because it's per-instance and optional. `data-f0-component-name` gives that stable anchor.

## Implementation details

- feat: the `Component` factory (`lib/component/component.tsx`) now stamps `data-f0-component-name={meta.name}` on the wrapped component's root node via a mount effect on the forwarded ref

  <details>
  <summary>Why the Component factory?</summary>

  It is already the F0 component-identity wrapper — it carries `{ name, type }` metadata and already stamped the node with `meta.name` for the XRay debug overlay (gated behind XRay). Making that stamp always-on, under the dedicated `data-f0-component-name` attribute, reuses the right abstraction instead of overloading the per-instance `data-testid` mechanism.
  </details>

- feat(F0AudioPlayer, F0AudioPlayerCard, F0VideoPlayer): route the media players through the factory so their roots carry the attribute

  <details>
  <summary>F0VideoPlayer note</summary>

  `F0VideoPlayerInternal` was a plain function component (no ref forwarding), so the factory's ref-stamp couldn't reach its DOM node. It's now a `forwardRef` that composes the forwarded ref with its existing wrapper ref (used for fullscreen/focus) via `useComposedRefs`.
  </details>

- test: cover stamping by `meta.name`, stamping without any XRay provider, ref forwarding, and per-component assertions on the audio/video roots
- chore: export `F0_COMPONENT_NAME_ATTRIBUTE` for reuse in tooling/tests

## How to verify

Rendered in Storybook and inspected the live DOM: `components-f0audioplayer--default` → root has `data-f0-component-name="F0AudioPlayer"`; `f0videoplayer--default` → root has `data-f0-component-name="F0VideoPlayer"`.

### Coverage / definition of done

Full coverage requires every public component to be built through the `Component(meta, …)` factory. Today only ~12 components use it, so this PR establishes the mechanism and migrates the media players; routing the remaining components through the factory (each needs a `type` classification) is tracked in #4854.

### DIFF
--- a/packages/react/src/components/F0AudioPlayer/__tests__/F0AudioPlayer.test.tsx
+++ b/packages/react/src/components/F0AudioPlayer/__tests__/F0AudioPlayer.test.tsx
@@ -36,6 +36,14 @@ describe("F0AudioPlayer", () => {
     expect(screen.getByRole("button", { name: "Play" })).toBeInTheDocument()
   })
 
+  it("marks the root node with data-f0-component-name", () => {
+    render(<F0AudioPlayer src="test.mp3" />)
+    expect(screen.getByRole("group", { name: "Audio player" })).toHaveAttribute(
+      "data-f0-component-name",
+      "F0AudioPlayer"
+    )
+  })
+
   it("renders the audio element with the given src and preload", () => {
     render(<F0AudioPlayer src="test.mp3" preload="auto" />)
     const audio = getAudio()
--- a/packages/react/src/components/F0AudioPlayer/index.tsx
+++ b/packages/react/src/components/F0AudioPlayer/index.tsx
@@ -1,3 +1,4 @@
+import { Component } from "@/lib/component/component"
 import { withDataTestId } from "@/lib/data-testid"
 import { experimentalComponent } from "@/lib/experimental"
 
@@ -21,7 +22,9 @@ export type { AudioPlayerState, AudioPlayerControls } from "./useAudioPlayer"
  */
 export const F0AudioPlayer = experimentalComponent(
   "F0AudioPlayer",
-  withDataTestId(F0AudioPlayerBase)
+  withDataTestId(
+    Component({ name: "F0AudioPlayer", type: "info" }, F0AudioPlayerBase)
+  )
 )
 
 /**
@@ -29,5 +32,10 @@ export const F0AudioPlayer = experimentalComponent(
  */
 export const F0AudioPlayerCard = experimentalComponent(
   "F0AudioPlayerCard",
-  withDataTestId(F0AudioPlayerCardBase)
+  withDataTestId(
+    Component(
+      { name: "F0AudioPlayerCard", type: "info" },
+      F0AudioPlayerCardBase
+    )
+  )
 )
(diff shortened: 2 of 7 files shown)


eliseo-juan (commented):
### 🔴 The stamp is lost if the root node changes identity after mount

In `packages/react/src/lib/component/component.tsx`:

```tsx
useEffect(() => {
  const element = ref.current
  if (element instanceof Element) element.setAttribute(F0_COMPONENT_NAME_ATTRIBUTE, meta.name)
}, [ref, meta.name])
```

`ref` is a stable `useRef` object and `meta.name` is constant, so **the effect runs exactly once**. If the wrapped component later replaces its root element (skeleton → content, conditional root, `asChild`, polymorphic `as`), React does `ref.current = null` → `ref.current = newNode` without re-running the effect, and the attribute silently disappears.

Reproduced on this branch — fails:

```
× keeps the marker after the root element is replaced
Expected: data-f0-component-name="F0Swap"   Received: null
```

```tsx
const Swapping = forwardRef<HTMLDivElement, {}>((_, ref) => {
  const [loaded, setLoaded] = useState(false)
  useEffect(() => { setLoaded(true) }, [])
  return loaded
    ? <div ref={ref}>Content</div>
    : <section ref={ref}>Skeleton</section>
})
```

This undercuts the premise of the PR: a marker that is *sometimes* missing is worse than no marker, because the analyzer can't tell "not an f0 component" from "the attribute got dropped". The media players in this PR happen to have a stable root, but the pattern is common enough that it will bite during the #4854 rollout.

**Suggested fix** — a callback ref instead of a mount effect, so it re-stamps whenever the node changes. I ran this locally: the repro passes and all 66 existing tests in `lib/component/`, `F0AudioPlayer` and `F0VideoPlayer` stay green.

```tsx
const { ref } = useComponentXRay(meta, forwardedRef)

const stampedRef = useCallback(
  (node: R | null) => {
    ref.current = node
    node?.setAttribute(F0_COMPONENT_NAME_ATTRIBUTE, meta.name)
  },
  [ref]
)

return <Component ref={stampedRef} {...props} />
```

It also drops the `instanceof Element` check, which is unnecessary here and fails across realms (e.g. a portal into another `document`).

PR #4865, opened 2026-07-28
chore(storybook): [FCT-59941] align the a11y addon's axe rules with CI

fleveque:
## Description

https://factorialmakers.atlassian.net/browse/FCT-59941

The Storybook a11y **addon** and the **CI** test-runner ran different axe rule sets, and nothing surfaced the difference. CI passes an explicit `runOnly` tag filter; the addon had no `options`, so it ran axe's **defaults**.

axe ships some rules `enabled: false`, and selecting them by tag switches them on. So CI ran rules the addon never did — **a story could show 0 violations in the panel and still fail CI.**

`target-size` (WCAG 2.2 SC 2.5.8) is the case that bit us: `enabled: false` in axe-core, activated only via the `wcag22aa` tag, which #4775 added to `test-runner.ts` without touching `preview.tsx`. That's why #4738 (Select) failed CI on `target-size` after removing its `skipCi`, on a component whose panel showed nothing — see the screenshots in that thread.

There was also a **third** independent copy of the tag list in `src/component-status/A11yRow.tsx` — the live audit behind the Accessibility row on every docs page, which also feeds the a11y gate in `meetsStableBar()`.

## Implementation details

- chore: add `src/lib/storybook-utils/a11yAxeConfig.ts` as the single source of truth (`A11Y_WCAG_TAGS`, `A11Y_RUN_ONLY`, `A11Y_CI_CONTEXT`), with the rationale and the known asymmetries documented inline.
- chore: `preview.tsx` sets `a11y.options.runOnly` so the addon runs CI's rules.
- chore: `test-runner.ts` imports the shared constant instead of its inline array.
- chore: `A11yRow.tsx` imports `A11Y_WCAG_TAGS` instead of its own copy.

| Surface | Before | After |
| --- | --- | --- |
| CI (`test-runner.ts`) | inline tag array | `A11Y_RUN_ONLY` |
| Addon (`preview.tsx`) | axe defaults (no filter) | `a11y.options.runOnly` |
| Docs panel (`A11yRow.tsx`) | own inline copy | `A11Y_WCAG_TAGS` |

**No enforcement change.** The global mode stays `test: "todo"`, so newly visible violations are non-blocking. The addon will surface more findings — that is the point, not a regression.

<details>
<summary>Note on the required <code>.ts</code> extension</summary>

`test-runner.ts` imports the constant with an explicit `.ts` extension. The test-runner's loader does not resolve extensionless sibling imports — Storybook warns *"extensionless imports detected"* and the module fails at runtime. This is the reason `wcagFromTags` was previously inlined there rather than shared with `A11yRow`. Typecheck does **not** catch it; only running `test-storybook` does. A comment records this so it is not re-broken.
</details>

## Net rule change for the addon

- **Gains 3** (were invisible, CI-enforced): `target-size` (WCAG 2.2 AA), `aria-roledescription` (2 A), `audio-caption` (2 A)
- **Loses 30**, all `best-practice` only, none WCAG-required: `heading-order`, `landmark-*`, `region`, `tabindex`, `empty-heading`, `aria-dialog-name`, `skip-link`, …

The best-practice rules were never a deliberate choice: the only mention of `best-practice` in the repo is the comment in `test-runner.ts` explaining they are *excluded*. They appeared in the panel purely because the addon was unconfigured, they gate nothing, and both intentional surfaces (CI and the docs a11y row) already excluded them. Tag filter selects **69 of 104** axe rules.

## Deliberately NOT aligned

- **Scope.** CI scopes axe to `#storybook-root`; the addon scans the whole document. Portaled content (dropdowns, tooltips, dialogs, the Select listbox) mounts **outside** `#storybook-root`, so **CI does not see it** and the addon does. Narrowing the addon would hide the only place portal violations are currently visible. Widening CI is a separate question — the document also holds Storybook's own chrome. Documented in `a11yAxeConfig.ts`.
- **Viewport.** The addon uses the canvas size, CI uses 1280×720. Not configurable away, and it is why viewport-sensitive rules like `scrollable-region-focusable` can still legitimately disagree between the two.

## Verification

- `tsc --noEmit` clean; oxlint 0 warnings; oxfmt clean.
- `vitest run src/component-status/` — **44 pass** (includes `A11yRow.test.tsx`, `ComponentStability.test.tsx`).
- `pnpm test-storybook --testPathPatterns F0BigNumber` — **7/7 pass**, confirming the shared module loads in the test-runner (this is what caught the extensionless-import bug).
- `pnpm test-storybook --testPathPatterns F0Card/__stories__/Card` — **29/29 pass** and still reports `target-size` (2 nodes) as a non-blocking warning, confirming the shared `runOnly` is genuinely applied rather than silently dropped.

## Note for review

⚠️ After this lands, **the addon panel will show more violations across the library** — `target-size` in particular. Nothing new is being enforced; those violations were always there and always counted in CI. Expect questions from anyone who reads the panel as a health check.

ℹ️ #4775's measured 2.2 baseline is a lower bound: it was taken while ~36 files still skipped axe, so `target-size` inside those files was invisible to it. Corrected figures (21 stories / 9 files, vs the 5 reported) posted as a comment on FCT-59727.

### DIFF
--- a/packages/react/.storybook/preview.tsx
+++ b/packages/react/.storybook/preview.tsx
@@ -16,6 +16,7 @@ import { dataCollectionLocalStorageHandler } from "@/lib/providers/datacollectio
 import { F0Provider } from "@/lib/providers/f0"
 import { buildTranslations, defaultTranslations } from "@/lib/providers/i18n"
 import { ThemeProvider } from "@/lib/providers/theme"
+import { A11Y_RUN_ONLY } from "@/lib/storybook-utils/a11yAxeConfig"
 
 import {
   getAllComponentStatuses,
@@ -137,6 +138,12 @@ const preview: Preview = {
           },
         ],
       },
+
+      // Same rule scope CI enforces (see src/lib/storybook-utils/a11yAxeConfig.ts).
+      // Without this the addon ran axe's defaults, which omit rules axe ships
+      // disabled — notably `target-size` (WCAG 2.2 SC 2.5.8) — so a story could
+      // show 0 violations in the panel and still fail CI.
+      options: { runOnly: A11Y_RUN_ONLY },
 
       // 'todo' - show a11y violations in the test UI only
       // 'error' - fail CI on a11y violations
--- a/packages/react/.storybook/test-runner.ts
+++ b/packages/react/.storybook/test-runner.ts
@@ -10,10 +10,13 @@ import type Reporter from "axe-playwright/dist/types"
 import { appendFileSync, readFileSync } from "fs"
 import { join } from "path"
 
+// NOTE: the `.ts` extension is required — the test-runner's loader does not
+// resolve extensionless sibling imports (Storybook warns "extensionless imports
+// detected" and the module fails to load at runtime).
 import {
   A11Y_CI_CONTEXT,
   A11Y_RUN_ONLY,
-} from "../src/lib/storybook-utils/a11yAxeConfig"
+} from "../src/lib/storybook-utils/a11yAxeConfig.ts"
 
 // Story files grandfathered to skip axe while their violations are burned
 // down (Path to AA). Maps file → number of allowed skip call-sites; counts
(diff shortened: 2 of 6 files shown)


eliseo-juan (approved):
LGTM!

PR #4866, opened 2026-07-28
fix(a11y): [FCT-59951] remove axe rule suppressions and gate future ones

fleveque:
## Description

https://factorialmakers.atlassian.net/browse/FCT-59951

#4772 gated `a11y: { skipCi: true }`, but a second, **ungated** escape hatch remained:

```tsx
parameters: { a11y: { config: { rules: [{ id: "color-contrast", enabled: false }] } } }
```

This is **strictly more invisible than `skipCi`**. A disabled rule:

- doesn't fail CI
- produces **no** job-summary line (`test-runner.ts` only emits those for `todo`/`warning`)
- produces **no** entry in `a11y-violations.jsonl`, so **no PR comment**
- shows **nothing** in the Storybook a11y panel

`skipCi` at least leaves a countable allowlist entry and fails CI when newly added. A disabled rule left no trace at all, and nothing stopped anyone adding one.

It was also **documented as a supported pattern** in this skill (`SKILL.md`, `references/storybook-test-patterns.md`), using `color-contrast` — a WCAG 1.4.3 **AA** rule — as the worked example. Very likely where the existing usages came from.

## The debt was zero

9 story files were suppressing rules, **none with a comment explaining why**. All 9 turned out to be dead weight — every affected story passes with them removed, **no component fixes needed**:

- **`color-contrast` (3)** — `F0AvatarEmoji`, `F0AvatarIcon`, `ResourceHeader`. Redundant: `BaseAvatar.tsx` already carries `data-a11y-color-contrast-ignore`, the element-scoped opt-out honoured by `preview.tsx`'s `color-contrast` selector. The story-level disables predate it.
- **`svg-img-alt` (6)** — `PieChartWidget`, `Widgets/Layout/Dashboard`, `WidgetStrip`, `kits/Charts/PieChart`, `layouts/Layout`, `TwoColumnLayout`. No longer reproduces.
- **`layouts/Layout`** was pure dead code — the file also has `skipCi` and is allowlisted, so axe never ran there anyway.

So the suppressions were only hiding the fact that there was nothing to hide.

## Implementation details

- fix: remove all 9 suppressions
- docs: drop the documented pattern from `SKILL.md` and `references/storybook-test-patterns.md`, replacing it with a **"Never disable an axe rule"** section pointing at `test: "todo"` and the element-scoped alternative
- test: add `src/lib/storybook-utils/a11yRuleSuppression.test.ts` — a static gate asserting no story disables an axe rule

- test: 12 unit tests for the detector itself, plus a widened scan (see below)

**No allowlist.** The population is zero after this PR, so the rule is simply "never" — no grandfathering, no counts to shrink. Much lighter than #4772's ratchet, because rule-disabling is detectable in source: one unit test, no `test-runner.ts` change, no runtime plumbing, no reporting.

*Enabling* or reconfiguring a rule (`enabled: true`, adding a `selector`) stays allowed — the gate only rejects disabling, so `preview.tsx`'s `color-contrast` selector config is unaffected.

## Verification

Ran the real test-runner against a fresh build, per affected story file — all green:

| File | Result |
| --- | --- |
| `F0AvatarEmoji` | 2/2 |
| `F0AvatarIcon` | 2/2 |
| `ResourceHeader` | 15/15 |
| `PieChartWidget` | 1/1 |
| `kits/Charts/PieChart` | 1/1 |
| `Widgets/Layout/Dashboard` | 3/3 |
| `WidgetStrip` | 5/5 |
| `TwoColumnLayout` | 5/5 |
| `layouts/Layout` | 7/7 |

### Testing the gate itself

A gate whose only assertion is "scan the repo, expect empty" is a false-negative risk: if the detection regex breaks, it keeps passing and silently stops gating. So the detector is unit-tested directly — **12 tests**:

Positive cases: compact single-line form, expanded multi-line form, several disabled rules in one array, single-quoted ids with tolerant spacing (`rules : [ { id : 'x' , enabled : false } ]`), and more than one `rules` array per file.

Negative cases that matter: `enabled: true` is allowed, reconfiguring an enabled rule **with a selector** is allowed (that's `preview.tsx`'s own pattern, so a regression here would break the global contrast config), and unrelated `enabled: false` config is not flagged — e.g. `ApplicationFrame`'s `aiPromotion` prop or a React context default.

Plus a guard asserting the walker sees >100 files, so a broken walker fails loudly instead of passing vacuously.

### Widened scope

The first version only scanned `*.stories.tsx`, which left a bigger hole than the one it closed: a disable in **`.storybook/preview.tsx` global parameters** silences a rule *library-wide* and would have been invisible. The scan now covers `.storybook/*.ts(x)` as well.

Verified by injection, both levels:

- a suppression added to `kits/Charts/PieChart` → fails, reporting `src/kits/Charts/PieChart/index.stories.tsx → svg-img-alt`
- a **global** disable added to `preview.tsx` → fails, reporting `.storybook/preview.tsx → color-contrast`

Both pass again once reverted.

Also: `src/lib/storybook-utils/` suite **15/15** (both a11y gates), `tsc --noEmit` clean, oxlint 0 warnings, oxfmt clean.

## Not in scope

`a11y: { disable: true }` also exists (2 usages in `Navigation/Dropdown`). It is **not** a CI loophole — `test-runner.ts` never consults it, so axe still runs in CI — and both usages already sit alongside a gated `skipCi`. It only suppresses the addon panel. Left alone deliberately.

### DIFF
--- a/packages/react/src/components/avatars/F0AvatarEmoji/__stories__/F0AvatarEmoji.stories.tsx
+++ b/packages/react/src/components/avatars/F0AvatarEmoji/__stories__/F0AvatarEmoji.stories.tsx
@@ -29,16 +29,6 @@ const meta = {
           .join(""),
       },
     },
-    a11y: {
-      config: {
-        rules: [
-          {
-            id: "color-contrast",
-            enabled: false,
-          },
-        ],
-      },
-    },
   },
 } satisfies Meta<typeof F0AvatarEmoji>
 
--- a/packages/react/src/components/avatars/F0AvatarIcon/__stories__/F0AvatarIcon.stories.tsx
+++ b/packages/react/src/components/avatars/F0AvatarIcon/__stories__/F0AvatarIcon.stories.tsx
@@ -36,16 +36,6 @@ const meta = {
           .join(""),
       },
     },
-    a11y: {
-      config: {
-        rules: [
-          {
-            id: "color-contrast",
-            enabled: false,
-          },
-        ],
-      },
-    },
   },
 } satisfies Meta<typeof F0AvatarIcon>
 
--- a/packages/react/src/experimental/Widgets/Charts/PieChartWidget/index.stories.tsx
+++ b/packages/react/src/experimental/Widgets/Charts/PieChartWidget/index.stories.tsx
@@ -10,11 +10,6 @@ const meta = {
   component: PieChartWidget,
   parameters: {
     layout: "centered",
-    a11y: {
-      config: {
-        rules: [{ id: "svg-img-alt", enabled: false }],
-      },
-    },
   },
   tags: ["autodocs", "experimental"],
   args: {
--- a/packages/react/src/experimental/Widgets/Layout/Dashboard/index.stories.tsx
+++ b/packages/react/src/experimental/Widgets/Layout/Dashboard/index.stories.tsx
@@ -50,13 +50,7 @@ const meta = {
     widgetWidth: "sm",
     children: Array.from({ length: 20 }, (_, i) => widgets[i % widgets.length]),
   },
-  parameters: {
-    a11y: {
-      config: {
-        rules: [{ id: "svg-img-alt", enabled: false }],
-      },
-    },
-  },
+  parameters: {},
 } satisfies Meta<typeof Dashboard>
 
 export default meta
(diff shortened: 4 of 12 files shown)


albertpmz (approved):
LGTM! Great job tightening these a11y checks and avoiding supressions 🚀

PR #4867, opened 2026-07-28
fix(ci): request team reviews with an org-scoped token

eliseo-juan:
## Description

The `Review policy` workflow classifies PRs correctly but has never managed to add the required teams as reviewers: every `POST /pulls/:n/requested_reviewers` answers `422 Could not resolve to a node`, because the default Actions token has no org scope and cannot resolve teams. The visible symptom is that adding the `needs-design-review` label (or opening a `feat:` PR) sets the status check to pending on an `f0-designers` approval, but no designer is ever assigned — [#4793](https://github.com/factorialco/f0/pull/4793) is the case that surfaced it, and the same failure is in the log of every run since the workflow landed.

## Implementation details

- fix: pass `REVIEW_REQUEST_TOKEN` (`RELEASE_PLEASE_GH_TOKEN`, already in the repo) to the review-request call only

  <details>
  <summary>Why only that call?</summary>

  The comment and the commit status work fine with the default token and stay authored by `github-actions`. Scoping the privileged token to the one call that needs org access keeps that blast radius small. If the secret is missing the code falls back to `GITHUB_TOKEN`, so nothing gets worse than today.
  </details>

- fix: request one team per call, so a team GitHub cannot resolve no longer drops the others from the same batch
- chore: log a hint pointing at `REVIEW_REQUEST_TOKEN` when a request fails
- docs: document the token requirement in `ownership/README.md`

## Verification

`DRY_RUN=1` against #4793 reproduces the same classification and comment as the live run. The review request itself can only be exercised once this is on `main` — the next run should log `Requested review from f0-designers` instead of the 422. If `RELEASE_PLEASE_GH_TOKEN` turns out to lack `read:org`, the follow-up is a GitHub App token via `actions/create-github-app-token`; the code needs no further change.

### DIFF
--- a/.github/workflows/review-policy.yaml
+++ b/.github/workflows/review-policy.yaml
@@ -30,5 +30,8 @@ jobs:
       - name: Evaluate review policy
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          # Only for requesting team reviews: the default Actions token has no
+          # org scope and 422s on every team ("Could not resolve to a node")
+          REVIEW_REQUEST_TOKEN: ${{ secrets.RELEASE_PLEASE_GH_TOKEN }}
           PR_NUMBER: ${{ github.event.pull_request.number }}
         run: pnpm --filter @factorialco/f0-ownership run review-policy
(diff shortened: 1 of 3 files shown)


sauldom102 (approved):

sauldom102:
Closing this as I'm tweaking this review process in a separate PR and I'm incorporating this on that PR

PR #4873, opened 2026-07-28
feat(OneDataCollection): reach select-all with no selectable rows on page

Rogermax:
## Description

A table page where `selectable` returns `undefined` for every record trapped the user: the header checkbox was enabled but inert, and the cross-page "select all" CTA was gated behind having a selection that page could never produce. On top of that, the number in that CTA came from `paginationInfo.total`, which counts every record — selectable or not — so it was simply wrong on any partially-selectable dataset. Sources can now declare the real count via `fetchSelectableTotal`, and when they don't, the labels drop the number rather than showing a false one.

Story: `Patterns/Data Collection/Miscellaneous → WithCrossPageSelectionAndNoSelectableRowsOnPage` (50 records, 10 non-selectable on page 1, `fetchSelectableTotal` resolving 40).

Screenshot 2026-07-28 at 13 59 50

Screenshot 2026-07-28 at 13 59 40

## Behavior change for existing consumers

No new required props and no API break — `fetchSelectableTotal` is optional. But one existing behavior changes without an opt-in, so it's worth being explicit about who sees it:

| Collection | Before | From now on |
| --- | --- | --- |
| Every row selectable (no conditional `selectable`) | "Select all 50 items" | **unchanged** — "Select all 50 items" |
| Some rows not selectable, no `fetchSelectableTotal` | "Select all 50 items" (wrong — only 40 are selectable) | "Select all items" |
| Some rows not selectable, `fetchSelectableTotal` provided | "Select all 50 items" (wrong) | "Select all 40 items" |

So collections without a conditional `selectable` are untouched: `paginationInfo.total` is still used whenever every loaded row is selectable. The number only disappears where it was provably wrong, and providing `fetchSelectableTotal` brings it back as a correct one. Same rule applies to the "all items selected" label in the bulk-action bar.

## Implementation details

- feat: offer the cross-page "select all" CTA when the current page has no selectable rows, so a fully non-selectable page is no longer a dead end
- feat: add `fetchSelectableTotal` to the data source — a consumer-provided count of selectable items for the current filters/search

  <details>
  <summary>Why a callback and not a field on the fetchData response?</summary>

  `selectable` is a client-side predicate, so F0 can only count the rows it has loaded. The honest cross-page total has to come from the backend either way, but a separate callback lets consumers add the count without touching a shared list query or resolver. It is resolved once per collection in `useDataSource` (so Table and the action bar read the same value), scoped to filters/search only — sortings don't change a count — with a staleness guard so a response for a previous filter set can never overwrite the current one, and a `catch` that falls back to no number. The effect deliberately does not depend on the callback's identity, so an inline arrow doesn't refetch on every render.
  </details>

- fix: never present `paginationInfo.total` as the selectable total once a non-selectable row is visible; fall back to `selectAllItemsUnknownTotal` / `allItemsSelectedUnknownTotal` instead
- fix: make `totalKnownItemsCount` use the selectable count, so `areAllKnownItemsSelected` can actually become true on partially-selectable datasets
- fix: disable the header checkbox while it has nothing to act on, and re-enable it once a cross-page selection exists so it can still clear it
- fix: give the "all items selected" label in the action bar an explicit `text-f1-foreground` — it had no color class inside the `dark` container and inherited the light-theme foreground, rendering black
- test: cover the CTA reachability, the disabled checkbox, the no-number fallback, and the consumer-provided total
- docs: add a story for a page with no selectable rows

## Known limitation

If the current page happens to be fully selectable while other pages are not, and no `fetchSelectableTotal` is provided, the CTA still shows `paginationInfo.total`. That case is undetectable client-side — the callback is the only fix.

### DIFF
--- a/packages/react/src/hooks/datasource/index.ts
+++ b/packages/react/src/hooks/datasource/index.ts
@@ -14,3 +14,4 @@ export type {
   UseSelectableReturn,
 } from "./useSelectable/typings"
 export * from "./useSelectable/useSelectable"
+export * from "./useSelectableTotal"
--- a/packages/react/src/hooks/datasource/itemNeighbors/useItemNeighbors.ts
+++ b/packages/react/src/hooks/datasource/itemNeighbors/useItemNeighbors.ts
@@ -14,6 +14,7 @@ import {
 import { RecordType } from "../types/records.typings"
 import { SortingsStateMultiple } from "../types/sortings.typings"
 import { DataError } from "../useData"
+import { stableStringify } from "../utils"
 import { resolveItemNeighbors } from "./resolveItemNeighbors"
 
 export interface UseItemNeighborsOptions<
@@ -53,22 +54,6 @@ export interface UseItemNeighborsReturn<R extends RecordType> {
   neighbors: ItemNeighborsResponse<R> | null
   isResolving: boolean
   error: DataError | null
-}
-
-/** JSON-stringify with sorted object keys, so logically equal filter states
- *  always produce the same request key. */
-const stableStringify = (value: unknown): string => {
-  if (Array.isArray(value)) {
-    return `[${value.map(stableStringify).join(",")}]`
-  }
-  if (value !== null && typeof value === "object") {
-    const record = value as Record<string, unknown>
-    return `{${Object.keys(record)
-      .sort()
-      .map((key) => `${JSON.stringify(key)}:${stableStringify(record[key])}`)
-      .join(",")}}`
-  }
-  return JSON.stringify(value) ?? "undefined"
 }
 
 const CACHE_MAX_ENTRIES = 50
--- a/packages/react/src/hooks/datasource/types/datasource.typings.ts
+++ b/packages/react/src/hooks/datasource/types/datasource.typings.ts
@@ -16,6 +16,15 @@ import { SelectedItemsState } from "./selection.typings"
 import { SortingsDefinition, SortingsState } from "./sortings.typings"
 
 /**
+ * Options handed to `fetchSelectableTotal`: enough to scope the count to what
+ * the user is currently looking at.
+ */
+export type SelectableTotalOptions<Filters extends FiltersDefinition> = {
+  filters: FiltersState<Filters>
+  search?: string
+}
+
+/**
  * Defines the structure and configuration of a data source for a collection.
  * @template R - The type of records in the collection
  * @template Filters - The available filter configurations for the collection
@@ -81,6 +90,19 @@ export type DataSourceDefinition<
 
   /** Selectable items value under the checkbox column (undefined if not selectable) */
   selectable?: (item: R) => string | number | undefined
+  /**
+   * Resolves how many items match `selectable` across the whole dataset for the
+   * current filters/search.
+   *
+   * `selectable` is a client-side predicate, so F0 can only count the rows it
+   * has loaded — `paginationInfo.total` counts every record, selectable or not.
+   * Without this callback, a collection with non-selectable rows falls back to
+   * a cross-page "select all" label with no number rather than showing a wrong
+   * one. Sortings are not passed: they don't change a count.
+   */
+  fetchSelectableTotal?: (
+    options: SelectableTotalOptions<Filters>
+  ) => Promise<number>
   /** Default selected items */
   defaultSelectedItems?: SelectedItemsState<R>
   /**
@@ -143,6 +165,13 @@ export type DataSource<
   Sortings extends SortingsDefinition,
   Grouping extends GroupingDefinition<R>,
 > = DataSourceDefinition<R, Filters, Sortings, Grouping> & {
+  /**
+   * Total selectable items resolved from `fetchSelectableTotal` for the current
+   * filters/search. Undefined while in flight, on failure, or when the source
+   * doesn't provide the callback.
+   */
+  selectableTotal?: number
+
   /** Current state of applied filters */
   currentFilters: FiltersState<Filters>
   /** Function to update the current filters state */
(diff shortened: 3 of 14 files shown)


Rogermax:
Thanks — all three were real. Fixed in f496bd090:

**1. Registry-based length comparison** — confirmed. `currentPageSelectableIds` falls back to `selectionRegistry.ids`, which can hold lazily-loaded nested children, so the comparison mixed two different populations: a nested table with one non-selectable top-level row and several selectable children would compare `3 > 7` and conclude every row was selectable. Detection now runs over the fetched records directly, via a shared `hasNonSelectableRecords(records, selectable)` used by both the table and the action bar so the rule lives in one place.

**2. Flag reset on page navigation** — confirmed, and worse than the limitation already documented in the description: there we simply don't know, here we knew and threw the knowledge away. The flag is now sticky within a filter/search combination (`useHasNonSelectableRows`) and only resets when the query changes, since that's a different dataset. Covered by two unit tests.

Implemented with a ref rather than state on purpose: the value is only read during a render that is already happening because the page or the query changed, and an effect-based reset would leave one commit where a stale `true` from the previous query was still in play — a one-frame wrong number is exactly what this PR is about.

**3. Stale total when the callback disappears** — confirmed. `isEnabled` flipping true → false re-ran the effect, which returned early without clearing, so the last resolved count survived the callback backing it. It now clears on the early return.

Rogermax:
Both suppressed comments were right, and the first one was my own regression — fixed in 5f0cbe897.

`useHasNonSelectableRows` reset on a new `queryKey` and then immediately took that same render's page flag, which is still derived from the *previous* query's records because the new fetch hasn't resolved. That re-armed the flag and stranded the new dataset on the unknown-total path permanently, since nothing resets it again until the query changes once more. Evidence from the render where the query changes is now ignored; the next render, once the new records land, provides the real answer.

The failure window this leaves is the opposite and much narrower: a dataset whose records arrive in the *same* commit as the filter change would have its evidence ignored for one render. Records always land through a state update after the fetch promise resolves, so that's a later commit in practice.

Also added the suggested test — `page: true` across a `queryKey` change must read `false`, and `true` again on the following render.

Rogermax:
All three are fixed in 74721c221 — the key-order one twice over, since it turned out to be a dedup too.

**Sync throw** — real. The signature says `Promise<number>`, so an `async` callback can only reject, but a plain function that throws while building its promise would throw inside the effect and take the whole collection down. Now resolved through `Promise.resolve().then(() => fetchTotal(...))`, so sync throws land in the existing `.catch` and fall back to no number. Covered by a test that passes a synchronously-throwing callback and asserts the unknown-total label renders.

**Unstable query key** — real, and worth noting the trigger is reachable: `setCurrentFilters` compares with `JSON.stringify` too, so a filter state rebuilt in a different key order is already treated as a change by the datasource. My key would then reset the sticky flag and hand the dataset back to per-page detection — the exact bug from your earlier comment, via a different door.

Rather than add a second stable-stringify, I moved the recursive one out of `useItemNeighbors.ts` into `hooks/datasource/utils.ts` and imported it in both places, so there's one implementation for three callers instead of a private copy plus two new ones. The key is now built inside `useHasNonSelectableRows` from `{ filters, search }` rather than by each caller, so the two callsites can't drift. Test asserts `{ a: 1, b: 2 }` → `{ b: 2, a: 1 }` keeps the flag.

`753 passed`, tsc/oxlint/oxfmt clean.

Rogermax:
This one doesn't hold as described, but it pointed next door at something that does — fixed in c1a646ef4.

**Why the described path doesn't happen:** `selectedNumber` is not 0 in that flow. `selectedItemsCount` short-circuits on `allSelectedCheck && selectAllTotal !== null` and returns `selectAllTotal - uncheckedCount` (`useSelectable.ts:228-232`), and `handleSelectAllItems(true)` sets `selectAllTotal = totalKnownItemsCount` before any of that renders. So triggering cross-page selection from a page with no selectable rows yields `selectedNumber = 40`, not 0, and `leftContent` renders normally.

Verified in the browser on the new story (page 1 entirely non-selectable, cross-page select-all triggered from the CTA) — the buttons present afterwards are:

```
["Unselect", "Delete all items", "Delete all items"]
```

Label, Unselect and the bulk action are all there.

**What is real:** the count itself can be 0, and then everything you describe does happen. A `fetchSelectableTotal` that resolves `0` — a legitimate answer meaning "nothing in this dataset is selectable" — produced a CTA reading "Select all 0 items", and clicking it set `selectAllTotal = 0`, so `selectedNumber` was 0, `leftContent` returned `null`, and there was no bar and no Unselect. The CTA is now hidden for a known total of 0, since there is no selection to offer. Test added.

`754 passed`.

Rogermax:
Real, and it was an inconsistency I introduced in the previous commit — fixed in 6e88f9a4c.

`useHasNonSelectableRows` treated a reordered filter state as the same dataset while `useSelectableTotal`, two functions above it in the same file, treated it as a different one. Two answers to "is this the same query?" in the same feature is exactly the kind of split that produces the next bug, so both now key on the same `stableStringify` identity.

Worth noting the trigger isn't hypothetical: `setCurrentFilters` bails out via `JSON.stringify` comparison, so a reordered state doesn't get filtered out upstream — it reaches the hooks as a change.

The effect reads `filters`/`search` through a ref so it always sees current values without depending on their identity. Behaviour-wise this removes a redundant count request and, more visibly, a flicker: the old effect ran `setTotal(undefined)` on every identity change, so the CTA dropped from "Select all 40 items" to "Select all items" and back for a filter state that hadn't logically changed.

Test asserts `{ a: 1, b: 2 }` → `{ b: 2, a: 1 }` spends no second request, and that a real value change still does. `755 passed`.

PR #4885, opened 2026-07-29
feat(OneDataCollection): optional group-header selection in Kanban

victorquiros98:
## Summary

Adds `selectableGroups?: boolean` (default `true`) to the Kanban visualization options. When `false`, a **grouped + selectable** Kanban keeps **per-card selection** (and bulk actions) but does **not** render the per-group-header checkbox.

Motivation: onboarding tracking groups the Kanban by policy version. "Select a whole version" isn't a meaningful action there, so product wants the group-header checkbox hidden while keeping onboardee (card) selection + bulk-remove. Today the group-header checkbox is gated only on `source.selectable`, same as Card/List, with no way to opt out.

## Changes
- `KanbanVisualizationOptions.selectableGroups?: boolean` (default `true`)
- Grouped Kanban gates the `GroupHeader` checkbox on `selectableGroups && source.selectable`
- The header's clickable styling (`cursor-pointer`/hover) now follows the same `collapsible || groupSelectable` condition as `GroupHeader`'s own `isInteractive`, so the header no longer looks interactive when it isn't
- Backwards compatible: default preserves current behaviour (parity with Card/List)
- Unit test: with `selectableGroups={false}` the group-header "Select all" checkbox is absent while exactly the per-card checkboxes still render
- Story `KanbanWithGroupingHiddenGroupSelect` demonstrates the opt-out (grouped + selectable, group checkbox hidden, cards still selectable) with a `play` assertion

## Testing
- `pnpm lint` 0/0 · `pnpm build:types` clean · Kanban unit suite 9/9
- New story is type-checked; its `play` mirrors the unit contract (no group "Select all", card checkboxes present)

## For Foundations
**Scope is intentionally Kanban-only.** The default is `true`, so out of the box all three visualizations behave identically (full parity) — the divergence only appears when a consumer deliberately opts out. The opt-out has a Kanban-specific rationale: a collapsed group unmounts its cards, so the group-header checkbox is otherwise the only way to bulk-select a collapsed group; onboarding (grouped by policy version) doesn't want "select a whole version" at all.

Scoped it to the **Kanban viz option** (smallest, additive, lives next to `getLanesForGroup`) rather than Card/List, which have no consumer need for it today. If you'd prefer a **grouping-level** flag (`GroupingDefinition.selectableGroups`) or a shared visualization-overrides option honoured consistently across Card/List/Kanban, happy to lift it there — flag your preference.

### DIFF
--- a/packages/react/src/patterns/OneDataCollection/__stories__/grouping.mdx
+++ b/packages/react/src/patterns/OneDataCollection/__stories__/grouping.mdx
@@ -202,3 +202,10 @@ options to hide the group-level checkbox while keeping per-card selection —
 useful when "select a whole group" isn't a meaningful action for the consumer.
 
 <Canvas of={KanbanStories.KanbanWithGrouping} />
+
+### Kanban grouping with the group checkbox hidden
+
+`selectableGroups={false}` on a grouped, selectable Kanban keeps per-card
+selection (and bulk actions) while hiding the per-group "Select all" checkbox.
+
+<Canvas of={KanbanStories.KanbanWithGroupingHiddenGroupSelect} />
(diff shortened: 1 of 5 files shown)


sauldom102 (approved):
Let's keep this behavior only in Kanban for now, thank you!

PR #4920, opened 2026-07-30
feat(RichText): support heading levels 4 to 6 and source indentation

manfredmmm:
## Description

The editor configured Tiptap with `heading: { levels: [1, 2, 3] }`, so a document holding an h4 to h6 was drawn as an h1: `level` is a plain attribute with no validation, so the stored value survives, but `renderHTML` clamps anything outside the configured list to the first entry. The visible result is a subsection appearing larger than the section that contains it. This widens the list to 1 through 6 and adds the CSS those levels never had, so they are legible rather than merely accepted. Nested list markers come along because they are the same class of gap: a sub-list repeated its parent's marker, leaving indentation as the only cue for depth.

## Type of change

f0-4920-heading-levels

The `components-rich-text-notestexteditor--heading-levels` story, showing every level with body text between for comparison, plus both nested list types. Values read off `getComputedStyle` in that story:

| Level | Notes editor | | Nesting | Bullet | Ordered |
|---|---|---|---|---|---|
| h1 | 29px / 600 | | depth 1 | `disc` | `decimal` |
| h2 | 24px / 600 | | depth 2 | `circle` | `lower-alpha` |
| h3 | 19px / 600 | | depth 3 | `square` | `lower-roman` |
| h4 | 17px / 600 | | | | |
| h5 | 15px / 600 | | | | |
| h6 | 13px / 600 | | | | |

Updated after design review: every level now differs by **size**, not weight. In the shared stylesheet h3 moves onto `text-xl` (18px) — the step the token ladder skipped between h2 and h3 — which frees `lg`/`base`/`sm` for h4/h5/h6; h1 and h2 are untouched. Previously h5 and h6 were both 12px there, and h5 was semibold against h4 medium, so a deeper level rendered heavier than its parent. In the notes editor h5/h6 go to 15/13, because h5 at 16px was exactly body size and read as bold body text.

| Level | Notes | Shared |
|---|---|---|
| h1 | 26px | 26px |
| h2 | 22px | 22px |
| h3 | 18px | 18px |
| h4 | 16px | 16px |
| h5 | 14px | 14px |
| h6 | 12px | 12px |

Both editors are now on the Foundations typography tokens (`3xl`/`2xl`/`xl`/`lg`/`base`/`sm`), so a heading renders identically in either. The notes editor previously carried a bespoke px ladder (29/24/19), which means **existing content re-typesets**: h1, h2 and h3 shrink by 1-3px in policies, meetings and trainings. Weights are unchanged (notes stays semibold throughout, shared keeps medium below h1).

On the token scale the notes editor's 16px body sits between h4 (16px) and h5 (14px): h4 matches body size and is distinguished by weight, while h5 and h6 fall below it. Levels stay distinct and the ladder stays monotonic — 26 / 22 / 18 / 16 / 14 / 12.

Full unification of the two scales onto the tokens is deliberately **not** in this PR: it would shrink the notes editor h1/h2/h3 across all five render sites (pages, meetings, trainings), which is a cross-team visual change that deserves its own review rather than riding along in a heading-levels fix.

## Implementation details

**Indent (added in review round): render-only `indent` attribute on paragraphs and headings.** A new `IndentExtension` (global attribute, no commands, no toolbar) stores `indent: 1..n` and renders 24px per level, capped at 8 styled levels. Same sanitizer constraint as elsewhere in RichText: read-only content passes through DOMPurify which drops `style`, so a level renders as an `f0-indent-N` class (which the CSS styles) plus an inline `padding-left` carrying the true depth for renderers outside F0. Registered in both editors; without the attribute declared, a stored indent is destroyed on the first open+save — the same failure mode fontSize had (#4922). 10 tests; the HeadingLevels story shows three indented paragraphs. **Backspace at the start of an indented block, and Shift-Tab anywhere in one, remove a level** — an import infers depth from page geometry and can be wrong, and previously the only way to clear an indent was converting the block to a heading and back (which happens to reset custom attributes). Both keys fall through when there is no indent, so joining a paragraph backwards and reverse tabbing out of the editor are untouched, and both stand aside inside list items where those keys belong to list nesting; `setIndent`/`unsetIndent`/`outdent` commands are there for a future toolbar. There is deliberately no way to ADD depth by keyboard — that would compete with list nesting for Tab. 19 tests, keymap precedence checked against the task-list bindings. The backend emitter is [factorial#108464](https://github.com/factorialco/factorial/pull/108464); it stays dormant behind its flag until this PR's release is bumped into the monolith.

- feat: widen `StarterKitExtension` heading levels to `[1, 2, 3, 4, 5, 6]`

  <details>
  <summary>This affects both editors</summary>

  `StarterKitExtension` is a module-level singleton shared by `F0NotesTextEditor`
  and `F0RichTextEditor`, so the wider range applies to the compact composer too.
  That looks right to me, since an h4 pasted into a comment was equally
  mis-drawn, but it is a deliberate choice rather than a side effect. Scoping it
  to one editor would mean splitting that export into two configurations.
  </details>

- feat: style h4/h5/h6 in `F0NotesTextEditor/index.css`, continuing the existing 29/24/19px ladder with 17/16/15px
- fix: give h4 a real step in `RichText/index.css` — it was a byte-identical copy of h3 — and add h5/h6

  <details>
  <summary>The sizes are a proposal, not a design decision</summary>

  In the notes editor the ladder is arbitrary-px and easy to continue. The shared
  stylesheet is token-based and already compressed: h1 `text-3xl` (26), h2
  `text-2xl` (22), h3 `text-lg` (16), which leaves little room below h3. h4 is
  `text-base` (14) and h5/h6 share `text-sm` (12) separated by weight, because
  the next token down is `text-xs` (10) and that reads too small for a heading.
  If Foundations would rather see a different ladder there, or would rather fix
  the h2 to h3 jump that skips `text-xl` (18), I am glad to change it.
  </details>

- feat: per-depth list markers in both stylesheets — circle then square for bullets, lower-alpha then lower-roman for ordered lists
- fix: select those markers by `.f1-bullet-list` / `.f1-ordered-list` rather than `ul ul`

  <details>
  <summary>Why the first attempt was dead code</summary>

  Both stylesheets already set a marker through those classes, which Tiptap puts
  on every list including nested ones. A class selector at specificity 0,2,0
  outranks `ul ul` at 0,1,2, so the element-based rules never applied: a
  second-level list still computed `disc`, and even the same rule injected
  manually into the page lost until given `!important`. Selecting by the same
  classes restores the intended cascade.
  </details>

- test: `internal/Extensions/__tests__/configured.test.ts` — 24 cases covering all six levels: rendered tag, JSON round-trip, HTML parse, and the toggle command

  <details>
  <summary>Confirmed the tests can actually fail</summary>

  Reverting `levels` to `[1, 2, 3]` fails exactly the three rendering cases for
  levels 4 to 6, while the JSON round-trip cases keep passing. That asymmetry is
  the bug this PR fixes, so the test discriminates rather than merely passing.
  </details>

- docs: `HeadingLevels` story showing every level against body text, plus both nested list types at three levels

## Notes for the reviewer

Full RichText suite is green (10 files, 64 tests), `format` and `lint` clean. `tsc` reports 36 errors on this branch and the same 36 on a clean `main` in my environment (unbuilt `@factorialco/f0-core`), none of them in RichText.

This comes from a PDF-to-policy import in the monolith ([factorial#107901](https://github.com/factorialco/factorial/pull/107901), merged), which now produces correct h1 to h6 structure the editor could not display, and is gaining source indentation (this PR makes it displayable and edit-safe ahead of the backend emitting it). The import is also gaining source indentation, which this PR makes displayable and edit-safe ahead of the backend emitting it. #4922 is the other half, adding a `fontSize` attribute to the `textStyle` mark. Toolbar and slash-menu entries for the new levels are deliberately excluded: they need new required i18n keys and Heading4-6 icons that do not exist, and they are about authoring rather than displaying what is already stored.


arnaub (approved):
LGTM

desiree-np:
Nice fix — the root cause (levels stored but clamped on render, so h4–h6 drew as h1) is well diagnosed, and I like that the tests fail when the range is narrowed again.

One thing on the notes editor: the three new levels land at 17/16/15px, so they're only 1px apart and hard to tell apart as distinct levels — and h5 at 16px is exactly the body size (`p` is 16px), so it reads as bold body text rather than a heading. Since this stylesheet is px-based anyway, could we widen the steps a bit so each level is clearly distinct and none collides with the body? Something like 17/15/13 works — no new tokens needed.

Approving the functional fix — the spacing tweak doesn't need to block it.

manfredmmm:
> Nice fix — the root cause (levels stored but clamped on render, so h4–h6 drew as h1) is well diagnosed, and I like that the tests fail when the range is narrowed again.
> 
> One thing on the notes editor: the three new levels land at 17/16/15px, so they're only 1px apart and hard to tell apart as distinct levels — and h5 at 16px is exactly the body size (`p` is 16px), so it reads as bold body text rather than a heading. Since this stylesheet is px-based anyway, could we widen the steps a bit so each level is clearly distinct and none collides with the body? Something like 17/15/13 works — no new tokens needed.
> 
> Approving the functional fix — the spacing tweak doesn't need to block it.

If you think it's better to have this scale factor 17/15/13 for h4/h5/h6, I can do it, no problem. TBH, I think we should use the same tokens [here](https://ds.factorial.dev/?path=/docs/foundations-typography--documentation), as we are doing for the font sizes. But IDK why we didn't do it from the beginning instead of creating another scale factor.

Screenshot 2026-08-03 at 09 31 04

desiree-np:
Agreed tokens are the cleaner answer, and honestly the better one, it would also unify the two editors, which today run on different scales. The cost is that it re-typesets the existing h1/h2/h3 (bespoke px from an earlier decision), which is a bit more than this bug strictly needs, but if you're up for it, here's how today's values would map onto the Foundations type tokens:

| Level | Today (notes / shared) | → Unified token |
|---|---|---|
| h1 | 29px / `text-3xl` 26px | `text-3xl` · 26px |
| h2 | 24px / `text-2xl` 22px | `text-2xl` · 22px |
| h3 | 19px / `text-lg` 16px | `text-xl` · 18px |
| h4 | 17px / `text-base` 14px | `text-lg` · 16px |
| h5 | 16px / `text-sm` 12px | `text-base` · 14px |
| h6 | 15px / `text-sm` 12px | `text-sm` · 12px |

What actually changes: in the notes editor everything moves from px to tokens (h1 29→26, h3 19→18…); in the shared stylesheet it fills the `text-xl` step the current scale skips (h3 16→18) and h5/h6 stop being the same size. Since every level now has a distinct *size*, h5/h6 no longer lean on weight to differ, which also removes the h4-heavier-than-h5 inversion. Weights can stay semibold.

If you'd rather keep this PR minimal and not touch the pre-existing levels, the px fallback is fine too, just space h4/h5/h6 to 17/15/13 so they read as distinct and h5 doesn't land on the 16px body.

manfredmmm:
> Agreed tokens are the cleaner answer, and honestly the better one, it would also unify the two editors, which today run on different scales. The cost is that it re-typesets the existing h1/h2/h3 (bespoke px from an earlier decision), which is a bit more than this bug strictly needs, but if you're up for it, here's how today's values would map onto the Foundations type tokens:
> 
> Level	Today (notes / shared)	→ Unified token
> h1	29px / `text-3xl` 26px	`text-3xl` · 26px
> h2	24px / `text-2xl` 22px	`text-2xl` · 22px
> h3	19px / `text-lg` 16px	`text-xl` · 18px
> h4	17px / `text-base` 14px	`text-lg` · 16px
> h5	16px / `text-sm` 12px	`text-base` · 14px
> h6	15px / `text-sm` 12px	`text-sm` · 12px
> What actually changes: in the notes editor everything moves from px to tokens (h1 29→26, h3 19→18…); in the shared stylesheet it fills the `text-xl` step the current scale skips (h3 16→18) and h5/h6 stop being the same size. Since every level now has a distinct _size_, h5/h6 no longer lean on weight to differ, which also removes the h4-heavier-than-h5 inversion. Weights can stay semibold.
> 
> If you'd rather keep this PR minimal and not touch the pre-existing levels, the px fallback is fine too, just space h4/h5/h6 to 17/15/13 so they read as distinct and h5 doesn't land on the 16px body.

done! 🚀
(I only updated the h4, h5, and h6 ones)

oskar-hernandez (approved):

desiree-np (approved):
The token pass looks great — h1–h6 now read as a clean, distinct ladder, and unifying the shared scale (plus filling the `text-xl` step at h3) is exactly right. Two minor notes, neither blocking:

1. **The body-size collision moved rather than disappeared.** In the notes editor, h4 is now `text-lg` (16px), which is the body size (`p` is 16px) — so there's still one heading level that matches body text, separated only by weight. It shifted from h5 to h4. This is pretty much inherent to fitting six levels around a 16px body, and h4 in semibold reads clearly enough as a heading, so I'm fine leaving it — just flagging it's not fully gone.

2. **The notes stylesheet now mixes px (h1–h3) and tokens (h4–h6).** That's the natural consequence of not re-typesetting the pre-existing levels, which I think is the right call for this PR — noting it so it's a conscious state rather than a surprise.

Approving from design — nice work.

PR #4921, opened 2026-07-30
feat(OneDataCollection): allow capping primary actions menu width

rocio-fdez-factorial:
## Description

When a `primaryActions` item defines a long `description`, the primary actions dropdown menu had no width limit and would stretch to fit the text on a single line instead of wrapping. This adds an opt-in `primaryActionsMenuClassName` prop on `OneDataCollection` so consumers can cap the menu's width for their own case, without changing the default sizing behavior of any existing dropdown.

Before
Screenshot 2026-07-30 at 13 40 39

After
Screenshot 2026-07-30 at 14 15 24

## Implementation details

- feat: add `contentClassName` prop to `DropdownInternal`, merged onto the Radix `DropdownMenuContent`
- feat: thread `contentClassName` through `F0ButtonDropdown`'s dropdown mode
- feat: add public `primaryActionsMenuClassName` prop to `OneDataCollectionProps`, passed down to `CollectionActions` and from there to the primary actions `F0ButtonDropdown`

  <details>
  <summary>Why opt-in and not a default max-width?</summary>

  The `DropdownMenuContent`/`DropdownItem` pair is shared by every dropdown built on `DropdownInternal` (item actions menus, the "Actions" overflow menu, any `F0ButtonDropdown`). Adding a default `max-width` would have changed sizing for all of them. Keeping the prop unset by default means every existing consumer keeps its current behavior, and only the specific case that needs a capped width (a primary action with a long description) opts in explicitly.
  </details>

### DIFF
--- a/packages/react/src/components/F0ButtonDropdown/F0ButtonDropdown.tsx
+++ b/packages/react/src/components/F0ButtonDropdown/F0ButtonDropdown.tsx
@@ -221,6 +221,7 @@ const DropdownMode = ({
   disabled,
   loading,
   tooltip,
+  contentClassName,
 }: {
   onClick: (value: string, item: ButtonDropdownItem<string>) => void
   trigger?: string
@@ -233,6 +234,7 @@ const DropdownMode = ({
   disabled?: boolean
   loading?: boolean
   tooltip?: string
+  contentClassName?: string
 }) => {
   const [isOpen, setIsOpen] = useState(false)
 
@@ -280,6 +282,7 @@ const DropdownMode = ({
         if (disabled) return
         setIsOpen(open)
       }}
+      contentClassName={contentClassName}
     >
       <Action
         variant={variant}
@@ -314,6 +317,9 @@ const _F0ButtonDropdown = (props: F0ButtonDropdownProps) => {
         disabled={props.disabled}
         loading={props.loading}
         tooltip={props.tooltip}
+        contentClassName={
+          "contentClassName" in props ? props.contentClassName : undefined
+        }
       />
     )
   }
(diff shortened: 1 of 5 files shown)


sauldom102 (changes requested):
I think it's a good and valid change, let's talk with @factorialco/f0-designers to align what is the exact width we should have here. I wouldn't expose `contentClassname`, it should be an internal thing instead of being open to modify outside of F0

warcos-fact:
1. From a Design point of view the proposal matches the F0 Dropdown documentation and design. I would recommend double checking that part.
2. The issue I see is with the application of the One icon, when applied in the dropdown it looks misaligned, and since it's too big it gets too close to the text. That being said, the icon issues are not directly related to the max-width code updates. I would recommend you address that in a different PR, adjusting at the very list the vertical alignment (should be top aligned in this case), and considering using a smaller version of the One icon, if that's available.

rocio-fdez-factorial:
Hi @sauldom102!

Regarding the concern about exposing `contentClassName`: design confirmed (see comment above from @warcos-fact, Aug 23) that capping the width via `max-width` matches the F0 Dropdown documentation/design — but they didn't give (nor were asked for) a single value to apply platform-wide. Forcing one default width for every dropdown with a `description` (item action menus, the overflow "..." menu, any `F0ButtonDropdown`) would be a much bigger design decision than what was reviewed here.

That's why we kept it opt-in: each consumer decides its own width when it needs one, and every existing dropdown keeps its current behavior unchanged by default — same reasoning as in the PR description ("Why opt-in and not a default max-width?").

With the answer from design, I think we can move on with this solution, WDYT?

rocio-fdez-factorial:
Superseded by #5648 — reworked the free-form `className` here into a closed `primaryActionsMenuVariant: "expanded" | "reduced"` variant per feedback from Rafa, so the design system keeps menu widths unified instead of each consumer picking its own value. Keeping this PR open for the discussion history; new work continues on #5648.

PR #4922, opened 2026-07-30
feat(RichText): carry a font size on textStyle

manfredmmm:
## Description

Documents can hold text at a size other than the body size, but the `textStyle` mark had nowhere to put it: it declares only `color`, so an unknown `fontSize` attribute was dropped when a document was parsed and the editor then saved the mark without it. Opening a document was enough to destroy the sizes it arrived with. This adds a `fontSize` attribute to that mark, so a stored size renders and survives being edited. There is no toolbar entry yet; the commands exist for a control to be added later.

## Type of change

f0-4922-font-sizes

Every step of the scale in the `components-rich-text-notestexteditor--font-sizes` story. 16px is the body size, shown for comparison; text at 16 normally carries no mark at all.

| Class | Computed size | Line height | Survives `style` being stripped |
|---|---|---|---|
| `f0-fs-12` | 12px | 16px | yes |
| `f0-fs-14` | 14px | 20px | yes |
| `f0-fs-16` | 16px | 24px | yes |
| `f0-fs-18` | 18px | 28px | yes |
| `f0-fs-20` | 20px | 28px | yes |
| `f0-fs-24` | 24px | 32px | yes |
| `f0-fs-29` | 29px | 39px | yes |

Line heights follow the core typography tokens where a token has the same size (12/14/16/18 are `sm`/`base`/`lg`/`xl`); 20/24/29 sit in gaps the token scale does not cover. The last column is the point of the class-based rendering: I removed the `style` attribute from every span in the page, as the read-only sanitizer does, and every size still computed correctly.

## Implementation details

- feat: `FontSizeExtension` adds a `fontSize` global attribute to the `textStyle` mark, following `BlockIdExtension`'s shape
- feat: register it in both editors' extension lists, and export `FONT_SIZE_SCALE` for a future toolbar control
- feat: `.f0-fs-*` rules in both stylesheets, with line heights matched to each step

  <details>
  <summary>Why a class and an inline style, rather than one or the other</summary>

  `F0RichTextDisplay` sanitizes with DOMPurify allowing `href`, `target`, `rel`
  and `class` — `style` is stripped, which is why the existing `color` attribute
  does not survive there either. So the size has to ride on a class.

  A class alone is not enough, though: `TextStyle.parseHTML` returns `false` for
  any `<span>` without a `style` attribute, so a class-only span is never
  recognised as a mark and the attribute's own `parseHTML` never runs. I found
  this from a failing test, not by reading the source first.

  Rendering both satisfies both constraints, and the class is treated as
  authoritative on parse since it holds the snapped value.
  </details>

- feat: `setFontSize` / `unsetFontSize` commands via module augmentation, as `EnhanceHighlight` does
- test: 24 cases — the load-and-edit survival that this PR exists to fix, class and inline rendering per step, parse precedence, and the commands
- docs: `FontSizes` story showing every step against body text

## Notes for the reviewer

**The scale is a proposal.** `FONT_SIZE_SCALE` is `[12, 14, 16, 18, 20, 24, 29]`. 16 is the body size and is never marked in practice; 24 and 29 line up with the notes editor's h2 and h1 so a size lifted from a document shares the rhythm of its headings. The core typography tokens have no 20, 24 or 29, so these use arbitrary values like the surrounding notes styles already do. If Foundations wants different steps, the constant and seven CSS rules are the only things to change — the importer feeding this snaps to whatever the scale says.

**One known limitation, recorded in a test rather than worked around:** HTML stripped of its styles parses back without the size, because of the `TextStyle.parseHTML` rule above. Supporting it would mean changing TextStyle's own parse rule for both editors, and nothing in the product feeds sanitized HTML back into an editor today.

**Verification.** 24 new tests pass; removing the extension from the test editor fails 22 of them, so they discriminate rather than merely pass. Full RichText suite green (10 files, 64 tests — 40 of those pre-existing). `format` and `lint` clean. `tsc` reports 36 errors on this branch and the same 36 on a clean `main` in my environment (unbuilt `@factorialco/f0-core`), none in RichText.

This is the second half of the editor work for a PDF-to-policy import in the monolith ([factorial#107901](https://github.com/factorialco/factorial/pull/107901), merged), which already emits these marks; the first half is #4920 (heading levels 4 to 6). The two are independent and can merge in either order.

Worth flagging for scheduling rather than for review: until this ships, an editor save rewrites the mark without its size, so any policy imported beforehand loses its sizes the first time someone opens and saves it, and no backfill recovers them once the source PDF is gone.


arnaub (approved):
Thanks ❤️

desiree-np:
One consistency note: in the shared `RichText/index.css`, the surrounding heading rules are token-based (`text-3xl`, `text-2xl`, `text-lg`…), and you used tokens there yourself in #4920. The new `.f0-fs-*` rules are written as raw px instead. Where a token matches the size (12/14/16/18 → `text-sm`/`text-base`/`text-lg`/`text-xl`), could they use the token class to stay consistent with the file? The remaining steps (20/24/29) have no token, so leave those as-is — and the notes-editor stylesheet is px throughout, so its `.f0-fs-*` rules are fine as they are.

manfredmmm:
> One consistency note: in the shared `RichText/index.css`, the surrounding heading rules are token-based (`text-3xl`, `text-2xl`, `text-lg`…), and you used tokens there yourself in #4920. The new `.f0-fs-*` rules are written as raw px instead. Where a token matches the size (12/14/16/18 → `text-sm`/`text-base`/`text-lg`/`text-xl`), could they use the token class to stay consistent with the file? The remaining steps (20/24/29) have no token, so leave those as-is — and the notes-editor stylesheet is px throughout, so its `.f0-fs-*` rules are fine as they are.

done

desiree-np (approved):

PR #4923, opened 2026-07-30
chore(a11y): enforce axe on 6 components meeting the DoD

albertpmz:
## Description

No component in F0 actually met the Definition of Done. All 41 components tagged `stable` failed the mechanical bar computed in `src/component-status/component-status.ts`, and every one of them failed on the same requirement — axe was never enforced, because `.storybook/preview.tsx` sets a global `a11y: { test: "todo" }` so axe runs everywhere and blocks nowhere. Six components were failing on that requirement *alone*, and a full axe run confirmed they were already clean, so this opts them into blocking enforcement and adds a CI ratchet so the win can't silently regress. Components truly meeting the DoD goes **0 → 6**.

## Type of change

- [x] Bug fix
- [x] Other: a11y enforcement + CI gate for the stable Definition of Done

Not a promotion — these six were already tagged `stable`. This closes the **mechanical** DoD gap that made the badge read `Experimental` despite the tag (see `packages/react/docs/definition-of-done.mdx`). No `#f0-support` thread applies. The two non-static promotion gates (≥3 product teams, no breaking changes 60d) are untouched and remain manual.

- Linked #f0-support thread: n/a (not a promotion)
- Phase exit criteria met: mechanical DoD now met for all six — verified 63/63 stories pass with `test: "error"`, zero axe violations

## Implementation details

- fix: opt six already-clean components into blocking axe (`a11y: { test: "error" }` at `meta.parameters`) — F0Alert, F0Avatar, F0Button, F0ButtonDropdown, F0Link, F0AiChatHeader

  <details>
  <summary>Why the flip alone was enough</summary>

  A full `test-storybook` run over all six *before* any edit returned zero axe violations across 63 stories. The gap was policy, not defects: `a11yTierOf` in `scripts/component-status-build.mjs` only reports `enforced` when a file contains `test: "error"` and no `todo`/`warning`/`skipCi`. Nothing opted in, so all six sat at `todo` and failed the DoD bar.
  </details>

- fix(F0AiChatHeader): stop the `WithHeaderAction` play test from flaking

  <details>
  <summary>Two independent races</summary>

  The module-scoped `openRoutines = fn()` mock is shared with `Compact`/`Snapshot` but was never reset, while the test asserts `toHaveBeenCalledOnce()` — a CI retry or repeat visit inflated the count. Now reset via `meta.beforeEach`.

  Separately, the tooltip lookup raced: `TooltipInternal` opens on a 700ms timer (`experimental/Overlays/Tooltip/index.tsx:45`) against testing-library's 1000ms default `findBy` timeout, leaving ~300ms of headroom. Now waits explicitly.

  Reproduced at `--maxWorkers=3` (1 failure) and passing at `--maxWorkers=1` before the fix; both configs green after.
  </details>

- fix(F0Avatar): remove invalid `aria-labelledby="John Doe"` from the PersonAvatar story — `aria-labelledby` takes an ID reference, not a literal string; the adjacent `aria-label` is the correct mechanism

- feat: add `pnpm check:stable-dod`, a CI ratchet that keeps the bar from rotting

  <details>
  <summary>How the ratchet works</summary>

  Mirrors the existing `.storybook/a11y-skip-allowlist.json` idiom. `.scripts/stable-dod-debt.json` lists the stable-tagged components still below the bar (starts at 35; the six graduated here are already removed). The check fails **both** ways: a stable-tagged component that drops below the bar fails, and so does a debt-listed component that starts passing — forcing its entry to be deleted so each win becomes permanent.

  The pass/fail verdict comes from the repo's own `meetsStableBar`, so the gate cannot drift from the policy it enforces. The per-requirement output is diagnostics only.
  </details>

- test: 14 unit tests for the ratchet logic (`.scripts/__tests__/check-stable-dod.test.ts`, 3ms)
- ci: run the gate in `quality.yaml` as `[⚛️ REACT] Stable Definition of Done` — a pure filesystem scan, no build or Storybook needed
- docs: document the mechanical gate and the ratchet in `definition-of-done.mdx`

## Verification

| Check | Result |
| --- | --- |
| `test-storybook` (6 files, `test: "error"`) | 63/63 pass, zero axe violations |
| Flake regression (`--maxWorkers=3`) | 63/63 pass (previously 1 failure) |
| `pnpm tsc` | pass |
| `oxlint` on changed files | 0 warnings, 0 errors |
| `pnpm format:check` | pass (3285 files) |
| `vitest --project=unit` (status/a11y invariants + all 6 components) | 277 pass, 1 skipped |
| `pnpm check:stable-dod` | ratchet holds, exit 0 |

## Note for reviewers

`pnpm lint` is currently a **vacuous pass**: the script's `"${@:-src/**/*}"` default expands to nothing, so with no arguments it lints 0 files and always succeeds — including in CI. Run properly (`pnpm exec oxlint src`), `main` has 3 pre-existing `display-name` errors in `src/lib/data-testid/__tests__/withDataTestId.test.tsx` (file untouched here, identical to `main`). Out of scope for this PR, but that lint job is guarding nothing today and is worth a follow-up.


sauldom102 (approved):
Supeer! We could focus now on Avatar components, we have many of them tagged as stable with just a few things missing in order to reach the DoD

PR #4933, opened 2026-07-30
fix(useSelectable): keep pre-selected multi-select values checked when their option loads first

sauldom102:
## Problem
In multi-select, a **pre-selected value gets stuck unchecked** when its option row loads *before* the controlled value resolves. `useSelectable`'s `updateLocalSelectedState` merges the external `selectedState` additively: for an item that already exists locally with a resolved record, it never flips `checked: false → true` to match the external state. The data-sync effect seeds loaded option rows as `checked: false`, so this is a **race**:

1. The options page loads → the selected item's row is added as `checked: false`.
2. The controlled value then resolves (`checked: true`) → the additive merge ignores it → the item stays unchecked.

Off-page selected items are never pre-seeded, so they arrive `checked: true` — only the on-page one is lost. The selector then reports **one fewer selected** than the controlled value contains, and committing the selection back drops it.

This reliably triggers with async form defaults (F0Form/form-definitions) and any paginated picker whose selected item is on the first page.

## Fix
Honor an external `checked: true` for already-present items in multi-select, while still ignoring external `checked: false` so cross-page manual selections stay additive. Single-select behavior is unchanged (external state remains the full source of truth).

## Tests
Added two regression tests to `useSelectable.test.tsx` (both fail before, pass after):
- **keeps every externally-selected id checked, even the one already on the page** — reproduces the race (options page → then external value); asserts all 3 stay selected. Before the fix it asserts `expected 2 to be 3`.
- **stays additive** — a local selection the external value omits is preserved; a newly-added external id whose row was pre-seeded still becomes selected.

Full suite: 31/31 `useSelectable` tests pass.

## Downstream
Reproduced in Factorial as **FCT-57267**: a task's *Assignees* editor showed "2 selected" for a 3-assignee task and dropped the third assignee on any edit. This is the shared-component root cause behind that ticket; the monorepo will bump `@factorialco/f0-react` once this releases.

### DIFF
--- a/packages/react/src/hooks/datasource/__test__/useSelectable.test.tsx
+++ b/packages/react/src/hooks/datasource/__test__/useSelectable.test.tsx
@@ -1,7 +1,7 @@
 import { act, renderHook, waitFor } from "@testing-library/react"
 import { describe, expect, it, vi } from "vitest"
 
-import { DataSource, PaginationInfo } from "../types"
+import { DataSource, PaginationInfo, SelectedItemsState } from "../types"
 import { Data, GROUP_ID_SYMBOL } from "../useData"
 import { useSelectable } from "../useSelectable/useSelectable"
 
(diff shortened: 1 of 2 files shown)


albertpmz:
**Rebased onto latest main and fixed the format failure**

The branch was 251 commits behind main (last commit Jul 30), so CI was red for reasons mostly unrelated to the change here. Rebased onto current main, no conflicts.

The one real failure was **Format**: `oxfmt` wanted the `needsItem` declaration wrapped, it was a line over. Fixed in `style(useSelectable): satisfy oxfmt line wrapping`, no behaviour change.

**Storybook Tests** was down to the staleness, not this change. Ran the full suite locally against a fresh static build before pushing (363 suites, 2317 tests, all green) and CI now agrees, all 8 shards passing.

The logic in this PR is untouched, only the rebase and the formatting fix.

albertpmz (approved):

PR #4946, opened 2026-07-31
feat(F0DataChart): render text-only empty state

igorsafonov-gif:
## Description

The F0DataChart empty state previously rendered a faded chart-type skeleton behind the message, which made empty charts read as "still loading" and added visual noise. This PR removes the skeleton background so the empty state is just the centered title and description.

Before: faded per-chart-type skeleton illustration behind the text. After: text only — every chart type (bar, line, pie, funnel, radar, gauge, heatmap) now shows the same centered "No data available / Try a different date or fewer filters" message.

Before
image

image

## Implementation details

- feat: remove the skeleton background layer and per-type skeleton map from `DataChartEmptyState`, keeping only the centered message
- chore: keep `chartType` as an optional `@deprecated` prop on `DataChartEmptyState` and `DataChartEmptyStateView` so existing consumers (e.g. the factorial monorepo) don't break on upgrade

  <details>
  <summary>Why not remove the prop?</summary>

  `DataChartEmptyStateView` is a public export of the F0DataChart kit and is
  called with `chartType` from the factorial monorepo. Dropping the prop would
  turn this into a breaking API change; deprecating it keeps the upgrade
  friction-free and the prop can be deleted in the next major.
  </details>

- refactor: stop passing `chartType` from `F0DataChart` and `F0AnalyticsDashboard`'s `ChartItem`
- docs: update the empty-state description in `EmptyStates.stories.tsx` autodocs and `F0DataChart.mdx` to match the new text-only behavior

### DIFF
--- a/packages/react/src/kits/F0DataChart/__stories__/F0DataChart.mdx
+++ b/packages/react/src/kits/F0DataChart/__stories__/F0DataChart.mdx
@@ -37,7 +37,7 @@ All variants share an optional `emptyState` prop — see [Empty states](#empty-s
 
 ## Empty states
 
-When data is empty (empty `series`, empty `data`, all-zero values, or a missing gauge `value`), `F0DataChart` auto-renders a faded chart skeleton with a centered message so you never see a bare axis. No prop is required — it is the default behavior.
+When data is empty (empty `series`, empty `data`, all-zero values, or a missing gauge `value`), `F0DataChart` auto-renders a centered message so you never see a bare axis. No prop is required — it is the default behavior.
 
 Default copy is **"No data available"** / **"Try a different date or fewer filters"**.
 
--- a/packages/react/src/kits/F0DataChart/__stories__/F0DataChart.mdx
+++ b/packages/react/src/kits/F0DataChart/__stories__/F0DataChart.mdx
@@ -37,7 +37,7 @@ All variants share an optional `emptyState` prop — see [Empty states](#empty-s
 
 ## Empty states
 
-When data is empty (empty `series`, empty `data`, all-zero values, or a missing gauge `value`), `F0DataChart` auto-renders a centered message so you never see a bare axis. No prop is required — it is the default behavior.
+When data is empty (empty `series`, empty `data`, or a missing gauge `value`), `F0DataChart` auto-renders a centered message so you never see a bare axis. No prop is required — it is the default behavior. All-zero datasets are **not** empty — `[0, 0, 0]` is a legitimate zero-valued chart and renders as usual.
 
 Default copy is **"No data available"** / **"Try a different date or fewer filters"**.
 
--- a/packages/react/src/kits/F0DataChart/components/EmptyState/DataChartEmptyStateView.tsx
+++ b/packages/react/src/kits/F0DataChart/components/EmptyState/DataChartEmptyStateView.tsx
@@ -5,7 +5,11 @@ import type { F0DataChartEmptyStateProps, F0DataChartProps } from "../../types"
 import { DataChartEmptyState } from "./EmptyState"
 
 interface DataChartEmptyStateViewProps {
-  /** @deprecated No longer used — the empty state renders text only. */
+  /**
+   * @deprecated No longer used — the empty state renders text only. Remove the prop.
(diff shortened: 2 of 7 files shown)


packages/react/src/kits/F0DataChart/components/EmptyState/DataChartEmptyStateView.tsx
```diff
@@ -5,8 +5,8 @@ import type { F0DataChartEmptyStateProps, F0DataChartProps } from "../../types"
 import { DataChartEmptyState } from "./EmptyState"
 
 interface DataChartEmptyStateViewProps {
-  /** The chart variant — drives the background skeleton illustration. */
-  chartType: F0DataChartProps["type"]
+  /** @deprecated No longer used — the empty state renders text only. */
```
zygisS22: [P2] Complete the public deprecation contract before merging. DataChartEmptyStateView is exported by the kit, so its inferred public props expose this deprecated chartType. F0 requires @deprecated, @removeIn, and a reachable @migration guide for a public prop. Please add the removal version and migration path, or keep the prop non-deprecated until that contract exists.

packages/react/src/kits/F0DataChart/__stories__/F0DataChart.mdx
```diff
@@ -37,7 +37,7 @@ All variants share an optional `emptyState` prop — see [Empty states](#empty-s
 
 ## Empty states
 
-When data is empty (empty `series`, empty `data`, all-zero values, or a missing gauge `value`), `F0DataChart` auto-renders a faded chart skeleton with a centered message so you never see a bare axis. No prop is required — it is the default behavior.
+When data is empty (empty `series`, empty `data`, all-zero values, or a missing gauge `value`), `F0DataChart` auto-renders a centered message so you never see a bare axis. No prop is required — it is the default behavior.
```
zygisS22: [P2] Align this contract with the implementation and tests. isDataChartEmpty deliberately treats an all-zero dataset as non-empty, and its unit coverage asserts that behavior, so this documentation currently promises an empty state that will not render. Please either change the detector and tests or remove all-zero values here and from the EmptyStates story description.

zygisS22 (changes requested):
Requesting changes for two public-contract issues. I reviewed the full diff and exercised the empty-state stories in Chrome: all seven text-only variants, custom copy, the render escape hatch, and disabled detection behaved as expected. Focused Vitest passed 40/40, and typecheck/format/diff checks are clean. The remaining blockers are the incomplete public deprecation contract and documentation that still disagrees with the tested all-zero behavior.

packages/react/src/kits/F0DataChart/components/EmptyState/DataChartEmptyStateView.tsx
igorsafonov-gif: Done in 640aaabee. Added `@removeIn 5.0.0` (matching the other live deprecations in the kit, e.g. `F0CanvasCard`) and a `@migration` link to a new guide at `packages/react/docs/migrations/f0-datachart-emptystate-charttype-removal.md`, on both `DataChartEmptyStateView` and `DataChartEmptyState`.

packages/react/src/kits/F0DataChart/__stories__/F0DataChart.mdx
igorsafonov-gif: Done in 640aaabee. You are right — `isDataChartEmpty` returns `false` for all-zero data and the unit tests assert it, so the docs were promising an empty state that never renders. Removed `all-zero values` from the trigger list in both `F0DataChart.mdx` and the `EmptyStates` story description, and stated the opposite explicitly instead ("All-zero datasets are **not** empty — `[0, 0, 0]` is a legitimate zero-valued chart and renders as usual"). Left the detector and its tests untouched.

zygisS22 (approved):
🚀

warcos-fact:
Do you have a before/after screenshots to see the change in action?

igorsafonov-gif:
@warcos-fact added example to the text of the post

siguenzaraul (approved):

PR #4947, opened 2026-07-31
feat(F0DataChart): bar labels by default and stacked bar polish

igorsafonov-gif:
## Description

Two related bar-chart changes. Bar charts inside `F0AnalyticsDashboard` rendered without value labels because `F0DataChart`'s `showLabels` defaults to `false`, so dashboard tiles needed per-item config to show readable values — labels are now the dashboard default. On top of that, stacked bars get visual polish: a hairline gap between segments, a hover highlight that isolates one series across every bar, and consistently white inside-labels.

**One deliberate behaviour change rides along**, called out separately below because it is a responsive-contract change rather than polish: horizontal bar charts now keep their category axis at the `md` breakpoint (220–519px), where the shared matrix previously hid it for every chart family.

Dashboard bar charts (vertical, stacked, horizontal) now show their value labels. Stacked segments are separated by a 0.5px hairline in the colour of the surface the chart sits on, and hovering a segment fades every other series down to 40% opacity over 500ms.

https://github.com/user-attachments/assets/16e15e1e-6358-469d-8d45-1d7c50f24a24

Showing labels on MD breakpoints
**Before**
md-category-axis-BEFORE
**After**
md-category-axis-AFTER

## Implementation details

- feat: default `showLabels: true` for bar charts in `buildNativeChartProps`, spread before the item config so explicit values win
- feat: include `showLabels: true` in `defaultChartConfig("bar")` so charts transformed to bar via the chart-type menu also get labels
- feat: preserve an explicitly-set `showLabels` across chart-type transforms, mirroring the existing `showLegend` handling
- feat: separate stacked segments with a `0.5px` border in the container's own background colour

  <details>
  <summary>Why a border rather than a gap?</summary>

  ECharts 6 has no native spacing between stacked segments — `barGap` and
  `barCategoryGap` only control space *between* bars, and the stack mixin
  exposes just `stack` / `stackStrategy` / `stackOrder`. A border in the
  container colour is the standard workaround. Both adjacent segments stroke
  the shared edge and canvas strokes are centred on the path, so the two cover
  the same band rather than adding to it: the separation equals the border
  width, not twice it. `0.5` is deliberate — a hairline at 1x, a crisp single
  device pixel at 2x.
  </details>

- feat: highlight the hovered series across all bars via `emphasis.focus: "series"`, dimming other series to 40% opacity, with a 500ms `stateAnimation` fade

  <details>
  <summary>Why the animation flags change</summary>

  `stateAnimation` is only honoured at the option root — the series-level one
  in the ECharts types is ignored for state transitions. It also requires the
  animation engine to be on, while the base options set `animation: false` to
  skip entrance animations. Stacked charts therefore re-enable animation with
  zero-duration entrance and update animations, so only the blur fade animates.
  </details>

- feat: always render inside-bar value labels in white

  <details>
  <summary>Contrast note</summary>

  This replaces a per-fill luminance check that flipped to black on lighter
  colours, which made one bar look like two label styles. Every selectable
  series colour is a shade-50 chromatic token, so white stays legible, but on
  the two lightest (`flubber`, `yellow`) it is below WCAG AA — a deliberate
  design call in favour of consistency.
  </details>

## Accepted behaviour change: horizontal category axis at `md`

Not label polish — a deliberate change to the shared responsive contract, kept in this PR because it is what makes labelled horizontal bars legible in a chat card. Reviewed on its own terms:

**What changed.** `resolveResponsiveDisplay` returned `showCategoryAxis: size === "lg"` for every bar chart, matching `LineChart`. It now also returns `true` for horizontal bars at `md`:

```ts
showCategoryAxis: size === "lg" || (size === "md" && !isVertical)
```

**Why.** A horizontal category label names the single row beside it. Hiding them leaves a stack of anonymous bars that the value axis alone can't explain — the reader sees four lengths and no subjects. Vertical bars keep the old rule: their categories run along the bottom competing with each other, and they crowd well before the plot runs out of room.

**What it costs.** Category labels take `min(80, width * 0.2)` (`yAxisMaxLabelWidth`), so a horizontal chart in the `md` band gives up 44px at 220px and 80px from 400px up — bars up to 20% shorter than before, in the same container. `md` is the wide-chat-card size, so this lands on existing consumers, not only the dashboard tiles this PR is about. That is the trade-off being accepted: naming the rows is worth a fifth of the bar length.

**Evidence.** The screenshots below are the new `CategoryAxisBoundaries` story rendered twice — once with the deviation reverted (before) and once as it ships (after) — so the width cost is visible rather than asserted:

| width | before | after |
| --- | --- | --- |
| 219px (`sm`) | anonymous bars | unchanged — `sm` drops all chrome |
| 220px (first `md`) | anonymous bars | row labels, truncated to the 44px cap; bars ~20% shorter |
| 519px (last `md`) | anonymous bars | full row labels; bars ~80px shorter |
| 520px (first `lg`) | full row labels | unchanged — `lg` always showed the axis |

The vertical row in each shot is the control: identical before and after, gaining its axis only at 520px.

Chromatic can't supply that pair on its own — `CategoryAxisBoundaries` is a new story, so it has no baseline and registers as *new* rather than as a diff. What it buys is forward coverage: this band previously had no image coverage at all, since the responsive matrix stories render 180/320/720 and aren't snapshotted. Boundary unit tests pin the flip to the exact pixel in both orientations, so a later breakpoint tweak fails loudly instead of silently reflowing every horizontal chart in a chat card.

## Review fixes

Applied after review — see the `fix(F0DataChart)` commits.

- fix: an explicitly-`undefined` `showLabels` no longer beats the bar default (`...chart` carries the key even when unset, so the default is resolved after the spread)
- fix: only bar targets inherit `showLabels` across a chart-type transform — pie and funnel default it to `true`, so transforming one to a line was showing labels the line default keeps off
- fix: the stacked animation block defers to consumer-provided `echartsOptions` instead of overwriting them
- fix: the target "ghost" series fades to the same 40% as its stack rather than the ECharts default
- fix: the separator uses the nearest painted ancestor background (`containerBackground`) instead of the page token, so charts on tinted cards, modals and fullscreen overlays blend correctly
- fix: the cross-fade honours `prefers-reduced-motion` via `useReducedMotion()` — the duration drops to 0, so the hover highlight still isolates the series but arrives instantly. The preference is the user's rather than the consumer's, so it also flattens an explicitly-provided `stateAnimation` duration
- test: the `md` category-axis deviation is now declared as an accepted behaviour change (section above) with boundary assertions at 219/220 and 519/520 for both orientations, driven off `SM_MAX_WIDTH` / `MD_MAX_WIDTH` rather than hardcoded widths, plus a Chromatic-snapshotted `CategoryAxisBoundaries` story
- test: 36 new unit tests across `BarChart`, the chart theme, and the dashboard label defaults

### DIFF
--- a/packages/react/src/kits/F0DataChart/components/BarChart/useBarChartOptions.ts
+++ b/packages/react/src/kits/F0DataChart/components/BarChart/useBarChartOptions.ts
@@ -1,6 +1,8 @@
 import * as echarts from "echarts"
 import { type RefObject, useMemo } from "react"
 
+import { useReducedMotion } from "@/lib/a11y"
+
 import type {
   F0DataChartBarDataPoint,
   F0DataChartBarProps,
@@ -52,7 +54,10 @@ const STACK_GAP_BORDER_WIDTH = 0.5
 /** Opacity of the series that are *not* hovered, while one series has focus. */
 const BLUR_OPACITY = 0.4
 
-/** Cross-fade duration (ms) in and out of the hover blur state. */
+/**
+ * Cross-fade duration (ms) in and out of the hover blur state. Drops to 0 when
+ * the user prefers reduced motion.
+ */
 const STATE_ANIMATION_DURATION = 500
 
 function resolveGridRightSpace(
@@ -454,6 +459,7 @@ export function useBarChartOptions(
   const theme = useChartTheme(containerRef)
   const { width: containerWidth, height: containerHeight } =
     useContainerSize(containerRef)
+  const prefersReducedMotion = useReducedMotion()
 
   return useMemo(() => {
     const isVertical = orientation === "vertical"
@@ -688,15 +694,22 @@ export function useBarChartOptions(
     // This runs after `buildBaseChartOptions` merged the consumer's
     // `echartsOptions`, so each key defers to an explicitly-provided value —
     // same convention as the `userGridRight` guard below.
+    //
+    // `prefers-reduced-motion` zeroes the duration rather than dropping the
+    // state: the hover highlight still isolates the series, it just arrives
+    // instantly. The preference belongs to the user, not the consumer, so it
+    // also flattens an explicitly-provided `stateAnimation` duration.
     if (stacked) {
       options.animation = echartsOptions?.animation ?? true
       options.animationDuration = echartsOptions?.animationDuration ?? 0
       options.animationDurationUpdate =
         echartsOptions?.animationDurationUpdate ?? 0
-      options.stateAnimation = echartsOptions?.stateAnimation ?? {
-        duration: STATE_ANIMATION_DURATION,
-        easing: "cubicOut",
-      }
+      options.stateAnimation = prefersReducedMotion
+        ? { ...echartsOptions?.stateAnimation, duration: 0 }
+        : (echartsOptions?.stateAnimation ?? {
+            duration: STATE_ANIMATION_DURATION,
+            easing: "cubicOut",
+          })
     }
 
     // Non-stacked horizontal bars render labels BESIDE the bar end, so the
@@ -733,5 +746,6 @@ export function useBarChartOptions(
     containerWidth,
     containerHeight,
     size,
+    prefersReducedMotion,
   ])
 }
--- a/packages/react/src/kits/F0DataChart/components/BarChart/useBarChartOptions.ts
+++ b/packages/react/src/kits/F0DataChart/components/BarChart/useBarChartOptions.ts
@@ -406,27 +406,28 @@ function buildSeriesEntries(
 export type BarChartSize = ChartResponsiveSize
 
 /**
- * Maps a discrete `size` to which chrome (legend, axes) is rendered. The
- * matrix mirrors `LineChart.resolveResponsiveDisplay` so the two chart
- * families behave the same at every breakpoint, with one bar-only exception
- * at `md`:
+ * Maps a discrete `size` to which chrome (legend, axes) is rendered:
  *
  * - `sm` → just the bars, no axes, no legend
- * - `md` → bars + legend + value axis; the category axis shows for horizontal
- *   bars only
- * - `lg` → bars + legend + both axes (with smart truncation on the category axis)
+ * - `md` / `lg` → bars + legend + both axes (with smart truncation on the
+ *   category axis)
  *
- * Horizontal bars carry their categories on the left, one label per row beside
- * the bar it names. Dropping them leaves a stack of anonymous bars that the
- * value axis alone can't explain, and they cost at most 20% of the width
- * (`yAxisMaxLabelWidth`) rather than competing with each other. Vertical bars
- * keep the plain `md` rule: their categories run along the bottom, where they
- * crowd each other well before the plot runs out of room.
+ * Bars deviate from `LineChart.resolveResponsiveDisplay`, which keeps the
+ * category axis for `lg` alone. A bar chart's categories are its subjects — a
+ * plot without them is a set of lengths with nothing to compare, which the
+ * value axis can't rescue. So bars show them wherever there is room for any
+ * chrome at all, and `sm` is the only size that drops them.
+ *
+ * Crowding is handled rather than avoided: `computeCategoryAxisLayout` runs
+ * whenever the axis is shown, fitting every label with ellipsis truncation and
+ * only skipping labels when even a 3-char stub won't fit. The cost is plot
+ * area — horizontal labels take `min(80, width * 0.2)` of the width, vertical
+ * ones a row of height.
  */
-function resolveResponsiveDisplay(size: BarChartSize, isVertical: boolean) {
+function resolveResponsiveDisplay(size: BarChartSize) {
   return {
     showLegend: size !== "sm",
-    showCategoryAxis: size === "lg" || (size === "md" && !isVertical),
+    showCategoryAxis: size !== "sm",
     showValueAxis: size !== "sm",
   }
 }
@@ -472,7 +473,7 @@ export function useBarChartOptions(
       containerWidth
     )
 
-    const responsive = resolveResponsiveDisplay(size, isVertical)
+    const responsive = resolveResponsiveDisplay(size)
     // The user-provided `showLegend` prop can still force the legend off,
     // but it can never override the `sm` rule.
     const effectiveShowLegend = responsive.showLegend && showLegend
--- a/packages/react/src/kits/F0DataChart/components/BarChart/useBarChartOptions.ts
+++ b/packages/react/src/kits/F0DataChart/components/BarChart/useBarChartOptions.ts
@@ -402,17 +402,26 @@ export type BarChartSize = ChartResponsiveSize
 
 /**
  * Maps a discrete `size` to which chrome (legend, axes) is rendered. The
- * matrix is identical to `LineChart.resolveResponsiveDisplay` so the two
- * chart families behave the same at every breakpoint:
+ * matrix mirrors `LineChart.resolveResponsiveDisplay` so the two chart
+ * families behave the same at every breakpoint, with one bar-only exception
+ * at `md`:
  *
  * - `sm` → just the bars, no axes, no legend
- * - `md` → bars + legend + value axis, no category axis
+ * - `md` → bars + legend + value axis; the category axis shows for horizontal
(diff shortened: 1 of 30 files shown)


packages/react/src/kits/F0DataChart/components/BarChart/useBarChartOptions.ts
```diff
@@ -29,34 +29,31 @@ const ARIA_MAX_SERIES = 10
...
-      Number.parseInt(normalized.slice(offset, offset + 2), 16) / 255
-    return value <= 0.04045
-      ? value / 12.92
-      : Math.pow((value + 0.055) / 1.055, 2.4)
-  })
-  const luminance =
-    0.2126 * (channels[0] ?? 0) +
-    0.7152 * (channels[1] ?? 0) +
-    0.0722 * (channels[2] ?? 0)
+const INSIDE_LABEL_COLOR = "#ffffff"
```
zygisS22, line 48: [P1] Restore a contrast-safe foreground instead of forcing white for every stacked label. These are 11px text, so WCAG AA requires 4.5:1; 12 of the 15 supported shade-50 tokens fail with white. Examples: flubber 1.96:1, yellow 2.02:1, viridian 3.17:1, and barbie 3.29:1. Canvas text is invisible to axe, so the green accessibility check cannot catch this. Please select text color from measured contrast or constrain the fills, and add ratio-based coverage.

packages/react/src/kits/F0DataChart/components/BarChart/useBarChartOptions.ts
```diff
@@ -388,17 +402,26 @@ export type BarChartSize = ChartResponsiveSize
...
+ * (`yAxisMaxLabelWidth`) rather than competing with each other. Vertical bars
+ * keep the plain `md` rule: their categories run along the bottom, where they
+ * crowd each other well before the plot runs out of room.
  */
-function resolveResponsiveDisplay(size: BarChartSize) {
+function resolveResponsiveDisplay(size: BarChartSize, isVertical: boolean) {
   return {
     showLegend: size !== "sm",
-    showCategoryAxis: size === "lg",
+    showCategoryAxis: size === "lg" || (size === "md" && !isVertical),
```
zygisS22: [P2] This changes the medium-breakpoint contract for horizontal bars: the category axis now consumes plot width from 220-519px instead of remaining hidden. Please make this an explicit accepted behavior with before/after evidence and visual assertions at the boundary widths, or split/revert it if it is outside the label-polish scope.

packages/react/src/kits/F0DataChart/components/BarChart/useBarChartOptions.ts
```diff
@@ -654,6 +678,27 @@ export function useBarChartOptions(
...
+    //
+    // This runs after `buildBaseChartOptions` merged the consumer's
+    // `echartsOptions`, so each key defers to an explicitly-provided value —
+    // same convention as the `userGridRight` guard below.
+    if (stacked) {
+      options.animation = echartsOptions?.animation ?? true
+      options.animationDuration = echartsOptions?.animationDuration ?? 0
+      options.animationDurationUpdate =
+        echartsOptions?.animationDurationUpdate ?? 0
+      options.stateAnimation = echartsOptions?.stateAnimation ?? {
```
zygisS22: [P2] Respect reduced-motion preferences for this new 500ms cross-fade. F0 requires animated components to use useReducedMotion and set duration to 0 when motion is reduced. Please branch the stateAnimation duration and add a focused reduced-motion assertion.

zygisS22 (changes requested):
Requesting changes for one accessibility blocker and two behavioral gaps. Focused Vitest passed 67/67; typecheck/format/diff checks are clean. In Chrome I verified the stacked and horizontal stories, the nine-chart responsive matrix, dashboard labels, and hover tooltips. Separately, this head merges cleanly by itself, but applying PR 4948 after it conflicts in useBarChartOptions.ts and utils/theme.ts, so the combined head must be resolved and retested.

packages/react/src/kits/F0DataChart/components/BarChart/useBarChartOptions.ts
igorsafonov-gif: Done on purpose. 
To fix this we need to change the actual chart colors, which is a complex case. All numbers are duplicated in tooltips, so it's okay for now.

packages/react/src/kits/F0DataChart/components/BarChart/useBarChartOptions.ts
igorsafonov-gif: Fixed in 817f7a80a. `useBarChartOptions` now calls `useReducedMotion()` and the cross-fade duration branches to 0:

```ts
options.stateAnimation = prefersReducedMotion
  ? { ...echartsOptions?.stateAnimation, duration: 0 }
  : (echartsOptions?.stateAnimation ?? { duration: 500, easing: "cubicOut" })
```

Zeroing the duration rather than dropping the state was deliberate — hovering still isolates the series and dims the rest to 40%, it just arrives instantly instead of fading, so the affordance survives without the motion.

One judgement call worth flagging: it also flattens an explicitly-provided `stateAnimation` duration (the `easing` key is preserved). The rest of that block defers to consumer `echartsOptions`, but `prefers-reduced-motion` is the end user's setting rather than the integrator's, so it wins. Happy to restore the defer-to-consumer symmetry there if you'd rather.

Two assertions in `BarChart.test.tsx` cover it — the zeroed default with the highlight intact, and the consumer-override case — driven by a query-scoped `matchMedia` stub (same pattern as `ChatHeader.fullscreen.test.tsx`, so it can't accidentally flip unrelated media queries). Both verified to fail without the source change.

packages/react/src/kits/F0DataChart/components/BarChart/useBarChartOptions.ts
igorsafonov-gif: Kept, and declared rather than split — you were right that it was riding along undocumented. The PR description now has an **Accepted behaviour change: bar category axis at `md`** section covering what changed, why, the measured cost, and the evidence.

Two things went further than your comment asked:

**It now covers vertical bars too.** Writing up the rationale exposed that the horizontal-only split didn't hold. The reason vertical hid its axis at `md` was crowding — categories along the bottom competing for the same pixels — but crowding is already handled rather than avoided: `computeCategoryAxisLayout` runs whenever the axis is shown, fitting every label with ellipsis truncation and only skipping labels when even a 3-char stub won't fit, with `hideOverlap` on top. So the rule collapses to `showCategoryAxis: size !== "sm"` for both orientations, and `sm` is the only size that drops categories. A bar chart's categories are its subjects; without them the chart is a set of lengths with nothing to compare.

**The cost is stated numerically, both directions.** Horizontal labels take `min(80, width * 0.2)` — 44px at 220px, 80px from 400px up, so bars up to 20% shorter in the same container. Vertical labels take a row of height and arrive truncated. `md` is the wide-chat-card size, so this lands on existing consumers, not just the dashboard tiles this PR is about.

Evidence, per your ask:

- **Boundary assertions** in `BarChart.test.tsx` at 219/220 and 519/520 for both orientations, driven off `SM_MAX_WIDTH` / `MD_MAX_WIDTH` rather than hardcoded widths, so a later breakpoint tweak fails loudly instead of silently reflowing every bar chart in a chat card.
- **A `CategoryAxisBoundaries` story** rendering both orientations at all four boundary widths, carrying `withSnapshot` — this band had no image coverage at all before, since the responsive matrix stories render 180/320/720 and aren't snapshotted. Worth knowing Chromatic can't produce the before/after by itself: a new story has no baseline, so it registers as *new* rather than as a diff.
- **A before/after pair** captured from that story — once with the deviation reverted, once as it ships.
- **The worst case checked by hand** rather than argued: `ResponsiveSnapshotMatrix` at 320px with 10 long month names across 1 / 3 / 5 series truncates and skips to every other label (`Sept…`, `Nov…`, `Janu…`) with no overlap in all three rows.

One knock-on for a separate decision: bars now diverge from `LineChart`, which still shows its category axis at `lg` only. Line has a defensible reason to differ — a time axis reads fine without every tick labelled — but if we want the families consistent, that's its own change.

zygisS22 (approved):
LGTM!

siguenzaraul (approved):

PR #4948, opened 2026-07-31
feat(F0DataChart): unify the tooltip across every chart type

igorsafonov-gif:
## Description

Each chart type had grown its own tooltip formatter, so the same hover produced a different layout depending on the visualization — pie showed `value (pct%)`, funnel had `Overall:` / `From X:` lines, radar an inline list, bar an axis summary of every series. This introduces one shared `renderValueTooltip` and routes all seven chart types through it, so a tooltip always reads the same: coloured dot + name, the category, a large value, then context rows.

https://github.com/user-attachments/assets/77e36fd1-095f-4456-965d-143d27219bec

image

image

image

image

image

image

image

## Implementation details

- feat: add `renderValueTooltip`, `deltaRow` and `renderMarker` to the chart options utils as the single tooltip renderer, with a 130px min width and escaping of everything except ECharts' own marker HTML
- feat: route bar, line, pie, funnel, radar, gauge and heatmap tooltips through the shared renderer
- feat: show full numbers in tooltips (`125,000`, not the compact `125k` used on axes); `tooltipValueFormatter` still overrides
- feat: show "from previous" only on line charts

  <details>
  <summary>Why line only?</summary>

  Bar categories are not necessarily a sequence — departments, locations,
  statuses — so comparing a bar with the one to its left is often meaningless.
  A line chart is a trend by definition, so the comparison always holds there.
  </details>

- feat: give bar charts an item-triggered tooltip describing the hovered bar or segment, replacing the axis tooltip that listed every series

  <details>
  <summary>Interaction change</summary>

  With `trigger: "item"` a bar tooltip appears only when the pointer is over a
  bar, not anywhere in the category column. That is what makes it possible to
  know *which* segment is hovered. Lines keep the axis trigger — a 2px line is
  too thin to hover reliably — and therefore list every series when there is
  more than one.
  </details>

- fix: build the gauge tooltip marker from the resolved arc colour, since ECharts' own `params.marker` uses the palette entry rather than the colour painted
- feat: add `positive` / `critical` to the chart theme colours for signed deltas
- test: cover the renderer's escaping, omitted sections, delta signing and marker sanitisation, plus per-type tooltip behaviour on bar charts

### After review (@zygisS22)

- **fix: never derive a bar's share from a signed net total.** A category mixing gains with losses has no total its parts add up to — 24 hires against a net of 19 read as "126.3% of total", and near-cancellation made the ratio arbitrarily large. Mixed-sign categories now show the value alone; all-negative keeps both rows (the ratio still reads as a positive share of a negative whole); all-zero keeps the true total and skips the division.
- **feat!: one tooltip number contract for every chart type.** `tooltipValueFormatter` was bar/line only, and the other five formatted their tooltip with `valueFormatter` — so the same hover still read differently per visualization. All seven now share one rule in one place (`tooltipValueFormat`): a tooltip shows the full localized number, and `tooltipValueFormatter` is how a caller changes it.

  <details>
  <summary>Breaking for five types</summary>

  Funnel, pie, radar, gauge and heatmap tooltips no longer apply
  `valueFormatter`. A chart that put a unit in `valueFormatter` and relied on
  seeing it on hover needs the same function passed as
  `tooltipValueFormatter`.
  </details>

- **fix: i18n falls back to the built-in translations for missing keys.** `I18nProvider` deep-merges the dictionary it is given over `defaultTranslations`, so a dictionary written against an older f0 renders English for keys it lacks instead of `undefined` — or throwing on a new subtree. The prop type stays exact, so a consumer still learns at compile time that new copy needs translating.
- **docs: correct the formatter contract.** `F0DataChart.mdx` claimed `valueFormatter` formatted tooltips; `Line/CustomFormatters` showed `M €` on the axis and raw values on hover. Both now describe and demonstrate the pair.
- **test: execute the five uncovered formatters.** Funnel, pie, radar, gauge and heatmap tooltip callbacks are now called with the params ECharts passes — values, missing values, escaping and each type's context rows. `FunnelChart.test.tsx` is new; funnel had no test file.
- **chore: rebased on #4947** (conflicted in `useBarChartOptions.ts` and `utils/theme.ts`), so the merge order is explicit. Base retargeted to `feat/dashboard-bar-labels-default`.

### DIFF
--- a/packages/react/src/kits/F0DataChart/components/BarChart/useBarChartOptions.ts
+++ b/packages/react/src/kits/F0DataChart/components/BarChart/useBarChartOptions.ts
@@ -593,7 +593,10 @@ export function useBarChartOptions(
     // numbers), otherwise fall back to the shared value formatter.
     const tooltipValue = tooltipValueFormatter ?? valueFormatter
 
-    const formatTooltipValue = tooltipValueFormat(tooltipValueFormatter)
+    const formatTooltipValue = tooltipValueFormat(
+      tooltipValueFormatter,
+      valueFormatter
+    )
 
     const options = buildBaseChartOptions({
       categories,
--- a/packages/react/src/kits/F0DataChart/components/BarChart/useBarChartOptions.ts
+++ b/packages/react/src/kits/F0DataChart/components/BarChart/useBarChartOptions.ts
@@ -2,6 +2,7 @@ import * as echarts from "echarts"
 import { type RefObject, useMemo } from "react"
 
 import { useReducedMotion } from "@/lib/a11y"
+import { useI18n } from "@/lib/providers/i18n"
 
 import type {
   F0DataChartBarDataPoint,
@@ -462,6 +463,7 @@ export function useBarChartOptions(
   size: BarChartSize
 ): echarts.EChartsOption {
   const theme = useChartTheme(containerRef)
+  const i18n = useI18n()
   const { width: containerWidth, height: containerHeight } =
     useContainerSize(containerRef)
   const prefersReducedMotion = useReducedMotion()
@@ -590,8 +592,10 @@ export function useBarChartOptions(
     // numbers), otherwise fall back to the shared value formatter.
     const tooltipValue = tooltipValueFormatter ?? valueFormatter
 
-    // Tooltips show full numbers, not the compact axis format ("125,000", not
-    // "125k"). `tooltipValueFormatter` exists precisely to override this.
+    // `valueFormatter` formats the value AXIS, which is usually compact
+    // ("125k"); tooltips show the full number instead ("125,000").
+    // `tooltipValueFormatter` overrides that — it is also the way to keep a
+    // unit (currency, "%") in the tooltip.
     const formatTooltipValue = (value: number) =>
       tooltipValueFormatter
         ? tooltipValueFormatter(value)
@@ -684,11 +688,14 @@ export function useBarChartOptions(
     }
 
     // Bar charts use an item-triggered tooltip about the hovered bar or
-    // segment (pairing with the stacked series highlight) instead of the
-    // axis tooltip listing every series: value large, then change vs. the
-    // previous category and — with multiple series — the category total
-    // and the hovered value's share of it.
-    {
+    // segment (pairing with the stacked series highlight) instead of the axis
+    // tooltip listing every series: value large, then — with multiple series —
+    // the hovered value's share of the category total, that total, and the
+    // target when there is one.
+    //
+    // `buildBaseChartOptions` has already merged `echartsOptions`, so a caller
+    // that passed its own tooltip keeps it: only build ours when it didn't.
+    if (echartsOptions?.tooltip === undefined) {
       options.tooltip = buildItemTooltip({
         theme,
         formatter: (params: unknown) => {
@@ -726,15 +733,15 @@ export function useBarChartOptions(
                 multiSeries &&
                   total > 0 && {
                     value: `${((value / total) * 100).toFixed(1)}%`,
-                    label: "of total",
+                    label: i18n.dataChart.tooltip.ofTotal,
                   },
                 multiSeries && {
                   value: formatTooltipValue(total),
-                  label: "total",
+                  label: i18n.dataChart.tooltip.total,
                 },
                 target !== undefined && {
                   value: formatTooltipValue(target),
-                  label: "target",
+                  label: i18n.dataChart.tooltip.target,
                 },
               ],
             },
@@ -775,6 +782,7 @@ export function useBarChartOptions(
     valueAxisSplitNumber,
     echartsOptions,
     theme,
+    i18n,
     containerWidth,
     containerHeight,
     size,
--- a/packages/react/src/kits/F0DataChart/components/BarChart/useBarChartOptions.ts
+++ b/packages/react/src/kits/F0DataChart/components/BarChart/useBarChartOptions.ts
@@ -10,7 +10,11 @@ import type {
 } from "../../types"
 
 import { paletteColor, resolveChartColorToken } from "../../utils/colors"
-import { buildBaseChartOptions, escapeTooltipText } from "../../utils/options"
+import {
+  buildBaseChartOptions,
+  buildItemTooltip,
+  renderValueTooltip,
+} from "../../utils/options"
 import type { ChartResponsiveSize } from "../../utils/responsive"
 import { useChartTheme } from "../../utils/useChartTheme"
 import { useContainerSize } from "../../utils/useContainerSize"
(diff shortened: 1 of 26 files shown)


packages/react/src/kits/F0DataChart/components/BarChart/useBarChartOptions.ts
```diff
@@ -654,6 +629,70 @@ export function useBarChartOptions(
...
+            value?: number
+            dataIndex?: number
+            marker?: string
+          }
+          const seriesName = String(p.seriesName ?? "")
+          if (seriesName.endsWith(" (target)")) return ""
+
+          const value = Number(p.value)
+          const dataIndex = p.dataIndex ?? 0
+          const total = series.reduce(
```
zygisS22: [P1] Do not compute part-of-total percentages from a signed net sum. In the existing Stacked Negative Values story, Q1 is 24 + 6 - 8 - 3 = 19, so Chrome reports 24 hires as 126.3% of total; near-cancellation can produce arbitrarily large percentages, and non-positive totals suppress only the percentage while retaining a misleading net total. Please omit total/share when signs are mixed or provide separate positive and negative subtotals, with mixed-sign, zero-total, and cancellation tests.

packages/react/src/lib/providers/i18n/i18n-provider-defaults.ts
```diff
@@ -672,6 +672,14 @@ export const defaultTranslations = {
       title: "No data available",
       description: "Try a different date or fewer filters",
     },
+    tooltip: {
```
zygisS22, line 675: [P1] This new required subtree is source- and runtime-breaking for consumers that pass a complete custom translations object. The public I18nProviderProps uses the exact inferred TranslationsType, and tooltip formatters directly dereference dataChart.tooltip, so older dictionaries either stop compiling or can fail at runtime. The breaking-changes label acknowledges the surface change, but please either document and ship it through an intentional major migration or preserve compatibility with a deep-partial/default merge or translation fallback.

packages/react/src/kits/F0DataChart/components/FunnelChart/useFunnelChartOptions.ts
```diff
@@ -139,26 +141,37 @@ export function useFunnelChartOptions(
...
+              value: formatPercent(val, prevData.value),
+              label: i18n.t("dataChart.tooltip.fromStage", {
+                stage: prevData.name,
+              }),
+            })
           }
         }
 
-        return `<div>${String(p.marker ?? "")} <strong>${name}</strong></div><div style="margin-top: 2px">${formattedValue}</div>${conversionHtml}`
+        return renderValueTooltip(
```
zygisS22, line 171: [P2] Add automated execution coverage for this formatter and the equivalent gauge, heatmap, pie, and radar callbacks. The current tests exercise the bar/line/shared-option paths, but these five callback bodies are not invoked, and their stories have no play/hover interaction. I manually verified the happy paths in Chrome, but regressions in ECharts parameter shape, zero/missing values, formatting, or escaping would currently pass CI.

zygisS22 (changes requested):
Requesting changes for tooltip correctness, compatibility, coverage, and documentation. Focused Vitest passed 73/73; typecheck/format/diff checks are clean. I manually hovered bar, line, pie, funnel, radar, gauge, and heatmap in Chrome. That reproduced the mixed-sign percentage defect and the Line / Custom Formatters contract drift: the axis shows M € while the tooltip shows raw values. Please also update F0DataChart.mdx around line 322 and Line.stories.tsx around line 169 so valueFormatter versus tooltipValueFormatter is documented and demonstrated. Finally, this head conflicts with PR 4947 in useBarChartOptions.ts and utils/theme.ts when applied after it; rebase on the intended merged order and rerun the combined matrix.

igorsafonov-gif:
Thanks — all four addressed, plus one change that goes further than you asked.

## [P1] Part-of-total from a signed net sum — `2f99f9698`

You're right, and the `total > 0` guard made it worse by hiding only the percentage while keeping the misleading net total on its own row.

A share of a total only means something when the parts push the same way, so a category with values on both sides of zero now shows **the value alone** — no share, no total. Q1 of Stacked Negative Values reads "24" and nothing else. An all-negative category keeps both rows (value and total share a sign, so the ratio is still a positive share of a negative whole: -8 of -10 → 80.0%). All-zero keeps the true total and skips the division.

Tests: mixed-sign, near-cancellation (1000 vs -999.9, which would have read 1,000,000%), all-negative, all-zero, and a series shorter than `categories` — that last one used to read past the end of the array.

## [P1] The new translation subtree — `d57faa86d`

Fixed the runtime half properly: `I18nProvider` now deep-merges the dictionary it is given over `defaultTranslations`, so a dictionary written against an older f0 renders English for keys it doesn't have instead of `undefined` — or throwing, when a whole subtree is new. Arrays stay leaves, and an explicit `undefined` means "not translated" rather than "erase the default". Two tests cover a stale dictionary and a partially-translated subtree.

I kept the prop type as the exact `TranslationsType` rather than widening it to a deep-partial: passing a complete object is how a consumer learns at compile time that new copy needs translating, and the (skipped) type test in the spec pins that intent. Worth noting this is how f0 has always shipped new copy — `feat(F0MeetingCard)` #4904, `feat(F0ProgressSeries)` #4842 and `feat(F0VideoPlayer)` #4754 each added a required subtree in a minor — so the fix above helps every one of those retroactively rather than being specific to this PR.

## [P2] The five uncovered formatters — `200c4f64d`

All five now execute in CI, called with the params ECharts passes: funnel (conversion rows read from *rendered* order, so `sort="ascending"` names the right previous stage), pie (share + whole), radar (per-indicator rows, including indicators the point omits), gauge (share of range, zero-width range, and that the dot uses the painted arc colour rather than the palette entry), heatmap (both dimensions, missing value tuple, out-of-range index). Escaping and zero/missing values are covered per type. `FunnelChart.test.tsx` is new — funnel had no test file at all.

## The `valueFormatter` / `tooltipValueFormatter` contract — `200c4f64d`

Docs were wrong, and the contract was worth fixing rather than just documenting. `tooltipValueFormatter` existed on bar and line only; the other five types formatted their tooltip with `valueFormatter`. So the same hover still read differently depending on the chart type, which is what this PR set out to remove.

Now all seven share one rule in one place (`tooltipValueFormat`): **a tooltip shows the full localized number, and `tooltipValueFormatter` is how you change it.** F0DataChart.mdx says so (it previously claimed `valueFormatter` formatted tooltips), and Line/CustomFormatters — the story where you saw `M €` on the axis and raw values on hover — now passes both and shows `4,000,000 €`.

⚠️ This is a behaviour change for funnel, pie, radar, gauge and heatmap: a chart that put a unit in `valueFormatter` and relied on seeing it on hover needs the same function passed as `tooltipValueFormatter`. Flagged as a BREAKING CHANGE footer on that commit.

## The #4947 conflict

Confirmed — `useBarChartOptions.ts` and `utils/theme.ts`. This branch is now **rebased on top of #4947** and the PR base retargeted to `feat/dashboard-bar-labels-default`, so the merge order is explicit rather than implied. Both conflicts were additive: `ChartThemeColors` gains `containerBackground` and `positive`/`critical` side by side, and the bar hook keeps #4947's `stateAnimation` block followed by this PR's item tooltip.

Combined matrix on the rebased head: 271 tests across F0DataChart + F0AnalyticsDashboard + i18n, all green (BarChart alone is 65 now — #4947's plus this PR's), lint and format clean, typecheck clean.

zygisS22 (approved):
🚀

siguenzaraul (approved):

PR #4950, opened 2026-08-01
feat(F0DataChart): add a scatter chart type

igorsafonov-gif:
## Description

Every chart type in the kit plots a category against a measure, so none of them can answer "does X relate to Y?". This adds `scatter` — two value axes, one point per entity, with an optional colour split by group — and wires it through `F0AnalyticsDashboard` so it works as a dashboard widget.

> [!NOTE]
> No longer stacked. #4948 has landed, and this branch is rebased onto `main` — the 18 commits it shared with #4947/#4948 are gone from the diff, which is now scatter-only. It uses `renderValueTooltip`/`tooltipValueFormat` from `main`.

**Safe to merge on its own, and the bump needs no migration.** No merge-order dependency, and nothing outside f0 has to change for the package to compile. Existing dashboards see no behaviour change: the chart-type picker change is scoped to scatter, and `ChartItem` now degrades an unknown or missing chart config to that one item's error state instead of throwing, where previously it would have taken the whole dashboard down.

> [!IMPORTANT]
> **The one compile-blocking change is gone.** @zygisS22 found that the branch's new `ai.dashboardItem.unsupportedChart` key was a real consumer migration: any addition to `TranslationsType` is compile-blocking, because `F0Provider.i18n` is `Omit<I18nProviderProps, "children">` and Factorial maintains a complete mirror in `frontend/src/modules/core/lib/f0-provider.tsx`. Rather than document it, `a6efd58` removes the key — the unrenderable-chart guard now passes a message-less `Error`, so the tile falls back to the already-translated `errorTitle` ("Error loading data") and `OneEmptyState` skips the optional description. The absent Retry button remains the tell that separates this from a fetch failure. **This PR now adds no translation key, so it contributes no migration of its own.** Note the narrow scope of that claim: the 5.0.1 -> 6.2.0 bump *does* need translation work, because `dataChart.tooltip` was added by other PRs in the same window — see factorial#108776, which hit exactly this `TS2741`. Dropping this key removes one instance; it does not fix the mechanism, which is why the partial-override follow-up below matters.
>
> **What's left is one non-compile-blocking change: a `case 'scatter'` in the widget mapper.** Scatter stays inert without it. `toDashboardChartConfig` in `frontend/src/modules/ai/copilotActions/canvas/dashboards/components/widgetItemMappers.ts` switches over `ChatDashboardChartConfig` with no `default`, and its declared return type is the local `type DashboardChartConfig = any` — so an eighth member compiles clean instead of raising TS2366, and a scatter item falls through to `undefined`. `isRenderableChart` catches that and degrades the one tile to its error state, so nothing crashes, but the widget renders as an error until the mapper adds:
>
> ```ts
> case 'scatter': {
>   const { valueFormat: _, xValueFormat: __, ...rest } = chatChart
>   // Both axes are measures, so X carries its own preset. It gets no runtime
>   // fallback: the resolved format describes a single measure, and the axes
>   // need not share a unit (salary in € against tenure in years).
>   const xValueFormatter = buildFormatter(chatChart.xValueFormat)
>   return {
>     ...rest,
>     ...(valueFormatter ? { valueFormatter } : {}),
>     ...(xValueFormatter ? { xValueFormatter } : {}),
>   }
> }
> ```
>
> Scatter is the first chart type with a measure on *both* axes, so it is the first to carry two presets — `valueFormat` for Y and the new `xValueFormat` for X — and `ScatterChartConfig` takes each as a formatter function. An earlier revision of this description spread `xValueFormat` through untouched, which silently left the X axis and its tooltip row rendering raw numbers; the snippet above is the corrected one, and is what factorial#108779 implements.
>
> That is the only remaining host site: nothing there narrows `computation`, and nothing uses `F0DataChartProps["type"]` as an exhaustive `Record`. factorial-agent cannot emit scatter yet, so nothing renders scatter until it does — the safe rollout order is f0 → Factorial → agent. The first follow-up below is the cleanup that would turn the *next* widening into a build error instead of a silent one.

<details>
<summary>Why this is a <code>feat:</code> minor, not <code>feat!:</code> — re: the API-surface bot</summary>

The bot's 5 entries reduced to 3 contracts. One is now resolved, one is real, one is a false positive:

- **Translations** — was real, now **removed**. `a6efd58` drops the key rather than migrating consumers, so there is no translation contract change left in this PR. Covered above.
- **`CanvasActions`** — real as a generic contract. F0 delivers the widened union *into* consumer-owned `dashboard.save`/`create` callbacks, and by parameter contravariance a callback written against the narrower union is no longer assignable. @zygisS22 verified this with a strict assignability probe on the workflow's own base/head `.d.ts` artifacts. Factorial's actual actions type `items` and `fetchSpecs` as `unknown`, so the same probe passes against the real implementation — a genuine semver concern, not a Factorial break.
- **`F0CanvasPanelProps`** — false positive. Here the union is an *input* the consumer supplies to f0, so every value accepted before is still accepted. Counted twice only because the symbol is exposed through both the `f0` and `ai` entrypoints; it is one change, not two.

The checker also has a blind spot in the other direction: it walks property paths on exported object types, so it flags `.computation` but not a top-level union alias gaining a member — it misses `ChatDashboardChartConfig` and `F0DataChartProps`. Making it variance-aware (inputs widen safely, callback params do not) and grouping duplicate entrypoint reports would cut the noise and catch what it currently misses.

On the bump: nothing is removed or renamed, and every type consumers reference still exists. Per [Release & Versioning](https://github.com/factorialco/f0/blob/main/packages/react/docs/development/release-and-versioning.mdx) this is the "add, don't replace" path — a normal `feat:` — and that doc's break path can't be followed for a purely additive change anyway, since it requires deprecating an old surface with `@deprecated`/`@removeIn`/`@migration` and there is no old surface here.

The durable fix for the remaining contract is to stop it being breakable. For translations, dropping the key sidesteps the problem for this PR but does not solve it — the next f0 key hits the same wall, so accepting partial overrides publicly while keeping the resolved dictionary complete internally is still the right change, and is filed as a follow-up. For `CanvasActions`, @zygisS22's stable-persistence-DTO proposal is the equivalent, also filed rather than bundled here.
</details>

## Screenshots

**Colour split, with the shared tooltip**

A point's label is the title, its series the subtitle, and both coordinates read as rows — the same shape radar uses, since a scatter point has no single headline value.

**Dashboard widget** — `patterns-analyticsdashboard--mixed-dashboard` now carries a "Salary vs Tenure" item, which exercises the chain end to end: skeleton → fetch → render → PNG/CSV export.

https://github.com/user-attachments/assets/bd2f64fa-b5e0-4d44-a8fe-baba990c61b0

## Implementation details

- feat: add the `scatter` type to `F0DataChart` — `ScatterChart` + `useScatterChartOptions`, skeleton, empty state, stories, MDX

  <details>
  <summary>Why it doesn't go through <code>buildBaseChartOptions</code></summary>

  That builder always pairs a category axis with a value axis, and a scatter
  needs value × value. It follows the heatmap pattern instead — the only other
  cartesian chart of the five that assemble their own options — calling
  `buildValueAxis` twice and reusing the shared legend, grid, tooltip and
  emphasis builders. Extracting a shared frame so three charts could delegate
  was considered and rejected as a larger refactor than this warrants; it stays
  available as a later cleanup covering all six non-delegating charts.
  </details>

- chore: let a value axis fit its data range via `scale`, and anchor its end labels

  <details>
  <summary>Both were needed to make a scaled axis render correctly</summary>

  ECharts defaults `scale` off, forcing every value axis through zero — right
  for a bar, but it collapses a plot whose values sit far from the origin. With
  it on, the axis bound lands exactly at the plot edge and the final tick label
  rendered half outside the canvas; `alignEdgeLabels` is the value-axis
  counterpart of the `edgeAligned` treatment `buildCategoryAxis` already
  applies. Both are omitted when unset, so bar and line serialise unchanged.
  </details>

- feat: route the tooltip through `renderValueTooltip`, adding an opt-in `size: "large"` to `ValueTooltipRow`

  <details>
  <summary>Why a row needs headline size</summary>

  A scatter point is two coordinates and no third measure, so there is no
  single headline value — but both numbers still deserve the emphasis every
  other chart's value gets. No other chart sets the flag and their output is
  unchanged.
  </details>

- feat: render scatter as a dashboard widget — config, `scatterSeries` data, shape detection, canonical conversions, skeleton, and tabular export for Excel/CSV

  <details>
  <summary>Why <code>scatterSeries</code> is its own field</summary>

  The two plausible reuses are both broken: `data` collides with the heatmap
  tuples, and `series` falls through to the bar branch, where `numericValue([x, y])`
  returns `0` for every point — silent total data loss. Detection keys on the
  new field's presence rather than its length, so an empty scatter still renders
  its own empty state.
  </details>

- feat: exclude scatter from the chart-type switcher — measure against measure has no category-axis equivalent, so converting either invents an x or discards one
- fix: hide the chart-type picker for scatter, whose only entry would be "Table"

  <details>
  <summary>Why hide it, and what that costs</summary>

  The group is `required`, so a picker whose only entry is "Table" lets the
  user switch to table view with no route back. Scatter converts to nothing —
  measure against measure has no category-axis equivalent — so that is the
  picker it would get.

  The consequence is that scatter has no in-place table view: `setViewMode`
  is only ever called from the picker, so `ChartTableView` is unreachable for
  it and the data is available as a table through Download → Excel/CSV only.
  Acceptable for a new type that never had it; a table route that isn't the
  type picker would be the fix if that changes.

  Gauge has the identical trap, but it also has that route today, so removing
  it is a change existing users would see — listed as a follow-up rather than
  bundled in here.
  </details>

- feat: add `ChatDashboardScatterChartConfig` and `ScatterComputation` to the chat-dashboard wire types

  <details>
  <summary>No aggregation on the computation</summary>

  Both axes are measures and each row is a point, rather than rows being
  grouped into categories — so unlike every other chart computation there is
  nothing to aggregate. Purely additive: f0 has no consumer that switches on
  either union.
  </details>

- test: 246 tests across the kit and pattern, including first-ever coverage for `chartDataAdapter` — the module where the data-shape collision lives

- fix: key tabular rows on stable ids so user-controlled axis names can't collide

  <details>
  <summary>Silent data loss, found in review</summary>

  Axis names were computed keys placed after the literal `Series` and `Label`
  ones, so a later key overwrote an earlier one. Two measures both called
  "Amount" exported the Y value twice and dropped X; an axis named "Series"
  destroyed the series name. Nothing threw. Likely rather than exotic on the
  chat path, where the names are LLM-generated. Fixed via the `keys` escape
  hatch `downloadHelpers` already documented and the collection export already
  used, threaded through the table view and all three export paths.
  </details>

- fix: contain an unrenderable chart config as one item's error state, reusing the shared `errorTitle` rather than adding a translation key

  <details>
  <summary>Why the error carries no message</summary>

  The guard first passed its own string, which meant a new
  `ai.dashboardItem.unsupportedChart` key — and any addition to
  `TranslationsType` is a compile break for consumers holding a complete
  dictionary. The message rendered as the description under the
  already-translated `errorTitle` ("Error loading data"), so dropping it costs
  one line of sub-text: `OneEmptyState.description` is optional and guarded, so
  an empty message renders nothing. The absent Retry button is what separates
  this state from a fetch failure. Reviewed alternatives — reusing
  `collections.emptyStates.error.description` (wrong namespace, and "while
  loading the data" is not what happened) or `dataChart.heatmapNotSupported`
  (means the responsive size fallback) — were both worse than no string.
  </details>

- fix: guard each tooltip coordinate, clamp `pointSize`, and reserve no skeleton legend by default

## Follow-ups, deliberately not here

- **Drop `type DashboardChartConfig = any`** in the host app's `widgetItemMappers`. Because `any` includes `undefined`, its exhaustive switch never raises TS2366 — so a widened wire union compiles clean and fails at runtime instead. Doing this converts every future widening into a build error. The guard above is f0's half of that; this is the other half.
- **Accept partial translation overrides.** Dropping the key sidesteps the break for this PR; it does not fix the mechanism. `I18nProviderProps` should take a deep-partial override type while `useI18n()` keeps returning the complete resolved dictionary, which makes every future f0 key additive forever. Same split needed for `AiChatTranslationsProviderProps` and `buildTranslations`. Per @zygisS22's proposal.
- **A stable persistence DTO behind the canvas actions.** The remaining real API contract: f0 passes its evolving render union into consumer-owned `dashboard.save`/`create`, so every new computation is a potential contravariant break. Also per @zygisS22's proposal, migrated additively rather than in place.
- **The same one-way "Table" trap on gauge.** Real, pre-existing, and out of scope here because fixing it removes a route users have today.
- **Unbounded point count.** `ScatterComputation` has no aggregation, so element count is row count, and `focus: "series"` restyles every point on hover. Fine at the ~450 of the `Dense` story; worth revisiting with `large`/`largeThreshold` or a required `limit` when real row counts are known.


siguenzaraul (approved):

zygisS22 (commented):
I checked the public-API warning against the exact head `6c8cd0c4`, the generated base/head declaration artifacts, and the current Factorial and factorial-agent consumers.

## Findings

### 1. The translation addition is a real compile-time consumer migration

`ai.dashboardItem.unsupportedChart` is newly required by `TranslationsType`. F0's recursive default merge keeps stale dictionaries safe at runtime, but consumers maintaining a complete typed dictionary still have to add the key.

Factorial is such a consumer: `frontend/src/modules/core/lib/f0-provider.tsx` supplies the full F0 dictionary and currently lacks this key. Its F0 translation check/type validation will fail when the package is bumped until the mapping is added.

Please call this out in the PR/release notes and coordinate the Factorial translation addition. This means the current claim that nothing outside F0 needs to change for a bump is too strong.

### 2. `CanvasActions` is technically breaking as a generic public callback contract, although the current Factorial implementation remains compatible

The new union is delivered by F0 into consumer-owned `dashboard.save` and `dashboard.create` callbacks. A consumer whose callback only accepts the previous dashboard config is no longer assignable once F0 may provide `ScatterComputation`.

I verified this with a strict TypeScript assignability probe using the API workflow's exact base/head `.d.ts` artifacts: old generic `CanvasActions` -> new `CanvasActions` fails. This matches the usual variance rule: widening what a package provides to a callback can break consumers.

However, Factorial's actual actions accept `items` and `fetchSpecs` as `unknown`; the same strict probe using that implementation passes. So this is a real semver/API-contract concern, but not an immediate Factorial compile break.

Please either treat/document this as an intentional breaking API change or adjust the public callback contract so adding a chart computation does not repeatedly break callback implementers.

### 3. The `F0CanvasPanelProps` reports are false positives

Here the widened computation union is an input supplied by the consumer to F0. Existing values accepted before are still accepted now. A strict old-props -> new-props assignment passes.

The warning is duplicated because the same symbol is exposed through both the `f0` and `ai` entrypoints; it is not two independent changes. The API checker should account for input variance and ideally group duplicate entrypoint reports.

## Consumer audit

- The five bot entries reduce to three underlying contracts: translations, `CanvasActions`, and `F0CanvasPanelProps`.
- Factorial's current Canvas action provider compiles against the new declaration.
- Factorial does not yet map scatter widgets, but this PR degrades an unsupported chart to a per-widget error instead of crashing the whole dashboard.
- factorial-agent's current dashboard contract cannot emit scatter charts, so merging this PR alone does not change existing dashboard behavior.
- I found no other indexed Factorial organization consumers of these affected Canvas types.

## Recommended design to keep future additions non-breaking

### A. Accept partial translation overrides; keep the resolved dictionary complete internally

The runtime already recursively merges consumer values over `defaultTranslations`, but the public input type still requires the entire dictionary. Align the type with that runtime contract:

```ts
type TranslationOverrides<T> = {
  [K in keyof T]?: T[K] extends string ? string : TranslationOverrides<T[K]>
}

type ResolvedTranslations = TranslationShape<typeof defaultTranslations>

type I18nProviderProps = {
  translations?: TranslationOverrides<ResolvedTranslations>
}
```

`I18nContext` and `useI18n()` should continue exposing the complete resolved type after merging defaults. If a product wants to enforce 100% locale coverage, keep that as a separate product-side completeness check rather than making every new F0 key a public API break. With this split, adding a default translation key becomes additive forever.

The same input/resolved distinction should be applied to `AiChatTranslationsProviderProps` and `buildTranslations`; otherwise the required full shape can keep leaking through other public exports.

### B. Separate the evolving render model from the stable persistence callback boundary

`ChatDashboardConfig` is an F0 render model and will naturally grow whenever a chart type is added. `DashboardCanvasActions` is a host persistence boundary. Passing the closed render union directly into that callback makes every new computation a potential contravariant break.

For a maintainable long-term contract, introduce a stable/open persistence DTO, for example:

```ts
type PersistableDashboardConfig = {
  title: string
  description?: string
  filters?: Record<string, unknown>
  navigationFilters?: Record<string, unknown>
  items: readonly unknown[]
  fetchSpecs: Record<string, unknown>
}
```

That matches what the current Factorial host actually needs: it serializes/persists the payload and already accepts `items` and `fetchSpecs` as `unknown`. F0 can keep the strongly typed closed union for rendering, validation, export, and internal exhaustiveness, while the host callback remains stable as chart kinds evolve.

Changing the existing callback to this open DTO immediately would itself be a type change, so the non-breaking migration for this PR should be additive:

1. Keep the existing `dashboard` actions and their legacy config contract unchanged.
2. Add an optional versioned/capability-scoped action boundary such as `dashboardV2`, accepting the stable persistence DTO.
3. Route scatter persistence only through the new boundary. If a host only supplies legacy actions, render scatter safely but disable or omit save/create for that unsupported config rather than passing it to an old callback.
4. Migrate Factorial to the stable boundary, then deprecate the legacy boundary. Remove it only in a future intentional major release.

An alternative is a discriminated capability union that preserves the legacy branch, e.g. legacy actions with `schemaVersion?: 1` and wider actions with `schemaVersion: 2`. F0 must narrow on the version before invoking the callback. Do not solve this with a bivariant callback or cast: that only hides the compiler warning while still allowing F0 to send a scatter payload to a consumer that cannot handle it.

### C. Make the API checker variance-aware and consumer-oriented

The checker should distinguish:

- package inputs: widening is compatible (`F0CanvasPanelProps.content`),
- package outputs/callback inputs: widening can break (`CanvasActions`), and
- the same symbol re-exported through multiple entrypoints: report once with the entrypoints listed.

Add compile fixtures for these three cases, plus a partial translation object. Those fixtures would protect the intended compatibility contract more reliably than a structural leaf diff alone.

### Suggested rollout order

1. Stabilize the F0 public boundaries above and land scatter with graceful fallback.
2. Adopt the partial translation input and stable persistence actions in Factorial; add the localized fallback text and scatter renderer.
3. Only then allow factorial-agent to emit scatter configs.

This keeps each deployment independently safe and prevents future chart additions from reopening the same translation and callback breakages.

## CI context

Build/types, lint, format, Storybook tests, Chromatic, accessibility, and coverage checks are green. The React unit job's only failure is the `non-scaling-stroke` icon test for `MaximizeHorizontal.tsx` and `MinimizeHorizontal.tsx`; the exact same failure is present on `main`, and this PR does not touch either file.

**Conclusion:** this is not five independent breaking changes and the current Factorial stack is not broken by the scatter addition. But the translation migration is real, and `CanvasActions` is a genuine generic semver concern, so I would update the release/PR claims before treating the package bump as entirely standalone.

igorsafonov-gif:
Thanks @zygisS22 — finding **#1 is resolved by removal rather than migration**, in `a6efd58`.

Since the only consumer of the string was the tile's error *description*, and the title above it (`ai.dashboardItem.errorTitle`, "Error loading data") is already translated, the guard now passes a message-less `Error`. `OneEmptyState.description` is optional and guarded by `{description && …}`, so nothing renders in that slot and there is no layout hole. The absent Retry button — already suppressed for this condition — stays the tell that separates it from a fetch failure.

I checked the existing dictionary for something reusable before deciding on no string: `collections.emptyStates.error.description` is the wrong namespace and says "while loading the data", which is not what happened; `dataChart.heatmapNotSupported` means the responsive size fallback. Both were worse than nothing.

**Net effect: the bump needs no translation work.** `TranslationsType` is unchanged against `main`, so the compile-blocking consumer migration is gone, and the remaining Factorial change is only the non-compile-blocking `case 'scatter'` in `widgetItemMappers`.

To be clear about what this does *not* fix: your point stands that the mechanism is the problem, not this key. The next f0 key hits the same wall. Your partial-override design (deep-partial input type, complete resolved type out of `useI18n()`, same split for `AiChatTranslationsProviderProps` and `buildTranslations`) is filed as a follow-up, as is the stable persistence DTO for `CanvasActions` — I've kept both out of this PR rather than reshaping public boundaries inside a chart-type addition, but I'd rather they land before the *next* one.

PR description updated: the API-surface section now reads 3 contracts → 1 resolved, 1 real (`CanvasActions`), 1 false positive (`F0CanvasPanelProps`).

zygisS22 (approved):
🚀

igorsafonov-gif:
@zygisS22 — correcting the scope of my own claim above, with evidence that strengthens your finding #1 rather than closing it.

I said "the bump now needs no translation work." That is true of *this PR's* key and misleading about the bump as a whole. I took the actual 5.0.1 → 6.2.0 bump through Factorial in factorial#108776, and it fails to typecheck:

```
f0-provider.tsx(1059,11): error TS2741: Property 'tooltip' is missing in type
  '{ heatmapNotSupported: string; ... }' but required in type 'TranslationShape<...>'
```

`dataChart.tooltip` (six keys, one interpolating `{{stage}}`) was added by other PRs in the same window. So removing `unsupportedChart` here removed **one instance** of the break, not the break — the bump needed a translation migration anyway, for keys nobody flagged, because no API-surface warning fires on the consumer's side of a `TranslationsType` addition.

That is your point demonstrated on a real bump within 24 hours of the release. Two independent additions, two compile breaks, one of which was only found by running the bump. Your section A is not a nice-to-have refactor — it's the thing that stops this recurring on every f0 release. I'd support landing it ahead of the next chart-type addition, and I'm happy to take it if you'd rather not.

One implementation note on your sketch, from having just written the mirror by hand: `useI18n()` must keep returning the complete resolved type, because consumers like Factorial's provider are generated *from* the shape — a deep-partial leaking into the resolved type would silently make every `translations.x.y` access optional at call sites inside f0. Your `TranslationOverrides<T>` / `ResolvedTranslations` split gets this right; just flagging that the two must not be collapsed into one alias for convenience.

Finding #2 (`CanvasActions`) is untouched by any of this and remains filed as-is.

PR #4963, opened 2026-08-03
feat(F0Coachmark): add experimental component

warcos-fact:
## Description

Adds `F0Coachmark`, a controlled, dismissible panel anchored to an element, used to point out a feature the user has not discovered yet. It carries a title, optional description, a single right-aligned CTA, an optional step indicator for walkthroughs, and an optional arrow pointing at its target. Built on Popover rather than Tooltip on purpose: `role="tooltip"` content is hover/focus driven, so a CTA and a dismiss button inside it are unreachable by keyboard — focus leaves the trigger and the tooltip closes.

## Type of change

- [x] New component (aligned in #f0-support, lands in `experimental/`)
- [ ] Component enhancement / variant (existing component, no breaking change)
- [ ] New pattern (composition of existing F0 components)
- [ ] Bug fix
- [ ] Documentation only
- [ ] Other:

- Linked #f0-support thread: **not linked yet — needs the alignment thread before merge**
- Phase exit criteria met: MDX docs page, 13 unit tests, consolidated Chromatic snapshot covering all four arrow sides

image

## Implementation details

- feat: add `F0Coachmark` under `experimental/Overlays/`, controlled via `open` / `onDismiss`, with `title`, optional `description`, a single `action`, optional `step`, and `arrow` on by default

- feat: optional `step: { current, total }`, rendered as `1/3` at the left of the action row

  <details>
  <summary>Why presentational only</summary>

  The component sequences nothing itself — it renders the position it is given. The consumer keeps control of what each step shows and when it advances, which keeps the coachmark a primitive rather than a tour engine.
  </details>

- feat: reuse the library's shared floating-dark-panel surface instead of inventing one

  <details>
  <summary>Which surface, and why it matters for the buttons</summary>

  Same recipe as `F0Toast` and `F0ActionBar`: `bg-f1-background-inverse` with a `dark:bg-f1-background-tertiary` branch, `text-f1-foreground-inverse`, `shadow-lg`, `backdrop-blur-sm`. Verified to resolve to the identical token as `F0Toast` in both themes.

  The important consequence is that the surface is dark in **both** themes rather than inverting. That means the content can sit in a plain `dark` wrapper — the pattern both of those components already use — so every control inside resolves the tokens that suit a dark surface on its own. Both buttons are therefore **stock variants carrying only layout classes**: no colour overrides anywhere in this component.

  The trade is dark mode: the panel is a lifted dark surface at 1.13:1 against the page, held off it by shadow and blur, rather than becoming light. That is what the other two look like today.
  </details>

- feat: expose `PopoverArrow` and the `PopoverContentProps` type from `@/ui/popover`

  <details>
  <summary>Arrow implementation, and one non-obvious fix</summary>

  No component in the library rendered a Radix popper arrow before this one, so the wrapper never re-exported it.

  The arrow's fill uses the panel's surface tokens, alpha included, so both composite over the same backdrop to the same colour. It deliberately does **not** overlap the panel: on a translucent surface an overlap stacks the two alphas and draws a darker line along the arrow's base.

  It also needs `overflow-visible` on the content, which is load-bearing rather than cosmetic. `PopoverContent` ships `overflow-auto`, and `backdrop-blur` makes the panel a containing block for absolutely-positioned descendants — which pulls the arrow inside that scroll box and clips it away entirely. A coachmark is a few lines tall and never needs to scroll.

  Verified on all four sides, including the flip that happens on viewport collision.
  </details>

- refactor: title keeps the panel's colour while description and step drop to `text-f1-foreground-inverse-secondary`, the same title/description pairing `F0Toast` uses

- refactor: the action is right aligned rather than full width, via `ml-auto` on the button rather than `justify-end` on the row, so it stays right aligned whether or not a step is present

- test: 13 unit tests covering render, dismissal via close button and Escape, no-dismiss on outside click, focus moving to the panel rather than the dismiss button, focus restoration, the arrow toggle, and the surface/`dark`-wrapper pairing

- docs: add the MDX documentation page, switch stories to `!autodocs`, and anchor every example at the top of its canvas so the portalled panel stays inside the example instead of covering the text beneath it

## Verification

Contrast measured from computed styles in both themes, since Storybook cannot exercise dark mode (see Notes):

| | light | dark |
|---|---|---|
| title vs panel | 18.1:1 | 18.1:1 |
| CTA label vs panel | 18.1:1 | 18.1:1 |
| dismiss icon vs panel | 18.1:1 | 18.1:1 |
| panel surface | identical token to `F0Toast` | identical token to `F0Toast` |

Also verified: arrow fill matches the panel exactly in both themes; each docs example's panel sits inside its own canvas (45–83px of slack); the action stays right aligned with and without a step.

## Notes for review

- **The tooltip padding fix that was originally bundled here has been split out into #4992.** It was unrelated to this component and touched all 15 tooltip call sites, so it is easier to review and to revert separately. This branch was rebased to drop it; `ui/tooltip.tsx` is no longer in this diff.
- **`ui/Action/variants.ts` is untouched.** An earlier iteration changed the `outline` variant to be surface-relative; it was reverted. Verified the library's own outline buttons still compute their original values, so nothing outside this component moves.
- **Accessibility**: the panel is a non-modal `dialog` named by `title` and described by `description`. Opening moves focus to the panel itself rather than the dismiss button, so a screen reader announces the message and Enter does not immediately discard it. Radix restores focus to a Trigger on close, but a coachmark anchors rather than triggers, so focus restoration is handled explicitly.
- Outside clicks deliberately do **not** dismiss — a coachmark stays until acknowledged, otherwise the close button is pointless.
- **Storybook does not apply a `dark` class to the story DOM** (`ThemeProvider` only sets React context), so Chromatic captures light mode only. Dark mode here was verified by measurement, not by CI. Worth fixing separately — it is the main gap for the follow-up below.
- **Unrelated pre-existing failure on `main`**: `src/icons/__tests__/noNonScalingStroke.test.ts` fails because `icons/app/MaximizeHorizontal.tsx` and `MinimizeHorizontal.tsx` ship `vector-effect="non-scaling-stroke"`. Those icons landed in c0c0c5780 and are byte-identical to `main` here, so this is not from this branch — but CI will be red until it is fixed.
- **Follow-up, tracked separately**: button variants resolve colour against the page theme rather than the surface they sit on. That is why `F0Toast`, `F0ActionBar` and now this component all need a hardcoded `dark` wrapper, and why none of them can have a light panel in dark mode. `outlinePromote` and ~24 icon compound variants carry the same assumption. A surface-relative prototype (`bg-transparent text-current` + `ring-current`) was measured correct in all four contexts with zero call-site overrides, but it changes 141 existing `variant="outline"` call sites and has no dark-mode CI to verify it, so it is out of scope here.


desiree-np:
Nice work on this. Leaving some design feedback and one technical note to align on before we lock the implementation, more to discuss than a request for changes.

1. **The full-width CTA isn't a card pattern in F0.** We don't use full-width buttons in any card. A natural-size button aligned to the right reads more in line with the system.

2. **There's no hierarchy between the title and the body.** They're the same size, and on the panel's surface the weight difference is barely perceptible, so the title doesn't lead. The body should step down in color, not just in weight.

3. **The full-width button doesn't scale.** It leaves no room for a button bar (Back / Next) or for secondary info like a step counter ("1/3"). A footer, secondary on the left, action on the right, handles those. And it matters because the docs already describe a multi-step use case ("guided tour… queue them and show the next after the previous is dismissed"): a single full-width CTA can't express that sequence.

I'm attaching an example from an exploration I did a while back, it doesn't have to look exactly like this, but it shows the title/body hierarchy, the natural-size button, and the footer with the step counter.

Screenshot 2026-08-04 at 14 41 12

4. **The tooltip padding fix is bundled into a new-component PR.** Evening out the shared tooltip surface (`px-2 py-1.5` → `p-2`) is a fine change on its own, but it's a change to Tooltip — unrelated to the new Coachmark — and it touches all 15 tooltip call sites. Worth splitting into its own small PR: easier to review, and easier to revert the Coachmark later without dragging every tooltip along with it.

packages/react/src/experimental/Overlays/F0Coachmark/types.ts
```diff
@@ -0,0 +1,58 @@
...
+   * consumer owns visibility so it can decide when to show it and persist the
+   * dismissal.
+   */
+  open: boolean
+  /**
+   * Called when the user dismisses the coachmark, either with the close button
+   * or the Escape key. Clicking outside does **not** dismiss it — a coachmark
+   * stays until it is explicitly acknowledged.
+   */
+  onDismiss: () => void
```
sauldom102: I wouldn't have it fully controllable with an `open` prop, I would even get rid of this `open` prop at all and would be more prescriptive. `onDismiss` is okay to have it in case we want to track when it's actually dismissed/closed but it should auto-close when we press the close button no matter if we pass this `onDismiss` or not.

We could have a similar approach to what we do with toasts, something like:

import { coachMarks } from "@factorialco/f0-react"

coachMarks.open({ id: "coachmark-1" })

Or even:

coachMarks.open({ targetElement: "#..." | "whatever css selector that returns exactly one element", title, description, action })

Same with steps, we should receive a list of steps if we want { title, description } | { steps: { title, description} }

warcos-fact: Done — the `open` prop is gone and there is no component to render anymore. `coachmarks.open()` mirrors `toasts.open()`:

```tsx
coachmarks.open({
  targetElement: "#filters-button", // selector matching one element, or the element
  title: "Filters got smarter",
  description: "Stack filters, then save the combination as a view.",
  action: { label: "Learn more", onClick: () => openDocs() },
})
```

The close button and Escape always close it; `onDismiss` is optional and only reports it, plus `onComplete` for "reached the end". Steps are a list and the component sequences them — the indicator counts up, the action reads `Next` until the last step then `Got it`, and each step can name its own `targetElement`. Only one is on screen: a second `open` queues, and the same `id` replaces in place.

One naming call: I went with `coachmarks` rather than `coachMarks` so it matches `F0Coachmark` and `toasts` — happy to switch.

sauldom102 (approved):

PR #4964, opened 2026-08-03
fix(F0AvatarList): make the +N popover keyboard reachable

albertpmz:
## Description

The names collapsed behind the `+N` counter were reachable by mouse only. This makes the counter a real disclosure button and moves its popover from `HoverCard` to `Popover` so the list can be scrolled from the keyboard. With that, `AvatarList` clears the mechanical Definition of Done and leaves the debt list.

This PR owns **all** AvatarList work — behaviour, stories, docs and tests. Its sibling #4962 does the other four Avatars and deliberately contains no AvatarList files, so the two are independent and can be reviewed in either order.

There were two independent breaks, and only one was visible to axe. The `+N` counter was a role-less `<div>`, so it never entered the tab order and the popover could not be opened by keyboard at all — axe cannot detect that. Its content was also a `ScrollArea` inside a hover card, which axe *did* flag as `scrollable-region-focusable` once a play function opened the card. Removing the scroll region alone would have turned the gate green and left the feature mouse-only.

<details>
<summary>Why the scroll region could not simply be made focusable, and why Popover</summary>

`ui/scrollarea.tsx` already sets `tabIndex={0}` on the Radix viewport — that is what makes `ScrollArea` compliant in the other 22 places it is used. Radix `HoverCardContent` then rewrites every tabbable node it contains to `tabindex="-1"`, on every render (an effect with no dependency array). That is the library's contract, not a bug: a hover card is meant to be read, not operated. So a fix inside `ScrollArea` is silently undone on the next render.

The list does need to scroll — 63 collapsed entries are 1936px of content — so the container had to change. `Popover` preserves tab stops, which was verified rather than assumed: with `Popover` + `ScrollArea` the viewport keeps `tabindex="0"`, takes focus and scrolls, and axe passes. Leaning on `ui/popover.tsx`'s own `overflow-auto` instead of a `ScrollArea` was also measured, and axe rejects it — the `ScrollArea` is load-bearing.
</details>

## Implementation details

- fix: make the `+N` counter a `<button>`; Radix supplies `aria-expanded` and `aria-haspopup`
- fix: move the popover from `HoverCard` to `Popover` so its scroll region is keyboard-operable
- fix: keep hover-to-open, with focus untouched

  <details>
  <summary>Three interaction bugs this surfaced</summary>

  Swapping the container means replacing one library's focus model by hand, and each of these was found by driving a real browser:

  1. **Click closed the card instead of pinning it.** `PopoverTrigger` has its own toggle `onClick`, and a pointer click is preceded by `pointerenter` — so hover opened it and the toggle immediately closed it. Radix composes handlers with `checkForDefaultPrevented`, so the handler calls `preventDefault()` when the card is already open and focuses the content instead.
  2. **Hover stole focus.** `Popover` moves focus into its content on open, which is right for Enter/click and hostile on hover. `onOpenAutoFocus` is suppressed for the pointer path.
  3. **Hover left the counter focus-ringed.** Radix returns focus to the trigger on close, so a single hover-and-leave left the counter looking as though it had been tabbed to. `onCloseAutoFocus` is suppressed for the pointer path too.

  A 150ms close grace period is also reimplemented, since `HoverCard` provided it for free and WCAG 1.4.13 requires hover-triggered content to stay reachable as the pointer travels onto it.
  </details>

- fix: give the `xs` counter a text alternative — it renders an ellipsis icon with no text, so the count was unavailable to screen readers (WCAG 1.1.1). An `sr-only` span rather than `aria-label`, because the non-disclosing counter is a role-less `<div>` where ARIA prohibits it
- fix: name the counter with the count itself, so the accessible name matches the visible label at every size (WCAG 2.5.3)

  <details>
  <summary>Why no translation key</summary>

  An earlier revision added an `avatarList.showMore` entry so the name could read "+12 more". That would have been a breaking change: `TranslationsType` is a complete shape, `I18nProvider` is publicly exported, and adding a key stops any consumer maintaining a full translations object from compiling. The count alone is a sufficient name — it is exactly what a sighted user reads — so the dictionary is untouched.
  </details>
- chore: deprecate `tooltipScroll` with `@removeIn 5.0` rather than removing it — it is public API on a stable component. It no longer has any effect and is no longer threaded through `F0AvatarList` or the `value-display` adapter. Call sites that still pass it are left alone so this PR stays inside the avatars/value-display ownership boundary
- test: add six unit tests covering the button role and name, Tab reachability, Enter-to-open, click-to-pin, focus staying put across a hover cycle, and the scroll region being present and focusable
- docs: rewrite the Accessibility section — two limitations remain rather than five
- chore: remove `Avatars/AvatarList` from `stable-dod-debt.json`

## Behaviour changes to review

Visual output is unchanged — that was verified by measuring both builds, not by eye:

| | before | after |
| --- | --- | --- |
| card | 172x200, `rgba(2, 11, 29, 0.92)`, 10px radius, no border or padding | identical |
| counter | 40x32, `rgba(5, 38, 87, 0.06)`, pill radius | identical |
| scroll viewport | 172 visible of 392, `tabindex="-1"` | 172 of 392, **`tabindex="0"`** |

What *does* change is keyboard and click behaviour:

- **the card opens on Enter, not on focus.** Unavoidable if the scroll region is to be reachable: `Popover` moves focus into its content, and that content is portalled to the end of `<body>`, so suppressing the focus move would leave the scroll unreachable by Tab. Hover is unchanged for pointer users.
- **clicking the counter no longer toggles it closed** when it is already open from hover. Escape and an outside click dismiss.

## What is deliberately not fixed

Two limitations remain, documented rather than papered over, and axe can see neither: the cluster is still silent to screen readers (every avatar is `aria-hidden`), and per-avatar tooltips are still hover-only.

Also out of scope: **the same defect exists in four other `HoverCardContent` call sites** — `F0TagList`'s `TagCounter` is the same `ScrollArea`-in-a-hover-card shape, `F0TableOfContentPopover` is worse (scroll region plus collapsible sections and drag-to-reorder, all unreachable), and `OneTable/TableHead` has an action button inside a card styled with a `focusRing()` that can never render. Those need per-call-site migration and are worth raising as shared work.

## Public API

The repo's API-surface check flagged two things on an earlier revision; both are now gone rather than argued away:

- **a translations key was added** — removed, see above. No consumer has to update a translations object.
- **`F0AvatarList`'s emitted signature changed**, because dropping `tooltipScroll` from the destructuring rewrites the `.d.ts`. The prop still exists in `F0AvatarListProps` and still typechecks, so this was pure churn. The binding is kept under its exact name with a `void` discard, and the destructuring pattern is now byte-identical to `main`.

## Verification

`tsc`, `oxlint`, `oxfmt --check`, 165 unit tests across `avatars` and `value-display`, and the Storybook test-runner against a built Storybook.

The axe result was checked in both directions rather than assumed: `OverflowPopover` passes at `test: "error"`, and restoring the scroll region inside a hover card makes the identical pipeline fail on `scrollable-region-focusable`. Every new unit test was mutation-tested — reverting the trigger to a `<div>` fails four, restoring the `ScrollArea` fails one, and removing the `onCloseAutoFocus` guard fails the hover-focus test.

### DIFF
--- a/packages/react/.scripts/stable-dod-debt.json
+++ b/packages/react/.scripts/stable-dod-debt.json
@@ -10,7 +10,6 @@
     "Avatars/AvatarFile",
     "Avatars/AvatarFlag",
     "Avatars/AvatarIcon",
-    "Avatars/AvatarList",
     "Avatars/AvatarModule",
     "Avatars/AvatarPerson",
     "Avatars/AvatarTeam",
--- a/packages/react/src/components/avatars/F0AvatarList/F0AvatarList.tsx
+++ b/packages/react/src/components/avatars/F0AvatarList/F0AvatarList.tsx
@@ -33,8 +33,15 @@ export const F0AvatarList = ({
   noTooltip = false,
   remainingCount: initialRemainingCount,
   max,
+  // Deprecated and intentionally ignored — see `F0AvatarListProps`. Kept in the
+  // destructuring, under this exact name, because the pattern is emitted
+  // verbatim into the public .d.ts signature of `F0AvatarList`: dropping or
+  // renaming it reads as a breaking public API change for a prop that still
+  // exists and still typechecks. `void` below satisfies `noUnusedLocals`.
   tooltipScroll,
 }: F0AvatarListProps) => {
+  void tooltipScroll
+
   // Check legacy size
   if (size && !avatarListSizes.includes(size)) {
     const sizesMappingList: Record<string, AvatarListSize> = {
@@ -135,7 +142,6 @@ export const F0AvatarList = ({
             size={size}
             type={type === "person" ? "rounded" : "base"}
             avatarType={type}
-            tooltipScroll={tooltipScroll}
             list={
               initialRemainingCount
                 ? undefined
--- a/packages/react/src/components/avatars/F0AvatarList/__stories__/F0AvatarList.mdx
+++ b/packages/react/src/components/avatars/F0AvatarList/__stories__/F0AvatarList.mdx
@@ -1,5 +1,8 @@
-import { Canvas, Meta, Controls, Unstyled } from "@storybook/addon-docs/blocks"
+import { Canvas, Meta, Controls } from "@storybook/addon-docs/blocks"
 import * as ListStories from "./F0AvatarList.stories"
+import { DoDonts } from "@/lib/storybook-utils/do-donts"
+
+import { F0AvatarList } from "../F0AvatarList"
 
 <Meta of={ListStories} />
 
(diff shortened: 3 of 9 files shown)


eliseo-juan (approved):
LGTM

PR #4975, opened 2026-08-04
feat(BaseBanner): [FCT-60673] add vertical card variant

fleveque:
## Description

Adds a `card` variant to `BaseBanner`: a vertical layout — media on top, content below — that keeps its stack at every viewport, for narrow surfaces such as popovers and side panels.

The existing `default` and `full-width` variants go horizontal at the `sm` breakpoint. Since that is a *viewport* breakpoint rather than a container query, they cannot be used in a narrow container on a wide screen — media and text end up crammed side by side. `card` is the same banner with that behaviour removed.

Driven by a concrete consumer: the `one_announcement` in-app communications widget ([factorial#108764](https://github.com/factorialco/factorial/pull/108764)), a 308px card announcing a One AI capability. It is an **extension of an existing component**, not a new one — per the contribution skill, "extend before you create".

## Type of change

- [ ] New component (aligned in #f0-support, lands in `experimental/`)
- [ ] New pattern (composition of existing F0 components)
- [ ] Bug fix
- [ ] Documentation only
- [ ] Other:

Backward-compatible addition to an experimental component, so per the Definition of Done this is a light thread rather than an up-front design review. `BaseBanner` was previously untagged (`apiStatus: "unknown"`); this PR declares it `experimental`, which is what it has effectively always been.

- Linked #f0-support thread: [AiAnnouncementCard component](https://factorialteam.slack.com/archives/C082ZNKS403/p1786343125188429) — raised 10 Aug. This PR is **the alternative** in that thread; [#5038](https://github.com/factorialco/f0/pull/5038) is the same card as its own component and is the preferred route. One of the two gets closed once Foundations decides.
- Phase exit criteria met: story tagged `["autodocs", "experimental"]` ✅ · unit tests covering the public API ✅ (the component had none before) · play function on the primary flow ✅ · axe passes with no new skips ✅ · MDX documentation ❌ — see Known gaps

[Link to Figma Design](https://www.figma.com/design/VTVKWL9OGmnTJDPOmQmVhI/One-AI-Kit?node-id=12559-19565) 

The card at its designed width — 308px, media on top, actions in a row, dismiss in the corner:

f0-card-CURRENT-default

And in its first consumer ([factorial#108764](https://github.com/factorialco/factorial/pull/108764)) — floating below the One toggle while the panel is closed, and as the assistant overlay while it is open. These are the two narrow surfaces the variant exists for:

| One closed | One open |
| --- | --- |
| in-situ-CURRENT-one-closed | in-situ-CURRENT-one-open |

## Implementation details

One `isCard` flag gates every branch, so `default` and `full-width` emit exactly the class strings they did before — the diff is additive by construction.

- **Layout:** `gap-0` and no `sm:flex-row`, so the vertical stack survives every viewport.
- **Media:** uniform `p-1` instead of the horizontal variants' asymmetric padding.
- **Title:** `font-medium text-lg` — f0's `lg` token is `1rem / 1.5rem`, the 16/24 in the Figma frame.
- **Subtitle:** `line-clamp-2`, so a long CMS-authored description cannot unbalance the card.
- **Actions:** `gap-2`.

Consumers compose the actions through `children` rather than `primaryAction`/`secondaryAction` where they need a button variant outside `BannerAction`'s `default | outline | ghost` — factorial's announcement does this to keep its "open One" button on the `ai` variant.

### Testing

- **Unit (Vitest):** first tests for this component — render, both actions, close behaviour, and a two-sided assertion that `card` never gets `sm:flex-row` while `default` still does.
- **Play function:** the card's flow end to end — readable, primary action fires, close notifies and removes it.
- **Chromatic:** one consolidated `Snapshot` story covering both orientations, since Chromatic bills per snapshot.

### Known gaps

- **No MDX documentation.** `BaseBanner` has never had any, and the Definition of Done asks for at least the *Acceptable* tier at Phase 3. Not introduced by this PR, and not mechanically gated while the component is experimental (`check-stable-dod.ts` only enforces stable-tagged components), but it blocks promotion. Happy to add it here if reviewers would rather not carry it forward.
- **`bg-white` and the border are dead classes.** Worked around by scoping `bg-f1-background border-solid` to `card`, to avoid changing shipped visuals in this PR. Full explanation and the real fix in **#5022** (draft, stacked on this branch).

### DIFF
--- a/packages/react/src/kits/ai/Banners/BaseBanner/__tests__/BaseBanner.test.tsx
+++ b/packages/react/src/kits/ai/Banners/BaseBanner/__tests__/BaseBanner.test.tsx
@@ -1,0 +1,106 @@
+import { describe, expect, it, vi } from "vitest"
+import "@testing-library/jest-dom/vitest"
+import { zeroRender as render, screen, userEvent } from "@/testing/test-utils"
+
+import { BaseBanner } from "../index"
+
+const baseProps = {
+  title: "Submit expenses in seconds",
+  subtitle: "Upload receipts. One organizes everything for you.",
+  mediaUrl: "https://example.com/media.png",
+}
+
+/** The element carrying the layout classes is the banner root. */
+const getRoot = () => document.querySelector("div.shadow-md") as HTMLElement
+
+describe("BaseBanner", () => {
+  it("renders title, subtitle and media", () => {
+    render(<BaseBanner {...baseProps} />)
+
+    expect(screen.getByText(baseProps.title)).toBeVisible()
+    expect(screen.getByText(baseProps.subtitle)).toBeVisible()
+    expect(
+      screen.getByRole("presentation", { hidden: true })
+    ).toBeInTheDocument()
+  })
+
+  it("renders both actions and calls their handlers", async () => {
+    const onPrimary = vi.fn()
+    const onSecondary = vi.fn()
+    const user = userEvent.setup()
+
+    render(
+      <BaseBanner
+        {...baseProps}
+        primaryAction={{
+          label: "Try it out",
+          onClick: onPrimary,
+          variant: "outline",
+        }}
+        secondaryAction={{
+          label: "Not now",
+          onClick: onSecondary,
+          variant: "ghost",
+        }}
+      />
+    )
+
+    await user.click(screen.getByRole("button", { name: "Try it out" }))
+    await user.click(screen.getByRole("button", { name: "Not now" }))
+
+    expect(onPrimary).toHaveBeenCalledOnce()
+    expect(onSecondary).toHaveBeenCalledOnce()
+  })
+
+  it("calls onClose and removes itself when dismissed", async () => {
+    const onClose = vi.fn()
+    const user = userEvent.setup()
+
+    render(<BaseBanner {...baseProps} onClose={onClose} />)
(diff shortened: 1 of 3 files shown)


fleveque:
Follow-up on the `bg-white` / `border-solid` workaround in this PR: **#5022** (draft, stacked on this branch).

To be explicit about what the comment in `index.tsx` is describing — both class names are dead, not merely wrong:

- `bg-white` is **not a class f0 emits at all.** `packages/core/src/tokens/colors.ts` defines `white` as numeric steps (`white-3` … `white-100`) with no `DEFAULT` key, so Tailwind generates `bg-white-100` but never a bare `bg-white`. `default` and `full-width` have been rendering with no background.
- `border` never draws because `corePlugins.preflight` is `false` in `packages/core/tailwind.ts`, so Tailwind's base reset — which sets `border-style: solid` globally — never lands. Hence the explicit `border-solid`.

Scoping the workaround to `card` here is deliberate and I think correct for this PR: fixing it properly changes how shipped banners look, which is a separate decision and wants a designer's eye. #5022 does that and calls out the expected Chromatic diff.

It also lists four other `bg-white` usages elsewhere in the repo with the same problem, which are left alone as they belong to other owners.

PR #4983, opened 2026-08-04
perf(F0Graph): cut the remaining F0GraphView render rate during gestures

factmarc10010:
## What

Two changes, plus the instrumentation used to find them (added and removed within the branch, so the tip is clean). Rebased onto `main` now that **#4978** has landed, so the diff is only the new work — three files.

#4978 cut the node re-render cascade. That left `F0GraphView` itself as the dominant React work: ~50 renders/s during a gesture, only a third of them explained.

## What the instrumentation established

A props-identity counter answered the question that decided where the fix belonged. React allocates a fresh props object when the *parent* renders and reuses it when a component re-renders from its own state, so:

| | pan | zoom |
|---|---|---|
| `F0GraphView` renders | 180 | 104 |
| props identity changed | **2** | **0** |
| windowing recomputes | 76 | 42 |

The renders are **internal**, not parent-driven — so this is F0Graph's to fix, not OneDataCollection's. And render count tracked windowing recomputes at a fixed **~2:1** ratio across every run, with no other tracked value contributing. That is what made the quantize step the lever rather than a guess.

## The changes

**`17e7434cb` — quantize step 200 → 400.** Halves cell crossings, and with them both the O(visible-tree) intersection and the renders they trigger. Measured: recomputes 26/s → 9/s, renders 52/s → 18/s, with the 2:1 ratio unchanged (2.01 → 2.06).

Safe by construction: the rect is snapped *outward*, so it always covers the viewport plus the padding, and a larger step only widens it. The cost is over-inclusion — the region grows by up to one step per side, so more nodes are materialized between recomputes.

The four rect tests move to the new grid. Their expectations stay spelled out rather than derived from the constant, so retuning the step fails them loudly instead of silently agreeing with the implementation. Added the step-agnostic property they were missing: the rect must always contain the viewport grown by the padding, checked across five cameras (including non-integer zooms) and four paddings. Without it, a node could sit inside the padding band — the band that exists so panning reveals already-materialized nodes — and still be windowed out.

**`ec245aa30` — stop writing unchanged viewport state every frame.** `handleViewportChange` called `setViewportReady(true)` and `setCurrentZoom(vp.zoom)` unconditionally on every camera frame. Both are now guarded by refs.

**This one produced no measured improvement.** It was made on the theory that React re-runs the component body once before bailing out of a no-op state write; the instrumentation refuted that — React was bailing out correctly. It is kept as a defensible micro-optimization (fewer setState calls, no behaviour change: `viewportReady` still flips exactly once and `currentZoom` still tracks every real zoom change), but it should not be credited with any of the numbers here.

## End-to-end result

Frame times during a drag, `main` **as it was before #4978 landed** versus this branch, no CPU throttling:

| | main | this branch | |
|---|---|---|---|
| median frame | 8.4 ms — but 16.5 ms in 2 of 7 drags | **8.3 ms in all 12** | |
| **p95 frame** | **25 ms** | **9 ms** | −64% |
| worst frame seen | 108 ms | **33 ms** | |
| frames over 50 ms | 4 in 740 | **0 in 1547** | |

Zero frames over 50 ms across 1547, including a 426-frame drag whose worst frame was 17 ms.

## What this does NOT establish

**The step change's individual contribution is not isolated.** This table compares `main` against #4978 + #4983 together. Pointing the same build at #4978's tip and re-measuring would separate them; until that is done, the case for `17e7434cb` rests on its counter numbers (65% fewer recomputes and renders) rather than on frame times. If it turns out to add nothing, reverting it is the better outcome — it trades a real cost (more nodes materialized per recompute) for it.

**Under 4× CPU throttling the two builds are indistinguishable on long frames** — ~7% of frames over 50 ms on both, medians 7.4% vs 6.4%. That refuted the model that each windowing recompute costs a dropped frame: a build with 65% fewer recomputes had the same long-frame rate. Something else causes those frames, it is present on `main` equally, and it is the actual remaining jank on slower hardware. The way in is profiling a single long frame, not reasoning about mechanisms — five hypotheses died that way over the course of this work.

**Nothing here was measured on a large org chart.** The test tree has ~530 visible nodes and 11–21 materialized. The intersection and layout scale with the full tree, so a several-thousand-employee chart is a different regime.

## Testing

249 F0Graph tests pass (28 files) at every commit; `tsc --noEmit` clean for F0Graph; `oxfmt` clean.


albertpmz (approved):
LGTM! Great work on optimizing this 💪🏼

PR #4984, opened 2026-08-04
feat(F0DataChart): make a dense horizontal bar chart readable

igorsafonov-gif:
## Description

A dense horizontal bar chart — 29 workplaces × 3 gender series, straight out of a One Analytics report — was unreadable in a dashboard widget: hairline bars, truncated names, and a scale nobody could use. This finishes the work started by the thickness floor and expanded view, covering the three things such a chart needs to be read: **room for its rows**, **its numbers on the bars**, and **its category names in full**.

Stacked on #4948 (`feat/chart-tooltip-unified`), so it's a sibling of #4950. Review that one first.

## Type of change

- [x] Bug fix

`F0DataChart/Bar` → **Horizontal** story, unchanged fixture, so this is purely the rendering change. Full names instead of `Sarah Johns…`, no value-axis ticks, thicker rows, and the plot reaching the end of the data.

chart-expanded-show-less
chart-stacked-totals-windowed
chart-grouped-windowed-long-labels

### Rows and framing

- **The row window is fixed.** `disabled` on the `inside` `dataZoom` drops drag-to-pan and wheel-pan. Panning duplicated the widget's own "show all" link as a hidden gesture, and wheel-panning swallowed the scroll of the page the chart sits on.
- **Expanded bars go to 24px**, and an expanded widget now grows with its chart instead of clipping it at the fold. The card carried `min-h-full` so a tall chart could push past the viewport, but a percentage `min-height` *replaces* a flex item's content-based automatic minimum, so flexbox was free to squeeze the card back — its border closed at 662px around 964px of content. `shrink-0` is what makes `min-h-full` stick.
- **"Show less"** appears where the "Show all" link was, so the control that got the reader into the expanded view takes them back out. Derived from props rather than remembered: expanding swaps `DashboardGrid` to a different tree, so `ChartItem` unmounts and any latched state is gone by the time the expanded view renders.

### Numbers

- **Stacked horizontal bars carry their category total** beside the bar. ECharts places a stacked series' label at that series' own position in the stack, so labelling the *end* of the whole bar takes something stacked there: a zero-value, transparent ghost that adds no length, kept out of the legend and the tooltip. Skipped where a total means nothing — a single series (the total would restate the segment label) or a category mixing signs (the parts don't sum to the length you see). Same rule as the tooltip's existing `showTotal`, so the two never disagree.
- **Stacked segment labels get 6px of fit padding** instead of 12. At 12 the height gate needed `labelHeight ≤ thickness − 24`, which no bar at any current thickness could satisfy, so those labels were silently dropped.
- **A labelled horizontal chart drops its value-axis ticks and takes its axis maximum from the data.** The reader takes the number off the bar, so the ticks only compete with it — and ECharts' rounded-up maximum (150 for a 106 max) spent ~29% of the plot on a scale that is no longer drawn. Grid lines stay: they're driven by `showGrid`, and they're all the axis was still contributing. Unlabelled charts keep their ticks — there, they're the only way to judge magnitude — and vertical charts keep theirs either way, since their labels sit above the bars rather than duplicating the axis.

### Category names

- **The axis is sized to its longest name**, capped at `min(400px, containerWidth × 0.3)`. Previously a fixed allowance: short names reserved a gutter they didn't need, long ones truncated regardless of available room.
- **Grouped categories are separated by more than the bars inside them** (24px vs 8px). With both at 8, six categories × three series read as one run of 18 evenly-spaced bars and the reader had to count in threes to recover the groups.
- **Two measurement bugs behind the truncation**, both found by scanning canvas pixels rather than by eye:
  - Widths were measured before the web font loaded, so canvas silently used the `sans-serif` fallback — ~6% narrower than Inter at the same size, an error that grows with the string. `useFontsReady` re-measures once `document.fonts.ready` resolves.
  - `containLabel` sizes the gutter from ECharts' own measurement of the labels, which comes out at fallback metrics while the canvas paints them in the real font (211.5px vs 225.7px for a 36-character name). Right-aligned text puts that difference past the left edge of the reservation, where the canvas clips it — the first letter of every long name went missing. Padded proportionally (8%) rather than corrected; the honest fix is to own the inset instead of relying on `containLabel`, which is a wider change than this needs. Called out in the code comment.

### Also fixed

A **vertical** column reaching the axis maximum had its value label cut off by the widget header: `containLabel` reserves nothing above the plot for series labels, and the fit pass only checks a column label's *width*. Pre-existing — every vertical bar chart with `showLabels` was affected — and surfaced by the new story below rather than by this work.

## Testing

- **57 new unit tests** across `BarChart` and `DashboardItem` (314 total in the kit + pattern, all green), pinning: the stacked-total ghost and its skip conditions, the value-axis extent, the label-width band, the gutter padding, the column headroom, the band geometry, and the `descriptionAction` link — which had no coverage at all before.
- Three fixture additions to the salary-dynamics story for cases nothing covered: a **three-row** horizontal bar (height to spare rather than too little), a **vertical** bar (the other orientation), and **three deliberately long workplace names**, salted to the top of the value band so they survive the sort and actually exercise the gutter.
- Verified in Storybook throughout, including canvas pixel scans to confirm the clipping is gone (label ink starts at 14.5px, was 0px).

## Notes for the reviewer

- **`grid.left` padding is a workaround**, not a fix, and it's calibrated to Inter vs the system fallback at these sizes. A theme whose font differs from its fallback by more than 8% would clip again.
- **The 400px label cap now only binds for genuinely long names**, since the axis asks for what it needs — so it's much cheaper to raise than it was as a fixed allowance.
- **Expanded height grows with the category gap**: a 29 × 3 chart is ~3,300px expanded. Nothing clamps it, so a 200-category chart would produce a ~17,000px widget. Pre-existing, worth a follow-up.
- **The stacked total is precomputed from props**, so isolating a series via the legend moves the ghost while the number still reads as the full total. The tooltip's total behaves identically, so they stay consistent with each other.


packages/react/src/kits/F0DataChart/components/BarChart/useBarChartOptions.ts
```diff
@@ -771,6 +1247,67 @@ export function useBarChartOptions(
...
+    if (categoryWindow !== undefined) {
+      options.dataZoom = [
+        {
+          type: "inside",
+          yAxisIndex: 0,
+          startValue: 0,
+          endValue: categoryWindow - 1,
+          filterMode: "none",
+          zoomLock: true,
+          disabled: true,
```
zygisS22, line 1336: [P1] Do not hide public chart data without a recovery control

This disabled fixed window is created for every horizontal `F0DataChart`, but the “Show all” control is only wired by `F0AnalyticsDashboard/ChartItem`. Any existing or external consumer with more rows than fit now renders only the first N categories and has no way to reach the rest. That changes the public default from compressed-but-complete data to silently incomplete data. Please make windowing explicit/opt-in, or let `BarChart` own an accessible recovery affordance. Add direct coverage for fixed-window creation, the hidden-count callback, expansion, and resize changes.

packages/react/src/kits/F0DataChart/components/BarChart/useBarChartOptions.ts
```diff
@@ -408,6 +647,151 @@ function buildSeriesEntries(
...
+  }
+
+  let widest = 0
+  for (let dataIndex = 0; dataIndex < categoryCount; dataIndex++) {
+    let extent = 0
+    for (const s of series) {
+      const point = s.data[dataIndex]
+      if (point === undefined) continue
+      const own = pointExtent(point)
+      extent = stacked ? extent + own : Math.max(extent, own)
```
zygisS22: [P1] Compute mixed-sign stack extents by sign

ECharts stacks positive and negative values independently around zero. For example, `+100, +100, -150` visibly reaches `+200`, but this signed sum produces `50` and pins the axis at `52.5`, clipping most of the positive stack. A temporary regression test on this head failed with exactly `received 52.5, expected 210`. For stacked data, accumulate the rendered positive and negative extents independently, including target stack grouping, and derive `max` from the positive side. Please add this mixed-sign regression case.

packages/react/src/kits/F0DataChart/components/BarChart/useBarChartOptions.ts
```diff
@@ -408,6 +647,151 @@ function buildSeriesEntries(
...
+ * A zero-height ghost stacked on top of the real segments, carrying the
+ * category total as a label beside the bar's end.
+ *
+ * The ghost draws nothing — it exists because ECharts places a stacked series'
+ * label at that series' own position in the stack, so the only way to label the
+ * *end* of the whole bar is to stack something there. Its own value is 0, so it
+ * adds no length; the label text comes from {@link stackTotals} instead of from
+ * the datum.
+ *
+ * Consequence of precomputing the text: isolating a series from the legend
```
zygisS22, line 763: [P1] Keep stacked totals synchronized with legend selection

Legend clicks intentionally isolate series. After isolation, ECharts moves this zero-value ghost to the end of the shorter visible stack, while the formatter still displays the precomputed total of all series; the tooltip also continues using all series. The visible bar and its number therefore contradict each other. Please recompute totals from the selected legend state, or remove the total while filtering, and cover the legend-isolation interaction.

packages/react/src/patterns/F0AnalyticsDashboard/components/ChartItem/ChartItem.tsx
```diff
@@ -539,10 +556,62 @@ export function ChartItem<Filters extends FiltersDefinition>({
...
+  // The cost of deriving it is that a chart short enough to have fitted whole
+  // also gets the link — harmless, since collapsing is exactly what it does, and
+  // it duplicates the header's collapse button rather than inventing anything.
+  const canCollapseCategories = canRevealCategories && !!isFullscreen
+
+  // While rows are windowed, the item's own description is replaced by what the
+  // reader is actually looking at: a subset. Both counts come from the data and
+  // the chart's reported hidden count, so they track resizes without a second
+  // source of truth. "Top" is accurate because the window takes the first rows in
+  // data order and dashboard series arrive sorted by amount — a chart fed in some
```
zygisS22: [P2] Do not label arbitrary input order as “Top”

`ChartItem` accepts any `DashboardChartData` order and does not enforce amount-descending sorting; this comment itself acknowledges that other orders make the wording false. Date, alphabetical, or custom-ordered data would still be presented as “Top N”. Please use neutral copy such as “Showing N of M”, or define and enforce a sorted-input contract before rendering. An unsorted fixture would protect this behavior.

zygisS22 (changes requested):
Thanks for tackling this dense-chart problem. I reviewed the exact head and am requesting changes for the three correctness blockers called out inline.

Suggested order:
1. Decide the public windowing contract: keep the default complete, make windowing explicit, or provide the recovery control in F0DataChart itself.
2. Fix mixed-sign stacked scale extents.
3. Keep total labels and tooltips synchronized with legend-selected series.
4. Use order-neutral window copy and add the missing window/expand/resize and unsorted-data tests.
5. After #4948 lands, rebase and coordinate integration order with sibling #4950; the two branches currently conflict in utils/options.ts.

Validation on this head:
- TypeScript, changed-file lint, formatting, and git diff --check pass.
- 314 F0DataChart and AnalyticsDashboard tests pass across 22 files.
- Browser checks pass for the real 29 categories x 3 series case, a 375px viewport, expand/collapse, and a temporary 200-category stress case (22,464px expanded canvas).
- The current Storybook job failure is in the unrelated F0Chat voice-attachment story; this PR changes no chat files. It still needs a clean rerun before merge.

igorsafonov-gif:
Thanks — all three P1s were real, and two of them were things I'd documented as "known limitations" rather than fixed, which wasn't good enough. Addressed in 93c9fb2.

**1. Windowing hid data without a recovery control.** Correct, and the pan removal earlier in this PR is what made it airtight: before it, panning was at least a (bad) way out. It's now an opt-in `windowCategories` prop, default off, so a bare `F0DataChart` keeps every category and compresses — the previous public contract. `ChartItem` opts in exactly where it can offer the way back (`canRevealCategories`, i.e. an expand handler exists), and subscribes to the hidden count alongside it rather than independently. I kept the affordance out of the kit: only the surrounding UI knows where a "see everything" control belongs, which is why the count is reported rather than rendered.

**2. Mixed-sign stack extents.** Reproduced your `+100, +100, -150` case exactly — 52.5 instead of 210. Each sign now accumulates separately and only the positive side can set `max`; the negative side is left to the automatic minimum. Target grouping keeps the existing conservative over-estimate, which errs toward slack rather than clipping.

**3. Totals vs legend selection.** Fixed properly rather than by dropping the total. A new `useLegendSelection` hook observes `legendselectchanged` (dispatching nothing, so it doesn't fight `useLegendInteraction`), and the totals, the tooltip's share-of-total, and the data-derived axis maximum are all computed from the selected series. So isolating a series re-totals the ghost, drops the total entirely once one series is left (where it would just restate the segment label), and rescales the axis the way ECharts would with an auto maximum — that last one was a second bug your comment led me to.

**4. "Top N of M".** Now "Showing {{count}} of {{total}} categories", with the comment rewritten to say why. Note for whoever picks this up in the monorepo: the app supplies this string itself via `f0.data_chart.windowed_categories` → `common.*`, so the wording there needs the same change or it'll keep saying "Top".

**5. Tests.** Added: window creation and its locked/non-interactive shape (`disabled`, `zoomLock`, `filterMode: "none"`), the hidden-count callback, resize (fewer rows hidden as the container grows), expansion dropping the window, the vertical no-op, the mixed-sign regression, four legend-isolation cases, and an alphabetically-ordered fixture pinning that the window is positional. 329 green across 22 files.

Two of those tests earned their keep immediately: the hidden count was being computed without the new flag, so it reported `0` while the chart was windowing, and the axis wasn't rescaling on isolation.

**On sequencing (your #5):** agreed. I'll rebase once #4948 lands and coordinate with #4950 before either merges — happy to take the conflict in `utils/options.ts` on this side since #4950 is further along.

**Still open, deliberately:** the `grid.left` pad is a proportional workaround for ECharts laying out with fallback-font metrics while painting with the real font, and expanded height is unbounded (~3,300px at 29×3; your 200-category stress case at 22,464px is the same issue). Both are called out in code comments and the PR description. Say the word if you want either turned into a blocker.

zygisS22 (approved):
🚀

siguenzaraul (approved):

PR #5005, opened 2026-08-05
fix(F0Wizard): reset step content scroll on step change

albertpmz:
## Description

The wizard's step-content pane is a single DOM node reused across every step, so its `scrollTop` survived the children swap and a step taller than the pane opened part-way down — guaranteed whenever the user had to scroll to reach "Continue". This resets the pane to the top on every step change, which covers all three navigation paths (next, previous, sidebar `goToStep`) since they all funnel through `currentStep`. Reported from the course-creation wizard migration in the product monorepo ([factorial#101686](https://github.com/factorialco/factorial/pull/101686)); consumers cannot fix it themselves because the container is created inside `F0Wizard` and no prop exposes it:

https://github.com/user-attachments/assets/365aa780-b40c-4056-ab92-e8187bfbeee7

## Implementation details

- fix: reset the step-content pane to the top whenever `currentStep` changes

  <details>
  <summary>Why a layout effect, and why "instant"</summary>

  `useLayoutEffect` runs after the new step is in the DOM but before paint, so
  the reset never flashes. `behavior: "instant"` keeps it immune to a
  `scroll-behavior: smooth` on the element — a smooth scroll on step change
  reads as a glitch. jsdom implements no `Element.scrollTo` at all, so there is
  a `scrollTop` fallback to avoid crashing consumers' unit tests.
  </details>

- test: unit tests for forward, backward, sidebar and blocked navigation, plus the `scrollTop` fallback path
- test: `ScrollResetOnStepChange` play test on `F0WizardForm`, which scrolls the pane to the bottom, advances, and asserts the same node is back at 0

  <details>
  <summary>Why a play test too</summary>

  jsdom does no layout, so `scrollHeight`/`clientHeight` are 0 and nothing is
  ever scrollable there — the unit tests can only assert the reset call reaches
  the container. The play test is what actually proves the visible behavior.
  Verified both ways: green with the fix, red with the effect disabled.
  </details>

### DIFF
--- a/packages/react/src/patterns/F0WizardForm/__stories__/F0WizardForm.stories.tsx
+++ b/packages/react/src/patterns/F0WizardForm/__stories__/F0WizardForm.stories.tsx
@@ -996,7 +996,14 @@ export const ScrollResetOnStepChange: Story = {
   render: () => <ManyFieldsStory extraDefaultValues={{ gender: "female" }} />,
   play: async ({ canvasElement, step }) => {
     const page = within(canvasElement.closest("body")!)
-    const content = getStepContent(canvasElement)
+
+    // The wizard dialog mounts through a portal and animates in, so the pane is
+    // not in the DOM yet when the play function starts.
+    const content = await page.findByTestId(
+      "wizard-step-content",
+      {},
+      { timeout: 10000 }
+    )
 
     await step("The step content pane actually overflows", async () => {
       await waitFor(() =>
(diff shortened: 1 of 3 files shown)


sauldom102 (changes requested):
Hey hey could you put some before/after video?

albertpmz:
> Hey hey could you put some before/after video?

Hey bud! Done, check around 0:40 for the exact moment when I go to the next step and have to scroll up again 🙃

PR #5012, opened 2026-08-05
fix(F0Graph): keep windowed-in nodes' edges from vanishing on pan

factmarc10010:
## Problem

With `enableNodeWindowing` on, connecting/reporting lines vanish on a large, flat or deep org chart — the node stays on screen but its line disappears. This is the "las conexiones se pierden al abrir mucho el árbol" report, and it survives the earlier #4742 fix (which only stopped React Flow's own culling). Two triggers:

- **Panning** a large/deep tree: a node's line flickers and drops as the camera moves.
- **Expanding** a node: expanding a wide node spreads its siblings past the screen edge, so the parent ends up looking connected only to the child that stayed on screen — its reporting fan-out disappears.

Screenshot 2026-08-05 at 16 10 07

## Root cause

Two distinct issues in `useGraphRenderModel`, both coming from how viewport windowing feeds React Flow:

1. **Un-measured nodes drop their edges.** A node that windowing pans/expands into view is handed to React Flow with no measured handle bounds for the commit it is added. `getEdgePosition` returns `null` for any edge touching an un-measured node, so those edges are dropped from the DOM that frame — the line vanishes (and stays gone while panning, since every frame re-adds nodes un-measured).

2. **Off-viewport edge endpoints are culled.** Windowing materializes only the nodes intersecting the viewport (plus the ancestry spine). Expanding a wide node pushes its siblings off-screen, so the parent's edges to them are dropped even though the parent is on screen and those lines cross the viewport.

## Fix

1. **Seed port handles + size on each windowed graph node.** React Flow falls back to `toHandleBounds(node.handles)` when a node has no measured handle bounds, so providing the source/target port offsets (plus `width`/`height`) lets `getEdgePosition` resolve the endpoints on the very commit the node is added — the lines never drop, and the real measured bounds take over once the DOM mounts. Applied only while windowing actually drives the render, so React Flow's own `onlyRenderVisibleElements` path (windowing off) is left untouched.

2. **Frontier edge inclusion.** For every visible edge with exactly one endpoint inside the viewport, pull the other endpoint into the windowed set too, so React Flow can route the edge even when the layout pushed that endpoint off-screen. Keyed on the viewport set (not the ancestry spine), so an off-screen ancestor doesn't drag in all of its children at every level; a collapsed parent contributes only its expander-stub edge (its real children aren't in `visibleEdges`), so closed subtrees stay windowed out, and a lone far node with no edge crossing the view is still culled. The ancestry walk already covered the reverse (a visible child's line up to an off-screen parent).

Net effect: every edge whose path crosses the window is drawn — parent → off-screen child and child → off-screen ancestor — with **no sticky/unbounded retention across a session** (nodes panned away are released, and never-seen subtrees stay windowed out). `F0Graph` unit tests pass (250/250); typecheck/lint clean for the touched files.

## Performance trade-off

The frontier set is recomputed per frame and doesn't accumulate over time, but its size **grows with the child count of the visible nodes**: when a node with a very large fan-out is on screen and expanded, *all* of its children are materialized — including the ones the layout pushed off-screen — so their connecting lines can be drawn. For such **extreme fan-out, FPS preservation is not guaranteed**.

This is a deliberate, accepted trade-off: we sacrifice some performance on pathologically wide nodes in order to always show the complete set of paths from a node to its children — the exact pain this PR sets out to fix. For typical fan-out the extra cost is a handful of off-screen nodes and is negligible. If extreme fan-out becomes a problem in practice, a follow-up can cap the number of off-screen children materialized per parent (drawing the first N lines + an indicator).

Screenshot 2026-08-05 at 18 09 15

## How to test

- Storybook → `Graph/F0Graph/DeepTreeWindowing` (or `ViewportWindowing`): pan far from the initial frame with windowing on; the connecting lines stay attached under the camera.
- Local app: install this branch's alpha and open the org chart (v3) on a flat/deep tree. Expand a wide node and pan — the parent keeps its lines to **all** its children, including the ones pushed off-screen, and nothing drops.

### DIFF
--- a/packages/react/src/patterns/F0Graph/hooks/useGraphRenderModel.ts
+++ b/packages/react/src/patterns/F0Graph/hooks/useGraphRenderModel.ts
@@ -529,6 +529,16 @@ export function useGraphRenderModel<T>({
           y: (pos?.y ?? 0) * yStretch,
         },
         width: BASE_W,
+        height: BASE_H,
+        // Seed the measured size from the layout so React Flow can route this
+        // node's edges on the SAME commit it is added, instead of waiting for
+        // its ResizeObserver. Without it, a node that windowing pans into view
+        // is handed to React Flow un-measured for one commit; edges touching it
+        // resolve to a null position and are dropped from the DOM that frame, so
+        // connecting lines flicker/vanish while panning a large or deep tree
+        // (the on-screen node stays, its reporting line disappears). The real
+        // DOM measurement still runs and overwrites this with the exact size.
+        measured: { width: BASE_W, height: BASE_H },
         sourcePosition: sourcePos,
         targetPosition: targetPos,
         data: {
--- a/packages/react/src/patterns/F0Graph/hooks/useGraphRenderModel.ts
+++ b/packages/react/src/patterns/F0Graph/hooks/useGraphRenderModel.ts
@@ -384,6 +384,11 @@ export function useGraphRenderModel<T>({
   ])
 
   // ── Node-array windowing ──
+  // Ids that have ever been materialized into React Flow. Once a node has been
+  // rendered (seen + hydrated), we keep it in the window so it never re-enters
+  // React Flow un-measured on a later pan — see the sticky union below.
+  const seenWindowedIdsRef = useRef<Set<string>>(new Set())
+
   // Ids of the LAYOUT nodes (graph pills + expanders) whose box intersects the
   // padded viewport. `null` means "no windowing" (feature off, or viewport not
   // yet measured) → render everything, which is also what the initial `fitView`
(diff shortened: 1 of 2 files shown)


packages/react/src/patterns/F0Graph/hooks/useGraphRenderModel.ts
```diff
@@ -430,10 +430,34 @@ export function useGraphRenderModel<T>({
...
+    // route the edge even when the layout pushed that endpoint off-viewport. This
+    // keeps a visible parent's line to an off-screen child: expanding a wide node
+    // spreads its siblings past the screen edge, and without this the parent
+    // would look connected only to the child that stayed on screen. The reverse
+    // (a visible child's line up to an off-screen parent) is already covered by
+    // the ancestry walk above. Keyed on the VIEWPORT set — not `ids` — so an
+    // off-screen spine ancestor does not drag in all of its children at every
+    // level. A collapsed parent contributes only its expander-stub edge (its real
+    // children aren't in `visibleEdges`), so closed subtrees stay windowed out.
+    const viewportIds = new Set(base)
```
albertpmz, line 447: If a parent node has many off-screen children, all of those children will be pulled to the rendered set to preserve their lines. This can potentially make the graph slow again. Can you take a look at it?

albertpmz (changes requested):
Great work! Just take a look to this comment below and it'll be good to go 👍🏼

albertpmz (approved):

PR #5025, opened 2026-08-06
fix(Table): only flash rows that were actually inserted

subinedward-boop:
## Description

The flash-on-add effect introduced in #4709 fires whenever rows appear on screen, not only when they are inserted. Searching to an empty result and then clearing the search makes every row flash; any filter change flashes whatever rows it reveals. This narrows the effect to genuine inserts, continuing the line of #4809 which fixed the same class of false positive for pagination.

## Type of change

- [x] Bug fix

Not attached — the change is the *absence* of an animation, which a still frame can't show. Reproduce by searching a `OneDataCollection` table for a term with no matches, then clearing the search: on `main` every row flashes green, on this branch none do.

## Implementation details

- fix: stop pruning the seen-keys set down to the currently rendered rows

  <details>
  <summary>Why?</summary>

  `useAddedRowKeys` kept its "already seen" memory trimmed to the rows on screen, so hiding a row erased its history and it counted as brand new when it returned. Searching to zero results wiped the memory entirely, which is why clearing the search flashed the whole table.

  The set now accumulates for as long as the query is unchanged — which is what the hook's own docstring already promised. The docstring and the code had drifted apart; the code was what shipped.
  </details>

- fix: widen the reset key from the pagination position to the whole query

  <details>
  <summary>Why?</summary>

  The hook already suppresses the flash when rows arrive by navigation rather than insertion, but only pagination counted. A search, filter, sort or grouping change is the same kind of event: the incoming rows are the answer to a different question. Without this, rows revealed by a filter that were never on screen before would still flash.
  </details>

- fix: wait for a query's own data before reseeding the baseline

  <details>
  <summary>Why?</summary>

  Resetting on a query change is not sufficient, and this was the hard part. Query state updates on the render the user acts, but the matching rows arrive later, so there is always at least one render carrying the new query and the *previous* query's rows. Reseeding there memorises the wrong rows and flashes the real results when they land.

  Gating on the datasource's `isLoading` does not close that window either — a first attempt at this fix did exactly that and still flashed the whole table in the browser. `useData` sets it from an effect, and it flickers back to `false` between a query change and its debounced fetch, so the wait resolves against the previous query's rows.

  Instead a query change marks the baseline *pending*: nothing flashes, and the baseline is adopted on the first render whose rows actually differ from it — the only reliable evidence the new query's data has arrived. The docstring records why `isLoading` must not be used here, since it is the obvious thing to reach for.
  </details>

- test: add integration coverage driving the real `useDataCollectionSource`

  <details>
  <summary>Why a second layer of tests?</summary>

  The hook-level unit tests call `useAddedRowKeys` directly, so they cannot reproduce the render ordering described above — and indeed they passed against a version of this fix that still flashed the whole table in the browser. The new tests render `TableCollection` against a real source with an async adapter and assert on the `animate-row-flash` class.

  Both search round-trip tests fail against `origin/main`. The insert test passes before *and* after, on purpose: it guards against over-correcting into "never flash at all".
  </details>

### Resulting behaviour

A row flashes only when the query is unchanged and a key appears that this table has not shown before. Initial load, paging, load-more, search, filters, sorting, grouping, refetch and reorder are all silent.

### Known ceilings

- A row genuinely deleted and then genuinely re-added under the same query no longer flashes the second time, since it stays in the seen set.
- A query change that returns exactly the same rows stays pending, so the next insert reseeds the baseline instead of flashing. One missed flash in a corner, versus flashing every row on every filter clear.

Both come from the table *inferring* inserts by diffing row keys. Removing the inference entirely means the insert path signalling the flash instead — new public API, and not worth it unless these cases come up. The existing `AddRowContext` only supplies the add-row buttons and carries no insert signal.

- fix: unregister exiting rows from the selection registry immediately

  <details>
  <summary>Why?</summary>

  Rows animated out by AnimatePresence stay mounted until their exit transition finishes, but their selection-registry cleanup only ran on unmount. During the fade a filtered-out or deleted row was still registered, so a select-all clicked in that window reached a row no longer in the result set.

  The row now watches its own presence (`useIsPresent`) and unregisters the moment its exit starts. Outside AnimatePresence the hook returns true, so nested rows keep the previous behaviour. Tested by supplying the PresenceContext value AnimatePresence gives an exiting child — jsdom completes exits within a frame, so a real AnimatePresence cannot hold the window open for an end-to-end test (which is also why the animation-skipping global test setup never saw this).
  </details>

### Not addressed here

Consumer test suites that assert on the DOM immediately after a delete/filter still need `MotionGlobalConfig.skipAnimations = true` in their test setup (as this repo's own harness does): AnimatePresence keeps exiting rows mounted until the transition ends, and that is the feature, not a bug.

### DIFF
--- a/packages/react/src/hooks/datasource/useData.ts
+++ b/packages/react/src/hooks/datasource/useData.ts
@@ -112,6 +112,19 @@ export interface UseDataReturn<R extends RecordType> {
 
   // Merged filters (default values and current values)
   mergedFilters: FiltersState<FiltersDefinition>
+
+  /**
+   * Opaque identity of the query whose response produced `data` — filters,
+   * search, sortings and pagination position, as they were when that fetch was
+   * issued. Undefined until the first response commits.
+   *
+   * Compare it across renders to tell "these rows answer a different question"
+   * from "these rows changed". The live filter/search state on the source can't
+   * do that: it moves a render (and a debounce) before the matching rows do, so
+   * there is always a window where it describes a query the rendered rows do
+   * not answer.
+   */
+  committedQuery: string | undefined
 }
 
 type DataType<T> = PromiseState<T>
@@ -315,6 +328,15 @@ export function useData<
   const [totalItems, setTotalItems] = useState<number | undefined>(undefined)
   const [isLoadingMore, setIsLoadingMore] = useState(false)
 
+  // Identifies the query whose response produced the records currently
+  // rendered — not the query the user has selected, which changes a render (or
+  // a debounce) before its data arrives. Consumers that must tell "these rows
+  // answer a different question" from "these rows changed" need this rather
+  // than the live filter/search state.
+  const [committedQuery, setCommittedQuery] = useState<string | undefined>(
+    undefined
+  )
+
   const isLoadingMoreRef = useRef(false)
 
   // Latest loaded records + the query/pageSize that produced them, so a
@@ -382,7 +404,8 @@ export function useData<
     (
       result: PaginatedResponse<R> | SimpleResult<R>,
       appendMode: boolean,
-      isLoadingYet?: boolean
+      isLoadingYet?: boolean,
+      query?: string
     ) => {
       /**
        * Call to the onResponse callback
@@ -452,6 +475,9 @@ export function useData<
       setIsLoading(!!isLoadingYet)
       setIsLoadingMore(false)
       isLoadingMoreRef.current = false
+      // Stamp the rendered records with the query they answer. Batched with
+      // the setters above, so this costs no extra render.
+      if (query !== undefined) setCommittedQuery(query)
     },
     // eslint-disable-next-line react-hooks/exhaustive-deps -- we don't want to re-run this callback when data.length changes
     [
@@ -464,6 +490,7 @@ export function useData<
       setIsLoadingMore,
       setTotalItems,
       isLoadingMoreRef,
+      setCommittedQuery,
     ]
   )
 
@@ -612,6 +639,18 @@ export function useData<
           }
         )
 
+        // What this particular fetch is asking for. Stamped onto the records it
+        // returns so consumers can tell a different question from a changed
+        // answer. The pagination position is part of it: paging and load-more
+        // swap the rows by navigation, not by insertion.
+        const query = JSON.stringify({
+          filters,
+          search,
+          sortings,
+          currentPage,
+          cursor,
+        })
+
         function fetcher(): PromiseOrObservable<ResultType> {
           setTotalItems(undefined)
 
@@ -648,7 +687,7 @@ export function useData<
 
         // Handle synchronous data
         if (!("then" in result || "subscribe" in result)) {
-          handleFetchSuccess(result, appendMode)
+          handleFetchSuccess(result, appendMode, undefined, query)
           return
         }
 
@@ -659,7 +698,7 @@ export function useData<
         const subscription = observable.subscribe({
           next: (state) => {
             if (state.data) {
-              handleFetchSuccess(state.data, appendMode, state.loading)
+              handleFetchSuccess(state.data, appendMode, state.loading, query)
             } else if (state.loading) {
               setIsLoading(true)
             } else if (state.error) {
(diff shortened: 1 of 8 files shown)


packages/react/src/patterns/OneDataCollection/visualizations/collection/Table/hooks/useAddedRowKeys.ts
```diff
@@ -40,29 +56,32 @@ export function useAddedRowKeys(
...
+      pendingRef.current = true
     }
 
-    // The first non-empty render seeds the baseline (initial load never
-    // flashes); after that every key we've ever seen is remembered so it can
-    // only flash the first time it appears.
-    if (!initializedRef.current && keys.length > 0) {
-      initializedRef.current = true
-    }
+    if (pendingRef.current) {
```
albertpmz: In the following scenario:
1. Q1 returns `[a, b]`.
2. A subsequent query, Q2, **also** returns `[a, b]`.
3. A third element, `c`, is added.

The last insertion wouldn't be properly animated with the flash-on-add effect. 

`sameKeys(seenRef.current, keys)` is not evidence that Q2 has not loaded: Q2 can validly return the same rows as Q1. In that case this `return` leaves `pendingRef` true. The next genuine insert is then treated as Q2’s first changed result, reseeded into `seenRef`, and does not receive `animate-row-flash`.

Please settle pending state from a datasource signal that the current query’s response has committed, rather than from a changed row-key set, and add this regression case to the spec.

albertpmz (changes requested):
Nice job fixing this! There's only one issue that needs to be taken care of 🙌🏼

packages/react/src/patterns/OneDataCollection/visualizations/collection/Table/hooks/useAddedRowKeys.ts
subinedward-boop: Confirmed — the Q1/Q2-same-rows scenario was real, and settling on a changed row-key set could never distinguish it from a not-yet-loaded response. Fixed in 0ca1d3a as you suggested, with a datasource signal:

`useData` now stamps every committed response with the query that produced it (filters, search, sortings, pagination position at fetch time) and exposes it as `committedQuery`. The hook keys its baseline on that instead of the live query state, so the reseed happens exactly when rows answering a different question are rendered — no waiting heuristic at all. The pending-latch and the `sameKeys` comparison this comment was anchored to are gone entirely; the hook is back to its pre-branch shape.

Your exact scenario is covered at both levels: a unit case in `useAddedRowKeys.test.ts` ("still flashes an insert after a query that returned the same rows") and an end-to-end case in `RowFlash.integration.test.tsx` driving the real `useDataCollectionSource` with a search that returns the same rows, followed by an insert that must flash. The integration version fails against the previous latch implementation.

One trade-off worth your eyes: `committedQuery` is a new field on the exported `UseDataReturn`, stamped in `handleFetchSuccess` (batched with the existing setters, so no extra render). Naming/placement suggestions welcome.

packages/react/src/hooks/datasource/useData.ts
```diff
@@ -112,6 +112,19 @@ export interface UseDataReturn<R extends RecordType> {
...
+   * search, sortings and pagination position, as they were when that fetch was
+   * issued. Undefined until the first response commits.
+   *
+   * Compare it across renders to tell "these rows answer a different question"
+   * from "these rows changed". The live filter/search state on the source can't
+   * do that: it moves a render (and a debounce) before the matching rows do, so
+   * there is always a window where it describes a query the rendered rows do
+   * not answer.
+   */
+  committedQuery: string | undefined
```
albertpmz: Making `committedQuery` required here is a breaking change for consumers. Make this optional (`committedQuery?: string`) or keep it internal to the collection-data path. `useData` can still always return the value, while existing consumer contracts remain valid.

albertpmz (changes requested):

packages/react/src/hooks/datasource/useData.ts
subinedward-boop: Right — required, it broke every existing constructor of the interface. Made optional in 745b9ad6 (`committedQuery?: string`); `useData` still always returns it, and the doc comment notes why it's optional. Typecheck and the full Table suite pass unchanged.

albertpmz (approved):
Now we're ready to go! Nice job 👏🏼

PR #5033, opened 2026-08-06
feat(ai-canvas): add AutofillTimesheetCanvasContent to the canvas union

GinnyThomas:
## Description

Add an `autofillTimesheet` variant to the `CanvasContent` discriminated union so host apps can render an **editable timesheet-autofill proposal in the center canvas panel** (via `openCanvas`) instead of inline in the narrow chat panel. Part of FCT-59357 (ONE smart timesheet autofill for contract employees).

## Type of change

Type-only additive change to the existing `canvas` kit — no new component, no breaking change, no API removal.

N/A — not a new component or promotion. Extends the existing canvas-content contract.

N/A — no visual/runtime change in this PR (see Implementation details).

## Implementation details

New exported types in `kits/ai/canvas/types.ts`:

- `AutofillTimesheetCanvasContent` (`CanvasContentBase & { type: "autofillTimesheet" }`) — carries `employeeId`, the `startOn`/`endOn` range, and the day-grouped proposed `shifts`. Inherits the required `title` from `CanvasContentBase`, same as every other variant.
- `AutofillTimesheetShift` — one proposed block: employee-local `clockIn`/`clockOut` (`HH:MM`, nullable), `workable`, and optional work location (`workplaceId` / `workAreaId` / `locationType`). Placed in its own labelled payload section below the entity contract, mirroring `DataDownloadDataset` / `ChatDashboardConfig`.

Added to the `CanvasContent` union and re-exported through the existing `canvas` and `F0AiChat` public barrels, matching the other content variants.

**No renderer in this PR by design.** The canvas registry (`useCanvasEntity`) is host-supplied — entities are resolved from `F0AiChatProvider.canvasEntities`, not from a built-in f0 map. So the renderer for `autofillTimesheet` is registered app-side (factorial), consistent with how `dashboard` / `form` / `dataDownload` are wired. This PR only unblocks `openCanvas({ type: "autofillTimesheet", … })` type-checking on the factorial side.

### Verification
- `oxfmt` + `oxlint` clean (lefthook pre-commit).
- `tsc` reports no errors in any touched file; remaining repo `tsc` errors pre-exist on `main` (missing optional dev deps + story typings), unrelated to this change.

### DIFF
--- a/packages/react/src/kits/ai/canvas/types.ts
+++ b/packages/react/src/kits/ai/canvas/types.ts
@@ -165,6 +165,11 @@ export type AutofillTimesheetShift = {
   workable: boolean
   workplaceId?: string | null
   workAreaId?: string | null
+  /**
+   * Host-defined work-location kind. Left as a loose string to keep the kit
+   * host-agnostic; the factorial consumer narrows it to its attendance
+   * location-type enum (today: "office" | "work_from_home" | "business_trip").
+   */
   locationType?: string | null
 }
 
--- a/packages/react/src/kits/ai/canvas/types.ts
+++ b/packages/react/src/kits/ai/canvas/types.ts
@@ -59,6 +59,32 @@ export type DataDownloadCanvasContent = CanvasContentBase & {
 }
 
 /**
+ * A single proposed shift block in a timesheet-autofill preview. Clock times are
+ * employee-local "HH:MM"; a null clock time marks an as-yet-unfilled bound.
+ */
+export type AutofillTimesheetShift = {
+  date: string
+  clockIn: string | null
+  clockOut: string | null
+  workable: boolean
+  workplaceId?: string | null
+  workAreaId?: string | null
+  locationType?: string | null
+}
+
+/**
+ * Autofill-timesheet canvas content — renders an editable timesheet proposal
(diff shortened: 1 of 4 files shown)


packages/react/src/kits/ai/canvas/types.ts
```diff
@@ -137,6 +150,29 @@ export type DataDownloadDataset = {
...
+// ---------------------------------------------------------------------------
+
+/**
+ * A single proposed shift block in a timesheet-autofill preview. Clock times are
+ * employee-local "HH:MM"; a null clock time marks an as-yet-unfilled bound.
+ */
+export type AutofillTimesheetShift = {
+  date: string
+  clockIn: string | null
+  clockOut: string | null
```
albertpmz: These fields are explicitly nullable, but the paired Factorial parser requires both values to be strings. One `null` bound causes it to reject the entire proposal and render nothing. Should these be non-`null`, or should the consumer accept `null` as this contract describes?

albertpmz (changes requested):
Great work! Let's just take a look at this thingy below 👀

albertpmz (approved):
LGTM!

PR #5036, opened 2026-08-07
fix(DatePickerPopup): keep the calendar open while its month and year dropdowns are used

MohamedHegab:
## Description

Inside a dialog, the calendar's **month and year dropdowns are unusable**: opening either one makes the whole calendar disappear, so the user can only reach other months with the prev/next arrows. Reported from the Factorial monolith (attendance "Add employees" modal, f0-react 6.5.0) as "the effective date field is not clickable and I can't select a date".

Repro on `main`: render an `F0DatePicker` inside a centered `F0Dialog`, open the calendar, click "August" (or the year). The dropdown flashes and the calendar closes.

## Type of change

- [x] Bug fix

## Implementation details

The header's month/year pickers are `F0Select`s, and a select always portals its listbox **out** of the calendar — into the surrounding dialog's container (or `document.body`). It is therefore never a DOM descendant of the calendar's `PopoverContent`, and Radix reports opening one as an interaction *outside* the calendar, dismissing it.

Verified in the browser rather than inferred — the dismissal fires as a `dismissableLayer.focusOutside` whose target is the dialog's own content root, because the dialog's focus scope parks focus there while the dropdown mounts:

```
pointerDownOutside  target=SPAN (month trigger, inside the popover)
focusin             target=DIV[role=dialog]      ← dialog content root
focusOutside        target=DIV[role=dialog]      → calendar dismissed
focusin             target=DIV[role=listbox]     (dropdown, in the dialog container)
```

So `DatePickerPopup` now declines to dismiss for two shapes that are not real outside interactions:

1. **Its own dropdowns.** Ownership follows the `aria-controls` link from the listbox back to its trigger, and the trigger must live inside this calendar. Matching on `[role="listbox"]` alone is not enough — unrelated selects in the same dialog render the same shape (e.g. an inline employee-selector list), and clicking those must still dismiss the calendar.
2. **An ancestor of the calendar** receiving focus (the dialog content root above), on `onFocusOutside` only. `onPointerDownOutside` keeps the default behavior there, so a genuine click elsewhere in the dialog still closes the calendar.

This complements #4966, which put the calendar *inside* the dialog container so the dropdowns paint above it; that fix addressed stacking, this one addresses dismissal.

## Testing

Browser-verified against the real monolith app (f0 built from this branch, linked into `frontend/node_modules`), attendance "Add employees" modal:

| Interaction | `main` | this branch |
|---|---|---|
| Open month dropdown | calendar closes | calendar stays open, dropdown visible |
| Pick "September" | unreachable | month switches, calendar stays open |
| Open year dropdown | calendar closes | calendar stays open |
| `Escape` | closes | closes |
| Click elsewhere in the dialog | closes | closes |

`vitest run src/ui/DatePickerPopup src/components/F0DatePicker src/components/OneCalendar` → 333 passed.

Added an `InsideDialog` interaction story rather than a unit test on purpose: the dismissal depends on real focus and pointer behavior, and a jsdom test of this flow passes with or without the fix, so it would guard nothing.

### DIFF
--- a/packages/react/src/components/F0DatePicker/__stories__/F0DatePicker.stories.tsx
+++ b/packages/react/src/components/F0DatePicker/__stories__/F0DatePicker.stories.tsx
@@ -3,8 +3,9 @@ import type { Meta, StoryObj } from "@storybook/react-vite"
 import { addMonths, subDays } from "date-fns"
 import MockDate from "mockdate"
 import { useState } from "react"
-import { expect, fn, within } from "storybook/test"
+import { expect, fn, screen, userEvent, within } from "storybook/test"
 
+import { F0Dialog } from "@/patterns/F0Dialog"
 import { Placeholder } from "@/icons/app"
 import { dataTestIdArgs } from "@/lib/data-testid/__stories__/args"
 import { withSkipA11y, withSnapshot } from "@/lib/storybook-utils/parameters"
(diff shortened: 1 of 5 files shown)


albertpmz (changes requested):
Great job! The fix and its ownership checks look well scoped 👏🏼 before approving, though, I’d like automated regression coverage for the behavior being fixed. 

As the story notes, its current play function only verifies that the dropdown opens and still passes without the `DatePickerPopup` changes, so **it cannot detect this regression returning**.

Could we add focused Vitest coverage around the dismissal handling that fails when this fix is removed, without needing to reproduce the browser focus transition itself?

albertpmz (approved):
LGTM!

PR #5045, opened 2026-08-07
feat(OneDataCollection): add defaultExpanded to nested tables

factmarc10010:
## Description

Nested tables always start fully collapsed, and today there is no way to ask for anything else — the expansion state lives in a private `useState` inside `NestedDataProvider`, with no prop, callback or handle to seed it.

This adds a `defaultExpanded` option to the table visualization. Two products need it with *different* policies: Teams wants the whole tree open, Job Catalog wants it open down to roles but not into their levels. So the option is a policy, not a flag.

```ts
defaultExpanded: true                            // everything
defaultExpanded: 2                               // down to depth 2
defaultExpanded: (node) => node.type !== "role"  // stop at roles
```

### Expanded until a certain level based on policies

https://github.com/user-attachments/assets/8b99b6fe-4ddf-4f19-967a-630a1d58b78e

### Expanded completely

https://github.com/user-attachments/assets/df5270f2-f7ed-4b02-b26a-4bcf3ae8f9d8

### Why a per-row policy and not a list of ids

Two reasons, and both rule out an `expandedIds` style API:

1. **A consumer cannot build the keys.** The row key is `` `${depth}-${item.id}-${index}` `` — it embeds the depth and the index within the parent's children, neither of which the consumer controls.
2. **The cascade is free.** A row that opens mounts its children, and those children evaluate the policy themselves. Recursion falls out of the component tree, so there is no traversal on the consumer side and the same code works whether the tree is already in memory or fetched lazily one level at a time.

A row opened by the policy never goes through `handleExpand`, so it requests its own children from an effect. That is guarded by a ref rather than by `children.length`, which stays `0` for a parent whose fetch comes back empty and would otherwise re-request on every render.

### The tri-state, and why it is load-bearing

Collapsing used to **delete** the entry:

```ts
if (expanded) return { ...prev, [rowId]: true }
const next = { ...prev }
delete next[rowId]      // ← "never seen" and "user closed it" became the same state
return next
```

With a default-expanded policy that is a bug generator: the two states are indistinguishable, so an opening policy reopens a row the instant the user collapses it. Recording the `false` gives three states — absent means "the user has not decided", and only then does the policy get a say.

It also makes a deliberate collapse survive the row unmounting. Collapse a grandchild, collapse its parent (the grandchild unmounts), reopen the parent: the grandchild comes back visible but still closed, because its `false` lives in the provider above the rows.

### `NestedDataProvider` is now mounted unconditionally

It used to be swapped for a `Fragment` on flat tables. Choosing the wrapper by branch makes it impossible to pass it props without rebuilding the component type on every render — which would remount the entire table whenever a consumer passed an inline `defaultExpanded` predicate. The provider only holds nested state that flat tables never read, so mounting it always is both cheaper and simpler than the alternatives.

### Interaction with filters

`clearFetchedData()` already resets the tree when filters, sortings or navigation filters change. With a policy in play that means a filtered view comes back expanded, which is the behaviour the consuming teams asked for. It also means manual collapses are discarded on a filter change — deliberate, and documented on the option.

### Cost

Opening a row loads its children, so a policy that opens a large tree costs one `fetchChildren` per opened row on first paint. Documented on the option. For the two consumers driving this the tree structure is already in memory, so the cost is per-node data hydration rather than structure fetching.

## Verification

Four stories with `play` assertions: collapsed, all, to-depth, and stop-at-roles.

Plus 11 unit tests in `Table/__tests__/defaultExpanded.test.tsx`, covering the policy (no policy, `true` cascading to depth 3, a depth, a predicate, and the `{ depth }` it receives), the two behaviours the feature leans on, and the flat table now that the provider is always mounted.

Every test was checked against the pre-fix code rather than only against the fix:

| Revert to the old code | What fails |
| --- | --- |
| `delete next[rowId]` on collapse | the two tri-state tests — the policy reopens the row the user just closed, and the collapse is lost when the row unmounts |
| guard the request on `children.length` instead of the ref | the empty-children test — 1 expected `fetchChildren` becomes **1145** |
| drop the self-fetch effect | 7 tests — a policy-opened row renders open with no children |
| ignore `isExpandedByDefault` in `open` | 8 tests |

Also manually verified in Storybook that the cascade reaches depth 3, that a manual collapse is not undone by the policy, and that a collapse survives its row unmounting.

### DIFF
--- a/packages/react/src/patterns/OneDataCollection/__stories__/visualizations/table/default-expanded.stories.tsx
+++ b/packages/react/src/patterns/OneDataCollection/__stories__/visualizations/table/default-expanded.stories.tsx
@@ -128,7 +128,9 @@ export const All: Story = {
   play: async ({ canvasElement }) => {
     const canvas = within(canvasElement)
     // Depth 3 proves the cascade: each level opened because its parent did.
-    await canvas.findByText("Senior")
+    // Both roles carry a "Senior", as levels repeat across roles, so this also
+    // pins that the cascade reached BOTH branches and not just the first.
+    expect(await canvas.findAllByText("Senior")).toHaveLength(2)
   },
 }
 
@@ -151,6 +153,9 @@ export const StopAtRoles: Story = {
   play: async ({ canvasElement }) => {
     const canvas = within(canvasElement)
     await canvas.findByText("Backend Engineer")
-    expect(canvas.queryByText("Junior")).toBeNull()
+    // `queryAllBy`, not `queryBy`: "Junior" exists under both roles, so if the
+    // levels ever did render, `queryByText` would throw on the duplicate rather
+    // than report the real failure.
+    expect(canvas.queryAllByText("Junior")).toHaveLength(0)
   },
 }
--- a/packages/react/src/patterns/OneDataCollection/__stories__/visualizations/table/default-expanded.stories.tsx
+++ b/packages/react/src/patterns/OneDataCollection/__stories__/visualizations/table/default-expanded.stories.tsx
@@ -1,0 +1,156 @@
+import { Meta, StoryObj } from "@storybook/react-vite"
+import { expect, within } from "storybook/test"
+
+import { useDataCollectionSource } from "../../../hooks/useDataCollectionSource"
+import { OneDataCollection } from "../../../index"
+
+/**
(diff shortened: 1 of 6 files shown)


sauldom102 (changes requested):
Please add some tests

sauldom102 (approved):

PR #5051, opened 2026-08-07
feat(F0Graph): opt-in stacked child nodes

factmarc10010:
## Description

Adds **opt-in stacked children** to F0Graph: a parent that sets `stackChildren` renders its children as a tight vertical column of compact rows directly beneath it, instead of the default horizontal fan-out.

This is the shape the pre-F0 job catalog uses for the job levels under a role — same x as the parent, ~40px per row, one trunk connector, and no horizontal space reserved so the parent's siblings close in around it. F0Graph had no way to express it: the layout engine is a strict cursor-based tree where every child consumes `nodeWidth` on the cross axis, `y` comes purely from `depth * (nodeHeight + rankSep)`, there is one global node size, and an edge is drawn to every visible child.

Jira: [FCT-60913](https://factorialmakers.atlassian.net/browse/FCT-60913)

## Type of change

## Screenshots

Screenshot 2026-08-11 at 14 27 33

## Implementation details

**1. Core (`patterns/F0Graph`)**

- `GraphNode.stackChildren` — set on the **parent** ("my children stack"), mirroring the legacy `hasLevelChildren` semantics. Propagated through `useTreeBuilder` onto `TreeNode`.
- `useLayoutEngine` — a stacked parent is laid out as a **leaf** on the cross axis, so its column costs no sibling space. After every cross-axis shift has settled, the children are placed as a column sharing the parent's final centre, each carrying a `stackIndex` that offsets it down the main axis in `stackedNodeHeight + stackedNodeGap` steps from the start of the next rank. Stacked rows materialize with their own (shorter) box. New `stackedNodeHeight` / `stackedNodeGap` props on `<F0Graph>` (defaults 40 / 8).
- `useGraphRenderModel` — `resolveStackedParents` is the single source of truth for "does this group stack?", shared by the layout input, the edge pass and the node render context. Only the trunk edge into the first row is drawn; the rest are filtered at the **rfEdges** stage, not out of `visibleEdges`, because the layout engine derives its parent→child adjacency from that same list (filtering earlier orphans every row below the first).
- `renderNode` receives `ctx.stacked`.

**2. Visual — `F0GraphStackedNode`**

Compact row: title plus a trailing slot (checkbox / count / small action), full node width, fixed height, `treeitem` semantics, selection/focus rings matching the node card. Clicks inside the trailing slot don't select the node (`stopPropagation` **and** `data-no-node-select`, since the canvas selects on `pointerup` regardless of React handlers). It deliberately does not morph with zoom — the layout reserves a fixed band per row.

**3. OneDataCollection graph visualization**

`stackChildren: (record) => boolean` and `stackedTrailing: (record) => ReactNode` in `GraphVisualizationOptions`, threaded through `useDataCollectionTreeData` (both node-construction paths), so a Data Collection can stack without dropping out of the pattern.

### Constraint worth knowing

Stacking only applies when **every child is a leaf**. A stacked row is a strip with no lane beneath it, so a group containing an expandable child silently falls back to the normal fan-out rather than rendering a subtree with nowhere to go. Same for a group whose children haven't loaded yet.

### Verification

- 21 new unit tests: the resolver (leaf-only rule, fallback, unloaded group), layout geometry (column x/y/height, no cross-axis cost, `BT` flip), edge suppression, the render-context flag, box sizing under windowing, and the row component (ARIA, height, click/Enter, trailing-slot isolation, loading).
- Full `F0Graph` + ODC `Graph` suites green (320 tests).
- `StackedChildren` story + a section in the F0Graph MDX docs.

[FCT-60913]: https://factorialmakers.atlassian.net/browse/FCT-60913?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


sauldom102 (approved):
We should take care of these accessibility issues in some other PR too https://github.com/factorialco/f0/pull/5051#issuecomment-5218395958

PR #5056, opened 2026-08-09
feat(F0RichTextEditor): add insertContent to the imperative handle

gergonzalez:
## Description

`F0RichTextEditorHandle` only exposes `setContent`, so consumers that insert variables into the body (e.g. ATS message templates in the monolith) have to append the variable to the full body and replace everything — which always inserts at the end instead of the cursor, and resets the selection and scroll position. This PR adds an `insertContent` method that inserts at the caret.

## Implementation details

- feat: add `insertContent(content)` to `F0RichTextEditorHandle`, inserting at the last known selection and falling back to the end of the document when the editor was never focused

  <details>
  <summary>Why the fallback?</summary>

  TipTap keeps the last selection when the editor loses focus (e.g. to a
  variables dropdown), so `focus()` restores it before inserting — exactly
  the interaction these consumers need. But a never-focused editor's
  default selection is the document start, which is not a position the
  user ever chose, so in that case the insert lands at the end instead.
  </details>

- test: cover both insert paths (caret insertion after focus, end-of-document fallback when never focused)

### DIFF
--- a/packages/react/src/components/RichText/F0RichTextEditor/F0RichTextEditor.tsx
+++ b/packages/react/src/components/RichText/F0RichTextEditor/F0RichTextEditor.tsx
@@ -100,11 +100,6 @@ export type F0RichTextEditorHandle = {
   focus: () => void
   setError: (error: string | null) => void
   setContent: (content: string) => void
-  /**
-   * Inserts content at the current cursor position (restoring the last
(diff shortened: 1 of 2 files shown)


sauldom102 (approved):

sauldom102 (commented):
Does this work okay as well if the rich text editor field/component is used from a F0Form? 👀

PR #5081, opened 2026-08-11
feat(valueDisplay): add neutral variant to tag cell

RafaMellado:
## Description

Adds an optional `variant` field to the `tag` value-display cell used by OneDataCollection so it can render as a grey filled tag. The HC Planning cost table shows variance values as plain grey pills (grey background, secondary text, no border, no dot), and no existing cell covered that: `tag` is always outlined and `statusTag`'s neutral variant always paints a dot.

[Link to Figma Design](https://www.figma.com/design/RptN7ZgQs5NT40f1ONdy0d/HC-Planning---Cost-Visualization?node-id=364-19249)

## Implementation details

- feat: add `variant?: "default" | "neutral"` to `TagCellValue`; `neutral` renders `bg-f1-background-secondary` + `text-f1-foreground-secondary` with no border

  <details>
  <summary>Why via className instead of a new F0TagRaw prop?</summary>

  The styling is applied through the `className` prop `F0TagRaw` already documents for this purpose ("e.g. to give it a background"), so the stable component's API is untouched and the change is fully additive — existing `{ label, icon }` consumers are unaffected. The variant name follows the system vocabulary: `neutral` is the same token pair `F0TagStatus` uses for its neutral status.
  </details>

- docs: document the neutral variant in the tag cell MDX with a Canvas example
- test: add a `NeutralTagType` story and drop a `color` value the old story passed that the cell type never accepted

Captura de pantalla 2026-08-12 a las 10 24 42


sauldom102 (changes requested):
Why are you changing this component if we have https://f0.factorial.dev/?path=/docs/components-tags-tagraw--documentation ?

sauldom102 (approved):

PR #5084, opened 2026-08-11
feat(OneApprovalHistory): add approval date to steps

gergonzalez:
## Description

Restores the approval date that the legacy approvals history showed before the migration to `OneApprovalHistory` ([FCT-60965](https://factorialmakers.atlassian.net/browse/FCT-60965)): clients reported they can no longer see when each step was approved. Steps now accept an optional, consumer-formatted `approvalDate` string rendered as secondary text below the approvers, keeping the component locale-agnostic. Aligned with Foundations in the ticket (owning team handles it as an improvement).

See the updated `Inbox/ApprovalHistory` → `MultipleApprovals` and `RejectedApproval` stories.

rejected-approval
multiple-approvals

## Implementation details

- feat: add optional `approvalDate` prop to `ApprovalStepProps` and render it under the avatar list for resolved steps
- fix: key the step list with `Fragment key` instead of unkeyed fragment shorthand
- chore: cover `approvalDate` in the `MultipleApprovals` and `RejectedApproval` stories

[FCT-60965]: https://factorialmakers.atlassian.net/browse/FCT-60965?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/react/src/sds/inbox/OneApprovalHistory/ApprovalStep/index.tsx
+++ b/packages/react/src/sds/inbox/OneApprovalHistory/ApprovalStep/index.tsx
@@ -1,3 +1,4 @@
+import { format } from "date-fns"
 import { FC, useMemo } from "react"
 
 import { F0AvatarList } from "@/components/avatars/F0AvatarList"
@@ -9,6 +10,7 @@ import {
   Question as QuestionIcon,
 } from "@/icons/app"
 import { useI18n } from "@/lib/providers/i18n"
+import { useDateFnsLocale } from "@/lib/providers/l10n"
 
 type Status = "waiting" | "pending" | "approved" | "rejected"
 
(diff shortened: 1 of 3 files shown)


packages/react/src/sds/inbox/OneApprovalHistory/ApprovalStep/index.tsx
```diff
@@ -24,6 +24,7 @@ export type ApprovalStepProps = {
   approvalsRequired?: number
   status: Status
   approvers: Approver[]
+  approvalDate?: string
```
sauldom102: I would rather just accept a Date object and format it here internally 👀

gergonzalez: Good point 👍 Done in c46027e14 — it now takes a `Date` and formats it internally with date-fns `PP` using the `L10nProvider` locale (`useDateFnsLocale`), same as `ActivityItem`.

sauldom102 (approved):

PR #5094, opened 2026-08-12
feat(Cocreation): make template selection its own step

warcos-fact:
## Description

Choosing what to start from is a step of its own, not something the user does while talking, so the template gallery and template previews now take the whole frame and cover the chat, which returns only once a resource actually exists. The same move folds the Triage flow into the shared entry actions instead of sending it down its own path, so all five Walkthrough flows now end at the same three choices — Empty Survey, Use a Template, or a named template — and closing a selection surface steps back to whatever opened it rather than abandoning creation.

The change is behavioral — static shots miss the step-back and the chat returning — so it is best reviewed by driving the stories:

- [Welcome Screen](https://one.factorial.dev/?path=/story/patterns-ai-cocreation-walkthrough--welcome-screen)
- [Welcome Screen · No Credits](https://one.factorial.dev/?path=/story/patterns-ai-cocreation-walkthrough--welcome-screen-no-credits)
- [Guided Chat · Triage](https://one.factorial.dev/?path=/story/patterns-ai-cocreation-walkthrough--guided-chat-triage)
- [Guided Chat · Entry Actions](https://one.factorial.dev/?path=/story/patterns-ai-cocreation-walkthrough--guided-chat-entry-actions)
- [Guided Chat · No Credits](https://one.factorial.dev/?path=/story/patterns-ai-cocreation-walkthrough--guided-chat-no-credits)

Worth walking in Triage: pick a type, pick **Use a Template**, then close the gallery — it returns to the entry question with the type still applied.

## Implementation details

- feat: add opt-in `fullscreen` canvas content to `ApplicationFrame`

  <details>
  <summary>Why a new field rather than reusing the chat's open state</summary>

  Closing the chat panel cannot express this: `AiChatStateProvider` drops `canvasContent` and falls back to `sidepanel` whenever the panel closes, so a canvas without a chat was unreachable. `fullscreen` on `CanvasContentBase` instead tells the frame to reserve no width for the seam, and the canvas layer already sits above the chat, so the panel keeps its state and its conversation while covered. The field is optional and unset everywhere else, so no existing consumer changes behavior.
  </details>

- feat: open the templates gallery full width, as cards, in every cocreation flow
- feat: open template previews full width too, so the chat appears only once a resource exists
- feat: step back on close instead of raising "Leave creation?" — the preview returns to the gallery, the gallery to the welcome screen or the entry question that opened it
- feat: fold Triage into the shared entry actions, scoped to the triaged type
- feat: ask Triage's two questions consecutively in one panel

  <details>
  <summary>How the second step depends on the first</summary>

  The entry actions are scoped to the type picked in step one, so `ClarifyingStep.options` in the story mock now also accepts a function receiving the option ids chosen in earlier steps. It is re-read as those change, so going back and picking another type re-derives what follows; a selection the new options no longer contain is dropped rather than left to resolve to an option that is not there.
  </details>

- docs: update the AI Cocreation pages for the fullscreen selection step and the two-question Triage shape
- docs: recapture both Walkthrough figures from the current mock
- test: cover the canvas inset in `ApplicationFrame`, and the new Triage step, its type scoping and its step-back, in a play test


sauldom102 (approved):

desiree-np (approved):
Approving from design. Making selection its own step is the right call, and "match the warning to whether backing out actually strands the user — where there is a step to return to, step back instead of asking" is a better rule than the one it replaces. Removing the "Leave creation?" gate from the gallery's Close is a clear improvement. I drove the flows in Storybook off this branch.

One thing I'd like resolved, here or as a follow-up: **this PR adds an accessibility rule that nothing implements.**

`accessibility.mdx` now says of the gallery and the preview:

> both take the full width and cover the chat, so focus must move into them: a user left behind on the covered conversation is focused on something they can no longer see

But the chat panel still renders unchanged when a `coversChat` canvas is up — no `inert`, no `aria-hidden` ([ApplicationFrame/index.tsx#L448](https://github.com/factorialco/f0/blob/f726a233fe76baf202aed4592ab58c9565e8e509/packages/react/src/patterns/ApplicationFrame/index.tsx#L448)) — and `F0CanvasPanel` has no `role`, `aria-*`, `tabIndex` or `focus()` call at all ([F0CanvasPanel.tsx](https://github.com/factorialco/f0/blob/f726a233fe76baf202aed4592ab58c9565e8e509/packages/react/src/kits/ai/F0CanvasPanel/F0CanvasPanel.tsx)).

So with the gallery open, a keyboard user can tab into the transcript, the composer, and the chat's own ✕ — all fully occluded. That ✕ is what raises "Leave creation?", so creation can be abandoned from a control the user cannot see. A screen-reader user is read a conversation and a text input that aren't on screen.

This is new surface area, not a pre-existing gap: before this PR the canvas always docked beside the chat, so "covered" never happened.

Suggested shape, both small and local to the `coversChat` branch you already added:

- `inert` on the chat panel container while `canvasContent?.coversChat` is true — one attribute, and it takes the whole subtree out of both the tab order and the a11y tree
- move focus into the canvas (its heading or the panel) when a `coversChat` canvas opens, and return it to whatever opened it on close — which is the step-back the docs already describe, so focus just follows the step

Happy for this to land as a follow-up if you'd rather not grow the diff — but I'd not leave the docs asserting a behaviour the implementation doesn't have, since that's what the next consumer of `coversChat` will build against.

PR #5095, opened 2026-08-12
feat(F0Select): add inline variant

pedroruizpareja:
## Description

Adds a compact, borderless inline variant to F0Select for row-level role and access controls. Inline uses one fixed 32px trigger aligned with F0 typography, icon, and radius tokens; its popup, option rows, selection indicator, and action footer reuse the existing F0Select implementation, leaving the field variant and Dropdown unchanged. Footer actions also honor their existing disabled state in both select variants.

## Type of change

- [Prototype reference](https://shareabstraction.vercel.app/p/policies/pol-011)
- Visual coverage is captured by the `Components/Select/Inline` Open, Dark Mode, and Snapshot stories.

## Implementation details

- feat: add a compact, borderless `variant="inline"` for single, non-clearable row controls
- fix: align the inline trigger with F0 using a fixed 32px height, 14px label typography, 10px radius, and the default icon token
- refactor: keep `size` exclusive to the field variant instead of redefining the field size scale for inline
- refactor: reuse the standard F0Select popup, option rows, selected indicator, and action footer without changing Dropdown
- feat: make inline popups content-sized by default while preserving the explicit `fitContentWidth=false` override
- fix: forward the existing disabled state to footer action buttons and prevent their callbacks from running
- docs: document inline usage, fixed trigger geometry, limitations, shared popup behavior, footer action states, keyboard interaction, and accessibility guidance
- test: cover field compatibility, the inline type contract, controlled selection, trigger tokens, popup sizing, standard popup reuse, enabled and disabled footer actions, and Storybook interactions

## Known accessibility debt

The Open and Snapshot stories keep axe running in `todo` mode for the shared popup's existing `aria-hidden-focus` violation. This PR adds no accessibility skip and does not change the shared popup implementation.

## Context

- [CIAM-989 — Enhance the sharing component UI](https://factorialteam.atlassian.net/browse/CIAM-989)

---
> - factorial-f0
> - factorial-dev-workflow
> - f0-pr
> - factorial-ci
> - f0-unit-testing
> - f0-component-patterns
> - f0-quality-gate


sauldom102 (changes requested):
We seem to be introducing another mdx file for the Select component instead of introducing that in existing docs, now we can't access https://66a7a8d7d124220c363457cc-icfinfcjuh.chromatic.com/?path=/docs/components-select--documentation

You can compare it with https://f0.factorial.dev/?path=/docs/components-select--documentation

sauldom102 (changes requested):
When having variant="inline" we're seeing it like this:

Image

sauldom102 (approved):

desiree-np (approved):

PR #5097, opened 2026-08-12
fix(F0DataChart): stop totalling grouped bar tooltips

igorsafonov-gif:
## Description

A bar chart's item tooltip showed the hovered value's share of the category total whenever more than one series was visible, grouped charts included — but the sum of side-by-side bars is only a real quantity when those bars happen to be additive, and nothing tells the component whether they are. On "average base salary by workplace (male vs female)" it added three averages into a €199,229.86 "total" and reported the hovered bar as "52.2% of total", making separate measurements read as slices of a pie. Both rows are now gated on `stacked`, where the total genuinely is the length of the bar being hovered; grouped tooltips show the value (and target) alone.

**This is a trade-off, not a strict improvement.** A grouped headcount by gender *is* parts of a whole — male and female bars sum to the category's real headcount — and that chart loses a correct, useful row here. `stacked` is a proxy: layout is the only signal available, since the component has no way to know whether a measure is additive. Giving up a right number on the additive groups is the better side of the trade against shipping a wrong one on averages, rates and medians, but it is a trade rather than a strictly better rule. If a consumer turns up needing the total back, the escape hatch is an opt-in (`showCategoryTotal`) rather than reverting the guard.

**Before** — `Bar/Multiple Series`, hovering New York's headcount. Headcount 145 is reported as "87.9% of total" of a 165 built from headcount + open positions + turnovers: three measures that share an axis and nothing else.

before-grouped

**After** — same bar, same hover. The invented total and its share are gone.

after-grouped

**Unchanged** — `Bar/Stacked` still shows both rows. Here the total is the height of the bar under the cursor, and the segments really are parts of it.

stacked-unchanged

## Implementation details

- fix: show the share-of-total and total rows only on stacked bar charts

  <details>
  <summary>Why?</summary>

  The guard was `visibleSeries.length > 1 && !hasMixedSigns` — it asked how many series there were, never how they were laid out. Adding `stacked &&` is the whole behavior change; the rest of the diff is the comment explaining which chart shapes have a total at all, and tests.

  Non-additive metrics make the bug obvious (averages, rates, medians), and the `Bar/Multiple Series` shape is wrong on its own terms too: headcount, open positions and turnovers are not parts of one another. Additive groups are the case this costs something — see the trade-off note above.
  </details>

- perf: compute the category sum only when a total can render

  <details>
  <summary>Why?</summary>

  `categoryValues` (a map over every visible series) and `total` fed a `showTotal` that is now always `false` on grouped charts, so both ran on every hover and were thrown away. Gating them on the same condition also makes the intent readable off the code rather than only out of the comment.
  </details>

- test: cover a grouped chart showing no share and no total, and pin the existing multi-series case to `stacked` — which is what it was actually testing

### DIFF
--- a/packages/react/src/kits/F0DataChart/__tests__/BarChart.test.tsx
+++ b/packages/react/src/kits/F0DataChart/__tests__/BarChart.test.tsx
@@ -1072,6 +1072,7 @@ describe("BarChart — item tooltip", () => {
     render(
       <F0DataChart
         type="bar"
+        stacked
         categories={["A", "B"]}
         series={[
           { name: "S1", data: [30, 10] },
@@ -1088,6 +1089,33 @@ describe("BarChart — item tooltip", () => {
     expect(html).toContain("30.0%")
     expect(html).toContain("of total")
     expect(html).toContain((100).toLocaleString())
+  })
+
+  // Side-by-side bars are independent measures, not parts of a whole: summing
+  // three average salaries produces a number that is not an average of
+  // anything, and a share of that sum makes the bars look like slices of a pie.
+  it("omits share and total on a grouped chart, however many series", () => {
+    render(
+      <F0DataChart
+        type="bar"
+        categories={["Barcelona"]}
+        series={[
+          { name: "Not specified", data: [104000] },
+          { name: "Male", data: [45700] },
+          { name: "Female", data: [49500] },
+        ]}
+      />
+    )
+    const html = getTooltipFormatter()?.({
+      name: "Barcelona",
+      seriesName: "Not specified",
+      value: 104000,
+      dataIndex: 0,
+    })
+    expect(html).toContain((104000).toLocaleString())
+    expect(html).not.toContain("of total")
+    expect(html).not.toContain("52.2%")
+    expect(html).not.toContain((199200).toLocaleString())
   })
 
   it("hides share of total for single-series charts", () => {
--- a/packages/react/src/kits/F0DataChart/__tests__/BarChart.test.tsx
+++ b/packages/react/src/kits/F0DataChart/__tests__/BarChart.test.tsx
@@ -1091,9 +1091,10 @@ describe("BarChart — item tooltip", () => {
     expect(html).toContain((100).toLocaleString())
   })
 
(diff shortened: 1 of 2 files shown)


sauldom102 (approved):

zygisS22 (commented):
The cleanest PR of the stack — one condition, a stated reason, and before/after/unchanged screenshots that make the case without argument. Gating on `stacked` is the right proxy, since the component has no way to know whether a measure is additive. Two notes.

## 🟠 The reasoning is a little stronger than the change supports, and it costs a real number

> Non-additive metrics make the bug obvious (averages, rates, medians), but it was wrong for additive ones too: a grouped headcount chart implied its three bars were parts of a whole they are not.

That's true of the `Bar/Multiple Series` story, where the three series are headcount, open positions and turnovers — genuinely unrelated quantities. It isn't true of a **grouped headcount by gender**: side-by-side male/female bars *are* parts of a whole, and their sum is the category's real headcount. That chart loses a correct and useful row here.

The fix is still right — layout is the only signal available, and shipping a wrong number on averages is worse than dropping a right one on counts. But it's a trade-off rather than a strict improvement, and worth saying so in the description so the next person doesn't read "grouped totals are always meaningless" as settled. If a consumer turns up needing it back, an opt-in (`showCategoryTotal`) would be the escape hatch rather than reverting the guard.

## 🟡 Dead work on the grouped path

`categoryValues` and `total` are now computed on every hover regardless, but both are only read under `showTotal`. On a grouped chart that's a `map` over every visible series plus a `reduce`, per tooltip render, thrown away. Trivial, but short-circuiting when `!stacked` would also make the intent read off the code.

## What's good

- The diff is one `stacked &&`; everything else is the comment and the tests. That ratio is the right one for a behaviour change to a shared component.
- Adding `stacked` to the existing multi-series test and saying plainly that it's "what it was actually testing" is honest about the fact that the old test was passing for the wrong reason.
- The comment now separates the two independent conditions — *does this shape have a total at all* (stacking) versus *is the total meaningful* (mixed signs, single series) — which is what stops someone re-adding the first one later.
- The screenshots include the **unchanged** stacked case, which is the one a reviewer would otherwise have to take on trust.

PR #5102, opened 2026-08-13
feat(F0DataChart): show how many entities a segment covers

igorsafonov-gif:
## The problem

A chart states its value but not the population behind it. An average salary of €52,400 describes a very different team over 12 people than over 1,200, and nothing on screen says which. Today the only way to find out is to build a second widget.

## What this adds

An optional `context` on **bar, line and pie** — counts carried alongside the plotted marks rather than as marks of their own:

```
● Average base salary        ● United States
  United States                €48,347.66
  €48,347.66                   12.4% of total
  256 people                   €390,237.56 total
                               256 people
```

Renderers draw `series` alone, so a count can never be mistaken for data. It reaches the reader three ways, all from the single `chartDataToTabular` that feeds them: the **tooltip row**, the **table view**, and the **CSV and Excel exports**.

`contextLabelFormatter` names what is being counted *given the count*, so the label agrees with it in number — **"1 person"** against **"256 people"**. It falls back to the entry's own `name`, which cannot inflect, so consumers that don't pass one still get a sensible label.

## Decisions worth reviewing

**Counts render as plain integers, never through `valueFormatter`.** That formatter carries the plotted measure's unit; a headcount shown as "€1,204.00" is a different quantity, not a differently styled one.

**Entries mirror `series` by index.** A chart split by a dimension carries a count per cell — the women in Engineering are not the people in Engineering. A single entry serves every series, since a category's population doesn't change with the measure plotted against it.

**Funnel and scatter are excluded.** A funnel's stages are sequential steps of one process, where the count is usually the plotted value itself. A scatter names each point with an entity dimension (the schema's own example is `employee_full_name`), so every count would be 1.

**The line multi-series card takes one shared row, not one per series.** That card already lists every series; per-series counts would double its length.

## One bug worth calling out

`buildChartProps` needed the counts on **both** of its routes.

`detectDataShape` reports any array-series data as `"bar"` — it honours the hint for pie-vs-funnel but not bar-vs-line. So a line item's detected shape never equals its target type, and **every line chart renders through the cross-type transform branch on every render**, not just after a user transform. That branch rebuilds props from `defaultChartConfig` plus a fixed allowlist.

Passing the counts on the native path alone worked on bar and pie and silently did nothing on line. `chartLabelDefaults.test.ts` now pins both routes — the line case fails without the fix.

I deliberately did **not** change `detectDataShape` to honour the hint. That would put line back on the native path and is probably the right long-term fix, but it changes which config every existing line chart is built from — a deliberate change, not a side effect of this one.

## Testing

- 8 new tooltip tests across bar, line and pie: label, inflection, the plain-integer rule, per-series lookup on a split, missing counts, and the no-context fallback.
- Table/export coverage for bar and pie, including that the column is keyed independently of its data-provided label.
- `buildChartProps` tests pinning both prop-building routes.
- Storybook: hover *Satisfaction Scores* in `patterns-analyticsdashboard--mixed-dashboard`, and switch it to the table view.

Verified end to end in the Factorial app against real cube data on all three chart types.

The `BarChart.test.tsx` label-width failures noted in an earlier revision of this description are gone: they came from the unreviewed base commit, and #5097's merged version fixes them. This branch is now rebased straight onto `main` and the whole `F0DataChart` / `F0AnalyticsDashboard` suite passes.

### DIFF
--- a/packages/react/src/kits/F0DataChart/__tests__/BarChart.test.tsx
+++ b/packages/react/src/kits/F0DataChart/__tests__/BarChart.test.tsx
@@ -1300,6 +1300,189 @@ describe("BarChart — item tooltip", () => {
     expect(description).not.toContain("Category 21")
   })
 
+  // The count of entities behind a bar: how many people an average was taken
+  // over, which the plotted value alone never says.
+  describe("segment context", () => {
+    const salary = {
+      categories: ["Engineering", "Sales"],
+      series: [{ name: "Average salary", data: [52400, 41000] }],
+    }
+
+    it("adds a row naming what it counts", () => {
+      render(
+        <F0DataChart
+          type="bar"
+          {...salary}
+          context={[{ name: "Active headcount", data: [1204, 380] }]}
+        />
+      )
+
+      const html = getTooltipFormatter()?.({
+        name: "Engineering",
+        seriesName: "Average salary",
+        value: 52400,
+        dataIndex: 0,
+      })
+      expect(html).toContain("1,204")
+      expect(html).toContain("Active headcount")
+    })
+
+    // The value formatter carries the plotted measure's unit. Running a
+    // headcount through it renders a different quantity, not a restyled one.
+    it("formats the count as a plain integer, never in the measure's unit", () => {
+      render(
+        <F0DataChart
+          type="bar"
+          {...salary}
+          valueFormatter={(v) => `€${v.toLocaleString("en-US")}`}
+          context={[{ name: "Active headcount", data: [1204, 380] }]}
+        />
+      )
+
+      const html = getTooltipFormatter()?.({
+        name: "Engineering",
+        seriesName: "Average salary",
+        value: 52400,
+        dataIndex: 0,
+      })
+      expect(html).toContain("€52,400")
+      expect(html).toContain("1,204")
+      expect(html).not.toContain("€1,204")
+    })
+
+    it("reads a single entry for every series", () => {
+      render(
+        <F0DataChart
+          type="bar"
+          categories={["Engineering"]}
+          series={[
+            { name: "Average salary", data: [52400] },
+            { name: "Median salary", data: [50000] },
+          ]}
+          context={[{ name: "Active headcount", data: [1204] }]}
+        />
+      )
+
+      const html = getTooltipFormatter()?.({
+        name: "Engineering",
+        seriesName: "Median salary",
+        value: 50000,
+        dataIndex: 0,
+      })
+      expect(html).toContain("1,204")
+    })
+
+    // A split chart's segment is a cell: the women in Engineering are not the
+    // people in Engineering, so each series carries its own counts.
+    it("picks the hovered series' entry when the chart is split", () => {
+      render(
+        <F0DataChart
+          type="bar"
+          categories={["Engineering"]}
+          series={[
+            { name: "Female", data: [48900] },
+            { name: "Male", data: [55100] },
+          ]}
+          context={[
+            { name: "Active headcount", data: [612] },
+            { name: "Active headcount", data: [592] },
+          ]}
+        />
+      )
+
+      const html = getTooltipFormatter()?.({
+        name: "Engineering",
+        seriesName: "Male",
+        value: 55100,
+        dataIndex: 0,
+      })
+      expect(html).toContain("592")
+      expect(html).not.toContain("612")
+    })
+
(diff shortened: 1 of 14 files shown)


zygisS22 (commented):
Reviewed together with factorial#109763. The `detectDataShape` finding is a real discovery and handling it by covering both prop-building routes — rather than "fixing" shape detection inside this PR — is the right restraint. Notes below.

## 🟠 A split line chart loses its counts entirely

`contextRow` is only reachable on the line's single-point branch. The multi-series card takes `sharedContext`, which is gated on `context?.length === 1`:

```ts
const sharedContext =
  context?.length === 1 && points[0] ? contextRow(...) : undefined
```

So when there is one entry *per series* — the split case — the multi-series card renders no context row at all. And because line uses an axis trigger, there is no single mark the reader can hover to get the per-cell count instead. The result is that the case the PR is built on ("the women in Engineering are not the people in Engineering") is unreachable on a line chart except through the table view.

The stated reason (a count per series would double the card's length) is fair for 8 series, less so for 2. Options: append the count into each series row's label, or render per-series rows when the count is small.

## 🟠 The transform path drops `context` for a pie target, but the table keeps it

`buildChartProps` allowlists only bar and line:

```ts
const contextForTarget =
  data.context && (targetType === "bar" || targetType === "line")
    ? { context: data.context } : {}
```

but `pieToTabular` *does* emit the context column. So after a user transforms a bar into a pie, the table view shows a "People" column that the tooltip no longer has. The indices still line up (pie slices come from the same categories the counts were aligned to), so passing it through is safe — and `buildNativeChartProps` already does exactly that for the native pie path, which makes the transform path the odd one out.

## 🟡 Two silent-wrong-value fallbacks

Both prefer showing a plausible wrong number over showing nothing:

- `context[series.findIndex((s) => s.name === seriesName)] ?? context[0]` — with N series and between 2 and N-1 context entries, an unmatched series silently reports entry 0's count. The `?? context[0]` fallback is right for the documented "one entry serves every series" case; suggest gating it on `context.length === 1` and returning `undefined` otherwise.
- Pie's `entry?.data[p.dataIndex ?? 0]` — if a params object ever arrives without `dataIndex`, every slice reports slice 0's count. Bailing out reads better than defaulting.

Neither is likely to fire today; both would be invisible when they did.

## 🟡 The counts never reach the aria description

`options.aria.label.description` already enumerates each series' values **and its targets** — the same rows the tooltip shows — but not `context`. So a screen-reader user gets the value and never the population. Worth adding for parity, and especially so given the sibling PR (#109630) explicitly sells "the exact figure on the tooltip and in the aria description" as the accessibility story.

## ℹ️ Notes

- **The failing-tests warning doesn't match CI.** The description says 7 `BarChart.test.tsx` tests fail on this branch and identically on `ce43d4d8e`, but the **⚛️ REACT Unit Tests** check is green on this PR. If they're local-only (font metrics / `measureText` differing between macOS and the CI container) that's worth saying, because #109630's whole argument rests on locally measured label widths — a local/CI divergence there would be load-bearing. Either way the note as written sends reviewers chasing a phantom.
- `count.toLocaleString()` uses the browser locale rather than the app's. I checked before flagging it: this matches the kit's own default formatter (`options.ts:573`) and the i18n provider exposes no locale, so it isn't a new deviation — just noting that once #109630 lands, a tooltip can show `46.390,86 €` (app locale) directly above `1,204` (browser locale).
- `shared.contextLabelFormatter` is spread into the funnel / radar / gauge / heatmap / scatter branches too, where the prop doesn't exist on the target type.

## What's good

- **Reusing `keys` rather than inventing a mechanism.** The "header labels come from data and can collide, row keys can't" problem was already solved for scatter, and this uses it as-is — including the test that pins the collision case explicitly.
- **The `detectDataShape` write-up.** Finding that every line chart takes the cross-type transform on every render is a genuinely non-obvious discovery, and `chartLabelDefaults.test.ts` pinning *both* routes is the right way to make it stick. Deliberately not changing `detectDataShape` in this PR is the correct call and well argued.
- Counts formatted as plain integers rather than through `valueFormatter` — the "€1,204.00 is a different quantity, not a differently styled one" framing is exactly right, and it's tested on all three chart types.
- `contextLabelFormatter` taking the count is a nice API shape: inflection is the consumer's problem and it can't be solved with a static string.
- `renderValueTooltip` already filtering falsy rows means the `false &&` and `undefined` rows are safe — good that this leaned on the existing contract instead of adding guards.
- Memo dependency arrays updated in all three hooks.

PR #5103, opened 2026-08-13
feat(F0AnalyticsDashboard): add info hint to widget headers

igorsafonov-gif:
> [!IMPORTANT]
> **The `breaking changes` label is a false positive — no consumer has to change anything.**
> The ⓘ type was renamed, `TableHeaderInfo` → `InfoHintContent`, with the same four fields and `export type TableHeaderInfo = InfoHintContent` left in place, so every existing import and every existing column config still compiles. The API surface check compares type names as plain text, so a rename behind an alias always reads as a retype — it flags the three CSV helpers whose signature mentions it. Nothing here warrants `feat!:`; this stays a minor release.

## Description

A dashboard widget can now carry `info` — help copy for what its figure actually measures, revealed by an ⓘ beside the title. The affordance already existed on table column headers, where a consumer can explain a column without stretching its label; charts, KPIs and collections had no equivalent, so a number's meaning lived only in the widget title. Its hover card was private to `TableHead`, so this PR lifts it into a shared internal component and points both surfaces at it, rather than growing a second implementation.

## Type of change

From the `Patterns/AnalyticsDashboard` → `MixedDashboard` story.

**At rest** — the ⓘ sits beside the title only on the widgets that carry `info` (the "Headcount by Department" chart and the "Avg. Salary" KPI); the others are untouched.

Dashboard with the info icon beside two widget titles

**Chart** — hovering the ⓘ reveals what the plotted figure means, with the link through to the full entry.

Hover card on the chart title explaining the Active headcount measure

**KPI** — the same affordance and the same copy a table column header already offers, so a measure reads identically wherever it appears.

Hover card on the Avg. Salary KPI explaining the Average base salary measure

## Implementation details

- feat: add `info?: string | InfoHintContent` to `DashboardItemBase`, rendered as an ⓘ beside the widget title by `DashboardItem` and passed through by `ChartItem`, `MetricItem` and `CollectionItem`

  <details>
  <summary>How it differs from the two fields next to it</summary>

  `description` states what the widget shows and `explanation` says how it was computed (behind the ⋯ menu). `info` answers a third question — what the figure itself means — and is the same copy a table column header already offers, so a measure reads identically whether it appears as a column, an axis or a KPI.

  The icon is `shrink-0`, so a long title truncates around it instead of squeezing it out of the row. It is also kept in the error header: a reader looking at a failed widget is exactly the one asking what it was meant to show.
  </details>

- refactor: extract `TableHead`'s private hover card into `lib/InfoHint`, shared by the column header and the widget header

  <details>
  <summary>Why?</summary>

  One component now owns both forms — a plain string renders a non-interactive `Tooltip`, an `InfoHintContent` renders a `HoverCard` that can carry a link action. `TableHeaderInfo` becomes an alias of `InfoHintContent`, so the public type and every existing consumer are unchanged, and the table keeps its own `stopPropagation` wrapper (reading a column's help is not asking to sort by it).
  </details>

- test: cover the widget header's ⓘ — accessible name, hover reveal, link action, and its presence in the error state
- test: add `Components/Primitives/InfoHint` stories — the structured card with and without a link, the plain-string tooltip, and the trigger's accessible name, plus a consolidated `Snapshot` that opens a card so Chromatic sees the state the copy lives in

  <details>
  <summary>Why the play functions focus instead of hover</summary>

  The synthetic pointer events `userEvent.hover` emits never reach Radix's `pointerenter` handler in a real browser, so the card stays closed and the story errors. Focusing the trigger opens the same card, and it is the keyboard path a reader takes anyway. The hover path stays covered in jsdom by `TableHead.test.tsx`.
  </details>

- chore: show the ⓘ on a chart and a KPI in the `MixedDashboard` story mocks

## Verification

`pnpm tsc` clean, `oxlint` clean, and 111 tests pass across `F0AnalyticsDashboard` + `OneTable` — including the existing `TableHead` suite unchanged, which is what proves the extraction preserved the column-header behaviour.

---
> - factorial-f0
> - factorial-ci
> - factorial-pr


zygisS22 (commented):
Clean extraction — one component now owns both forms, `TableHead` keeps its own `infoIcon = InfoCircleLine` default and its `stopPropagation` wrapper, and leaving the existing `TableHead` suite untouched is the right proof that the move preserved behaviour. Two accessibility points are worth settling before this becomes the affordance on every widget in the product.

## 🟠 The trigger's accessible name is the widget title, so it never says what it does

```tsx
aria-label={info.label ?? label}   // InfoHint
<InfoHint info={info} label={title} />   // DashboardItem
```

A screen-reader user hears **"Headcount by team, button"** — no indication it reveals help copy, and a duplicate of the `<h3>` immediately before it. The new test codifies it:

```tsx
screen.getByRole("button", { name: "Headcount by team" })
```

`InfoHintContent.label` exists precisely for this, but nothing sets it, so it always falls through to the bare title. Something like "More information about {title}" (via `useI18n`) would fix both the ambiguity and the duplication. This is inherited from `TableHead` rather than introduced here — but a column header is one control among many, whereas this puts the same pattern on every chart, KPI and collection.

## 🟠 Worth confirming the ⓘ works on touch

`StructuredHint` renders a `<button type="button">` with no `onClick` — the card is opened by the HoverCard's own pointer/focus handling. On a tablet that's likely to be a button that visibly does nothing when tapped. Radix's HoverCard is documented as a mouse affordance, so I'd want this checked on a real touch device rather than reasoned about; if it doesn't open, toggling `open` from `onClick` is a small addition and the `open` state is already lifted for the link case.

Same caveat as above: pre-existing in the table, but dashboards get read on tablets more than admin tables do.

## 🟡 The canonical type isn't exported under its canonical name

`InfoHintContent` is now the shape for column headers *and* widget headers, but it's only reachable from outside the package as `TableHeaderInfo` — a table-specific alias. A host typing a dashboard widget's `info` has to import the table's type to do it. Worth exporting `InfoHintContent` publicly and keeping `TableHeaderInfo` as the back-compat alias it now is.

## 🟡 Error-state header alignment

The error branch wraps a non-truncating `<h3>` in `flex min-w-0 items-center`. The success branch is fine because `OneEllipsis` keeps the title on one line, but in the error state a long title wraps, and `items-center` then centres the ⓘ against the full two-line block instead of sitting beside the first line. `items-start` (or reusing `OneEllipsis` here too) would keep it anchored.

## What's good

- **Extraction rather than a second implementation.** The obvious shortcut was to copy `HeaderInfo` into `DashboardItem`; lifting it into `lib/InfoHint` and pointing both surfaces at it is the version that doesn't drift.
- Keeping the ⓘ in the error header, with the reason written down — "a reader looking at a failed widget is exactly the one asking what it was meant to show" — is a good call that would otherwise get silently dropped by whoever touched that branch next.
- `shrink-0` on the icon wrapper so a long title truncates around it, again with the reason in the comment.
- The distinction between `description` / `explanation` / `info` is stated in the type docs, which is what stops the next person adding a fourth.
- Test coverage hits the accessible name, hover reveal, link action and the error state.

sauldom102 (approved):

PR #5104, opened 2026-08-13
feat(ai): mention charts and data points in chat

igorsafonov-gif:
## Description

Lets a user bring a dashboard widget — or a single data point — into a conversation with One, so they can ask about what they're looking at without retyping it.

Three routes, all landing in the composer's existing quote chip:

| Gesture | Quote |
|---|---|
| Drag a widget onto the chat panel | `Headcount by workplace` |
| ⋯ menu → **Ask One** | `Headcount by workplace` |
| Click a mark → **Ask One** in the popover | `Headcount by workplace — Barcelona office`<br>`Male: 18` |

The drop invitation appears the **moment a drag starts**, not when the cursor reaches the chat, so the target is visible before the user aims for it.

Nothing here is on by default: every entry point is gated on an AI chat actually being mounted, so `F0AnalyticsDashboard` used standalone is unchanged.

## Type of change

Best seen live — Storybook story **`AI/F0AiChat/Widget drop`** puts an editable dashboard beside an open chat and exercises all three routes. Screenshots to follow on the PR.

### Why the file-drop plumbing couldn't be reused

The chat already has a drop state for files, but the two gestures are different browser mechanisms:

| | File drop | Widget drag |
|---|---|---|
| Mechanism | Native HTML5 drag-and-drop | Plain pointer gesture |
| Events | `dragenter`/`dragover`/`drop` + `dataTransfer` | `pointerdown` on grip → `document` `pointermove`/`pointerup` |
| Hit-test | Browser names the element | Manual `getBoundingClientRect()` math |

A pointer drag fires no `dragenter`, so the existing handlers never see it. `DropOverlay` gains a `mode` (`"files" | "discuss"`) and is reused; the wiring is new.

### How the drag and the chat talk

`DashboardGrid` dispatches `f0:widget-drag-start` / `f0:widget-drag-end` on `window` (names shared via `lib/dnd/widgetDragEvents.ts`); `ChatWindow` listens. Plain DOM events rather than shared state because the producer is a `pattern` and the consumer a `kit` with no useful common ancestor, and neither should import the other. The chat marks itself with `data-ai-chat-dropzone`, which the grid hit-tests.

### `onPointClick` on the chart kit

`F0DataChartCommonProps` gains `onPointClick`, so all nine variants get it from one addition. `usePointClick` binds once at the ECharts instance level — item-level click params are uniform across series types, so no per-chart variant is needed — and follows the existing `useLegendSelection` shape. Charts stay inert unless a handler is passed.

Quoted values format through the chart's own `tooltipValueFormatter`/`valueFormatter`, so the number reads as the tooltip did rather than as a raw float.

### Two bugs worth calling out

**A stale-closure drop, only reproducible in a real browser.** The release handler first read the pending title from state. A fast drag delivers the drag-start announcement and the release inside one React batch, so no re-render separates them and the handler saw the pre-drag `null` — the quote was silently discarded. The payload now lives in a ref; state only drives the render. There's a regression test that puts both events in a single `act`, and it fails if the read goes back to state.

**The grid would reorder *and* quote.** `resolveDropTarget` matches rows on Y alone (X only picks a slot within the row), and the chat shares that Y band — so releasing over the chat painted a drop indicator behind the overlay *and* committed a reorder. Guarded at the top of the function; returning `null` also makes the existing `if (draggedId && target)` skip `commitDrop`. Uses rect containment rather than `document.elementFromPoint`, which jsdom 26 doesn't implement.

### Notes for review

- `patterns/F0AnalyticsDashboard` now imports `useAiChat` (`patterns → kits`). `check-cycle-dependencies` reports no new cycles; `patterns/ApplicationFrame` sets the precedent.
- `PointActionPopover` mirrors `ReplyPopover` (the quote-on-text-selection affordance) so quoting looks the same wherever it starts. It must portal to `document.body` — charts sit inside `overflow-hidden` cards and scroll containers that would clip it.
- One new i18n key, `ai.dashboardItem.askOne`. Deliberately **not** `ai.ask`: hosts override that one (factorial renders it as plain "Ask") and it drives two other surfaces. `ai.dropWidgetToDiscuss` is the other new key.
- Clicking a mark also dismisses the hover tooltip, so two floating panels never stack over one mark.

### Verification

- Full unit suite: 6253 passed, 13 skipped. `tsc`, `oxlint` and `oxfmt` clean.
- 25 new tests across `usePointClick`, `PointActionPopover`, the widget-drop flow and the ⋯ menu.
- Both guard tests confirmed non-vacuous — each fails when its fix is reverted.
- Manually exercised in Storybook (`AI/F0AiChat/Widget drop`) with real mouse input, and end-to-end in a local Factorial against this build.

### Known gaps

- The click popover fires on any mark; if that reads as noisy in use, the alternative is requiring the chat to be open already.
- Not exercised on touch, where the grid's own drag gesture is the prior constraint.
- The `—` and `:` separators in the data-point quote are literal, not translated.

---
> - factorial-pr

### DIFF
--- a/packages/react/src/kits/F0DataChart/__tests__/usePointClick.test.tsx
+++ b/packages/react/src/kits/F0DataChart/__tests__/usePointClick.test.tsx
@@ -1,0 +1,214 @@
+import type * as echarts from "echarts"
+import { useRef } from "react"
+import { describe, expect, it, vi } from "vitest"
+
+import { zeroRender as render } from "@/testing/test-utils"
+
+import type { F0DataChartPointClick } from "../types"
+
+import { usePointClick } from "../utils/usePointClick"
+
+/** Minimal stand-in for an ECharts instance: records handlers so tests can fire them. */
+function makeChartStub() {
+  const handlers: Record<string, ((params: unknown) => void)[]> = {}
+  const dispatched: { type: string }[] = []
+  let disposed = false
+  return {
+    handlers,
+    dispatched,
+    dispose: () => {
+      disposed = true
+    },
+    instance: {
+      on: (event: string, fn: (params: unknown) => void) => {
+        handlers[event] = [...(handlers[event] ?? []), fn]
+      },
+      off: (event: string, fn: (params: unknown) => void) => {
+        handlers[event] = (handlers[event] ?? []).filter((h) => h !== fn)
+      },
+      isDisposed: () => disposed,
+      dispatchAction: (action: { type: string }) => {
+        dispatched.push(action)
+      },
+    } as unknown as echarts.ECharts,
+  }
+}
+
+const Harness = ({
+  chart,
+  onPointClick,
+}: {
+  chart: echarts.ECharts
+  onPointClick?: (point: F0DataChartPointClick) => void
+}) => {
+  const ref = useRef<echarts.ECharts | null>(chart)
+  usePointClick(ref, onPointClick)
+  return null
+}
+
+const fire = (stub: ReturnType<typeof makeChartStub>, params: unknown) =>
+  stub.handlers["click"]?.forEach((h) => h(params))
+
+describe("usePointClick", () => {
+  it("reports the clicked mark, normalised", () => {
+    const stub = makeChartStub()
+    const onPointClick = vi.fn()
+    render(<Harness chart={stub.instance} onPointClick={onPointClick} />)
+
+    fire(stub, {
+      componentType: "series",
+      seriesName: "Male",
+      seriesIndex: 1,
+      dataIndex: 0,
+      name: "Barcelona office",
+      value: 18,
+      event: { event: { clientX: 420, clientY: 260 } },
+    })
+
+    expect(onPointClick).toHaveBeenCalledWith({
+      seriesName: "Male",
+      category: "Barcelona office",
+      value: 18,
+      dataIndex: 0,
+      seriesIndex: 1,
+      clientX: 420,
+      clientY: 260,
+    })
+  })
+
+  it("takes the measure from the last entry of a tuple value (scatter)", () => {
+    const stub = makeChartStub()
+    const onPointClick = vi.fn()
+    render(<Harness chart={stub.instance} onPointClick={onPointClick} />)
+
+    fire(stub, {
+      componentType: "series",
+      seriesName: "Salary vs tenure",
+      seriesIndex: 0,
+      dataIndex: 4,
+      name: "",
+      value: [3, 52000],
+    })
+
+    expect(onPointClick).toHaveBeenCalledWith(
+      expect.objectContaining({ value: 52000, category: "" })
+    )
+  })
+
+  it("dismisses the hover tooltip on a pick", () => {
+    const stub = makeChartStub()
+    render(<Harness chart={stub.instance} onPointClick={vi.fn()} />)
+
+    // Otherwise the tooltip and whatever answers the click stack over one mark.
+    fire(stub, { componentType: "series", seriesName: "Male", value: 18 })
+
+    expect(stub.dispatched).toEqual([{ type: "hideTip" }])
+  })
+
+  it("leaves the tooltip alone when the click is not a pick", () => {
+    const stub = makeChartStub()
+    render(<Harness chart={stub.instance} onPointClick={vi.fn()} />)
+
+    fire(stub, { componentType: "legend", name: "Male" })
+    fire(stub, { componentType: "series", seriesName: "Male", value: null })
+
+    expect(stub.dispatched).toEqual([])
+  })
+
+  it("falls back to 0,0 when the event carries no position", () => {
+    const stub = makeChartStub()
+    const onPointClick = vi.fn()
+    render(<Harness chart={stub.instance} onPointClick={onPointClick} />)
+
+    fire(stub, { componentType: "series", seriesName: "Male", value: 18 })
+
+    expect(onPointClick).toHaveBeenCalledWith(
+      expect.objectContaining({ clientX: 0, clientY: 0 })
+    )
+  })
+
+  it("ignores clicks that are not on a series", () => {
+    const stub = makeChartStub()
+    const onPointClick = vi.fn()
+    render(<Harness chart={stub.instance} onPointClick={onPointClick} />)
+
+    // Axis labels and legend entries raise `click` too, and carry no value.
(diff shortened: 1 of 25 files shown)


zygisS22 (commented):
Big surface, and the two "bugs worth calling out" are both real finds — the stale-closure drop in particular is the kind of thing that only shows up in a browser and then never gets a test. Before the notes, three claims I verified rather than assumed, all of which hold:

- **`useAiChat()` without a provider is safe.** It returns the `NO_PROVIDER_CONTEXT` Proxy, and `enabled` is in `FALSE_KEYS` — so it's `false`, not a truthy `noop`. Standalone `F0AnalyticsDashboard` really is unchanged.
- **`usePointClick`'s single bind can't go stale.** `useEChartsInstance` inits once (`useEffect` deps `[ref]`) and only `setOption`s afterwards, so there's no re-created instance to rebind to. Hook ordering also guarantees `chartRef.current` is populated before this effect runs.
- **No reorder regression when the chat is closed.** `data-ai-chat-dropzone` sits inside the `AnimatePresence` block gated on `isVisible`, so it's unmounted and `resolveDropTarget` finds nothing.

## 🔴 The two new i18n keys have no host-side companion

`ai.dashboardItem.askOne` and `ai.dropWidgetToDiscuss` are added to f0's defaults, but Factorial's bridge (`frontend/src/modules/core/lib/f0-provider.tsx`) maps every key explicitly, and its `ai.dashboardItem` block is:

```ts
dashboardItem: {
  chartType: i18n.t('common.ai.dashboard_item.chart_type'),
  errorTitle: i18n.t('common.ai.dashboard_item.error_title'),
  retry: i18n.t('common.ai.dashboard_item.retry'),
  dataExplanation: i18n.t('common.ai.dashboard_item.data_explanation'),
},
```

f0's `withDefaults` merges recursively, so nothing breaks — but **both strings ship in English in every locale**. This needs a paired factorial PR adding the `f0-provider.tsx` entries and the `translations/*.json` keys, otherwise a Spanish or German user gets "Drop here to discuss with One" on the primary affordance.

## 🟠 The dashboard pattern now knows the product name

`useAiChat` is imported into both `DashboardItem` and `ChartItem`, so `F0AnalyticsDashboard` has a hard dependency on the chat kit and hardcodes "One" in a design-system pattern. `check-cycle-dependencies` passing tells us there's no cycle; it doesn't tell us the coupling is right.

An injected handler — `onAskAi?` on the dashboard, or an item-level action — would keep the pattern agnostic, let the host own the copy and the icon, and **dissolve the i18n problem above**, since the string would live where the translations already do. Worth weighing now: this is the entry point that will get copied the next time something wants to talk to the chat from a pattern.

## 🟠 `PointActionPopover` never reacts to scroll or resize

It's `position: fixed` at the click's viewport coordinates, and the only dismissals are Escape and an outside `pointerdown`. A wheel scroll doesn't fire either, so scrolling the dashboard leaves the popover parked over whatever slid under it, still pointing at a mark that's moved. Either dismiss on `scroll`/`resize` (capture, so container scrolls count) or recompute `coords`.

## 🟠 Every widget gains a ⋯ menu

```ts
const showMenu = hasAskOne || hasDownloads || hasDelete || hasChartTypes || hasExplanation
```

`hasAskOne` is true for any titled widget once the chat is enabled, so widgets that deliberately had no menu — KPIs, most collections — now render one. That's a visible chrome change on every dashboard in the app, not just the ones with charts. Worth an explicit yes rather than arriving as a side effect.

## 🟠 The invitation fires on every reorder

`WIDGET_DRAG_START` is dispatched from `pointerdown` on the grip with no movement threshold, and the overlay stays up for the whole gesture. So in edit mode — where reordering is the *primary* gesture — the chat panel gets covered by "Drop here to discuss with One" every time someone moves a widget, and a bare click on the handle flashes it. The ghost has the same no-threshold behaviour today, so it's consistent, but the ghost is a small translucent card and this is a full-panel overlay with a message.

If the intent is "the target is visible before the user aims for it", a small movement threshold (a few px) would keep that and drop the flash.

## 🟠 A scatter quote loses half the point

```ts
const raw = Array.isArray(p.value) ? p.value[p.value.length - 1] : p.value
```

For a scatter that's the **y** value only, and the quote comes out as `series: y`. But a scatter point *is* the relationship between two measures — that's the entire reason the type exists (see factorialco/factorial-agent#2477, which teaches the agent to author them). Quoting "Female: 48900" from a salary-vs-tenure plot drops the tenure. Worth special-casing scatter to carry both, using `xAxisName`/`yAxisName` for the labels.

## 🟡 Smaller

- **`PointActionPopover` a11y** — no `role`/`aria-label`, and focus never moves into it. The ⋯ menu is the stated keyboard path, which is fair, but a screen-reader user who *does* click a mark gets a silent DOM insertion. `role="dialog"` + a label is cheap.
- **Empty titles announce anyway** — `detail.title ?? ""` sets `dragQuote` to `""`, which is `!== null`, so the overlay invites the drop; then `if (title)` in `handlePointerUp` silently does nothing. Don't announce when the title is empty.
- **`resolveDropTarget` queries the document on every `pointermove`** — `document.querySelector("[data-ai-chat-dropzone]")` per move event. Resolving it once at drag start would do.
- **`onDismiss={() => setPickedPoint(null)}`** is a fresh closure each render, so the popover's listener effect (deps `[anchor, onDismiss]`) tears down and rebinds on every `ChartItem` render.
- The description says "all nine variants"; there are eight (`F0DataChartProps` union). All eight do get it via `F0DataChartCommonProps`.

## What's good

- **Both bug write-ups are real and correctly diagnosed**, and both guards are pinned by tests confirmed non-vacuous. Putting the drag-start and the release inside a single `act()` to reproduce the batching is exactly the right way to capture that one.
- `usePointClick`'s value guards — rejecting `null`/`undefined`/`""` before `Number()` so a data gap isn't reported as a real zero — is the detail most people would miss, and the comment says why.
- `chart.isDisposed?.()` before `off`, and the `componentType !== "series"` filter so axis labels and legend entries don't fire.
- Rect containment instead of `document.elementFromPoint` because jsdom 26 doesn't implement it — good that this is written down rather than discovered later.
- Reusing `DropOverlay` with a `mode` rather than growing a second overlay, and the `canDrop` split so the widget path isn't gated behind an upload handler.
- The reasoning for `ai.dashboardItem.askOne` over `ai.ask` (hosts override the latter, and it drives two other surfaces) is right — it just needs the host side to actually exist.
- Quoted values formatted through the chart's own `tooltipValueFormatter`, so the number reads as it did on screen.

PR #5120, opened 2026-08-13
feat(l10n): honor a 12h/24h hourCycle in the time field

josiola95:
## What & why

Adds an optional **`time.hourCycle`** (`"12h" | "24h"`) to `L10nContextValue` so apps can drive the F0 **time field**'s format from `F0Provider` instead of the browser/OS locale — mirroring the existing `date.weekStartsOn`.

Motivation: Factorial users have a 12h/24h preference (`access.timeFormat`). Today the F0 time field renders a native `<input type="time">`, whose 12h/24h display is **browser/OS-controlled** and can't be overridden from JS — so a 12h user sees `13:00` instead of `1:00 PM`. This blocks adopting the built-in time field in shift-management (we currently keep a legacy custom time picker).

## Changes

- **`l10n` contract:** `L10nContextValue.time?.hourCycle` + default (`"24h"`), exported from the `l10n` barrel.
- **`TimeFieldRenderer`:** when `l10n.time.hourCycle` is set, render/parse via `date-fns` (12h → `hh:mm a`, 24h → `HH:mm`) using a text input; when `l10n.time` is **unset**, keep the current native `<input type="time">`.
- **`utils`:** `dateToDisplayTime` / `displayTimeToDate` helpers (+ unit tests).

The value is still stored as a `Date` in both modes.

## Back-compat / opt-in

Fully additive. Default `l10n` has no consumer-set `time`, so existing forms keep the native input and behave exactly as before. Only apps that pass `l10n.time.hourCycle` get the formatted input.

Consumer side becomes a one-liner:
```tsx
<F0Provider l10n={{ l10n: { locale, time: { hourCycle: is24h ? "24h" : "12h" } } }} … />
```

## Notes for reviewers

- A native `<input type="time">` genuinely can't honor an app-chosen hour cycle, hence the switch to a controllable text input in the opt-in path. If you'd prefer a different UX for the formatted mode (segmented input, or a stepped time **select** à la Google Calendar), happy to iterate — the `l10n` plumbing is the important part and is independent of that choice.
- Tests: `utils.test.ts` covers 12h/24h format + parse. `oxlint` + `tsc` clean.

image
image

### DIFF
--- a/packages/react/src/lib/providers/l10n/types.ts
+++ b/packages/react/src/lib/providers/l10n/types.ts
@@ -1,8 +1,19 @@
 import { WeekStartsOn } from "@/components/OneCalendar/types"
 
+/** How times are displayed and edited: 12-hour (AM/PM) or 24-hour. */
+export type HourCycle = "12h" | "24h"
+
 export type L10nContextValue = {
   locale: string
   date?: {
     weekStartsOn: WeekStartsOn
+  }
+  time?: {
+    /**
+     * Hour cycle used by time fields. Defaults to "24h".
+     * When set, the F0 time field renders/parses in this format so apps can
+     * honor a user's 12h/24h preference instead of the browser locale.
+     */
+    hourCycle: HourCycle
   }
 }
--- a/packages/react/src/lib/providers/l10n/types.ts
+++ b/packages/react/src/lib/providers/l10n/types.ts
@@ -1,19 +1,8 @@
 import { WeekStartsOn } from "@/components/OneCalendar/types"
 
-/** How times are displayed and edited: 12-hour (AM/PM) or 24-hour. */
-export type HourCycle = "12h" | "24h"
-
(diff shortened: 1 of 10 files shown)


packages/react/src/lib/providers/l10n/types.ts
```diff
@@ -1,8 +1,19 @@
...
   date?: {
     weekStartsOn: WeekStartsOn
   }
+  time?: {
+    /**
+     * Hour cycle used by time fields. Defaults to "24h".
+     * When set, the F0 time field renders/parses in this format so apps can
+     * honor a user's 12h/24h preference instead of the browser locale.
+     */
+    hourCycle: HourCycle
```
sauldom102: I wouldn't put this here, this should only come from f0 provider as a global user setting

sauldom102 (approved):

PR #5123, opened 2026-08-14
feat(F0DataChart): report the clicked mark via onPointClick

igorsafonov-gif:
## Description

Adds `onPointClick` to `F0DataChart`, so a host can respond to the exact chart mark a reader selected: one bar segment, slice, cell, radar value, scatter point, gauge value, or line category.

Nothing changes visually unless a handler is passed. The callback is deliberately mark-level rather than tooltip-level: it reports the selected datum, raw numeric payload, chart indices, the literal `pointer` source, and viewport coordinates a host can use to anchor its own action.

The canvas interaction is intentionally pointer-only. Any host that exposes `onPointClick` must provide an equivalent keyboard-accessible data/action surface outside the ECharts canvas. A keyboard source can be added later without breaking this contract.

**Split out of #5104.** First in the stack; #5124 and #5125 build on this public chart contract.

## Type of change

## Visual review

Point click payload after selecting a chart mark

Line chart point click payload

- [Interactive story - mark-sized target](https://66a7a8d7d124220c363457cc-tfxtqdfoix.chromatic.com/?path=/story/kits-f0datachart-point-click--on-a-mark): click a bar segment and compare the visible value with the emitted payload. Clicking between bars must keep **Nothing picked yet**.
- [Interactive story - line plot target](https://66a7a8d7d124220c363457cc-tfxtqdfoix.chromatic.com/?path=/story/kits-f0datachart-point-click--anywhere-in-a-plot-area): move across the plot, axes, and legend. Only the clickable plot uses the pointer cursor; axis labels stay default and the legend keeps its native pointer. Click a plot column and inspect every visible series at that category.
- [Interactive story - scatter tuple](https://66a7a8d7d124220c363457cc-tfxtqdfoix.chromatic.com/?path=/story/kits-f0datachart-point-click--both-measures-of-a-scatter-point): verify that both measures survive in `values`.
- **Legend callback:** the current-head Storybook deployment adds `kits-f0datachart-point-click--legend-selection-changes`. Isolate a legend item, restore it, and compare the full `selected` mapping printed below the chart. Chromatic publishes the exact deployment link in this PR once the current build finishes.
- [MDX documentation - F0DataChart point click](https://66a7a8d7d124220c363457cc-tfxtqdfoix.chromatic.com/?path=/docs/kits-f0datachart-overview--documentation#point-click): public contract, code example, chart-specific behavior, accessibility limitation, cursor behavior, legend support matrix, and linked Canvas stories.

This PR intentionally has no resting visual change. The useful screenshots are the story before selection and after selecting one mark, because the review target is the callback payload and hit area rather than new chrome.

## Implementation details

- **Pointer-only public source:** `F0DataChartPointClick.source` is the literal `"pointer"`; the unreachable `"keyboard"` union member and keyboard-coordinate wording were removed.
- **Truthful payload semantics:** `category` documents the ECharts mark name across category, pie, funnel, scatter, gauge, and radar charts. `seriesName` documents the configured series name and its possible empty fallback.
- **Grid-aware line affordance:** zrender adds a pointer only inside a handler-backed grid. Axis labels remain default across repeated movement, while legends and other native ECharts targets keep their own cursor. All added listeners and cursor state are cleaned up on global-out and unmount.
- **Scoped legend observer:** `onLegendSelectionChange` is accepted only by the six variants that wire interactive legends: bar, line, funnel, pie, radar, and scatter. Gauge and heatmap do not advertise an unsupported prop.
- **Normalized payload:** genuine `0` is reported, while `null`, `undefined`, blank, non-numeric, and non-finite tuple entries are rejected. `values` preserves scatter and heatmap tuples, and `series` preserves every visible line-series value at one category.
- **Host-owned action placement:** pointer coordinates let a host anchor an action beside the selected mark; a successful pick dismisses the tooltip so both surfaces do not compete.

## Verification

- 84/84 focused point-click, line, scatter, and heatmap tests passed.
- 2/2 focused bar/funnel legend callback tests passed.
- TypeScript, formatting, targeted lint, cycle-dependency hook, and diff checks passed.
- Production Storybook build passed.
- Exact Chrome check passed: plot pointer, persistent default axis cursor, native legend pointer, `source: pointer` payload, and legend selection readout.
- Two independent correctness/accessibility reviews found no remaining blockers.
- Exact-head GitHub CI is green, including React unit tests, all eight Storybook shards, Chromatic, accessibility, API surface, lint, formatting, and package build.

---

### The stack

| | PR | What it adds |
|---|---|---|
| 1 | #5123 | `F0DataChart` reports the selected mark - kit API, no resting visual change |
| 2 | #5124 | Widget menu -> **Ask One** |
| 3 | #5125 | Chart point -> **Ask One** |

---
> - factorial-f0
> - factorial-ci
> - factorial-pr

### DIFF
--- a/packages/react/src/kits/F0DataChart/components/LineChart/LineChart.tsx
+++ b/packages/react/src/kits/F0DataChart/components/LineChart/LineChart.tsx
@@ -23,7 +23,7 @@ export const LineChart = (props: F0DataChartLineProps) => {
   usePointClick(chartRef, props.onPointClick, "plot")
   const theme = useChartTheme(ref)
   useAxisLabelTooltip(chartRef, ref, theme)
-  useLegendInteraction(chartRef)
+  useLegendInteraction(chartRef, props.onLegendSelectionChange)
 
   // ECharts (zrender) sets `cursor: pointer` on the canvas element via inline
   // style whenever an axis label has `triggerEvent: true` (which we need for
--- a/packages/react/src/kits/F0DataChart/components/LineChart/LineChart.tsx
+++ b/packages/react/src/kits/F0DataChart/components/LineChart/LineChart.tsx
@@ -25,18 +25,5 @@ export const LineChart = (props: F0DataChartLineProps) => {
   useAxisLabelTooltip(chartRef, ref, theme)
   useLegendInteraction(chartRef, props.onLegendSelectionChange)
 
-  // ECharts (zrender) sets `cursor: pointer` on the canvas element via inline
-  // style whenever an axis label has `triggerEvent: true` (which we need for
-  // the truncation tooltip). `axisLabel.cursor` is ignored for canvas-rendered
-  // labels, so `useAxisLabelTooltip` toggles `data-axis-hover="true"` on the
-  // container while the cursor is over an axis label, and the Tailwind
-  // arbitrary variant below scopes the `cursor: default !important` reset to
-  // exactly that window — leaving the rest of the chart with its normal
-  // pointer cursor over data points.
-  return (
-    <div
-      ref={ref}
-      className="h-full w-full data-[axis-hover=true]:[&_canvas]:!cursor-default"
-    />
-  )
+  return <div ref={ref} className="h-full w-full" />
 }
--- a/packages/react/src/kits/F0DataChart/components/LineChart/LineChart.tsx
+++ b/packages/react/src/kits/F0DataChart/components/LineChart/LineChart.tsx
@@ -17,7 +17,10 @@ export const LineChart = (props: F0DataChartLineProps) => {
   const size = resolveChartSize(width)
   const options = useLineChartOptions(ref, props, size)
   const chartRef = useEChartsInstance(ref, options)
-  usePointClick(chartRef, props.onPointClick)
+  // "plot", not "mark": a line is a few pixels wide, so requiring a hit on the
+  // line itself makes the action unusable. Same reason the tooltip here is
+  // axis-triggered — the whole column is one target.
+  usePointClick(chartRef, props.onPointClick, "plot")
   const theme = useChartTheme(ref)
   useAxisLabelTooltip(chartRef, ref, theme)
   useLegendInteraction(chartRef)
--- a/packages/react/src/kits/F0DataChart/components/LineChart/LineChart.tsx
+++ b/packages/react/src/kits/F0DataChart/components/LineChart/LineChart.tsx
@@ -7,6 +7,7 @@ import { useAxisLabelTooltip } from "../../utils/useAxisLabelTooltip"
 import { useChartTheme } from "../../utils/useChartTheme"
 import { useContainerSize } from "../../utils/useContainerSize"
 import { useEChartsInstance } from "../../utils/useEChartsInstance"
+import { usePointClick } from "../../utils/usePointClick"
 import { useLegendInteraction } from "../../utils/useLegendInteraction"
 import { useLineChartOptions } from "./useLineChartOptions"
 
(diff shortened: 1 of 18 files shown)


packages/react/src/kits/F0DataChart/types.ts
```diff
@@ -32,12 +32,92 @@ export interface F0DataChartEmptyStateProps {
...
+ * line, where the click can land anywhere in the plot area) a whole category.
+ *
+ * The top-level fields always name a single series, so a consumer that wants
+ * one number has one. {@link F0DataChartPointClick.series} carries everything
+ * the click resolved, which for a line is every series at that category — the
+ * same rows its tooltip shows.
+ */
+export interface F0DataChartPointClick {
+  /** Interaction surface that resolved the point. */
+  source: "pointer" | "keyboard"
```
albertpmz: `usePointClick` only ever sets `source: "pointer"`, on both the mark path and the plot path. So `"keyboard"` is a case a consumer can branch on and never reach.

The mouse is also the only way to reach `onPointClick`. No `tabIndex`, no `onKeyDown`, no focus handling on marks anywhere in the diff. The green a11y job doesn't cover this either. ECharts draws every mark into a single `<canvas>` element, so there are no per-mark DOM nodes for axe to inspect. It sees one element with an aria-label and has nothing to report.

Either drop `"keyboard"` from the union and ship pointer-only, or land the keyboard path here. Adding the union member back later is additive, so nothing breaks. I'd take the first if #5124 and #5125 are waiting on this.

packages/react/src/kits/F0DataChart/types.ts
```diff
@@ -32,12 +32,92 @@ export interface F0DataChartEmptyStateProps {
...
  */
 interface F0DataChartCommonProps {
   /** Customize or opt out of the empty state shown when data is empty. */
   emptyState?: F0DataChartEmptyStateProps
+  /**
+   * Reports the chart's live legend visibility after an interactive toggle.
+   * Primarily used by accessible companion surfaces that must expose the same
+   * data currently shown on the canvas.
+   */
+  onLegendSelectionChange?: (selected: Record<string, boolean>) => void
```
albertpmz: `onLegendSelectionChange` is new here too, and I couldn't find it in the title, the description, or the MDX. It's tested in BarChart and LineChart, so I'm assuming it's intended rather than stray from the #5104 split. Can you confirm?

It also sits on `F0DataChartCommonProps`, and neither `GaugeChart` nor `HeatmapChart` calls `useLegendInteraction`, so on those two the prop typechecks and never fires.

Options: move it to the variants that have a legend, or keep it common and wire the other two. Either way a line in the MDX next to the point-click section would help. Splitting it into its own PR works too if that suits the stack better.

packages/react/src/kits/F0DataChart/components/LineChart/LineChart.tsx
```diff
@@ -16,9 +17,13 @@ export const LineChart = (props: F0DataChartLineProps) => {
   const size = resolveChartSize(width)
   const options = useLineChartOptions(ref, props, size)
   const chartRef = useEChartsInstance(ref, options)
+  // "plot", not "mark": a line is a few pixels wide, so requiring a hit on the
+  // line itself makes the action unusable. Same reason the tooltip here is
+  // axis-triggered — the whole column is one target.
+  usePointClick(chartRef, props.onPointClick, "plot")
```
albertpmz, line 23: The `"plot"` choice makes sense. A line is a few pixels wide, and matching the axis tooltip's target is the right call. What I'm missing is how a user knows the column is clickable.

Once a host passes `onPointClick`, the whole plot column responds to a click, but the cursor stays default. zrender only sets `pointer` over real interactive elements, and this listener sits on the canvas underneath them. So a clickable chart looks the same as one that isn't. Same on the mark charts, though there ECharts at least changes the cursor over the mark itself.

#5124 and #5125 build the Ask One action on top of this, so whatever lands here is where users learn the gesture. Could be `cursor: pointer` on the container while a handler is present, and the `data-axis-hover` reset already in this file would keep axis labels excluded. Could be hover emphasis on the resolved column, or the tooltip as the only cue with a note in the MDX saying so.

albertpmz (changes requested):
Good job! Let's work on these comments below 😃

albertpmz (approved):

PR #5124, opened 2026-08-14
feat(F0AnalyticsDashboard): offer Ask One in the widget menu

igorsafonov-gif:
## Description

Adds **Ask One** to a dashboard widget's actions menu. Choosing it lets a reader carry the widget context into One without retyping the title: the built-in path exits widget fullscreen, opens the chat, attaches the title as a quote, and focuses the composer.

The action is only offered when it can work. A host can provide `onAskAi` to own the handoff; otherwise an enabled F0 AI chat handles it. Untitled widgets and standalone dashboards without either responder remain unchanged.

**Split out of #5104.** Second in the stack, based on #5123.

## Type of change

- [Interactive story - menu action](https://66a7a8d7d124220c363457cc-wpapwoctjg.chromatic.com/?path=/story/patterns-analyticsdashboard-ask-one--widget-menu-action): leaves the widget menu open so reviewers can inspect the label, icon, and placement.
- [Interactive story - completed handoff](https://66a7a8d7d124220c363457cc-wpapwoctjg.chromatic.com/?path=/story/patterns-analyticsdashboard-ask-one--widget-quoted-in-chat): selects **Ask One** and leaves the real chat open with the widget quote visible and the composer focused.
- [MDX documentation - Ask One from a widget](https://66a7a8d7d124220c363457cc-wpapwoctjg.chromatic.com/?path=/docs/patterns-analyticsdashboard--documentation#ask-one-from-a-widget): usage, host-owned API, behavioral rules, and both linked Canvas stories.

**Menu action** - **Ask One** appears with the existing widget actions only when a chat or host handler is available.

Widget actions menu with Ask One between provenance and delete actions

image

## Implementation details

- feat: expose **Ask One** from every titled dashboard widget when an enabled chat or `onAskAi` responder is available

  <details>
  <summary>Why is it gated?</summary>

  `F0AnalyticsDashboard` is also used without the AI shell. Showing a command backed only by inert context setters would create a dead action, and an untitled widget has no context to quote. Both cases therefore keep the previous menu unchanged.
  </details>

- feat: support built-in and host-owned handoffs

  <details>
  <summary>Who owns which state?</summary>

  The built-in action exits widget fullscreen, attaches the title, opens the chat, and transfers focus after the Radix menu closes. When `onAskAi` is supplied, the host receives `{ id, title }` and F0 does not mutate chat or fullscreen state or steal host-managed focus.
  </details>

- feat: preserve safe focus while the real textarea mounts

  <details>
  <summary>Why?</summary>

  Opening the chat and mounting its composer are separate React steps. While the composer is absent, Radix restores focus to the widget trigger and the bridge queues the intended transfer. Closing the chat cancels that request so a later, unrelated mount cannot steal focus.
  </details>

- chore: add the `ai.dashboardItem.askOne` translation key and document the Factorial-side override required with the package bump
- fix: keep Ask One available in widget error headers, and hide host-owned actions when a directly rendered item has no ID rather than emitting an empty ID\n- test: cover provider, title, ID, and error-state gating; built-in quote/open/fullscreen/focus behavior; stale-focus cancellation; host ownership; public callback propagation; and the real Storybook menu-to-composer flow\n- story: opt the open-menu review state into Chromatic snapshots

## Verification

- Focused Ask One, dashboard, and focus-bridge suites: 35/35 passed.
- TypeScript, targeted lint, format, and production Storybook build passed.
- Production Storybook and exact Chrome checks passed for the open-menu and completed quote/focus stories. Exact-head GitHub CI is green, including React unit tests, all eight Storybook shards, Chromatic, and the package build.

## Note for the bump

This adds an F0 translation key, so Factorial's `f0-provider.tsx` and `en.json` need `common.ai.dashboard_item.ask_one` before or with the version bump. Overridden translation sections are exhaustive.

---

### The stack

| | PR | What it adds |
|---|---|---|
| 2 | #5124 | Widget menu -> **Ask One** |
| 3 | #5125 | Chart point -> **Ask One** |

---
> - factorial-f0
> - factorial-ci
> - factorial-pr

### DIFF
--- a/packages/react/src/kits/ai/F0AiChat/providers/AiChatStateProvider.tsx
+++ b/packages/react/src/kits/ai/F0AiChat/providers/AiChatStateProvider.tsx
@@ -216,12 +216,14 @@ export const AiChatStateProvider: FC<PropsWithChildren<AiChatState>> = ({
   // request as soon as it becomes available.
   const focusChatInputRef = useRef<(() => void) | null>(null)
   const pendingChatInputFocusRef = useRef(false)
-  const focusChatInput = useCallback(() => {
+  const focusChatInput = useCallback((): boolean => {
     if (focusChatInputRef.current) {
       focusChatInputRef.current()
-    } else {
-      pendingChatInputFocusRef.current = true
+      return true
     }
+
+    pendingChatInputFocusRef.current = true
+    return false
   }, [])
   const setFocusChatInputFunction = useCallback((fn: (() => void) | null) => {
     focusChatInputRef.current = fn
@@ -241,6 +243,10 @@ export const AiChatStateProvider: FC<PropsWithChildren<AiChatState>> = ({
 
   useEffect(() => {
     if (!open) {
+      // A focus request belongs to the interaction that opened the panel.
+      // Once that panel closes, replaying it on a later mount would steal
+      // focus from whatever the user moved on to.
+      pendingChatInputFocusRef.current = false
       setCanvasContent(null)
       setVisualizationMode("sidepanel")
       const prefersReducedMotion = window.matchMedia(
@@ -502,6 +508,7 @@ const REAL_VALUES: Partial<AiChatProviderReturnValue> = {
   placeholders: [],
   welcomeScreenSuggestions: [],
   welcomeScreenCards: [],
+  focusChatInput: () => false,
 }
 
 const NO_PROVIDER_CONTEXT = new Proxy({} as AiChatProviderReturnValue, {
--- a/packages/react/src/kits/ai/F0AiChat/providers/AiChatStateProvider.tsx
+++ b/packages/react/src/kits/ai/F0AiChat/providers/AiChatStateProvider.tsx
@@ -211,6 +211,26 @@ export const AiChatStateProvider: FC<PropsWithChildren<AiChatState>> = ({
     []
   )
 
+  // Focus bridge — callers can request focus before the chat has mounted.
+  // The textarea registers its local ref callback and consumes one buffered
+  // request as soon as it becomes available.
+  const focusChatInputRef = useRef<(() => void) | null>(null)
+  const pendingChatInputFocusRef = useRef(false)
+  const focusChatInput = useCallback(() => {
+    if (focusChatInputRef.current) {
+      focusChatInputRef.current()
+    } else {
+      pendingChatInputFocusRef.current = true
+    }
+  }, [])
+  const setFocusChatInputFunction = useCallback((fn: (() => void) | null) => {
+    focusChatInputRef.current = fn
+    if (fn && pendingChatInputFocusRef.current) {
+      pendingChatInputFocusRef.current = false
+      fn()
+    }
+  }, [])
+
   const resetChatWidth = () => {
     setChatWidth(DEFAULT_CHAT_WIDTH)
   }
@@ -399,6 +419,8 @@ export const AiChatStateProvider: FC<PropsWithChildren<AiChatState>> = ({
         setFileDragOver,
         processDroppedFiles,
         setProcessDroppedFilesFunction,
+        focusChatInput,
+        setFocusChatInputFunction,
         pendingContext,
         setPendingContext,
         pendingQuote,
--- a/packages/react/src/patterns/F0AnalyticsDashboard/components/DashboardItem/DashboardItem.tsx
+++ b/packages/react/src/patterns/F0AnalyticsDashboard/components/DashboardItem/DashboardItem.tsx
@@ -18,6 +18,8 @@ import {
   InfoCircleLine,
 } from "@/icons/app"
 import { InfoHint, type InfoHintContent } from "@/lib/InfoHint"
+import { One as OneIcon } from "@/icons/ai"
+import { useAiChat } from "@/kits/ai/F0AiChat/providers/AiChatStateProvider"
 import { OneEllipsis } from "@/lib/OneEllipsis"
 import { useI18n } from "@/lib/providers/i18n"
 import { cn } from "@/lib/utils"
@@ -130,6 +132,11 @@ export function DashboardItem({
    */
   const [isExplanationView, setIsExplanationView] = useState(false)
   const translations = useI18n()
+  const {
+    enabled: aiEnabled,
+    setPendingQuote,
+    setOpen: setAiChatOpen,
+  } = useAiChat()
 
   const handleDropdownOpenChange = (open: boolean) => {
     setIsDropdownOpen(open)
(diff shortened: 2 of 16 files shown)


packages/react/src/patterns/F0AnalyticsDashboard/components/DashboardItem/DashboardItem.tsx
```diff
@@ -304,6 +326,12 @@ export function DashboardItem({
               <DropdownMenuContent
                 align="end"
                 className={cn("py-1", isExplanationView && "w-96 max-w-[90vw]")}
+                onCloseAutoFocus={(event) => {
+                  if (!shouldFocusChatAfterMenuRef.current) return
+                  event.preventDefault()
+                  shouldFocusChatAfterMenuRef.current = false
+                  focusChatInput()
+                }}
```
albertpmz: `preventDefault()` runs before we know there's anywhere to send the focus. When no composer is registered, `focusChatInput()` only buffers, so Radix's restore-to-trigger is cancelled and nothing takes over. Focus lands on `document.body`.

That gap always exists. `ChatWindow` mounts `children` only when the panel is visible (`ChatWindow.tsx:152`), so `setOpen(true)` and the composer registering happen in separate commits. It never closes for a host that mounts `F0AiChat` with its own composer instead of `F0AiChatTextArea`, or for a user who dismisses the panel mid-open. A keyboard or screen reader user ends up on `body` with no way back to the widget.

Have the buffer report back:

```ts
// provider
const focusChatInput = useCallback((): boolean => {
  if (focusChatInputRef.current) {
    focusChatInputRef.current()
    return true
  }
  pendingChatInputFocusRef.current = true
  return false
}, [])
```

```ts
// here
onCloseAutoFocus={(event) => {
  if (!shouldFocusChatAfterMenuRef.current) return
  shouldFocusChatAfterMenuRef.current = false
  // Only take focus off the trigger if the composer can receive it.
  if (focusChatInput()) event.preventDefault()
}}
```

The open-then-mount path still buffers: the trigger holds focus for a beat, then the composer takes it. When nothing will ever answer, Radix does what it does today. Add a test with no `F0AiChatTextArea` mounted, asserting focus ends on the trigger.

packages/react/src/kits/ai/F0AiChat/providers/AiChatStateProvider.tsx
```diff
@@ -211,6 +211,26 @@ export const AiChatStateProvider: FC<PropsWithChildren<AiChatState>> = ({
...
+  // The textarea registers its local ref callback and consumes one buffered
+  // request as soon as it becomes available.
+  const focusChatInputRef = useRef<(() => void) | null>(null)
+  const pendingChatInputFocusRef = useRef(false)
+  const focusChatInput = useCallback(() => {
+    if (focusChatInputRef.current) {
+      focusChatInputRef.current()
+    } else {
+      pendingChatInputFocusRef.current = true
+    }
```
albertpmz, line 223: `pendingChatInputFocusRef` is cleared only when it's consumed, so an unresolved request stays armed for the life of the provider. With the mount gating in `ChatWindow`, this sequence works: pick **Ask One**, dismiss the panel before the composer mounts, work for ten minutes, then open the chat for something else. The composer registers, consumes the stale request, and pulls focus out of whatever the user was doing.

The file-drop bridge above survives the same race because a replayed drop still does what the user asked for. A replayed focus lands in a context they've left.

Clearing on close covers it:

```ts
useEffect(() => {
  if (!open) pendingChatInputFocusRef.current = false
}, [open])
```

packages/react/src/patterns/F0AnalyticsDashboard/components/DashboardItem/DashboardItem.tsx
```diff
@@ -358,6 +386,36 @@ export function DashboardItem({
...
 
+                    {hasAskOne && (
+                      <DropdownMenuGroup>
+                        <DropdownMenuItem
+                          onClick={() => {
+                            if (onAskAi) {
+                              // The host answers this one. Nothing else here
+                              // applies: it may not open the chat at all, so
+                              // leaving fullscreen would be a guess.
+                              onAskAi({ id: itemId ?? "", title })
```
albertpmz: `itemId ?? ""` hands the host an empty string while the public type promises `id: string` (`types.ts:530`), and the MDX invites keying off it (`openProductChat({ widgetId: id })`). The host can't tell `""` from a real id, so it opens a chat about widget `""`.

Every grid path does pass `item.id`. But `DashboardItem` is reachable directly, and two of the new tests render it without `itemId` and assert only `toHaveBeenCalledTimes(1)`, so nothing pins the id down.

Either let the type say the id can be missing:

```ts
onAskAi?: (item: { id?: string; title: string }) => void
// ...
onAskAi({ id: itemId, title })
```

or keep `id` required and drop `hasAskOne` when `onAskAi` arrives without an `itemId`, so the entry never appears in a state it can't describe. Either way, assert the whole payload in the no-chat test instead of the call count.

packages/react/src/patterns/F0AnalyticsDashboard/components/DashboardItem/DashboardItem.tsx
```diff
@@ -146,7 +162,13 @@ export function DashboardItem({
   const hasChartTypes = chartTypeOptions && chartTypeOptions.length > 0
   const hasExplanation = !!explanation && explanation.trim().length > 0
   const hasFullscreen = !!onFullscreenChange
-  const showMenu = hasDownloads || hasDelete || hasChartTypes || hasExplanation
+  // Keyboard-reachable twin of dragging the widget onto the chat. Gated on
+  // something being able to answer it: a host handler, or failing that a
+  // mounted chat — with no provider `useAiChat()` returns an inert context
+  // whose setters are no-ops, so offering the action would do nothing.
+  const hasAskOne = (!!onAskAi || aiEnabled) && title.trim().length > 0
```
albertpmz: `hasAskOne` is computed above the `if (error)` return on line 173, so it's dead in the error branch. A failed widget renders the title, the ⓘ, and the empty state, with no actions menu and no **Ask One**.

Your own comment in that branch argues the other way: "The help copy survives the failure: a reader looking at an error is exactly the one asking what the widget was meant to show." The title, the only thing Ask One quotes, is right there in the error header.

If this is a deliberate cut, add it to the behavior rules in the MDX (`WHEN the widget is in an error state → THEN ...`). The neighbouring rule promises the info trigger survives an error, so a reader will assume Ask One does too. If it isn't deliberate, the error header already has everything the action needs.

albertpmz (changes requested):
Good job, but you're missing a Snapshot story, which is required by the checklist. `preview.tsx:144` disables snapshots globally and neither new story opts back in, so add `parameters: withSnapshot({})` to `WidgetMenuAction`.

I've also left some comments below – let's take care of them 👇🏼

albertpmz (approved):

PR #5125, opened 2026-08-14
feat(F0AnalyticsDashboard): offer Ask One on a clicked chart mark

igorsafonov-gif:
## Description

Extends **Ask One** from widget-level context to one selected chart point. A pointer user clicks a mark and confirms the anchored action; a keyboard or screen-reader user opens the chart's named **Ask One** trigger and chooses the same formatted point from a menu.

The confirmation step is intentional: chart clicks are common during exploration, so selecting a mark must not open chat by itself. The built-in action quotes the chart title, category, series, and displayed value, then opens chat and focuses the composer. Hosts using `onAskAi` receive the raw point instead.

**Third PR in the #5123 → #5124 → #5125 stack.**

## Type of change

## Visual review

Chart point Ask One flow

- [Interactive story — complete chart-point flow](https://66a7a8d7d124220c363457cc-cekxzmvkdt.chromatic.com/?path=/story/patterns-analyticsdashboard-ask-one--chart-point-flow): the play function uses the keyboard route and leaves the formatted point quote visible in the focused composer. Reset it to exercise the pointer route by clicking the Engineering bar.
- [MDX documentation — Ask One from a chart point](https://66a7a8d7d124220c363457cc-cekxzmvkdt.chromatic.com/?path=/docs/patterns-analyticsdashboard--documentation#ask-one-from-a-chart-point): pointer and keyboard behavior, host ownership, formatting, pagination, dismissal, title requirements, and screen-reader naming.

Review these states:

1. **Pointer confirmation:** click the Engineering bar. The anchored **Ask One** action appears beside the mark and chat stays closed.
2. **Keyboard point picker:** reset, Tab to **Ask One: Headcount by Department**, then press Enter. The menu exposes complete formatted point context.
3. **Completed handoff:** choose Engineering. The quote contains title, category, series, and formatted value; focus lands in the composer.
4. **Ordinary host rerender:** while the anchored action or keyboard menu is open, an unrelated parent render must not dismiss it.

## Behavior and implementation

- Chart selection is a two-step offer, so ordinary exploration cannot unexpectedly open One.
- Quotes use the chart's tooltip/value formatters, including scatter axes, heatmap labels, and all visible line series at a category.
- The pointer action portals above clipped widgets and dismisses on Escape, outside press, scroll, resize, or genuinely stale data.
- The keyboard companion is named **Ask One: chart title**, exposes the same visible/legend-filtered points, and pages large charts at 100 items.
- `F0DataChart` remains truthfully pointer-only. The dashboard's real keyboard companion adds `source: "keyboard"` through `F0AnalyticsDashboardPointClick`; keyboard points use zero viewport coordinates because they have no pointer anchor.
- Host callbacks own chat/focus/fullscreen state and receive `{ id, title, point }`; the built-in action exits fullscreen, quotes, opens chat, and focuses its composer.
- Inline host config and callback identities may change on normal React renders. Point actions reset only for semantic invalidation, and an open action resolves the latest host callback when selected.
- Charts with blank titles expose neither pointer nor keyboard point actions, matching widget-level Ask One.
- The public props JSDoc remains attached to `F0AnalyticsDashboardProps`.

## Verification

Exact local candidate:

- Focused point-click, point-action, dashboard integration, and public callback suites: **84/84 passed**.
- TypeScript, format check, targeted lint, diff check, and production Storybook build passed.
- Real Chrome completed the point story with the formatted quote visible and the textarea focused.
- The pre-push changed-path run passed 245 tests; its six failures are inherited BarChart font-metric assertions reproduced on the exact lower stack and untouched by this PR.
- Exact-head GitHub CI is green, including React unit tests, all eight Storybook shards, Chromatic, accessibility, and the package build.

---

### The stack

| | PR | What it adds |
|---|---|---|
| 2 | #5124 | Widget menu → **Ask One** |
| 3 | #5125 | Chart point → **Ask One** |

---
> - factorial-f0
> - factorial-ci
> - factorial-pr

### DIFF
--- a/packages/react/src/patterns/F0AnalyticsDashboard/components/ChartItem/AccessiblePointActions.tsx
+++ b/packages/react/src/patterns/F0AnalyticsDashboard/components/ChartItem/AccessiblePointActions.tsx
@@ -1,0 +1,218 @@
+import { useEffect, useRef, useState } from "react"
+
+import { ButtonInternal } from "@/components/F0Button/internal"
+import { One as OneIcon } from "@/icons/ai"
+import { cn } from "@/lib/utils"
+import {
+  DropdownMenu,
+  DropdownMenuContent,
+  DropdownMenuItem,
+  DropdownMenuTrigger,
+} from "@/ui/dropdown-menu"
+
+export type AccessiblePointAction = {
+  key: string
+  getLabel: () => string
+  onSelect: () => void
+}
+
+type AccessiblePointActionsProps = {
+  hasActions: boolean
+  getActions: () => AccessiblePointAction[]
+  label: string
+  triggerLabel: string
+  previousLabel: string
+  nextLabel: string
+  setTrigger: (element: HTMLButtonElement | null) => void
+  focusChatAfterSelect: boolean
+  focusChatInput: () => void
+}
+
+/**
+ * Keyboard and screen-reader surface for canvas marks. It stays visually quiet
+ * until its trigger receives focus, then exposes the same point actions in a
+ * native Radix menu with arrow-key navigation.
+ */
+export function AccessiblePointActions({
+  hasActions,
+  getActions,
+  label,
+  triggerLabel,
+  previousLabel,
+  nextLabel,
+  setTrigger,
+  focusChatAfterSelect,
+  focusChatInput,
+}: AccessiblePointActionsProps) {
+  const [open, setOpen] = useState(false)
+  const [actions, setActions] = useState<AccessiblePointAction[] | null>(null)
+  const [page, setPage] = useState(0)
+  const shouldFocusChatRef = useRef(false)
+  const selectedActionRef = useRef(false)
+  const pendingActionRef = useRef<(() => void) | null>(null)
+  const triggerRef = useRef<HTMLButtonElement | null>(null)
+  const contentRef = useRef<HTMLDivElement>(null)
+  const shouldFocusPageRef = useRef(false)
+  const shouldFocusInitialActionRef = useRef(false)
+
+  useEffect(() => {
+    setActions(null)
+    setPage(0)
+  }, [getActions])
+
+  useEffect(() => {
+    if (!shouldFocusPageRef.current) return
+    shouldFocusPageRef.current = false
+    let focusFrame = 0
+    const frame = requestAnimationFrame(() => {
+      // Radix moves focus back to the content after an item prevents close.
+      // Run one frame later so the new page's first data action wins.
+      focusFrame = requestAnimationFrame(() => {
+        contentRef.current
+          ?.querySelector<HTMLElement>("[data-point-action]")
+          ?.focus()
+      })
+    })
+    return () => {
+      cancelAnimationFrame(frame)
+      cancelAnimationFrame(focusFrame)
+    }
+  }, [page])
+
+  useEffect(() => {
+    if (!open || actions === null || !shouldFocusInitialActionRef.current)
+      return
+    shouldFocusInitialActionRef.current = false
+    let focusFrame = 0
+    const frame = requestAnimationFrame(() => {
+      focusFrame = requestAnimationFrame(() => {
+        contentRef.current
+          ?.querySelector<HTMLElement>("[data-point-action]")
+          ?.focus()
+      })
+    })
+    return () => {
+      cancelAnimationFrame(frame)
+      cancelAnimationFrame(focusFrame)
+    }
+  }, [actions, open])
+
+  if (!hasActions) return null
+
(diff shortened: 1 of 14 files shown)


packages/react/src/patterns/F0AnalyticsDashboard/components/ChartItem/ChartItem.tsx
```diff
@@ -493,27 +826,119 @@ export function ChartItem<Filters extends FiltersDefinition>({
...
+    [item, data, unrenderableChart]
+  )
+
+  // A point belongs to one exact render. A filter/config/refetch transition
+  // can retain old data while loading; never let that stale mark reappear or
+  // resolve its tuple indexes against the next result.
+  useEffect(() => {
+    setPickedPoint(null)
+    setLegendSelection(undefined)
+  }, [data, isLoading, item.chart])
```
albertpmz: `item.chart` is host-supplied. Hosts that build `items` inline get a fresh object on every render, so this effect fires on every render and clears the picked point before the user can act on it. That is the normal React idiom, and the stories in this folder do it too.

I reproduced it locally: render `ChartItem` inside a host that rebuilds its item each render, click a mark, force one unrelated re-render, and the anchored Ask One button is gone. Any host state update while a mark is picked kills the flow, with no way for the user to tell why.

The invalidation itself is right, a mark does belong to one render of the data. It just needs to key on what the mark resolves against, so `data` or `chartProps`, rather than on the identity of the config object.

packages/react/src/patterns/F0AnalyticsDashboard/components/ChartItem/AccessiblePointActions.tsx
```diff
@@ -0,0 +1,219 @@
...
+  const triggerRef = useRef<HTMLButtonElement | null>(null)
+  const contentRef = useRef<HTMLDivElement>(null)
+  const shouldFocusPageRef = useRef(false)
+  const shouldFocusInitialActionRef = useRef(false)
+
+  useEffect(() => {
+    setActions(null)
+    setPage(0)
+    setOpen(false)
+  }, [getActions])
```
albertpmz: Same root cause as my comment on `ChartItem.tsx:881`, one level down. `getActions` chains through `chartProps` to `item`, so it churns on every parent render and this closes an open menu.

Reproduced the same way: open the menu, force one unrelated parent re-render, `queryByRole("menu")` returns null.

Worth keying on the data rather than on a callback identity this component does not control.

packages/react/src/patterns/F0AnalyticsDashboard/types.ts
```diff
@@ -452,6 +453,19 @@ export type DashboardItemLayout = {
  * An LLM can generate the full `items` array as JSON (minus the `fetchData`
  * functions) to build dashboards on the fly.
  */
+/**
```
albertpmz, line 454: Inserting the new interface here splits `F0AnalyticsDashboardProps` from its JSDoc. Only the closest block attaches, so the "Props for the F0AnalyticsDashboard component ... An LLM can generate the full `items` array as JSON" doc now documents `F0AnalyticsDashboardAskAiTarget` instead, and drops out of editor tooltips and the generated API docs for the props type.

Moving `F0AnalyticsDashboardAskAiTarget` above that comment block fixes it.

packages/react/src/patterns/F0AnalyticsDashboard/components/ChartItem/ChartItem.tsx
```diff
@@ -726,6 +1181,26 @@ export function ChartItem<Filters extends FiltersDefinition>({
               // category at a fixed row height, growing the widget.
               {...(fitContent ? { showAllCategories: true } : {})}
             />
+            <AccessiblePointActions
+              hasActions={hasAccessiblePointActions}
+              getActions={getAccessiblePointActions}
+              label={translations.ai.dashboardItem.askOne}
+              triggerLabel={`${translations.ai.dashboardItem.askOne}: ${item.title}`}
```
albertpmz, line 1201: `DashboardItem.tsx:171` hides widget-level Ask One when the title is blank (`title.trim().length > 0`), and the MDX documents the rule: "WHEN a widget has no title THEN the action is hidden because there is nothing to quote."

The point surface has no equivalent guard, so a chart with `title: ""` names its trigger `"Ask One: "` and quotes `" — Engineering"`. Either guard it the same way or note the exception in the doc.

albertpmz (changes requested):
Good job! Let's take a look at the comments below, especially the re-render one 👀

albertpmz (approved):

PR #5126, opened 2026-08-14
feat(ai): drop a dashboard widget into the chat to quote it

igorsafonov-gif:
## Description

Lets a reader drag an editable dashboard widget by its reorder grip and release it over the visible One chat. The widget title becomes a pending quote, the composer receives focus, and the dashboard keeps its original order.

The invitation appears on the first meaningful pointer movement, while the reader is still near the widget. A plain grip click stays quiet. Dragging remains a pointer enhancement; the widget menu's **Ask One** action is the keyboard and single-pointer equivalent.

**Final PR in the #5123 → #5124 → #5125 → #5126 stack.** This branch is based on #5125 because widget drops share its Ask One ownership and accessible-fallback contract.

## Type of change

## Visual review

- [Interactive story — drag widget to quote](https://66a7a8d7d124220c363457cc-zvijnhlhyr.chromatic.com/?path=/story/kits-ai-f0aichat-widget-drop--drag-widget-to-quote): exercises meaningful movement → invitation → drop → quote → composer focus and asserts that widget order is unchanged.
- [MDX documentation — discuss a dashboard widget](https://66a7a8d7d124220c363457cc-zvijnhlhyr.chromatic.com/?path=/docs/kits-ai-f0aichat--documentation#discuss-a-dashboard-widget): behavior, click threshold, cancellation, host ownership, excluded/closing chat surfaces, responsive constraints, and the accessible alternative.

Review these states:

1. **Invitation while still over the dashboard:** move a reorder grip more than 4px. Chat offers **Drop here to discuss with One** before the pointer reaches it.
2. **Pointer over chat:** the dashboard reorder indicator disappears, so one release cannot both quote and reorder.
3. **Released:** the title is attached to the focused composer and the widget remains in its original slot.
4. **Plain click:** press and release the grip without moving; no chat overlay or drag-end action should fire.

https://github.com/user-attachments/assets/0e624143-3040-4d7f-af22-c75a60e07a56

**Invitation**

Dragging a widget over the dashboard while chat shows the drop invitation

**Pointer over chat**

Dragging over chat with the dashboard reorder indicator removed

**Quoted in the composer**

The widget title quoted in the focused composer without a dashboard reorder

## Behavior and implementation

- The grid uses its existing pointer gesture and scoped `window` events; it does not pretend this is an HTML5 file drag.
- A 4px threshold separates a grip click from a real drag. The invitation is still early, but no longer flashes for a click.
- Chat eligibility is limited to the visible real composer: closing, voice, hosted content, host overlays, Pong, and clarifying flows cannot consume an invisible quote.
- Because `AnimatePresence` retains exiting DOM props, closing chat explicitly retracts the drop-zone marker and release reads the latest eligibility.
- The grid caches active chat zones while moving, then queries them again at pointer release. A chat that becomes ineligible mid-gesture cannot swallow the reorder.
- Final release coordinates are authoritative, and every active chat zone suppresses dashboard reorder.
- Host-owned drops call `onAskAi({ id, title })`; the built-in path quotes the title and focuses the composer.
- Pointer cancellation, producer unmount, surface changes, and game/clarifying transitions clear the active invitation.
- The invitation and completed quote are polite status messages. The widget menu's **Ask One** action provides the keyboard and reliable single-pointer route.

## Verification

Exact local candidate:

- Focused widget-drop, dashboard-grid, and widget-menu suites: **61/61 passed**.
- TypeScript, format check, targeted lint, diff check, and production Storybook build passed.
- Real Chrome completed the interaction story: `Total Headcount` quoted, textarea focused, invitation removed, first widget unchanged, and no console errors.
- Exact-head GitHub CI is green, including React unit tests, all eight Storybook shards, Chromatic, accessibility, code review, test coverage, API surface, lint, formatting, and the package build.

## Constraints

- Dashboard dragging is disabled below the grid's 640px editing breakpoint.
- Touch drag support depends on the host dashboard's existing reorder gesture.
- **Ask One** in the widget menu is the documented keyboard and reliable single-pointer alternative.

## Note for the bump

This adds `ai.dropWidgetToDiscuss`, so Factorial's `f0-provider.tsx` and `en.json` need `common.ai.drop_widget_to_discuss` before or with the version bump.

---

### The stack

| | PR | What it adds |
|---|---|---|
| 2 | #5124 | Widget menu → **Ask One** |
| 3 | #5125 | Chart point → **Ask One** |

---
> - factorial-f0
> - factorial-ci
> - factorial-pr

### DIFF
--- a/packages/react/src/kits/ai/F0AiChat/components/layout/ChatWindow.tsx
+++ b/packages/react/src/kits/ai/F0AiChat/components/layout/ChatWindow.tsx
@@ -22,6 +22,7 @@ export const SidebarWindow = ({
   visible,
   side,
   exitStyle = "shrink",
+  acceptsWidgetDrop = false,
 }: {
   children?: ReactNode
   /** Overrides the context `open` as the mount condition — lets the frame
@@ -36,6 +37,8 @@ export const SidebarWindow = ({
    * the panels feel like they were always there.
    */
   exitStyle?: "shrink" | "hold"
+  /** Enables dashboard-widget quoting for the real chat/composer view only. */
+  acceptsWidgetDrop?: boolean
 }) => {
   const {
     open,
@@ -137,9 +140,10 @@ export const SidebarWindow = ({
   useEffect(() => {
     const onStart = (e: Event) => {
       // Same freeze as the file drop: a clarifying flow owns the panel.
-      if (isClarifying) return
+      if (!acceptsWidgetDrop || isClarifying) return
       const detail = (e as CustomEvent<WidgetDragStartDetail>).detail
-      setDragQuoteBoth(detail?.title ?? "")
+      if (typeof detail?.title !== "string" || !detail.title.trim()) return
+      setDragQuoteBoth(detail.title)
     }
     const onEnd = () => setDragQuoteBoth(null)
 
@@ -149,17 +153,28 @@ export const SidebarWindow = ({
       window.removeEventListener(WIDGET_DRAG_START, onStart)
       window.removeEventListener(WIDGET_DRAG_END, onEnd)
     }
-  }, [isClarifying, setDragQuoteBoth])
+  }, [acceptsWidgetDrop, isClarifying, setDragQuoteBoth])
+
+  // A view change, host overlay, or clarifying flow can take ownership of the
+  // shell while a pointer is still down. Retract the invitation immediately
+  // so releasing over non-chat content can never create an invisible quote.
+  useEffect(() => {
+    if (!acceptsWidgetDrop || isClarifying) setDragQuoteBoth(null)
+  }, [acceptsWidgetDrop, isClarifying, setDragQuoteBoth])
 
   // Releasing over the chat quotes the widget. This is a handler on the card,
   // so a release anywhere else simply never reaches it — the grid's own
   // `pointerup` clears the invitation via WIDGET_DRAG_END.
   const handlePointerUp = useCallback(() => {
+    if (!acceptsWidgetDrop || isClarifying) {
+      setDragQuoteBoth(null)
+      return
+    }
     const title = dragQuoteRef.current
     if (title === null) return
     setDragQuoteBoth(null)
     if (title) setPendingQuote({ text: title })
-  }, [setDragQuoteBoth, setPendingQuote])
+  }, [acceptsWidgetDrop, isClarifying, setDragQuoteBoth, setPendingQuote])
 
   const fullscreen = visualizationMode === "fullscreen"
   // Stays LOCAL: it gates this handle's document mousemove listener and its
@@ -273,7 +288,7 @@ export const SidebarWindow = ({
             // Marks this card as a drop target for pointer-driven drags
             // elsewhere in the app (the dashboard grid hit-tests for it to
             // suppress its own reorder while the cursor is over the chat).
-            data-ai-chat-dropzone=""
+            data-ai-chat-dropzone={acceptsWidgetDrop ? "" : undefined}
             onDragEnter={handleDragEnter}
             onDragOver={handleDragOver}
             onDragLeave={handleDragLeave}
@@ -285,7 +300,7 @@ export const SidebarWindow = ({
             </div>
             {/* `canDrop` gates only the file drop — quoting a dragged widget
                 needs no upload handler, so it renders on its own. */}
-            {(canDrop || dragQuote !== null) && (
+            {(canDrop || (acceptsWidgetDrop && dragQuote !== null)) && (
               <DropOverlay
                 visible={(canDrop && fileDragOver) || dragQuote !== null}
                 mode={dragQuote !== null ? "discuss" : "files"}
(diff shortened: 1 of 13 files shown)


packages/react/src/kits/ai/F0AiChat/components/layout/ChatWindow.tsx
```diff
@@ -48,10 +54,13 @@ export const SidebarWindow = ({
     fileDragOver,
     setFileDragOver,
     processDroppedFiles,
+    setPendingQuote,
+    focusChatInput,
     activeGame,
     closeGame,
     panelSide,
   } = useAiChat()
+  const canAcceptWidgetDrop = acceptsWidgetDrop && activeGame === null
```
albertpmz: `canAcceptWidgetDrop` doesn't consult `isVisible` (line 69). `AnimatePresence` keeps this card mounted through the exit (300ms shrink, or 250 + 50ms for the split swap) and the exit is a `clipPath` inset, so the rect and pointer events stay live.

Repro: close the chat, grab a grip within 300ms, drag over where the panel was. The grid caches the card as a zone, suppresses the reorder across that whole Y band, and the release quotes into a panel that's leaving. That's the invisible quote the retract effect below is meant to prevent.

`aria-hidden` on line 302 already gates on `isVisible`. Folding it in here covers the attribute, the listener effect and the retract effect at once:

```ts
const canAcceptWidgetDrop = acceptsWidgetDrop && activeGame === null && isVisible
```

packages/react/src/patterns/F0AnalyticsDashboard/components/DashboardGrid/DashboardGrid.tsx
```diff
@@ -345,6 +373,25 @@ export function DashboardGrid<Filters extends FiltersDefinition>({
       dropTargetRef.current = null
       setDragId(id)
       setDropTarget(null)
+      chatDropZonesRef.current = Array.from(
```
albertpmz: Zones are snapshotted at pointerdown, but `ChatWindow` re-evaluates eligibility every render, so they can disagree mid-gesture.

Repro: start a drag, then start a clarifying flow. The chat retracts the invitation and refuses the quote, but this stale ref still matches the rect, so `resolveDropTarget` returns `null`. The release does nothing at all: no quote, no reorder, no feedback.

Re-read in `up`, once per release rather than per move, so the caching still pays for itself:

```ts
const up = (ev: PointerEvent) => {
  chatDropZonesRef.current = Array.from(
    document.querySelectorAll("[data-ai-chat-dropzone]")
  )
  dropTargetRef.current = resolveDropTarget(ev.clientX, ev.clientY)
  endDrag(true)
}
```

packages/react/src/patterns/F0AnalyticsDashboard/components/DashboardGrid/DashboardGrid.tsx
```diff
@@ -345,6 +373,25 @@ export function DashboardGrid<Filters extends FiltersDefinition>({
...
+
+      // Announce the drag so other drop targets can invite it immediately,
+      // rather than waiting for the cursor to reach them. The AI chat listens
+      // for this to offer "discuss this widget" from the first moment of the
+      // gesture. Fired on `window` so the listener needs no shared ancestor.
+      //
+      // Nothing to announce without a title: the title *is* the quote, so the
+      // chat would invite a drop that could only do nothing.
+      const title = itemMapRef.current.get(id)?.title ?? ""
+      if (title) {
```
albertpmz: No movement threshold, so a plain click on a grip flashes the full-panel overlay across the chat and retracts it. The ghost behaves the same way, but a ghost and the whole panel changing state aren't the same amount of noise. Waiting for the first `pointermove` past a few px would also stop `WIDGET_DRAG_END` firing for gestures that were never drags. Non-blocking, but worth a designer's read since the early invitation is the design intent.

albertpmz (changes requested):

packages/react/src/kits/ai/F0AiChat/components/layout/ChatWindow.tsx
```diff
@@ -48,16 +54,23 @@ export const SidebarWindow = ({
...
     processDroppedFiles,
+    setPendingQuote,
+    focusChatInput,
     activeGame,
     closeGame,
     panelSide,
   } = useAiChat()
+  const isVisible = visible ?? open
+  const canAcceptWidgetDrop =
+    acceptsWidgetDrop && activeGame === null && isVisible
```
albertpmz: The re-read in `up` fixed the timing half. The clarifying case is still open, and it isn't staleness.

`isClarifying` isn't in `canAcceptWidgetDrop`, so `data-ai-chat-dropzone` (327) stays on through a clarifying flow and the removal effect (89) never fires. But `onStart` (161), the retract effect (186) and `handlePointerUp` (193) all check it. The marker advertises a drop the handlers refuse: the grid re-reads, still matches, `resolveDropTarget` returns null, and the release does nothing. No quote, no reorder, no feedback.

Repro: open the chat, trigger a clarifying question, drag a widget over the panel and release. **Ask One** in the menu still works, so the gesture is the only thing that dies.

```ts
const canAcceptWidgetDrop =
  acceptsWidgetDrop && activeGame === null && isVisible && !isClarifying
```

Then the marker is the only statement of eligibility, the removal effect fires when clarifying starts mid-drag, and the release falls through to a normal reorder like it does over Pong. The `|| isClarifying` checks on 161/186/193 become redundant.

Pong is already folded in here, which is why this was easy to miss.

albertpmz (changes requested):
Good job! Let's check this nit below

albertpmz (approved):

PR #5150, opened 2026-08-17
fix(OneTable): keep the hover background on sticky rows

factmarc10010:
## Problem

In a nested table, the root row greys on hover while collapsed and **stops greying once you expand it**. An expanded root row becomes sticky (`useStickyParentRow`), and sticky rows had no hover state at all. Sticky group header rows had the same gap.

## Root cause

The sticky branch in `TableRow` asked for the hover background in Tailwind **v4** trailing-`!` important syntax:

```ts
sticky && "hover:bg-f1-background-hover! sticky z-20 bg-f1-background"
```

Two things go wrong at once:

1. `packages/react` is on **Tailwind v3.4**, which compiles no such class — verified against the built CSS: `.hover\:bg-f1-background-hover:hover` exists, `.hover\:bg-f1-background-hover\!:hover` does not.
2. **tailwind-merge 2.6 still reads it as conflicting** with the plain `hover:bg-f1-background-hover` the row inherits from `ui/table`, and drops that one:

```
twMerge(base, rowCls, sticky)
→ "… hover:bg-f1-background-hover! sticky z-20 bg-f1-background"
```

So a sticky row ended up with only the opaque `bg-f1-background` that sticky needs, and no hover utility whatsoever.

## Fix

Express it in the v3 form, `hover:!bg-f1-background-hover`, which does compile and which tailwind-merge keeps. This was the only occurrence of the v4 trailing-`!` syntax in `packages/react`.

## Testing

New `TableRow.test.tsx`: a plain row and a sticky row each keep a hover utility that Tailwind v3 can compile (a regex that rejects the trailing-`!` form). The sticky case is red on `main` and green here.

`src/experimental/OneTable` + `src/patterns/OneDataCollection/visualizations/collection/Table` — 162 tests green. oxfmt / oxlint clean.

Also verified by compiling a probe through the package's own Tailwind config: `hover:!bg-f1-background-hover` generates a selector, `hover:bg-f1-background-hover!` generates nothing.


albertpmz (approved):
LGTM

PR #4463, opened 2026-06-16
fix(F0Dialog): clip tab divider and refine resource side-panel header

ReddyWish:
## Changes

- **Tab divider overflow.** The tab strip's `-mx-2` — a long-standing shim
  that realigns tab labels to the dialog's 16px gutter (the shared `Tabs`
  component ships a 24px page gutter) — widened the nav by 8px per side, so
  `TabNavigation`'s full-bleed bottom divider overran the rounded side-panel
  container. Wrap the strip in a container-width `overflow-hidden` so the
  bleed is clipped and the divider stops at the rounded edge. Tab alignment
  is unchanged.
- **Back control icon.** `ChevronLeft` → `ArrowLeft` (variant stays `outline`).
- **Resource header gutter.** Scope the dialog's `BaseHeader` to `px-4`
  (16px) via `[&_.resource-header]:px-4`, so the identity band aligns with
  the controls row. Scoped to the dialog — the full-page `ResourceHeader`
  keeps `px-page` (24px).

## Testing

- `oxlint` clean on the changed file.
- Verified locally (built via `f0-build.sh`, linked into factorial): on the candidate side panel the divider sits flush to the rounded edge, the back button shows the outlined arrow, and the resource header is at the 16px gutter; the full-page `ResourceHeader` is visually unchanged.

**Before:**

https://github.com/user-attachments/assets/ac3999cf-9386-439d-b22c-2e15e293822d

**After:**

https://github.com/user-attachments/assets/8e31a3c9-20c9-4161-84f5-f74ce9014b3c

PR #5160, opened 2026-08-18
fix(F0Dialog): stop scrollable content from squeezing the tabs strip

JaRomero95:
## 🚪 Why?

When an `F0Dialog` has `tabs` and the content is tall enough to scroll, the tab strip gets squeezed. The taller the content, the smaller the strip. With a long list it disappears and the list is drawn where the tabs used to be.

This is not app specific. On plain `main`, in this repo's own `Patterns/Dialog > With Person List Items` story:

image

## 🔑 What?

`shrink-0` on that div, so the strip keeps its height and the scroll area takes the overflow. The divider clipping from #4463 still works.

[FCT-61502]: https://factorialmakers.atlassian.net/browse/FCT-61502?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Before

image

## After

image

### DIFF
--- a/packages/react/src/patterns/F0Dialog/__tests__/F0Dialog.tabs-strip.test.tsx
+++ b/packages/react/src/patterns/F0Dialog/__tests__/F0Dialog.tabs-strip.test.tsx
@@ -1,0 +1,35 @@
+import { describe, expect, it, vi } from "vitest"
+
+import { zeroRender as render, screen } from "@/testing/test-utils"
+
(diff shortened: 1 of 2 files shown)


albertpmz (approved):
Super nice! Approved! 👍🏼

PR #5161, opened 2026-08-18
feat(DateNavigator): select consumer-defined periods

MohamedHegab:
## Description

Some collections navigate by ranges no calendar granularity can express — payroll cycles that run from the 25th to the 24th, academic terms, fiscal periods — and until now the date navigator could only offer day/week/month/quarter/half-year/year/range, so those teams had no way to let a user pick their own cycle. `OneDateNavigator` and the DataCollection `date-navigator` navigation filter now accept a `periods` definition whose ranges become an extra entry in the "Selected by" list, named by the consumer, listing every period with its own label and dates instead of a calendar grid.

## Type of change

New stories to review (Chromatic covers both):

- `Patterns/Filters/DateNavigator` → **With Periods**
image

- `Data Collection/Navigation Filters` → **With Periods**

The popup keeps the existing two-column layout: the granularity list on the left gains a `Payroll` entry, and the right side lists each period as its label ("August 2026") over its actual range ("Jul 25 → Aug 24"), with the active one highlighted and an optional header above the list.

## Implementation details

- feat: add a `periods` granularity built from a consumer-supplied list of ranges

  <details>
  <summary>Why a granularity rather than presets?</summary>

  Presets replace the whole popup with a flat list and are a one-shot jump, whereas these ranges must behave like a unit of navigation: snapping a value, labelling the trigger, and bounding prev/next. Modelling them as a granularity means every existing consumer of `GranularityDefinition` — the trigger, the calendar, the collection filter — works unchanged. `createPeriodsGranularity` binds the list, mirroring how `createWeekGranularity` binds `weekStartsOn`.
  </details>

- feat: offer the periods entry in the granularity selector, named by the definition's `label`

  <details>
  <summary>Discovery rule</summary>

  Supplying `periods` is what makes the entry selectable; listing `"periods"` in `granularity` only controls where it sits in the list. That keeps the common case to a single prop while still allowing explicit ordering.
  </details>

- feat: step the navigation arrows from one period to the next, bounded by `min`/`max`

  <details>
  <summary>Overlap, not containment</summary>

  Consumer periods rarely line up with the bounds, so a period is reachable as soon as it overlaps them. Full containment — the rule the calendar granularities use — would make the current payroll cycle unreachable under a `max` of `endOfMonth(today)`, which is exactly how time tracking configures the filter. The grid disables cells under the same rule.
  </details>

- feat: accept `periods` on the DataCollection `date-navigator` navigation filter
- feat: hide the calendar header and date input in the periods view, since the list is the full set of selectable values
- docs: document the option and add stories for the navigator and for a collection
- test: cover snapping, labelling, prev/next bounds, selection and the collection filter (27 new tests)

### API surface

The API-diff bot flags `granularityDefinitions` gaining a `periods` key and the `compareTo`/preset unions widening as breaking. They are union widenings, not real breaks: existing keys, call signatures and behavior are unchanged, and `getGranularityDefinitions` still accepts its original positional `weekStartsOn` alongside the new options object.

The flag comes from registering the empty `periodsGranularity` in the static record, which is what makes `"periods"` a valid `GranularityDefinitionKey` everywhere — including places it cannot do anything useful, such as form-field presets. That is the cost of keeping the key a closed union derived from the record; a `"periods"` granularity with no data renders its empty state rather than failing, so the widening is safe.

### Degradation

A value that falls outside every period keeps its own day range rather than throwing, so a persisted or out-of-range value still renders and the trigger falls back to the formatted date. A picker given no periods renders an empty-state message instead of an empty grid.


albertpmz (changes requested):
`granularities/index.tsx` has a raw NUL byte on [line 45](https://github.com/factorialco/f0/blob/3cbea2d445bb853d4ad958a697a76f83ac5793a8/packages/react/src/components/OneCalendar/granularities/index.tsx#L45), in the cache key separator. `cat -v` renders it as `^@`:

```
].join("^@")
```

Git and GitHub both read that as a binary file, so all 113 lines diff as "Binary file not shown": no inline comments, no blame, and the `+1,120` headline undercounts the change.

Runtime is fine, a NUL is a valid string char. The fix is to make the separator printable, since the fields are already `|` delimited:

```diff
-  ].join(<the literal NUL>)
+  ].join("|~|")
```

Worth avoiding the `\u0000` escape as the fix: it works, but copying it around is how the raw byte gets reintroduced.

albertpmz (approved):
LGTM!

PR #5166, opened 2026-08-18
fix(Avatar): anchor the badge tooltip to the whole avatar

hugohajaali-ops:
## Description

`badge.tooltip` on avatars has never rendered: `BaseAvatar` handed the badge to `Tooltip` as a Fragment, and Radix's `asChild` Slot silently drops the hover handlers and anchoring ref on a child that can't carry props. This PR fixes it at both levels — `Tooltip` now wraps unslottable children (Fragment/string/array) in a real inline element so the Slot always gets a trigger, and `BaseAvatar` anchors the badge tooltip to the avatar's outer wrapper element, so the badge renders in its exact normal position and the hover target is the whole avatar rather than the tiny corner badge.

Replaces the earlier avatar-level `tooltip` prop approach on this branch: with the badge tooltip anchored to the avatar, the extra prop is no longer needed. Same content as #5315.

## Implementation details

- fix: wrap Fragment/string/array Tooltip children in an inline-flex span so the Radix Slot has a real trigger element
  <details>
  <summary>Why inline-flex and not display: contents?</summary>

  Radix positions the bubble from the trigger's `getBoundingClientRect`. A `display: contents` element generates no box, so the tooltip would anchor to an empty rect.
  </details>
- fix: anchor the avatar badge tooltip to the whole avatar wrapper — a real slottable element, so no wrapper span is needed and the badge keeps its exact position
- test: pin both behaviors — unslottable Tooltip children open on hover, and the badge tooltip opens when hovering the avatar

### DIFF
--- a/packages/react/src/components/avatars/F0Avatar/types.ts
+++ b/packages/react/src/components/avatars/F0Avatar/types.ts
@@ -20,7 +20,6 @@ export type AvatarBadge = (
       icon: BadgeProps["icon"]
     }
 ) & {
-  /** Shown when hovering the avatar */
   tooltip?: string
 }
 
--- a/packages/react/src/components/avatars/F0Avatar/types.ts
+++ b/packages/react/src/components/avatars/F0Avatar/types.ts
@@ -20,6 +20,7 @@ export type AvatarBadge = (
       icon: BadgeProps["icon"]
     }
 ) & {
+  /** Shown when hovering the avatar */
   tooltip?: string
 }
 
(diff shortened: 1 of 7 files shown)


sauldom102 (changes requested):
Probably you can just use https://f0.factorial.dev/?path=/docs/components-tooltip--documentation in the monorepo wrapping your avatar without needing to add this change in F0

albertpmz (approved):
LGTM!

sauldom102 (approved):

PR #5169, opened 2026-08-19
feat(OneDataCollection): select a node on graph entry via initialSelectedNodeId

AlbertGarcia-fact:
## Description

Add an opt-in `initialSelectedNodeId` option to the OneDataCollection **graph** visualization so a consumer can make the graph arrive with a node already **selected** (the click-selection ring), not just framed.

Until now the graph could frame a node on entry (`focusOnEntry`) and highlight a search reveal (`revealNodeId`), but there was no way to land in the `selected` state a user's own click leaves behind. Deep links (e.g. Job Catalog's "View Graph" from a node's detail page — FCT-60797) arrived framed but visually neutral, so it wasn't obvious which node the side panel belonged to.

## Type of change

## Implementation details

- **`collection/Graph/types.ts`** — new optional `initialSelectedNodeId?: string` on `GraphVisualizationOptions`, documented next to `focusOnEntry` (its viewport-framing sibling).
- **`collection/Graph/index.tsx`** — when the option is set, the graph runs F0Graph's selection in **controlled** mode, seeded with that node on the first render, and mirrors every later change back via `onSelectedNodesChange` so clicks/keyboard keep moving the selection normally. The control mode is decided **once at mount** (a stable boolean), so consumers that omit the option keep the previous **uncontrolled** behavior unchanged. No `F0Graph` changes — it already supports controlled `selectedNodes`/`onSelectedNodesChange`.
- **`__stories__/visualizations/graph.stories.tsx`** — new `SelectedOnEntry` story (paired with `focusOnEntry` on the same node).

Pair `initialSelectedNodeId` with `focusOnEntry` (same id) so the node's branch is expanded/framed and the ring is visible.

### Non-breaking

New **optional** prop; when omitted, `selectedNodes` stays `undefined` and selection is uncontrolled exactly as before. `check:api-surface` classifies this as SAFE.

### Verification

- oxfmt ✅ · oxlint ✅ · `tsc` clean for the edited files
- graph collection unit tests: 30/30 ✅

## Context

- Unblocks factorial **FCT-60797** ("[Job Catalog] Mark the deep-linked node as selected when arriving at the graph from View Graph"), the follow-up to FCT-60313.

### DIFF
--- a/packages/react/src/patterns/OneDataCollection/__stories__/visualizations/graph.stories.tsx
+++ b/packages/react/src/patterns/OneDataCollection/__stories__/visualizations/graph.stories.tsx
@@ -711,12 +711,15 @@ const DetailField = ({ label, value }: { label: string; value: string }) => (
 const OrgChartExample = ({
   defaultExpandDepth,
   focusOnEntry,
+  initialSelectedNodeId,
   graphLabel,
   tableLabel,
   lockedTagTypes,
 }: {
   defaultExpandDepth: number
   focusOnEntry?: string
+  /** Node marked as selected (the click ring) on entry — see the graph options. */
+  initialSelectedNodeId?: string
   /** Custom label for the graph chip; falls back to the localized "Graph". */
   graphLabel?: string
   /** Custom label for the table chip; falls back to the localized "Table". */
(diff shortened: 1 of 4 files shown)


eliseo-juan (approved):
LGTM

PR #5180, opened 2026-08-19
fix(OneDataCollection): make the all-items-selected label visible on the dark action bar

hugohajaali-ops:
## What

When a data collection's bulk-selection spans all pages ("select all N items"), the action bar's **"All N items selected"** label rendered black-on-black and was unreadable. Reported from the bulk compensation flow (screenshot showed *"133 éléments sélectionnés"* invisible on the dark bar).

## Why

The action bar's left content lives inside a `dark`-themed wrapper. The partial-selection branch sets `text-f1-foreground` explicitly (resolving to the light foreground inside `dark`), but the all-items-selected branch didn't — so the label inherited the page's default dark text.

## How

- Add `text-f1-foreground` to the all-items-selected span in `OneDataCollectionActionBar`.
- New `AllItemsSelected` story on `Data Collection/ActionBar` covering this state.
- Regression unit test (red on main, green with the fix): asserts the label carries `text-f1-foreground`.

## Verification

Reproduced the bug in Storybook (label invisible), then confirmed the fix renders "All 133 items selected" in light text on the dark bar. `oxfmt`, `oxlint` and the new vitest spec pass; `tsc` errors are pre-existing on main (39 before/after, none in these files).


albertpmz (approved):
LGTM!

PR #5183, opened 2026-08-19
fix(F0Graph): frame the entry target (initialFocusNodeId) reliably on arrival

AlbertGarcia-fact:
## Description

Arriving at a graph via `initialFocusNodeId` (OneDataCollection `focusOnEntry`) did **not** frame the target: the graph opened fit-to-all with the node un-centered — only re-clicking the node later centered it. This fixes the entry framing so it lands like a node click (centered, panel-aware).

Surfaced by Job Catalog's "View Graph" deep link (FCT-60797): the node arrived selected (f0#5169 / FCT-61650) but never framed.

## Type of change

- [x] Bug fix

## Root cause (two timing issues)

1. **Framed before measurement.** The initial frame ran at mount, before React Flow measured its container, so the imperative `fitView`/`setCenter` applied against zero dimensions and never took. The click / `focusedNode` / "Find me" flies work only because they defer past measurement (`FOCUS_SETTLE_DELAY_MS`). So the entry frame now flies through the **same click path** (`flyToNodeClickRef` — centered at the node-click zoom, `viewportInset`-aware), deferred by the same delay.
2. **The deferred fly was cancelled by a re-render.** It was scheduled in an effect whose cleanup cleared the timer, keyed on `renderedNodeIds`. Real consumers churn that set on the first renders (two-phase hydration, an entry side panel opening), so the cleanup cancelled the pending fly while the one-shot guard blocked re-scheduling — it silently never ran (static-node tests didn't churn, so they passed). It is now scheduled **once** and torn down **only on unmount**.

A whole-graph entry (no focus target) still fits immediately — unchanged.

## Implementation

* `packages/react/src/patterns/F0Graph/components/F0GraphView/F0GraphView.tsx` — entry frame flies via the click path, deferred, scheduled once (cleared on unmount).
* Tests: new `F0Graph.initialFocusFrame.test.tsx` (deferred entry frame + the re-render-during-delay regression); `F0Graph.focusNoRefire.test.tsx` and `F0Graph.clickCenter.test.tsx` updated for the deferred timing.

## Verification

* 274 F0Graph unit tests pass · oxfmt · oxlint clean.
* Verified in the Factorial app (Job Catalog "View Graph") via an alpha build: the node now arrives selected + centered/framed + panel, like a direct click.

## Context

* Blocks factorial **FCT-60797**. Split out of the selection PR **#5169** (FCT-61650) — they are independent (this touches `F0Graph`, that touches the OneDataCollection graph visualization).
* Also fixes the employees Org chart "open looking at me" (`focusOnEntry`) entry framing.

### DIFF
--- a/packages/react/src/patterns/F0Graph/__tests__/F0Graph.clickCenter.test.tsx
+++ b/packages/react/src/patterns/F0Graph/__tests__/F0Graph.clickCenter.test.tsx
@@ -96,8 +96,11 @@ async function clickNode(id: string) {
 }
 
 // Mount, then clear the camera spies so only the click's calls are counted
-// (isolates the click behavior from any mount-time framing).
-const mount = (props: Partial<React.ComponentProps<typeof F0Graph>> = {}) => {
+// (isolates the click behavior from any mount-time framing). The one-shot entry
+// frame runs from React Flow's `onInit` (deferred), so flush it before clearing.
+const mount = async (
+  props: Partial<React.ComponentProps<typeof F0Graph>> = {}
+) => {
   const view = zeroRender(
     <div style={{ width: 800, height: 600 }}>
       <F0Graph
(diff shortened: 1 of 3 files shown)


eliseo-juan (approved):
LGTM

PR #5187, opened 2026-08-20
fix(EditableTable): hide select chevron on disabled cells

rocio-fdez-factorial:
## Description

Disabled `EditableTable` cells with a `selectConfig` showed the select dropdown chevron even though the cell isn't clickable, which reads as broken affordance. Adds an opt-in `disabledConfig.hideSelectChevron` column option so a consumer can hide it where it applies, without changing the default behavior for existing disabled cells.

Jira: https://factorialmakers.atlassian.net/browse/FCT-60852

## Screenshots

**Before** (chevron 
Screenshot 2026-08-19 at 18 34 23

**After** (`disabledConfig: { hideSelectChevron: true }`):
Screenshot 2026-08-20 at 11 02 37

## Implementation details

- feat: add `disabledConfig.hideSelectChevron` to `EditableTableColumnDefinition`
- fix: skip rendering the select chevron in `ReadOnlyCellContent` when `disabledConfig.hideSelectChevron` is set
- test: cover both the default (chevron shown) and opt-in (chevron hidden) cases in `DisabledCell.test.tsx`

### DIFF
--- a/packages/react/src/patterns/OneDataCollection/visualizations/collection/EditableTable/__tests__/DisabledCell.test.tsx
+++ b/packages/react/src/patterns/OneDataCollection/visualizations/collection/EditableTable/__tests__/DisabledCell.test.tsx
@@ -93,6 +93,33 @@ describe("DisabledCell", () => {
     expect(screen.queryByRole("button")).not.toBeInTheDocument()
   })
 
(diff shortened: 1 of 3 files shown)


eliseo-juan (approved):
LGTM

PR #5188, opened 2026-08-20
feat(OneDataCollection): freeze table columns from settings

zygisS22:
## Description

Semantic-layer tables need an Excel-like way to keep whichever columns matter visible while scrolling, rather than relying only on the legacy implicit first column. This change adds an opt-in, settings-popover-only API for freezing and unfreezing multiple columns.

The scope is deliberately limited to Table settings: there are no new controls in column headers, so existing header information and sorting affordances are unchanged.

## Type of change

## Behavior

- Users can freeze multiple columns independently from the Table settings popover while one visible managed column remains unlocked as the scrollable region.
- If a controlled `lockedColumnIds` value requests every managed column, the table normalizes the final ordered column back to visible and unlocked.
- The final visible unlocked column cannot be hidden or locked. Bulk Hide all preserves it.
- Freezing a column moves it into the sticky group on the left, after any permanent `frozenColumns`, in `lockedColumnIds` order.
- Unfreezing restores the column to its saved logical position.
- Frozen columns stay visible and cannot be hidden, reordered, or removed.
- Locked columns remain visible and keep stable sticky offsets when their header group collapses; collapse also retains a scrollable column.
- Locked rows use the closed-lock icon. Unlocked rows reveal a named closed-lock action on hover or keyboard focus.
- Pointer activation does not leave the replacement row action focused or visually stuck. Keyboard activation keeps focus on the corresponding replacement control.

## Compatibility and ownership

- Adds `lockedColumnIds?: readonly ColId[]` and `onLockedColumnIdsChange?: (columnIds: ColId[]) => void`.
- The feature is opt-in. Omitting both props preserves the existing first-column behavior and UI.
- Passing only `lockedColumnIds` creates read-only managed locks without exposing controls.
- Passing the callback enables the lock/unlock controls and leaves state ownership with the host.
- Existing `frozenColumns` remain permanently frozen before the managed group and cannot be unlocked.
- Reordering unlocked rows preserves locked rows in the saved order, so later unfreezing restores the correct position.

## Storybook and visual review

https://github.com/user-attachments/assets/7b67cb2f-7470-40e1-b168-67218fb08270

With horizontal scroll

https://github.com/user-attachments/assets/500a3e45-4b1a-4a4a-8b23-cee8469f0698

The behavior is documented and executable in Storybook:

- `Patterns / Data Collection / Visualizations / Table` → `With Lockable Columns`
  - Renders 15 columns in a 720 px viewport, producing a 2,600 px-wide table.
  - Opens Table settings and unlocks/relocks the first column by keyboard.
  - Freezes Role beside Name and verifies both sticky headers and body cells.
  - Scrolls to the exact 1,880 px maximum offset and proves the far-right Status column is visible without being covered by the frozen group.
  - Verifies the restored saved position after unfreezing.
  - Covers pointer focus cleanup and keyboard focus handoff.
  - Leaves the final multi-lock, far-right scroll state and an unlocked row's actions visible for visual review.
- `Patterns / Data Collection / Visualizations / Table` → `With Lockable Columns And Collapsed Group`
  - Starts with Email and Role locked inside a collapsed Employment details group.
  - Verifies that locked headers stay sticky at stable, non-overlapping offsets while Department collapses.
  - Expands and collapses by keyboard, retaining focus, `aria-expanded`, sticky offsets, and visible column order.
  - Leaves the group collapsed with Name as the scrollable boundary for visual review.
- `Patterns / Data Collection / Internal / SortAndHideList` → `With Lockable Items`
- `Patterns / Data Collection / Internal / SortAndHideList` → `Snapshot`

The public and internal MDX pages explain the API, opt-in boundary, remaining-scrollable-column invariant, saved-position behavior, collapse interaction, pointer behavior, keyboard path, and horizontal freeze behavior. Chromatic will attach the hosted preview to this PR.

Suggested PR evidence:

1. Settings open after locking all but one managed column, showing that the final visible unlocked row has no Lock action and its visibility switch is disabled.
2. The 15-column table scrolled fully right, with Name and Role still sticky on the left and Status visible.
3. The collapsed Employment details story showing Email and Role sticky, Department hidden, and Name still scrollable.
4. Optional short recording: freeze Role, scroll sideways, unfreeze it, and show it returning to its original position without the row actions staying focused.

## Verification

- Focused Vitest after the feedback fix: 4 files, 99 tests passed.
- Repository pre-push changed-test gate: 5 files, 112 tests passed.
- React TypeScript: passed.
- Full React format and focused lint: passed with 0 warnings/errors.
- Pre-commit cycle-dependency, format, lint, and file-size checks: passed.
- Production Storybook build: passed.
- Original public lock/scroll story interaction run in Chrome: PASS, 50 assertions.
- New collapse/lock story in Chrome: PASS; Email and Role remained sticky at `0px` and `240px`, Department collapsed, Name remained scrollable, keyboard focus stayed on the toggle, and the console was clean.
- Settings interaction in Chrome: PASS; after locking every other managed column, the final Lock action was unavailable and the remaining visibility switch was disabled.
- Horizontal-scroll geometry: 2,600 px table / 720 px viewport / 1,880 px exact maximum offset.
- Storybook accessibility panel: 0 violations, 20 passes, 2 inconclusive.
- Independent correctness, compatibility, accessibility, documentation, Storybook, and coverage reviews: no blocking findings.

Design review: pending for the new stable Table-settings affordance.

---
> - factorial-f0
> - factorial-f0-component-documentation
> - factorial-pr

### DIFF
--- a/packages/react/src/patterns/OneDataCollection/visualizations/collection/Table/Table.tsx
+++ b/packages/react/src/patterns/OneDataCollection/visualizations/collection/Table/Table.tsx
@@ -157,7 +157,7 @@ export const TableCollection = <
     lockedColumnIds !== undefined || !!onLockedColumnIdsChange
 
   // Sorted and hidden columns
-  const { columns: orderedColumns } = useColumns(
+  const { columns: orderedColumns, stickyColumnIds } = useColumns(
     originalColumns,
     frozenColumns,
     visualizationSettings ?? settings.visualization?.table,
@@ -165,6 +165,10 @@ export const TableCollection = <
     allowColumnHiding,
     lockedColumnIds,
     usesExplicitColumnLocking
+  )
+  const stickyColumnIdSet = useMemo(
+    () => new Set(stickyColumnIds),
+    [stickyColumnIds]
   )
 
   // Header groups own the collapsed state and drop the columns hidden by a
@@ -179,6 +183,7 @@ export const TableCollection = <
   } = useHeaderGroups(orderedColumns, {
     headerGroups: headerGroupsOption,
     onCollapsedChange: onHeaderGroupCollapsedChange,
+    preservedColumnIds: stickyColumnIdSet,
   })
 
   const tableContainerRef = useRef<HTMLDivElement>(null)
@@ -250,22 +255,7 @@ export const TableCollection = <
     // eslint-disable-next-line react-hooks/exhaustive-deps --  we don't want to re-run this effect when the filters change, just when the data changes
   }, [paginationInfo?.total, data.records])
 
-  const frozenColumnIds = useMemo(
-    () =>
-      new Set([
-        ...originalColumns.slice(0, frozenColumns).map(getColumnId),
-        ...(lockedColumnIds ?? []),
-      ]),
-    [originalColumns, frozenColumns, lockedColumnIds]
-  )
-  const frozenColumnsLeft = useMemo(() => {
-    const firstUnlockedColumnIndex = columns.findIndex(
-      (column) => !frozenColumnIds.has(getColumnId(column))
-    )
-    return firstUnlockedColumnIndex === -1
-      ? columns.length
-      : firstUnlockedColumnIndex
-  }, [columns, frozenColumnIds])
+  const frozenColumnsLeft = stickyColumnIds.length
   const getRowKey = (item: R, index: number) => {
     if ("id" in item && item.id !== undefined && item.id !== null) {
       return `id:${String(item.id)}`
--- a/packages/react/src/patterns/OneDataCollection/visualizations/collection/Table/Table.tsx
+++ b/packages/react/src/patterns/OneDataCollection/visualizations/collection/Table/Table.tsx
@@ -111,8 +111,8 @@ export const TableCollection = <
   onLoadError,
   allowColumnHiding,
   allowColumnReordering,
-  lockedColumnId,
-  onLockedColumnChange,
+  lockedColumnIds,
+  onLockedColumnIdsChange,
   referenceRowType,
   boldRootRows,
   headerGroups: headerGroupsOption,
@@ -153,6 +153,8 @@ export const TableCollection = <
   )
 
   const { settings } = useDataCollectionSettings()
+  const usesExplicitColumnLocking =
+    lockedColumnIds !== undefined || !!onLockedColumnIdsChange
 
   // Sorted and hidden columns
   const { columns: orderedColumns } = useColumns(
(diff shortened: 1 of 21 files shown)


packages/react/src/patterns/OneDataCollection/visualizations/collection/Table/Table.tsx
```diff
@@ -244,7 +250,22 @@ export const TableCollection = <
...
+        ...originalColumns.slice(0, frozenColumns).map(getColumnId),
+        ...(lockedColumnIds ?? []),
+      ]),
+    [originalColumns, frozenColumns, lockedColumnIds]
+  )
+  const frozenColumnsLeft = useMemo(() => {
+    const firstUnlockedColumnIndex = columns.findIndex(
+      (column) => !frozenColumnIds.has(getColumnId(column))
+    )
+    return firstUnlockedColumnIndex === -1
```
albertpmz: When every column is locked, `findIndex` returns `-1` and this falls through to `columns.length`, so `frozenColumnsLeft` pins the entire table: every column goes `position: sticky`, the "rest" header cell at line 806 drops out (`columns.length - columns.length > 0` is false), and there's no scrollable region left. Nothing stops a user reaching this since each non-frozen column keeps its lock button. Can we drop the lock affordance on the last remaining unlocked column, or cap the managed frozen group so at least one scrollable column survives?

packages/react/src/patterns/OneDataCollection/visualizations/collection/Table/Table.tsx
```diff
@@ -244,7 +250,22 @@ export const TableCollection = <
...
+  const frozenColumnIds = useMemo(
+    () =>
+      new Set([
+        ...originalColumns.slice(0, frozenColumns).map(getColumnId),
+        ...(lockedColumnIds ?? []),
+      ]),
+    [originalColumns, frozenColumns, lockedColumnIds]
+  )
+  const frozenColumnsLeft = useMemo(() => {
+    const firstUnlockedColumnIndex = columns.findIndex(
```
albertpmz: `columns` here is the `useHeaderGroups` output, which drops columns hidden by a collapsed group (comment at line 170). Before this PR `frozenColumnsLeft` was the constant `frozenColumns` prop; now it's a `findIndex` over that collapse-mutated array. If a frozen or locked column sits inside a collapsible group, collapsing it recomputes the count and the sticky offsets can jump mid-animation. Probably fine in current usage, but the count is no longer invariant to collapse state. Worth a story that collapses a group containing a locked column, or deriving the frozen count from the pinned set directly rather than scanning the rendered array.

albertpmz (changes requested):
Nice work! Let's check these 2 thingies below :)

albertpmz (approved):
Both review comments are properly addressed.

`frozenColumnsLeft` is now derived from the pinned set instead of scanning the collapse-mutated array, and `preservedColumnIds` keeps sticky columns rendered through a collapse so the count stays aligned with the rendered prefix. The remaining-scrollable-column invariant is enforced in `useColumns` (normalizing controlled input), in the settings affordances, and in bulk hide.

Verified locally: 190 tests pass across the touched suites, `tsc --noEmit` clean on packages/react.

PR #5193, opened 2026-08-20
feat(CardSelectable): add an opt-in compact density

ReddyWish:
## What

Adds an opt-in `compact` prop to `CardSelectableContainer`: standalone card padding `p-4` → `p-3` (16px → 12px).

Opt-in, so all current call sites keep today's density. In `grouped` layout the rows already use `px-4 py-3`, so `compact` is a no-op there.

## This follows the existing F0Card pattern

Not a new concept: `F0Card` already exposes exactly this knob, and this PR mirrors it rather than inventing an API.

```tsx
// F0Card — CardInternalProps
/** Whether the card has a compact layout */
compact?: boolean
...
compact && "p-3",   // CardInternal.tsx:267 and :500, over a default p-4
```

Same prop name, same boolean shape, same `p-4` → `p-3` flip. Across `packages/react/src/components`, 16px is the card family's default (`F0Card`, `CardSelectable`, `F0Select`'s `SelectionPreview`) and 12px is the compact density (`F0Card` compact, dialog wrapper, `F0NumberInput`, audio/video player cards). `CardSelectable` was simply the one card primitive missing the density knob its sibling has.

## Why opt-in rather than changing the default

`CardSelectableContainer` is not only used directly — inside f0 it also backs `F0Form`'s `cardSelect` field renderer and `SwitchGroupRenderer`. In the Factorial monorepo alone there are 27 direct call sites plus two `fieldType: 'cardSelect'` forms, spanning trainings, expenses, banking, procurement, benefits, employees, IT, compensations and job catalog. Changing the default would silently restyle all of them.

## Why it is needed

The form designs now landing across modules — ATS "Schedule an interview" ([FCT-61677](https://factorialmakers.atlassian.net/browse/FCT-61677)) and the trainings session form among them — specify 12px padding on these cards. The value was hardcoded in `CardSelectable.tsx` with no prop reaching it, so a consumer's only alternative was overriding f0 internals with CSS, which the boundary policy rules out.

**Question for Design:** if 12px is meant to be the density for *every* selectable card rather than a per-screen choice, that is a design-system decision of its own and would want the same treatment on `F0Card`. Happy to turn this into a default flip plus a Chromatic sweep instead — say the word, and this PR can be dropped.

A secondary effect worth noting: with three horizontal cards the label box is `min-w-0 flex-1`, and f0's global `div[role=dialog] { overflow-wrap: anywhere }` lets a single long word break mid-word once that box is narrower than the word. `compact` gives each card 8px more room for text, which clears the threshold in the ATS wizard's layout. It does not remove the underlying behaviour, so a follow-up on the title's break behaviour (`break-keep`, or nowrap + truncate with a tooltip) may still be worth doing.

## Tests

- Four unit tests in `CardSelectable.test.tsx`: the default keeps `p-4`; `compact` yields `p-3`; `grouped` + `compact` is a no-op; and `compact` keeps the `moreInfoLink` row aligned with the tightened card. That last one needed a fix — the link row is a sibling of the interactive header (to avoid `nested-interactive`) and hardcoded `px-4 pb-4`, so under `compact` the link sat 4px past the title with 16px of bottom padding against 12px sides.
- 14 tests in the file pass; `oxlint` and `oxfmt` clean; `tsc --project tsconfig-build.json` clean.
- `Compact` story next to `Horizontal`, covering both the horizontal row and the `moreInfoLink` case.

[FCT-61677]: https://factorialmakers.atlassian.net/browse/FCT-61677?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/react/src/components/CardSelectable/CardSelectable.tsx
+++ b/packages/react/src/components/CardSelectable/CardSelectable.tsx
@@ -94,6 +94,8 @@ interface CardSelectableProps<T extends CardSelectableValue> {
   isToggle?: boolean
   /** When true, renders without individual card borders (for grouped layout) */
   grouped?: boolean
+  /** When true, uses 12px padding instead of 16px (standalone cards only) */
+  compact?: boolean
 }
 
 function _CardSelectable<T extends CardSelectableValue>({
@@ -104,6 +106,7 @@ function _CardSelectable<T extends CardSelectableValue>({
   onSelect,
   isToggle,
   grouped,
+  compact,
 }: CardSelectableProps<T>) {
   const { forms } = useI18n()
   const shouldReduceMotion = useReducedMotion()
@@ -166,7 +169,7 @@ function _CardSelectable<T extends CardSelectableValue>({
         className={cn(
           "flex cursor-pointer items-center gap-3",
           "focus-visible:outline-none focus-visible:ring-1 focus-visible:ring-inset focus-visible:ring-f1-special-ring",
-          grouped ? "px-4 py-3" : "p-4",
+          grouped ? "px-4 py-3" : compact ? "p-3" : "p-4",
           // The link row below supplies the bottom padding so the gap between
           // description and link matches the old in-column `gap-2`.
           moreInfoLink && "pb-0"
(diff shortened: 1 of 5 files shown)


eliseo-juan (approved):
LGTM

PR #5225, opened 2026-08-25
feat(F0Button): show a counter on buttons, styled right

AngelMsaa:
## What

Makes the button's counter pill look right: a neutral grey count next to the label, spaced and centered, and dark on the primary variant so it doesn't wash out.

- Neutral **default** (grey) counter instead of the `selected` teal — it reads as a count, not a toggled state. No prop to pick another type.
- 4px gap before it, vertically centered in the button height.
- On the **default (primary)** variant — a solid dark-red field — the counter renders as a dark pill (dark theme scoped to just the counter), regardless of the app's light/dark theme.

image

## Note

Touches the shared `F0Button`, so it also restyles the **OneFilterPicker applied-filters counter** (the only other caller) from teal to grey.

Letting a `OneDataCollection` secondary action actually request a counter is the paired PR #5249.


sauldom102 (changes requested):
Doesn't look a bit wrong in sm? 🤔 

Image

sauldom102 (approved):

desiree-np (approved):

PR #5228, opened 2026-08-25
fix(F0Graph): restore ARIA tree semantics flagged by axe

AngelMsaa:
## Why

The axe bot on #5051 reported **40 accessibility findings** across the F0Graph / F0GraphNode stories, in three rules:

- `aria-required-children` (WCAG 1.3.1 A, critical) — a `role="tree"` must own a `role="treeitem"`.
- `aria-required-parent` (WCAG 1.3.1 A, critical) — a `role="treeitem"` must be owned by a `tree`/`group`.
- `nested-interactive` (WCAG 4.1.2 A, serious) — on the Lazy story.

These are **pre-existing structural issues** surfaced by that PR touching the folder, so they're fixed here on their own branch off `main`.

### Root cause

React Flow renders its content inside a hardcoded `<div class="react-flow" role="application">` (can't be overridden via props). That element sits **between** F0's `role="tree"` container and the `role="treeitem"` nodes, severing the ARIA ownership chain: the tree sees no owned treeitem, and each root node sees `application` (not `tree`/`group`) as its nearest owner. The design already re-owns *nested* levels via per-node `aria-owns`; only the **roots** lacked an owner.

## What

- **Tree owns its forest roots** — the `role="tree"` container now `aria-owns` the rendered forest-root treeitems (roots, plus any node whose parent is windowed out). Combined with the existing per-node `aria-owns` for in-window children, every rendered treeitem has **exactly one owner** — no orphans (fixes the windowing stories too), no double-ownership. New `treeRootNodeIds` memo in `useGraphRenderModel`.
- **`aria-busy` on the empty tree** — while the tree renders no treeitems (deferred / staged / viewport-data loading, or an empty graph) it's marked `aria-busy`, which axe accepts in place of a required treeitem child.
- **Flatten expander/collapser** — the native `<button>` is now the single interactive control (roving `tabIndex`, `aria-expanded`, `aria-label`, and the focus-system ref), dropping the wrapping `role="button"` div that nested two interactive controls (`nested-interactive` on Lazy).
- **F0GraphNode stories** — a `role="tree"` decorator gives the isolated `role="treeitem"` a valid parent, with an `aria-owns` escape hatch (via a story parameter) for the ReactFlow-embedded `WithToolbar` story.

## Testing

- `pnpm vitest run --project=unit F0Graph` — 278 pass (added 3 `treeRootNodeIds` cases + 1 `aria-busy` case).
- `pnpm tsc` and `pnpm lint` clean.
- axe: relies on this PR's a11y bot re-running against the changed stories.

> [!NOTE]
> The expander/collapser change touches the roving-tabindex/focus wiring. Existing keyboard behavior is preserved by design, but a manual keyboard expand/collapse smoke test on the Lazy story is worth a look during review.

### DIFF
--- a/packages/react/src/patterns/F0Graph/__tests__/F0Graph.a11y.test.tsx
+++ b/packages/react/src/patterns/F0Graph/__tests__/F0Graph.a11y.test.tsx
@@ -377,6 +377,30 @@ describe("F0Graph a11y — tree structure roles", () => {
     expect(tree.getAttribute("aria-label")).toBe("Graph view")
   })
 
+  it("marks the tree aria-busy while it renders no treeitems", () => {
+    // React Flow renders 0 nodes in jsdom, so the tree owns nothing. Without a
+    // treeitem child a `role="tree"` fails axe `aria-required-children`; the
+    // empty tree is instead marked `aria-busy` so it stays valid (this is also
+    // the deferred / staged / viewport-data-loading state in the browser).
+    zeroRender(
+      <div style={{ width: 800, height: 600 }}>
+        <F0Graph nodes={makeNodes()} renderNode={renderNodeFn} />
+      </div>
+    )
+
+    const tree = screen.getByRole("tree", { name: "Graph view" })
+    if (tree.querySelectorAll("[role='treeitem']").length === 0) {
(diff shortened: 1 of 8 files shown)


sauldom102 (approved):
Thank you!! 💯

PR #5229, opened 2026-08-25
feat(storybook): add on-demand performance metrics scripts

sauldom102:
## Description

Follow-up to #5223, which added the Storybook performance panel. That panel only shows its numbers to someone with Storybook open; this exposes the same measurements as JSON, so an agent or a developer can check a component while building or changing it. Nothing here gates anything — no thresholds, no baseline, no required check.

```bash
pnpm perf-metrics F0Button --snapshot
pnpm perf-metrics F0Button | jq '.stories[0].deterministic'
```

📖 **Full documentation: [`packages/react/.scripts/perf-metrics.md`](packages/react/.scripts/perf-metrics.md)**

## What you can measure

Every story is measured on two groups of metrics, and the split is the point.

**`deterministic` — counts of work.** Identical on every run and machine, so they're safe to compare between runs, branches or components:

| | |
| --- | --- |
| `mounts` / `renders` / `updates` | how many times the tree rendered, and how much of that was after mount with no interaction |
| `cascades` | renders that scheduled another render |
| `slowUpdates` | updates that blew a 16ms frame |
| `domElements` | size of the rendered tree |
| `styleWrites` | inline style / CSS-var writes during mount |
| `forcedReflows` | layout read back synchronously after a style write |

**`timing` — wall-clock samples.** `mountMs`, `p95UpdateMs`, `cls`, `layoutShifts`, `totalBlockingTime`, `longTasks`, `fps`, `heapMB`. Useful as a rough signal, but they move between identical runs and include page-wide work — in Storybook `longTasks` is frequently axe-core rather than your component.

Measuring one story five times on an idle machine: every `deterministic` field was identical each time, while `totalBlockingTime` ranged 42–47ms. That is why the output is split, and why only `deterministic` metrics (plus the `cls` score) can raise a highlight.

## Implementation details

- feat: add `perf-metrics` — resolve stories by id, component name or substring, measure headlessly, print JSON to stdout
- feat: add `perf-changed.ts` — measure the stories a PR affects and reduce them to the few facts worth attention
- feat: add the `performance` agent prompt that turns that report into a PR comment
- docs: add `.scripts/perf-metrics.md`
- chore: add the `perf-metrics` package script

## Notes for reviewers

<details>
<summary>Thresholds come from the measured distribution, not intuition</summary>

The obvious rule — "flag any render cascade" — is useless here. **102 of 102 snapshot stories record at least one cascade** (median 3), because Storybook's own decorators and providers render around the story. The first version of this flagged 40 out of 40 measured stories.

Thresholds now sit at roughly p95 of the library:

| metric | p50 | p90 | p95 | max | threshold |
| --- | --- | --- | --- | --- | --- |
| cascades | 3 | 6 | 8 | 18 | 8 |
| updates | 6 | 10 | 13 | 26 | 13 |
| domElements | 89 | 678 | 884 | 5250 | 900 |

`forcedReflows > 0` and `slowUpdates > 0` stay absolute — base rates 0/40 and 2/40. On a real changed-set this highlights 4 stories out of 40.
</details>

<details>
<summary>Layout shift is judged by CLS score, never by shift count</summary>

Shift *count* is not reproducible — the same five F0Card stories measured three times gave three different sets, because a shift is only recorded when the browser happens to paint between the two layouts.

The *score* is stable once it clears the noise floor: `ApplicationFrame` scored 0.1943 / 0.1932 / 0.1936 across three runs, while every story that flickered scored ~0.0001. So the rule is `cls >= 0.01`. Across the library it selects 4 of 102 snapshot stories, worst being `patterns-app-shell-applicationframe--snapshot` at **0.198** — approaching the 0.25 Core Web Vitals calls "poor", and worth a look independently of this PR.
</details>

<details>
<summary>Which stories a PR measures</summary>

`perf-changed.ts` diffs every source file under `packages/react/src` and maps each to the stories that render it, walking up to the owning component directory. So a component-only change is measured — verified against `bf41fa6e4` (`fix(F0Chat)`), which reports 39 stories where a story-file-only rule reported none.

Test, snapshot and docs files are excluded: on that same commit the two `__tests__/` files alone attributed 38 stories, so a PR that only adjusted assertions would have triggered a full performance comment.

Each story records `measuredBecause` — `"story"` if its own file changed, `"source"` if only its component did — so the comment attributes a finding to the component rather than implying the author edited a story they never opened.
</details>

<details>
<summary>Two traps when editing these scripts</summary>

- **stdout is the JSON payload.** Progress goes through a `note()` helper writing to stderr, because `consola.info`/`consola.success` write to **stdout** — routing progress through them corrupts the payload and `| jq` fails with "Invalid numeric literal".
- **git pathspecs are cwd-relative.** `perf-changed.ts` anchors its globs with `:(top)`; without it, running from `packages/react` (where `pnpm --filter` puts you) matches nothing and the script reports "no files changed" on a PR that changed plenty.
</details>

### DIFF
--- a/.github/agent-prompts/performance.md
+++ b/.github/agent-prompts/performance.md
@@ -18,7 +18,8 @@ Turn a machine-generated performance measurement into a short, human-readable PR
 Each entry in `stories[]` has:
 
 - `id`, `title`, `name` — which story was measured
-- `storyFile` — the file it came from
+- `changedFile` — the changed file that pulled this story into the report
+- `measuredBecause` — `"story"` if the story's own file changed, `"source"` if a source file in its component changed but the story did not
 - `isNew` — whether this PR adds the story
 - `deterministic` — counts of work done: `mounts`, `renders`, `updates`, `cascades`, `slowUpdates`, `domElements`, `styleWrites`, `forcedReflows`, `layoutShifts`
 - `timing` — wall-clock samples
@@ -41,6 +42,7 @@ An empty `highlights` array means nothing crossed a threshold. That is the norma
 - **A cascade count above zero is normal here.** Every story in this library records at least one render cascade (median 3) because Storybook's own decorators and providers render around the story. The report only raises a cascade highlight well above that norm. Never tell an author to "eliminate render cascades" on the basis of a number the report did not highlight.
 - **A new story has no "before" to compare against.** The report measures this PR only; there is no baseline from `{{BASE_BRANCH}}`. So do not claim a change made something "worse", "slower", or "a regression" — you cannot know that. Describe what the numbers are, not how they moved.
 - **Do not comment on unchanged components.** Only stories this PR adds or changes are measured, and only those are in scope.
+- **Mind `measuredBecause`.** When it is `"source"` the author changed the component, not the story — they may never have opened that story file. Attribute the finding to the component ("F0Button's stories show…"), and never say or imply they edited the story. When it is `"story"` the story file itself changed and you can refer to it directly.
 - `truncated: true` means more stories were affected than were measured — mention that the report is partial.
 
 ## Writing the comment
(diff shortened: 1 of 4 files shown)


dmitry-zaets (approved):

sauldom102:
Closed because I want to change the approach

PR #5231, opened 2026-08-25
fix(one-data-collection): keep defaultExpanded across revisits and tree resets

AlbertGarcia-fact:
Fixes [FCT-62014](https://factorialmakers.atlassian.net/browse/FCT-62014).

## Symptom

A nested table with `defaultExpanded` opens as specified **the first time** the page is entered, but comes back **fully collapsed** on every later entry (list → detail → browser back). Observed on the Teams v2 list and the Job Catalog roles table (both behind dev flags).

## Root cause

Three pieces, each defensible alone:

1. `useDataCollectionUrlSync` writes `dc_sort` into the URL **with no user interaction** — a collection with `defaultSortings` (e.g. Teams' `{ field: 'name', order: 'asc' }`) stamps `?dc_sort=name-asc` on first paint.
2. On the next entry that param is re-applied via `setCurrentSortings` as a **new object carrying the same value**. `setCurrentFilters` guards with a value compare before setting; `setCurrentSortings` did not.
3. `useLoadChildren` compares sortings **by identity**, so it treated that as a real change and reset the tree (`setChildren([])` + `clearFetchedData()`). The row the policy left open never re-requested its children, because `NestedRow`'s "already asked" flag was a one-shot boolean — an open row with zero children, which renders as collapsed.

The reset path was the deeper bug: **any** genuine filter/sorting change killed the default expansion the same way, contradicting `defaultExpanded`'s own docstring.

## Fix

- **`useDataSource`** — give `setCurrentSortings` the same value-equality guard `setCurrentFilters` has, so a value-identical `dc_sort` neither resets the tree nor triggers a needless refetch.
- **`NestedProvider`** — bump a `resetGeneration` on `clearFetchedData`; `NestedRow` keys its default-children request on it, so a reset re-arms the fetch instead of leaving an opened row empty.
- **`useLoadChildren` / `NestedProvider`** — stop force-collapsing rows on reset, and **detect the reset in the provider instead of per row**. The provider is mounted once at table level and survives rows unmounting/remounting as the list re-renders around an async fetch; a per-row check re-seeds its "previous filters" ref on remount and could miss the change, leaving an opened parent showing the previous filter's cached children after the filter was cleared. Rows now drop their local children when the generation moves and re-arm their request, so a filtered view comes back expanded with a freshly fetched full child set.
- **`useDataCollectionUrlSync`** (the ticket's optional follow-up) — omit `dc_sort` from the URL while the current sorting still equals `defaultSortings`, so a collection that starts already sorted does not stamp the param on the first paint at all.

## Tests

- `defaultExpanded.test.tsx` — a sorting change after mount re-applies the policy and re-fetches children; a filter applied then cleared restores the full child set.
- `defaultExpandedFilterReset.integration.test.tsx` — full-component path (real source, real `setCurrentFilters`, async fetches, static-children observable that re-emits via `refreshChildren`, like the Teams list) restoring the full child set after a filter is cleared.
- `useDataSourceSortings.test.tsx` — a value-equal `setCurrentSortings` keeps the same object identity.
- `useDataCollectionUrlSync.test.tsx` — `dc_sort` is left out while the sorting equals the default, written when it moves away, and dropped again when it returns.

All OneDataCollection + datasource suites pass. The Teams v2 stale-children regression (an opened group showing only the previously-matching sub-team after a filter is cleared) was verified fixed against this alpha on a local slot.

[FCT-62014]: https://factorialmakers.atlassian.net/browse/FCT-62014?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


sauldom102 (changes requested):
Can you reflect this on the docs?

sauldom102 (approved):

PR #5241, opened 2026-08-25
fix(DateNavigator): keep periods out of the shared granularity key

MohamedHegab:
## Description

#5161 registered the empty `periods` definition in `granularityDefinitions`, which made `"periods"` a member of `GranularityDefinitionKey`. Every consumer of that key inherited it — including code that can never render periods. The API-diff bot flagged this and the PR argued it was a safe union widening. It isn't:

```ts
// A consumer's exhaustive map, compiling on 6.52.0, broken by 6.53.0
const PERIOD_RULE_BY_GRANULARITY: Record<GranularityDefinitionKey, Rule> = {
  year: ..., halfyear: ..., quarter: ..., month: ..., week: ..., day: ..., range: ...,
}   // ✗ Property 'periods' is missing
```

That consumer's navigator offers no consumer-defined periods; the entry it has to add is unreachable. Form-field presets and compare-to gained an option they cannot act on for the same reason.

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)

## Implementation details

- `GranularityDefinitionKey` is the seven calendar granularities again — exactly its 6.52.0 shape.
- `NavigationGranularityKey` adds `"periods"` and is used only where periods can appear: the picker popup's `granularities`, `DatePickerValue`, the calendar view helpers, the navigator trigger, and the DataCollection `date-navigator` filter.
- `DatePreset` and the compare-to `units` stay narrow, so `F0FormField`'s preset surface is unchanged — two of the flags on #5161 retire with this.
- `DatePickerCompareTo` widens, but it is `Partial<Record<…>>`, so no consumer gains a required key.

<details>
<summary>Why a lookup helper instead of leaving the record alone</summary>

Nothing holds a definition for `"periods"` in the static record any more, so the places that resolve a key they were handed (`getGranularityDefinition`, the navigator trigger's fallback, `F0DatePicker`) go through `resolveGranularityDefinition`. For `"periods"` it returns the same empty definition the record used to hold.

The three render paths that index the *dynamic* record instead — `DatePickerPopup`, `OneCalendar`, `OneDateNavigator` — are covered by `getGranularityDefinitions()`, which now always carries a `periods` entry and falls back to the empty definition when the consumer supplies none. The first cut of this PR let it return without one, which crashed all three on a `"periods"` value with no periods (caught in review). Either way a periods value with no periods renders the empty state rather than throwing.
</details>

`F0DatePicker` used to pick `granularities` straight from `DatePickerPopupProps`; it now declares its own `GranularityDefinitionKey[]`, since a form field has no `periods` prop to populate the entry with.

The three preset stories annotated their granularity `as CalendarView`, which type-checked only while both unions carried `periods`. They now say `GranularityDefinitionKey`, which is what a preset actually takes.

## Testing

`granularities/__tests__/index.test.ts` — 4 examples. Verified red against `main` (`expected [ 'day', 'week', … 8 keys ] to deeply equal [ … 7 keys ]`, and the resolver does not exist there) and green here.

The dynamic-record regression has a render assertion at each of the three sites (`OneCalendar.test.tsx`, `DatePickerPopupPeriods.test.tsx`, `OneDateNavigator.test.tsx`), each verified red without the `getGranularityDefinitions` fallback — `TypeError: Cannot read properties of undefined (reading 'calendarView' | 'calendarMode')` — and green with it. 393 examples pass across OneCalendar, DatePickerPopup, OneDateNavigator, F0DatePicker and the navigation filters; `tsc` clean on the touched files.

## API surface

The bot will flag `granularityDefinitions` losing a required `periods` key and `GranularityDefinitionKey` narrowing. Both restore the 6.52.0 shape: this reverts a widening that only existed in 6.53.0, released today. Anything written against 6.53.0's wider key keeps compiling unless it stored `"periods"` in a `GranularityDefinitionKey` variable, which only the navigator types would do — and those widened instead.

### DIFF
--- a/packages/react/src/components/OneCalendar/OneCalendar.tsx
+++ b/packages/react/src/components/OneCalendar/OneCalendar.tsx
@@ -11,10 +11,10 @@ import { Input } from "@/ui/input"
 import {
   DatePeriodsDefinition,
   GranularityDefinition,
-  GranularityDefinitionKey,
+  NavigationGranularityKey,
+  resolveGranularityDefinition,
   GranularityDefinitionSimple,
   getGranularityDefinitions,
-  granularityDefinitions,
 } from "./granularities/index"
 import {
   CalendarMode,
@@ -56,14 +56,9 @@ export type OneCalendarProps = Omit<
 >
 
 export const getGranularitySimpleDefinition = (
-  granularityKey: GranularityDefinitionKey
+  granularityKey: NavigationGranularityKey
 ): GranularityDefinitionSimple => {
-  const granularity = granularityDefinitions[granularityKey]
-  if (!granularity) {
-    throw new Error(
-      `Granularity simple definition for view ${granularityKey} not found`
-    )
-  }
+  const granularity = resolveGranularityDefinition(granularityKey)
   return {
     toRangeString: granularity.toRangeString,
     toString: granularity.toString,
@@ -71,14 +66,8 @@ export const getGranularitySimpleDefinition = (
 }
 
 export const getGranularityDefinition = (
-  granularityKey: GranularityDefinitionKey
-): GranularityDefinition => {
-  const granularity = granularityDefinitions[granularityKey]
-  if (!granularity) {
-    throw new Error(`Granularity definition ${granularityKey} not found`)
-  }
-  return granularity
-}
+  granularityKey: NavigationGranularityKey
+): GranularityDefinition => resolveGranularityDefinition(granularityKey)
 
 const OneCalendarInternal = ({
   mode = "single",
--- a/packages/react/src/components/OneCalendar/granularities/index.tsx
+++ b/packages/react/src/components/OneCalendar/granularities/index.tsx
@@ -121,12 +121,8 @@ export function getGranularityDefinitions(
           week: createWeekGranularity(effectiveWeekStartsOn),
         }
 
-  if (!periods) {
-    return definitions
-  }
-
   return {
     ...definitions,
-    periods: getPeriodsGranularity(periods),
+    periods: periods ? getPeriodsGranularity(periods) : periodsGranularity,
   }
 }
--- a/packages/react/src/components/OneCalendar/granularities/index.tsx
+++ b/packages/react/src/components/OneCalendar/granularities/index.tsx
@@ -24,10 +24,29 @@ export const granularityDefinitions = {
   halfyear: halfyearGranularity,
   year: yearGranularity,
   range: rangeGranularity,
-  periods: periodsGranularity,
 } as const satisfies Record<string, GranularityDefinition>
 
 export type GranularityDefinitionKey = keyof typeof granularityDefinitions
+
+/**
+ * The keys a date navigation can be set to. `periods` is not a member of the
+ * static record — it has no definition until a consumer supplies its ranges —
+ * so it widens only the types that can actually render it. Keeping it out of
+ * `GranularityDefinitionKey` is what stops it leaking into every exhaustive map
+ * over that key, in places (form-field presets, compare-to) where it can do
+ * nothing.
+ */
+export type NavigationGranularityKey = GranularityDefinitionKey | "periods"
+
+/**
+ * The definition behind a key with no consumer data to build it from. Only
+ * `periods` has one: its empty definition renders the "no periods" state, which
+ * is what a periods value without periods means.
+ */
+export const resolveGranularityDefinition = (
+  key: NavigationGranularityKey
+): GranularityDefinition =>
+  key === "periods" ? periodsGranularity : granularityDefinitions[key]
 
 export type GranularityDefinitionsOptions = {
   weekStartsOn?: WeekStartsOn
--- a/packages/react/src/ui/DatePickerPopup/DatePickerPopup.tsx
+++ b/packages/react/src/ui/DatePickerPopup/DatePickerPopup.tsx
@@ -2,10 +2,12 @@ import { useContext, useEffect, useMemo, useRef, useState } from "react"
 
 import { F0Button } from "@/components/F0Button"
 import { F0Select } from "@/components/F0Select"
-import { GranularityDefinitionKey, OneCalendar } from "@/components/OneCalendar"
+import { OneCalendar } from "@/components/OneCalendar"
 import {
   DatePeriodsDefinition,
   getGranularityDefinitions,
+  GranularityDefinitionKey,
+  NavigationGranularityKey,
 } from "@/components/OneCalendar/granularities"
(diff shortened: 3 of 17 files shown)


packages/react/src/components/OneCalendar/granularities/index.tsx
```diff
@@ -24,11 +24,30 @@ export const granularityDefinitions = {
...
+
+/**
+ * The definition behind a key with no consumer data to build it from. Only
+ * `periods` has one: its empty definition renders the "no periods" state, which
+ * is what a periods value without periods means.
+ */
+export const resolveGranularityDefinition = (
+  key: NavigationGranularityKey
+): GranularityDefinition =>
+  key === "periods" ? periodsGranularity : granularityDefinitions[key]
```
albertpmz, line 49: Dropping `periods` from the static record also drops it from what `getGranularityDefinitions()` returns when no periods are supplied (the early return at line 124 below). `resolveGranularityDefinition` covers the sites this PR rewrote, but not the three that index the dynamic record, and those are the render paths: `DatePickerPopup.tsx:142`, `OneCalendar.tsx:118`, `OneDateNavigator.tsx:82`. They now get `undefined` where base gave them the empty definition. `noUncheckedIndexedAccess` is off in `packages/react/tsconfig.json`, so `tsc` stays clean.

This PR widens `DatePickerValue.granularity` and `DateNavigationOptions.defaultGranularity` to `NavigationGranularityKey`, so `"periods"` without periods data is a type-legal input on all three. Verified red at `fdacb4b6d`, green at the base `4b370fdf1`:

```
DatePickerPopup value={{ granularity: "periods", value: {…} }}, no periods prop
  → TypeError: Cannot read properties of undefined (reading 'calendarMode')
OneCalendar view="periods", no periods prop
  → TypeError: Cannot read properties of undefined (reading 'calendarView')
OneDateNavigator granularities={["periods"]} value.granularity="periods", no periods prop
  → TypeError: Cannot read properties of undefined (reading 'calendarMode')
```

Dropping the early return fixes it:

```ts
return {
  ...definitions,
  periods: periods ? getPeriodsGranularity(periods) : periodsGranularity,
}
```

`GranularityDefinitionKey` stays at seven keys, and `granularityOptions` at `DatePickerPopup.tsx:150` already gates the selector entry on `periods`, so nothing changes in the UI. Applied locally: 389 pass, your 386 plus the three above.

Two notes. The description's "a periods value with no periods still renders the empty state rather than throwing" holds for the resolver call sites only. And since all 386 existing examples pass on the broken branch, the regression test has to be a render assertion at each site, not a shape assertion on the record.

packages/react/src/ui/DatePickerPopup/DatePickerPopup.tsx
```diff
@@ -35,15 +37,15 @@ export type CompareToDef = {
...
+  Record<NavigationGranularityKey, CompareToDef[]>
 >
 
 export interface DatePickerPopupProps {
   onSelect?: (value: DatePickerValue | undefined) => void
   value?: DatePickerValue
   defaultValue?: DatePickerValue
   presets?: DatePreset[]
-  granularities?: GranularityDefinitionKey[]
+  granularities?: NavigationGranularityKey[]
```
albertpmz, line 48: Crash site 1 of 3 for the record change above, at line 142 of this file: `definitions[localGranularity]`.

`localGranularity` is `NavigationGranularityKey` via `localValue?.granularity`, and `definitions` no longer holds `periods` unless the consumer passed the prop. `granularityDefinition` is `undefined`, and line 159 dereferences it on first render. Fixed by `granularities/index.tsx:124`.

packages/react/src/components/OneCalendar/OneCalendar.tsx
```diff
@@ -56,29 +56,18 @@ export type OneCalendarProps = Omit<
...
 export const getGranularityDefinition = (
-  granularityKey: GranularityDefinitionKey
-): GranularityDefinition => {
-  const granularity = granularityDefinitions[granularityKey]
-  if (!granularity) {
-    throw new Error(`Granularity definition ${granularityKey} not found`)
-  }
-  return granularity
-}
+  granularityKey: NavigationGranularityKey
```
albertpmz, line 69: Crash site 2 of 3 for the record change, at line 118 of this file: `return definitions[view]`.

`view` is `CalendarView`, which has always included `"periods"`, so `<OneCalendar view="periods" />` with no `periods` prop lands on `undefined` here and throws on `granularity.calendarView`. Fixed by `granularities/index.tsx:124`.

Crash site 3 of 3 is `OneDateNavigator.tsx:82`, which this PR doesn't touch so it can't take an inline comment: same shape, `granularityKey` comes from `localValue?.granularity`, now widened to include `"periods"`, and the record may not carry the key.

albertpmz (changes requested):

packages/react/src/components/OneCalendar/granularities/index.tsx
MohamedHegab: Confirmed and fixed in b13c39f — `getGranularityDefinitions()` now always carries `periods`, falling back to `periodsGranularity` when none are supplied, exactly as proposed. Taking your second note too: the regression tests are render assertions at each of the three sites (`OneCalendar.test.tsx` new, `DatePickerPopupPeriods.test.tsx`, `OneDateNavigator.test.tsx`), each verified red without the fallback with the same `TypeError`s you listed, plus a shape check on `getGranularityDefinitions()`. 393 pass. Description updated to drop the over-broad "still renders the empty state" claim. Also narrowed `F0DatePicker`'s `granularities` back to `GranularityDefinitionKey[]` in 72308f7, since a form field has no `periods` prop to populate the entry.

packages/react/src/ui/DatePickerPopup/DatePickerPopup.tsx
MohamedHegab: Fixed by the fallback in b13c39f; render test added in `DatePickerPopupPeriods.test.tsx` (`value.granularity: "periods"`, no `periods` prop → "No periods available").

packages/react/src/components/OneCalendar/OneCalendar.tsx
MohamedHegab: Fixed by the fallback in b13c39f; render test added in a new `OneCalendar/__tests__/OneCalendar.test.tsx` (`view="periods"`, no `periods` prop). Crash site 3 (`OneDateNavigator.tsx:82`) is covered by a case in `OneDateNavigator.test.tsx`.

albertpmz (approved):
LGTM!

PR #5247, opened 2026-08-26
fix(react): stop team and company names overflowing their column

larizzatg:
## Description

`TeamCell` and `CompanyCell` rendered the name as a plain `span` with no `min-w-0` inside a flex row that never shrank. In a fixed-width column (e.g. the goals Assignees column, `ASSIGNEE_COLUMN_WIDTH = 200`) the name spilled out of the column and got clipped mid-character.

Measuring `main`'s markup at a range of widths shows the failure is width-dependent: a flex item can shrink down to its longest word, so a 200px column already **wraps** to three lines and clips nothing — clipping only starts once the column is narrower than the name's longest word (around 80px, where it spills ~30px past the column).

So the fix keeps wrapping as the default and makes the name breakable, rather than capping it at one line:

- `min-w-0 flex-1` so the name shrinks to the column instead of overflowing it
- `break-words` so a name too long to fit even one line of its own breaks mid-word and stays inside the column, instead of spilling out and clipping
- a new optional `lines` in the value opts into a cap where a caller wants one

Capping by default was the first approach here, and it was wrong: it would have turned every currently-wrapping name into a single truncated line, in consumers we can't audit from this repo. See the commits for the flip.

## Value type

```ts
value: { name, src }            // default — wraps as far as the name needs, hiding nothing
value: { name, src, lines: 2 }  // two lines, then an ellipsis + tooltip
value: { name, src, lines: 1 }  // one line, ellipsis + tooltip (PersonCell's behavior)
```

## Screenshots

**Before** — the name overflows the 200px column with no ellipsis:

team cell before fix, overflowing column
company cell before fix, overflowing column

**After** — see the `Team Type With Long Name` / `Company Type With Long Name` stories in Chromatic for the default, plus `… With Lines` (`lines: 2`) and `… Truncated To One Line` (`lines: 1`).
CleanShot 2026-09-18 at 10 48 32
CleanShot 2026-09-18 at 10 48 53

## Verification

Measured in a browser against `main`'s own markup at the same widths:

| column width | `main` | this branch (default) |
|---|---|---|
| 200px | wraps to 3 lines, nothing clipped | wraps to 3 lines — identical |
| 80px | spills 30px past the column, clips mid-letter | stays inside the column, wraps |

## Implementation details

- fix: add `min-w-0 flex-1` and `break-words` so the name shrinks and breaks inside its column instead of overflowing it
- feat: optional `lines` in the `team`/`company` value caps the name at N lines, truncating with `OneEllipsis` (ellipsis + F0 tooltip) past them
- fix: a name that can wrap top-aligns its row and uses `tableDisplayClassNames.multiline`, so the avatar sits on the first line and that line shares a baseline with the single-line cells beside it
- test: unit tests for both cells covering the uncapped default, `lines: 1`/`lines: 2`, tooltip presence and absence, and avatar alignment
- test: Storybook stories per mode, in a fixed-width column, with source snippets
- docs: document `lines` in `team.mdx`/`company.mdx`; drop the "supports the same badge as the person type" claim, since `TeamCell`/`CompanyCell` never forward `badge` to `F0Avatar` today (pre-existing, separate from this fix)

Fixes FCT-61911

### DIFF
--- a/packages/react/src/ui/value-display/types/company/__stories__/company.mdx
+++ b/packages/react/src/ui/value-display/types/company/__stories__/company.mdx
@@ -5,8 +5,7 @@ import * as CompanyStories from "./company.stories"
 
 ### company
 
-Renders a company avatar. It supports the same badge as the person type (check
-it for more details).
+Renders a company avatar.
 
 #### Value type
 
@@ -14,9 +13,7 @@ it for more details).
 type value = {
     name: string
     src?: string // the avatar image
-    badge?: AvatarBadge
 }
-// Check PersonType for the AvatarBadge type
 ```
 
 <Canvas
@@ -33,3 +30,10 @@ render: (item) => ({
 `,
   }}
 />
+
+**Long name inside a fixed-width column**
+
+A name wider than its cell truncates with an ellipsis and shows the full name
+in a tooltip on hover, the same as the person type.
+
+<Canvas of={CompanyStories.CompanyTypeWithLongName} />
--- a/packages/react/src/ui/value-display/types/company/__stories__/company.stories.tsx
+++ b/packages/react/src/ui/value-display/types/company/__stories__/company.stories.tsx
@@ -8,8 +8,7 @@ const meta = {
     layout: "padded",
     docs: {
       description: {
-        component:
-          "Renders a company avatar with name. Supports the same badge system as the person type.",
+        component: "Renders a company avatar with name.",
       },
       source: {
         code: null,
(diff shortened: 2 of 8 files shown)


sauldom102 (changes requested):
Are we showing the native f0 tooltip when hovering or native one from the browser?

sauldom102 (approved):

larizzatg:
@sauldom102 heads up before this merges — the behaviour changed since your approval, so the approved version isn't what's on the branch now.

You approved it truncating to a single line with a tooltip, matching `PersonCell`. I've flipped the default to wrapping instead, because capping at one line would have been a regression.

The reason: on `main` the name is a flex item that can shrink down to its longest word, so it doesn't clip everywhere — it *wraps*. Measured against `main`'s markup:

| column width | `main` today | truncate-by-default | this branch now |
|---|---|---|---|
| 200px | wraps to 3 lines, fully readable | 1 line + ellipsis | wraps to 3 lines — same as `main` |
| 80px | spills 30px past the column, clips mid-letter | 1 line + ellipsis | stays inside, wraps |

So clipping only happens below ~80px, and capping at one line would have silently turned every wider column that currently shows a wrapped full name into a truncated one — in consumers we can't audit from this repo.

The default now matches `main` exactly, and `lines` opts into a cap:

```ts
value: { name, src }            // wraps, hides nothing (default)
value: { name, src, lines: 2 }  // two lines, then ellipsis + tooltip
value: { name, src, lines: 1 }  // the behaviour you approved
```

The tradeoff is that an uncapped name in a very narrow column gets tall (10 lines at 80px in my test) — that's what `lines` is for in a table where rows should stay even.

Worth a second look given it's a different behaviour than you signed off on, but happy to merge as is if you're good with it — it fixes the bug without changing anything that currently renders correctly.

PR #5268, opened 2026-08-26
feat(F0AiChatTextArea): neutral credit warning with CTA icon

daviz-fct:
## Description

The soft credit warning banner that sits above the chat composer was painted with the info palette, which made a passive nudge read like a system notice. It now uses the neutral tokens, and the "Get credits" CTA takes an optional icon so hosts can mark the action as an upsell.

## Type of change

image

## Implementation details

- feat: swap the credit warning banner from the info palette to the neutral one

  <details>
  <summary>Which tokens</summary>

  `bg-f1-background-info` → `bg-f1-background-secondary` and `text-f1-foreground-info` → `text-f1-foreground-secondary`, matching the neutral status pattern already used elsewhere (e.g. `F0Tag` status `neutral`).

  The `formBorder` override drops from `border-f1-border-info` to `border-f1-border`, which is the border the composer form already carries by default, so the banner no longer tints it. The key is kept in the config so a future `hard` level can still override it.
  </details>

- feat: add an optional `getCreditsIcon` to `AiChatCreditWarning`, rendered to the left of the "Get credits" label

  <details>
  <summary>Why a prop instead of hardcoding the icon</summary>

  The banner is host-driven: the CTA can point at billing, at a plan upgrade, or at an admin asking to raise an employee cap, so the icon belongs to the host. The prop is optional, so current consumers are unaffected and the button keeps rendering label-only when it is omitted.
  </details>

- chore: use the `Upsell` icon in the `F0AiChatTextArea` and Cocreation credit warning stories

### DIFF
--- a/packages/react/src/kits/ai/F0AiChat/types.ts
+++ b/packages/react/src/kits/ai/F0AiChat/types.ts
@@ -173,6 +173,11 @@ export type AiChatCreditWarning = {
   onDismiss?: () => void
   /** Called when the user clicks the "Get Credits" button. */
   onGetCredits?: () => void
+  /**
+   * Icon rendered to the left of the "Get Credits" label. Only used when
+   * `onGetCredits` is provided. Hosts typically pass the `Upsell` icon.
+   */
+  getCreditsIcon?: IconType
 }
 
 /**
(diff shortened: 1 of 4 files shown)


warcos-fact (approved):
LGTM

sauldom102 (approved):

sauldom102 (changes requested):
@daviz-fct we only have some checks in the CI failing, but once you make them pass we could merge this

PR #5270, opened 2026-08-26
fix(OneFilterPicker): track nested filter keys when options load async

factmarc10010:
## Problem

A nested `in` filter exposed its child filter keys only when its `options` were a literal array. Any filter that loads its options from a function (or a `source`) resolved to **no** nested keys, so the picker treated the parent as empty everywhere it is summarised outside the option list:

| Where | Symptom |
| --- | --- |
| `FilterList` | Parent gets no active dot when only a child is selected — the reported bug |
| `InFilter` | Child selections missing from the "N selected" count and the select-all indeterminate state |
| `InFilter` | Parent's clear / select-all toggle leaves the child values behind |
| `OneFilterPicker.removeFilterValue` | Removing the parent chip orphans the child values: chip gone, list still filtered |
| `getClearedFiltersValue` | Same call — masked today only when the child key is also declared as its own filter |

Repro in the product: **People → Filters → Workplace**, select only a work area under an office. The office row gets its dot (that path reads the already-loaded options), the *Workplace* entry in the filter list does not.

Screenshot 2026-08-26 at 16 57 44

## Fix

The nested keys have to be knowable **without** the options at hand, because the filter list and the chips render before the filter is ever opened — and still render that way after a reload with filters in the URL.

- New optional `nestedFilterKeys` on the in-filter options, declaring the child keys up front. Authoritative and always available, including for `source`-backed options.
- `collectNestedFilterKeys` now takes the whole schema, so when `options` is not a literal array it can also fall back to the options a previous load already resolved.
- Inside `InFilter` the keys additionally union with the currently loaded options, so they pick up children that were still unresolved when the schema was read.

Consumers with literal arrays keep working untouched; `nestedFilterKeys` is only needed for async/`source` options.

```ts
options: {
  cache: true,
  nestedFilterKeys: ["workArea"],
  options: loadWorkplaces,
}
```

Screenshot 2026-08-26 at 17 04 27

Screenshot 2026-08-26 at 17 03 09

## Testing

`packages/react/src/patterns/OneFilterPicker/__test__/nestedAsyncOptions.test.tsx` — nested filter with async options, covering the four behaviours above end to end. All four fail on `main`.

`.../InFilter/components/__tests__/option-utils.test.ts` — unit coverage for the three key sources (declared, literal, resolved-from-cache) and their merge.

The existing nested tests all use literal arrays, which is why this slipped through.

Ref: https://factorialmakers.atlassian.net/browse/FCT-61539

### DIFF
--- a/packages/react/src/patterns/OneFilterPicker/filterTypes/InFilter/components/option-utils.ts
+++ b/packages/react/src/patterns/OneFilterPicker/filterTypes/InFilter/components/option-utils.ts
@@ -1,4 +1,6 @@
+import { FilterTypeSchema } from "../../types"
 import { InFilterOptionItem, InFilterOptions } from "../types"
+import { getCacheKey, getCachedOptions } from "../useLoadOptions"
 
 /**
  * Recursively checks whether an option or any of its nested children
@@ -36,12 +38,10 @@ export function hasSelectedDescendant<T>(
 }
 
 /**
- * Collects all nested child filter keys from an InFilter's options.
- * Used to determine if a parent filter should show an active indicator
- * when any of its nested children have selections.
+ * Collects the nested child filter keys reachable from an option tree.
  */
-export function collectNestedFilterKeys<T>(
-  filterOptions: InFilterOptions<T>
+export function collectNestedFilterKeysFromOptions<T>(
+  options: InFilterOptionItem<T>[] | undefined
 ): string[] {
   const keys = new Set<string>()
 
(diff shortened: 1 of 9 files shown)


packages/react/src/patterns/OneFilterPicker/filterTypes/InFilter/components/option-utils.ts
```diff
@@ -54,8 +52,30 @@ export function collectNestedFilterKeys<T>(
...
+export function collectNestedFilterKeys<T>(
+  schema: FilterTypeSchema<InFilterOptions<T>>
+): string[] {
+  const filterOptions = schema.options
+  const keys = new Set<string>(filterOptions.nestedFilterKeys ?? [])
+
+  const resolvedOptions =
+    "options" in filterOptions && Array.isArray(filterOptions.options)
+      ? filterOptions.options
+      : getCachedOptions<T>(getCacheKey(schema))
```
albertpmz, line 75: This fallback doesn't fire where it's needed. With async options and no `nestedFilterKeys`, the dot stays off after the options load and only appears on a later mount: `nestedKeysMap` in `FilterList.tsx` is a `useMemo` keyed on `definition`, so the cache filling later never invalidates it. Same filter, two behaviors, depending on history the user can't see. It also makes render impure, since this reads a mutable module-level `Map` during render.

Suggest dropping the fallback and letting `nestedFilterKeys` be the single answer for non-literal options. It's deterministic and already carries the fix on its own: all four integration tests pass on the declared keys alone.

Heads up that `option-utils.test.ts:76` can't catch this. It calls `loadOptions(getCacheKey(schema), ...)` with the same object and asserts the read back, so it covers the Map round-trip, not the component path.

albertpmz (commented):
Nice work, let's check this issue below

PR #5271, opened 2026-08-27
ci(release): publish to npm via trusted publishing

eliseo-juan:
## Description

The `NPMJS_TOKEN` used to publish `@factorialco/f0-react`, `@factorialco/f0-core` and `@factorialco/f0-react-native` has to be rotated every 90 days, which is recurring manual work and a standing leak risk. This PR replaces it with [npm Trusted Publishers](https://docs.npmjs.com/trusted-publishers), which authenticates the publish job through GitHub's OIDC identity instead of a long-lived credential. The Trusted Publisher is already configured on npmjs.com for all three packages, pointing at `npm-publish.yaml`. Alpha builds are unaffected — they publish as git branches (`npm/alpha-pr-N`), never to a registry.

## Implementation details

- ci: authenticate `npm-publish.yaml` with OIDC instead of `NPMJS_TOKEN`

  <details>
  <summary>What changed in the job</summary>

  Each of the three jobs now grants `id-token: write` and drops the unused `packages: write` (nothing is published to GitHub Packages). The `registry_token` input is gone from the composite action.

  Two details are load-bearing and easy to miss:

  - `pnpm publish` shells out to `npm publish`, and OIDC needs npm >= 11.5.1. `ubuntu-latest` still ships npm 10.x with Node 22, so the action upgrades npm first.
  - `actions/setup-node` writes `_authToken=${NODE_AUTH_TOKEN}` into the npmrc. With no token set, npm sends that placeholder verbatim and OIDC fails with a misleading `E404`. The action strips the line before publishing.
  </details>

- chore: bump pnpm from 9.15.4 to 10.34.5

  <details>
  <summary>Why 10 and not 11</summary>

  OIDC support starts in the pnpm 10 line, so 9.15.4 cannot do trusted publishing at all.

  pnpm 11 is deliberately skipped: it silently stops reading the `pnpm` field from `package.json`, and this repo pins `react@18.3.1` and `@types/react` through `pnpm.overrides` there. The install would succeed with the wrong versions and no warning. It also enables `minimumReleaseAge` and `blockExoticSubdeps` by default. Moving to 11 means migrating that config to `pnpm-workspace.yaml` and is worth its own PR.

  `pnpm-lock.yaml` is unchanged — pnpm 10 reads the existing 9.0 lockfile as-is.
  </details>

- chore: allow build scripts blocked by pnpm 10's new default

  <details>
  <summary>Which ones</summary>

  pnpm 10 no longer runs dependency lifecycle scripts by default. `pnpm.onlyBuiltDependencies` re-enables the ones the builds actually need: `@swc/core`, `esbuild`, `lefthook`, `msw`, `unrs-resolver`. `canvas` is intentionally left out — it is an optional native dep of jsdom that nothing here uses.
  </details>

- chore: add `repository` to the three published packages

  Required for provenance, which npm validates against the building repo. `--provenance` is now passed on publish, so releases get the verified badge on npmjs.

### Verified locally with pnpm 10.34.5

- `pnpm install` clean, `pnpm-lock.yaml` untouched
- `f0-core` and `f0-react` build, including api-extractor and the `.d.ts` output
- `f0-react-native`: `tsc --noEmit` passes
- `pnpm pack` of `f0-react` resolves `workspace:*` to `"@factorialco/f0-core": "2.1.0"`, and `repository` reaches the published manifest
- `pnpm --filter ... publish --dry-run` still prints `There are no new packages that should be published`, so the `published` output the Slack steps depend on keeps working

### Follow-up after the first green release

Delete the `NPMJS_TOKEN` secret, revoke the token on npmjs.com, and restrict each package to trusted publishing under Settings → Publishing access.

### DIFF
--- a/.github/actions/publish-to-npmjs/action.yaml
+++ b/.github/actions/publish-to-npmjs/action.yaml
@@ -1,11 +1,13 @@
 name: "Publish to npmjs"
 
+# Authentication is done with npm Trusted Publishers (OIDC), not with a token.
+# The calling job MUST grant `id-token: write`, and the package must have a
+# Trusted Publisher configured on npmjs.com pointing at `npm-publish.yaml`.
+# See https://docs.npmjs.com/trusted-publishers
+
 inputs:
   workspace:
     description: "Workspace"
-    required: true
-  registry_token:
-    description: "Registry token"
     required: true
 
 outputs:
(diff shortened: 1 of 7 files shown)


developerdanx (approved):
Finally ❤️  Thank you

PR #5279, opened 2026-08-27
fix(lint): lint the packages, and enforce correctness ruleset

dmitry-zaets:
## Description

**The `Linting` checks on every react and react-native PR since 2026-01-12 have been passing without reading a single file.** oxlint ran, found nothing to lint, and exited 0. This is what CI printed on every PR for roughly seven and a half months:

```
> sh -c 'oxlint "${@:-src/**/*}" --max-warnings 0' --

Found 0 warnings and 0 errors.
Finished in 66ms on 0 files with 0 rules using 4 threads.
```

`on 0 files`. Both jobs feed the required `Quality` gate, so a green tick there carried no information about lint. Nothing has been enforced in CI across 3375 react files and 536 react-native files.

Two things limit how bad this was, and both are worth stating plainly. Lefthook's pre-commit hook passes staged paths as arguments, so the same script did lint changed files locally — developers who ran it were covered for what they touched. And the rules it enforced were a small subset either way, because `correctness` was separately switched off. What was lost is CI enforcement: nothing caught what a hook was skipped for, what landed from a machine without hooks, or what an existing file drifted into.

This PR restores the gate, fixes what it surfaces, turns `correctness` back on, and adds a regression test so the same failure cannot return silently.

## Type of change

- [x] Bug fix
- [x] Other: CI / tooling

## How it broke

`cfc7092ba9` (2026-01-12) wrapped the plain `oxlint src/**/*` in `sh -c '… "${@:-src/**/*}"' --` so lefthook could pass staged paths. The wrapper is correct and still needed. The quoting around the default is not: it stops the shell expanding the glob, so oxlint receives the literal string `src/**/*`, resolves it as a path, and matches nothing.

Verified under `dash`, which is `/bin/sh` on the CI runner:

| default | files linted |
| --- | --- |
| `"${@:-src/**/*}"` | **0** |
| `"${@:-src/}"` | **3368** |

The CI lint job itself is older and did work before this. It has existed since `ad840b0f25` (2024-08-02), and the script covered the package through the eslint era and the first four days of oxlint.

Confirmed fixed on this PR's own CI run: react **3375 files**, react-native **536 files**.

## Implementation details

- fix: default the `lint` and `lint:fix` scripts to `src/` in both packages

  <details>
  <summary>Why <code>src/</code></summary>

  It matches the `format` and `format:check` scripts a few lines below in the same files, which have used `"${@:-src/}"` all along and always worked. Passing explicit paths still works, so lefthook is unaffected.
  </details>

- test: add a regression test asserting the `${@:-…}` default resolves to a real path

  <details>
  <summary>Red-green</summary>

  A quoted glob lints nothing silently — there is no error to notice, only a suspiciously small file count in a log nobody reads. So the guard checks the property that actually matters: the default must be a path oxlint can resolve. Verified red on `main` (where it is `src/**/*`) and green here.
  </details>

- fix: set `"correctness": "error"` in both `.oxlintrc.json` files

  <details>
  <summary>Why it was off, and what it costs</summary>

  `9bb4e5a83f` disabled the category and re-enabled all 93 correctness rules by name — the normal shape of an eslint→oxlint migration. The next day `5d95bb5afc` deleted those 93 lines and kept the `categories` line three lines above them, so the opt-out outlived the enumeration that made it harmless. Three later commits touched the file and left it at 0.

  react now runs **97 of 98** correctness rules over 3375 files; react-native runs all **83** over 536 and was already clean.
  </details>

- fix: move `key` onto the array elements in `OneDataCollection/Settings` (`react/jsx-key`)

  <details>
  <summary>A real bug, not lint noise</summary>

  The array passed to `PopoverContent` has four entries. Two put `key` on the array element; the other two put it on the inner child, where React never sees it. Those two remount whenever the array's composition changes — which is exactly when grouping or sortings toggle.
  </details>

- fix: pass JSX children instead of a `children` prop in `F0Toast` and one story (`react/no-children-prop`)
- fix: replace six unsafe optional chains in tests with non-null assertions (`no-unsafe-optional-chaining`)

  <details>
  <summary>Scope</summary>

  Each dereferences an optional chain on the line after an assertion proving it is defined. The four sibling expressions in `widget-utils.test.tsx` that only assign the value are untouched — the rule correctly leaves them alone.
  </details>

- chore: turn `react-hooks/exhaustive-deps` off in react, explicitly rather than by category

  <details>
  <summary>Deliberately deferred</summary>

  It reports 75 violations. Each fix changes what an effect or memo re-runs on, so they need case-by-case review and their own PR. Naming the rule keeps the other 97 enforced and makes the debt visible instead of hiding it behind a disabled category. It stays **on** in react-native, which has none.
  </details>

- chore: suppress `react/no-this-in-sfc` at each of the 8 sites, with a disable comment

  <details>
  <summary>Why per-site and not per file type</summary>

  All 8 reports are the same echarts `LinearGradient` constructor mock. The rule flags it because the mocked class is capitalized and the rule reads any capitalized function using `this` as a component — a false positive. Turning the rule off for all test and story files would also give it up where it still applies: a component in a test file that really does use `this` would go unreported. A disable comment at each mock keeps the reason visible and the rule enforced everywhere else.
  </details>

- fix: point the react-native pre-commit hook at its `format` script

  <details>
  <summary>Why this is in the same PR</summary>

  `lefthook.yml` ran `pnpm --filter @factorialco/f0-react-native prettier --write`, but `dcf3854a3d` (2026-02-10) moved that package to oxfmt and removed the `prettier` script. Since then, staging any react-native file failed pre-commit with `ERR_PNPM_RECURSIVE_RUN_NO_SCRIPT`. It blocked committing this PR, so it is fixed here rather than separately.
  </details>

- fix: exempt `**/*.spec.*` from `no-console` in `packages/react-native/.oxlintrc.json`

  <details>
  <summary>Why the config, not the test</summary>

  The override exempts `**/*.test.*`, but react-native names every test file `.spec.*` — 76 of them, and zero `.test.*`. That override was copied from the react package, where `.test.*` is correct, so it had been matching nothing.
  </details>

- fix: add `displayName` to three test components in `withDataTestId.test.tsx` (`react/display-name`)
- fix: drop two unused type imports in `F0Avatar` (`IconType`, `F0AvatarModuleSize`)

## Verification

| | react | react-native |
| --- | --- | --- |
| `lint`, no args | 3375 files, 97 rules, exit 0 | 536 files, 83 rules, exit 0 |
| `lint <path>` | 8 files | 29 files |
| typecheck | 0 errors | exit 0 |
| `format:check` | clean (3575) | clean (558) |
| tests | 559 files / 7077 tests pass | 42 suites / 487 tests / 318 snapshots pass |
| red-green gate | verified locally and on CI | n/a |

## Known, pre-existing, not from this PR

The react unit suite can exit 1 on unhandled `ReferenceError: document is not defined` errors from prosemirror timers firing after jsdom teardown, in `RichText/.../EnhanceHighlight`. It reproduces on clean `main` locally, the count varies run to run, and the file passes in isolation — a teardown race. It has not fired on the CI runner. Worth its own ticket.

## Follow-up

`react-hooks/exhaustive-deps` in react: 75 violations, deferred above.


packages/react/.oxlintrc.json
```diff
@@ -2,7 +2,10 @@
   "$schema": "./node_modules/oxlint/configuration_schema.json",
   "plugins": ["react", "typescript", "vitest"],
   "categories": {
-    "correctness": "off"
+    "correctness": "error"
+  },
+  "rules": {
+    "react-hooks/exhaustive-deps": "off"
```
dmitry-zaets, line 8: This one needs to be fixed case by case, couldn't fix them automatically without the knowledge about the components

sauldom102 (approved):
Thank you Dmitry! 🙌🏼

PR #5312, opened 2026-08-30
feat(react)!: publish tree-shakable ESM and slim packaged CSS

somebody32:
## What

F0 was publishing two runtime graphs: a bundled legacy graph and a preserved ESM graph. This PR makes preserved ESM the only runtime and deliberately removes backward compatibility for legacy deep imports. Factorial will migrate in one coordinated integration.

- removes legacy `dist/*.js` runtime entries and unrestricted `./dist/*` exports
- publishes the root and supported component subpaths from `dist/esm` only
- keeps React and React DOM 18.3.1 and leaves TipTap runtime versions unchanged
- scopes Inter publication to the F0-owned stylesheet and emits 16 Unicode-ranged variable WOFF2 subsets
- removes the broad semantic-color safelist and prevents source-font/config duplication in the tarball
- validates the real packed artifact, every runtime export, declarations, ESM closure, CSS assets, and browser consumption
- compares API declarations with one compiler/config/dependency closure, varies only React source, includes every public declaration entry, and fails closed on missing public type edges
- freezes the 129 inherited cyclic runtime edges and rejects additions to the baseline, while allowing debt to shrink
- removes the retired API Extractor dependency/config and unused legacy Vite modes

## Breaking migration

Consumers must replace legacy imports with public package subpaths:

- `@factorialco/f0-react/dist/experimental` to `@factorialco/f0-react/experimental`
- `@factorialco/f0-react/dist/i18n-provider-defaults` to `@factorialco/f0-react/i18n-provider-defaults`
- `@factorialco/f0-react/dist/styles.css` to `@factorialco/f0-react/styles.css`
- any other `@factorialco/f0-react/dist/*` import must move to a supported public export or be added intentionally

There is no CommonJS fallback. The package is ESM-only.

`@factorialco/f0-core/tailwind` no longer safelists broad `bg-f1-*`, `text-f1-*`, or `border-f1-*` families. Consumers that construct class names dynamically must safelist the exact classes they generate.

## Packed-artifact evidence

All measurements use the generated npm tarball outside workspace aliases.

| Probe | Minified raw JS | Retained inputs |
| --- | ---: | ---: |
| `F0Button` component subpath | 581.7 KiB | 534 |
| package root importing `F0Button` | 596.8 KiB | 571 |
| root overhead | 15.1 KiB | 37 |

The 15.1 KiB root/subpath delta is below the enforced 20 KiB ceiling. The absolute number includes React and shared consumer runtime; it is not presented as the size of the button itself.

Published global CSS is 331,963 B raw and 33,133 B Brotli. Published fonts total 796,296 B across 16 range/style subsets; browsers fetch only subsets matching page text and style. No top-level legacy runtime, source-font duplicate, or internal build config is present in the tarball.

A Cloudflare Worker rehearsal with React 18.3.1 and Wrangler 4.127.1 previously passed `wrangler deploy --dry-run --minify` at 672.4 KiB raw / 220.7 KiB gzip, including React runtime.

## Honest remaining debt

- six pre-existing runtime-cycle groups remain; 129 exact cyclic edges are now frozen
- individual components still retain shared Tooltip, Motion, i18n, emoji, avatar, and flag capabilities
- PDF, XLSX, maps, canvas, AI chat, and other optional graphs remain component-by-component work
- root declarations still inherit third-party TipTap, LiveKit, and ECharts conflicts; explicit component subpaths compile strictly

The measured 54-file component untangling experiment saved 57,337 Brotli bytes but was removed from this PR. That work should proceed later component by component, prioritized by production usage.

## Validation

- merged and validated against `main` at `0c69a7c39` (`@factorialco/f0-react` 6.82.1)
- Node 24.20.0 via Mise/Corepack; frozen pnpm 10.34.5 install passed
- exact core and React production builds passed
- packed publication and consumer checks passed; root/subpath delta remains 15.1 KiB
- TypeScript 6 typecheck, lint, formatting, actionlint, and monotonic cycle ratchet passed
- full React unit suite passed: 602 files passed, 1 skipped; 7,562 tests passed, 3 expected failures, 13 skipped
- merged-state focused packaging/API/font/cycle/Home tests passed: 174 tests plus 3 intentional red-green fixtures
- all five public API declaration entries analyzed successfully
- repository pre-commit and pre-push gates passed
- final code, accessibility, Storybook/docs, and behavioral-coverage reviews found zero blockers or suggestions

No screenshots: there is no intended visual behavior change.


packages/core/tailwind.ts
```diff
@@ -111,18 +111,4 @@ export const baseConfig: Omit<Config, "content"> = {
     require("tailwindcss-animate"),
     require("@tailwindcss/container-queries"),
   ],
-  safelist: [
```
somebody32, line 114: that part was extremely bad: 20k classes only for 70 used

.github/workflows/api-surface.yaml
```diff
@@ -10,22 +10,18 @@ jobs:
...
+  # Emit the public declaration graph for both sides in parallel. Both sides
+  # use the compiler, config, and dependency graph from the merge commit; only
+  # React source varies. This keeps the check focused on React source changes
+  # instead of reporting dependency-version drift as an API change.
   build-dts:
     needs: detect-changes
-    # React source or core (whose types ripple into the React surface) changed.
     if: |
-      (needs.detect-changes.outputs.package_react == 'true' ||
-       needs.detect-changes.outputs.package_core == 'true') &&
```
dmitry-zaets, line 21: Do we skip the core package here on purpose?

packages/changelog-summarizer/package.json
```diff
@@ -17,8 +17,8 @@
     "commander": "^14.0.3"
   },
   "devDependencies": {
-    "@types/node": "^22.13.9",
+    "@types/node": "^24.13.3",
     "tsx": "^4.19.2",
-    "typescript": "^5.7.3"
+    "typescript": "npm:@typescript/typescript6@6.0.2"
```
dmitry-zaets, line 22: Is TS7 breaking anything here?

packages/core/tailwind.ts
dmitry-zaets: some of the factorial web uses tailwind classes in the code, even thought tailwind is not properly wired in the monorepo. We need to doublecheck that this will not break the monorepo code (or even better - to fix the monorepo to not to use non-safelisted tailwind

packages/react/package.json
```diff
@@ -128,6 +240,7 @@
     "@tiptap/suggestion": "^2.12.0",
     "@vueless/storybook-dark-mode": "^10.0.7",
     "aria-hidden": "^1.2.6",
+    "axe-core": "^4.11.1",
```
albertpmz, line 243: Should this be a runtime dep? It's only used in a Storybook a11y feature. Maybe we can declare it as optional in `peerDependencies`?

packages/react/package.json
albertpmz, line 181: We can remove this one, it's not imported by anything else/required by any peer

albertpmz, line 214: Same as above, not imported by anything else/required by any peer, removable

packages/react/package.json
```diff
@@ -128,6 +240,7 @@
     "@tiptap/suggestion": "^2.12.0",
     "@vueless/storybook-dark-mode": "^10.0.7",
```
albertpmz, line 241: We can move it to `devDependencies`, only Storybook imports it

packages/react/package.json
albertpmz, line 249: Same as some of the above, not imported by anything else/required by any peer, removable

albertpmz (commented):
Found some libs that we can cleanup and/or move 😊

PR #5378, opened 2026-09-02
perf(OneDataCollection): stop a page append re-rendering every row already on screen

factmarc10010:
## Why

A collection on infinite scroll gets heavier the further you scroll. The row count is not what makes it heavy: loading a page re-renders every row already on screen, and re-does bookkeeping once per row on top, so each page costs more than the last.

Nothing here changes what the collection looks like or how it behaves. Whether anything further is needed is [FCT-62647](https://factorialmakers.atlassian.net/browse/FCT-62647), which starts by measuring what is left.

## What was making a page append cost more than the page

**The record of every row was rebuilt on each fetch.** `useData` decorated every record with a group id through a spread — including when the collection is not grouped, which is when nothing ever reads it. `mergeItems` keeps the object of a record that is already loaded, so the only thing giving already-rendered rows a new identity was that decoration. Now ungrouped records pass through untouched, and a grouped collection reuses the object it built for a record the last time it saw it.

**Rows had no memo boundary, and three props that would have defeated one.** The animation variants were rebuilt per row per render, `onCheckedChange` was an inline arrow per row, and the whole selection `Map` was handed to every row so any selection change altered every row's props. Rows now take their own `isSelected` boolean, and the memo sits outside `motion.create` so a skipped row also skips framer's layout measurement.

**The selection registry was rebuilt once per rendered row.** Registration was keyed on the row's record, so a page append took every already-rendered row out of the registry and put it straight back in — and each of those membership changes re-materialized the entire id list. Rows now register on their id, refresh their entry separately, and the registry counts membership changes instead of rebuilding the list inside each one.

## Behaviour

Unchanged. The set of registered rows is still the selectable rows on screen; the group id is still attached wherever anything reads it; rows still flash, animate and select as before.

## Testing

`InfiniteScrollRowCommits.integration.test.tsx` drives the real source with an infinite-scroll adapter and fires the sentinel's `IntersectionObserver` to load each page, counting renders per row.

Red on `main` (source reverted, test kept):

```
× does not re-render them when the next page is appended
  AssertionError: expected 3 to be 2

× keeps the cost of a page flat as pages accumulate
  AssertionError: expected 18 to be less than or equal to 12
```

The second number is the shape of the problem: the third page cost 18 row renders where the second cost 12, with nothing added but rows already on screen.

Green here, and the full unit suite passes (7552 tests).

`SourceStaleness.integration.test.tsx` guards the other side of a memo boundary: `Row` calls six of the source's callbacks during its own render, so a consumer flipping state behind one of them — `selectionDisabled`, with hoisted columns, nothing else to compare on — has to reach the rendered checkbox. A row that skips its render never reads a new value, so any future attempt to make the source's identity stable has to answer this test.

Also validated by hand against the product: the alpha installed in a local slot, with the employees list switched from `pages` to `infinite-scroll` so rows accumulate.

## Not in this PR

- Windowing the rows. Considered and dropped: the collection renders a real `<table>`, so rows outside the DOM stop being reachable by the browser's find-in-page, and it would put the accessibility skip ratchet under pressure. FCT-62647 measures the linear remainder first; capping accumulation is the cheaper answer if it turns out to matter.
- The `layout` prop on every row: it still puts a framer projection node on each one. Scoping it changes reorder animations, so it wants design input.
- **A consumer render still re-renders every row.** `useDataCollectionSource` returns a new object each render — it carries the live filter, search and sorting state — and rows take it as a prop. This PR only fixes the append path, where the consumer never renders at all. #5411, stacked on this one, splits the definition out and hands rows that instead.
- Deriving the selection registry from the loaded data instead of from the rendered rows. Its docstring says it exists to reach loaded rows absent from `data.records`, so "loaded" was always the intent — but today the two sets coincide, so this is a design cleanup rather than a bug.

[FCT-62647]: https://factorialmakers.atlassian.net/browse/FCT-62647?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/react/src/patterns/OneDataCollection/visualizations/collection/Table/Table.tsx
+++ b/packages/react/src/patterns/OneDataCollection/visualizations/collection/Table/Table.tsx
@@ -344,6 +344,30 @@ export const TableCollection = <
     getRenderedSelectableEntries: selectionRegistry.getEntries,
     renderedSelectableCount: selectionRegistry.ids.length,
   })
+
+  // `handleSelectItemChange` is rebuilt whenever the consumer's `selectable`
+  // changes identity, which for an inline definition is every render. Rows get a
+  // stable wrapper instead, so the latest handler still runs.
+  const latestSelectItemChangeRef = useRef(handleSelectItemChange)
+  latestSelectItemChangeRef.current = handleSelectItemChange
+  const stableSelectItemChange = useMemo(
+    () => (item: R, checked: boolean) =>
+      latestSelectItemChangeRef.current(item, checked),
+    []
+  )
+
+  // Selection is handed to each row as its own boolean. Passing the whole Map made
+  // every row's props differ on any selection change (measured: 25 mismatches for
+  // 25 rows), so one checkbox re-rendered the entire table.
+  const isItemSelected = (record: R): boolean => {
+    const id = stableSource.selectable?.(record)
+    return id !== undefined && selectedItems.has(id)
+  }
+
+  // Only rows that render nested children need the Map, to derive their own
+  // children's state. Flat rows get `undefined`, which is stable.
+  const nestedSelectedItems = (record: R): typeof selectedItems | undefined =>
+    stableSource.itemsWithChildren?.(record) ? selectedItems : undefined
   const summaryData = useMemo(() => {
     // Early return if no summaries configuration or summaries data is available
 
@@ -886,11 +910,9 @@ export const TableCollection = <
                                 item={item}
                                 index={index}
                                 groupIndex={groupIndex}
-                                onItemCheckedChange={handleSelectItemChange}
-                                onCheckedChange={(checked) =>
-                                  handleSelectItemChange(item, checked)
-                                }
-                                selectedItems={selectedItems}
+                                onItemCheckedChange={stableSelectItemChange}
+                                isSelected={isItemSelected(item)}
+                                selectedItems={nestedSelectedItems(item)}
                                 columns={columns}
                                 frozenColumnsLeft={frozenColumnsLeft}
                                 checkColumnWidth={checkColumnWidth}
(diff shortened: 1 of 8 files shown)


packages/react/src/patterns/OneDataCollection/visualizations/collection/Table/Table.tsx
```diff
@@ -236,6 +248,45 @@ export const TableCollection = <
...
+        getOwnPropertyDescriptor: (_target, prop) => {
+          const descriptor = Reflect.getOwnPropertyDescriptor(
+            latestSourceRef.current,
+            prop
+          )
+          return descriptor && { ...descriptor, configurable: true }
+        },
+      }),
+    []
+  )
```
albertpmz: I don't think this Proxy earns its place.

A page append never renders the consumer. The records live in `useData` inside this component, so `effectiveSource` keeps its identity across appends. I replaced both `source={stableSource}` with `source={effectiveSource}` and both tests in `InfiniteScrollRowCommits.integration.test.tsx` still pass.

The Proxy only matters when the consumer re-renders, and then it goes stale. `Row.tsx` reads `source.selectionDisabled`, `source.itemUrl`, `source.itemsWithChildren` and `source.itemActions` during render. A consumer with hoisted `columns` and `selectionDisabled: (item) => !canEdit` flips `canEdit`, this component renders, and every row is skipped because no prop changed identity. The checkbox keeps the old disabled state. Before this PR that row re-rendered.

I'd drop the Proxy and pass `effectiveSource`. If you want to keep it, add a test that flips consumer state behind `selectionDisabled` and asserts the row updates. It fails today.

`stableSelectItemChange` below is fine. It only wraps an event handler, nothing is read at render.

albertpmz (changes requested):
Let's check this issue below 👀

packages/react/src/patterns/OneDataCollection/visualizations/collection/Table/Table.tsx
factmarc10010: okay, TY,  addressed on the following commit

albertpmz (approved):
LGTM!

PR #5381, opened 2026-09-03
feat(OneDataCollection): add a map visualization

AlbertGarcia-fact:
> [!NOTE]
> **Stacked on [#5463](https://github.com/factorialco/f0/pull/5463)** (`feat(icons): add PinOff app icon`), which this branch's "Not on map" work uses. Review and merge that one first; this PR targets `feat/pin-off-icon`, so its diff shows only the map work. Once #5463 lands, this retargets to `main` on its own.
>
> Also rebased onto current `main`, which brought in `refactor(lint): enable oxlint's unicorn, oxc and promise rules` (#5448). The last commit, `style: conform the map work to main's oxlint rules`, is the mechanical fallout: braces on single-statement branches and `cond && <x/>` becoming `cond ? <x/> : null`.

## What

Adds `map` as a built-in Data Collection visualization, so a collection can offer a map beside its table under the same toolbar — filters, search and the view switcher drive both, exactly as they do for the graph view.

Today a product wanting a map has to reach for `type: "custom"` and hand-build the whole view: its own query, its own filter plumbing, its own bridge for search selections, and its own camera work. That is what Workplaces did in the monorepo, and it is the code this replaces.

```tsx
{
  type: 'map',
  options: {
    coordinates: (office) => [office.longitude, office.latitude],
    label: (office) => office.name,
    selectedRecordId: selectedOffice?.id ?? null,
    onSelect: setSelectedOffice,
    revealRecordId: searchPickedOfficeId,
    viewportInset: panelWidth > 0 ? { right: panelWidth } : undefined,
  },
}
```

## How the camera behaves

This is most of the value, and most of the work. A map that does not follow its collection is a map the user has to re-aim by hand after every interaction.

| The user… | The camera… |
|---|---|
| opens the page | frames every marker |
| applies a filter | reframes to the markers left — with several filters active, the set matching all of them |
| clicks a marker | centres it **at the current zoom**, in the area left free by `viewportInset` — "show me this in context" |
| picks a search result | **flies** to that marker and zooms in — "take me there" |
| opens / resizes / closes the panel | redoes the last camera move with the new free area, so the selection stays clear of the panel |
| dismisses a panel it had flown to | zooms back out to the overview |
| dismisses a panel it had only clicked | stays put — a click never changed the zoom, so there is nothing to undo |
| switches to another view | drops the selection, so a panel never lingers over the table |

### One idea holds most of that together

Camera padding is part of MapLibre's transform, so "the panel covers this strip" and "fly here" are the same kind of instruction — and a naive `easeTo({ padding })` **targets whatever the zoom happens to be when it runs**. Because the panel opens a beat after the selection, that padding change kept landing mid-animation and freezing the camera there: it cut the search's zoom-in short, and later it froze the zoom-out part-way through.

So the map remembers what the camera was last told to do, and a padding change *redoes* it rather than re-applying padding on top of it:

```ts
type CameraIntent =
  | { kind: "fit" }                    // frame every marker
  | { kind: "flight"; zoom: number }   // fly to a point at a known zoom
  | null                               // a click: only re-centre
```

## Design

- **Reading a record's position is the only required option.** Everything about how a marker looks and behaves stays `F0Map`'s concern, so maps across the product stay uniform.
- **Records without coordinates draw no marker** rather than being pinned at `[0, 0]`. The map does no geocoding.
- **Selecting a marker hands the whole record back** through `onSelect`, so a consumer can open a side panel. `F0Map` itself only ever deals in ids.
- **The selection is controlled** via `selectedRecordId`. Without it the map keeps its own, and nothing outside could end a selection its panel started — the marker stayed marked after the panel was dismissed.
- **A reveal is a one-shot event, not state** — paired with the collection's existing `searchSelectionNonce`, so re-picking the same record flies again, and it can never be replayed when the view remounts.
- **One page, sized for markers.** A map is a single canvas, not a paginated surface, so it requests one page capped at `RECOMMENDED_MAX_MARKERS` instead of a table's page size.

## Two new `F0Map` props

Both additive and off by default, so no existing consumer changes.

- **`viewportInset`** — which region of the map external chrome covers, in screen px. Every camera path (fit, focus, reveal, locate) offsets its target by it. Mirrors `F0Graph`'s prop of the same name.
- **`centerOnMarkerClick`** — re-centre a clicked marker at the current zoom inside that free region. Deliberately distinct from the `focusMarker` handle, which flies and zooms.

## Loaded on demand

The visualization is behind `React.lazy`. It pulls in the MapLibre GL engine, and *every* collection imports the visualization registry — a static import would have put the map engine in the bundle of consumers that only ever render a table. That was not hypothetical: the static version broke 25 unrelated test files by loading MapLibre wherever the registry was imported.

## Tests

**54 tests across the two units** (23 `MapCollection`, 31 `F0Map`), and the full `OneDataCollection` + `F0Map` suites are green at **791**.

Four of the camera rules were written after the behaviour was reported broken, so each one has a test that was **verified to fail without its fix** — not just to pass with it:

| Fix | The test, run against the unfixed code |
|---|---|
| frame the markers a filter leaves, not the ones it removed | `expected "vi.fn()" to not be called at all, but actually been called 1 times` |
| carry a flight through to its zoom when a panel opens mid-way | `expected undefined to be 15` |
| redo the last camera move when the covered region changes | `expected 1 to be 2` |

The first needed a source whose `fetchData` resolves by hand, because the bug only exists in the window where the filter has changed and the previous markers are still drawn — synchronous mocks passed against the broken code.

The `F0Map` test's MapLibre stub now attaches marker elements the way MapLibre does, so a pin click is reachable from a test. Pins are `aria-hidden` — keyboard users get `F0MapList` — so they never collide with the accessible list's buttons.

## Stories

`Data Collection/Visualizations/Map` — the map/table switch with filters, and a side-panel story for `viewportInset`, click-centring and the zoom-out on close.

## Follow-up

The Workplaces map in the monorepo migrates to this in factorialco/factorial#112511 (the view) and factorialco/factorial#112512 (its side panel), both verified against this branch's alpha. The consumer side is 5 files and +63/−7: no map component, no second query, no context bridge, and none of the camera behaviour above needed an extra option.


sauldom102 (changes requested):
Hi Albert! Loving the fact that we're connecting the map visualization to the data collection, it's so cool, what's your exact use case? (just because I'm interested)

Can you try the alpha version in frontend to double-check we're not introducing any type regression?

What exactly can be placed in the sidebar? Is it for showing attributes of each item? I would like to see in the story a case of maximum with many different attribute types

How this map visualization would work when we have a sidebar and it's placed inside a narrow layout? Imagine the map has 1/4 of the user viewport width, can we see the attributes properly in that case? Could we reuse the F0Drawer here?

AngelMsaa (approved):

PR #5383, opened 2026-09-03
fix(OneEmptyState): stack actions in a narrow container

sauldom102:
## Description

Adding an empty-state widget to `NewHomeLayout`'s rail story turned up two overflow bugs, both of which this PR fixes at the source: an empty state's actions laid themselves out from the *viewport's* width rather than their own container's, so two full-sentence buttons in Home's 396px rail sat side by side and hung past the card's edge on both sides; and the rail, being a `overflow-y-auto` scrollport, computed its x axis to `auto` as well, so that overflow turned a fixed-width column into a sideways scroller with nothing to scroll to.


eliseo-juan (approved):
LGTM

PR #5407, opened 2026-09-07
feat(SurveyFormBuilder): freeze a question's wording without locking the question

GuillenSanti:
## Description

Adds `lockedFields` to `SurveyFormBuilder`'s questions: freeze a question's **wording** (title, description) while leaving everything else editable — the actions menu included, so the question can still be made optional, duplicated or removed.

Today `locked` is the only tool for this and it is all or nothing. It disables the title and the description **and** it replaces the actions menu with a lock, which is right for a question the consumer owns outright and wrong for one whose wording it merely depends on.

### Why we need it

Procurement's purchase request form is **generated, then handed to the admin to edit**. The backend creates a default set of questions and gives each one a `field_identifier` binding its answer to a column on the request — `cost`, `vendor`, `url`, `due_date`. From then on the form lives in `SurveyFormBuilder` and the admin can reshape it freely.

One of those bindings is load-bearing in a way the editor cannot see. `Procurement::PurchaseRequest` refuses a deadline that is already in the past, and it reads that deadline from whichever question carries `due_date`. **The rule holds only while that question is still asking for a future date** — and nothing stops an admin renaming it into one that isn't.

When that happens the question quietly stops meaning what the backend still assumes. Asked *"When did you spend that amount?"*, requesters answer with a past date, the request is refused, and there is no answer they could give that would work. One customer renamed it exactly like that and has had **zero** purchase requests since.

So Procurement needs the wording of that one question frozen. What it must **not** freeze is everything else:

- deleting it is a perfectly fair choice — no question, no deadline, no rule to break;
- so is duplicating it, or making it optional.

`locked` cannot express that. It disables the wording *and* replaces the actions menu, so the question can no longer be deleted or duplicated either. We built it that way first, and the product team turned it down for precisely that reason: *"like the lock, but it must still be possible to delete and copy it"*.

The **type** half of the same problem is already covered by `hiddenActions: ["questionType"]` — retyping `due_date` to text sends free prose where an `ISO8601Date` is expected, and every submission of that form then dies at GraphQL coercion. `lockedFields` is the **wording** half, and the two compose on the same question:

```tsx
{
  ...question,
  hiddenActions: ["questionType"],
  lockedFields: ["title", "description"],
  lockedNote: { description: "…why this question asks for a future date…" },
}
```

### Why it belongs here and not in the consumer

The restriction is about a builder-level affordance — which parts of a question the author can edit — and there is no way to express it from outside. A consumer can only reach for `locked` and accept the collateral, or reimplement the editor's input handling.

## Type of change

## Demo

**Before** — the deadline question can be renamed and retyped freely:

https://github.com/user-attachments/assets/5a5dbfac-00b5-4f49-9780-c805ed20908c

**After** — the wording is frozen; delete, duplicate and required still work:

https://github.com/user-attachments/assets/bd1c5430-0450-47bf-a870-458f3f78b286

## Implementation details

**`lockedFields?: ReadonlyArray<"title" | "description">`** on `BaseQuestionProps`. `locked` is the stronger form and wins: a locked question (or one inside a locked section) still freezes everything regardless.

`inputDisabled` splits into `titleDisabled` / `descriptionDisabled`; the actions menu keeps its own gate (`!locked`) and is untouched.

Three deliberate choices worth reviewing:

1. **No lock icon for a partial lock.** The lock means *the card is frozen* — it takes the actions menu's place precisely because a fully locked question has no actions left. Beside a menu that still offers delete and duplicate it would claim more than is true, so a partially locked question gets none.

2. **The notice moves to the frozen field.** `lockedNote` still carries the explanation, but it hangs off whichever field is frozen instead of the lock, so the author meets it at the moment they reach for the thing they cannot do. A disabled `textarea` fires no hover events, hence the wrapper — the same reason the existing lock button hangs its tooltip off a `span`.

3. **The blocked cursor needs the important variant.** A disabled control takes `cursor: default` from a rule the plain utility does not outrank; `!cursor-not-allowed` is what the whole-card lock already reaches for. Scoped to the two fields only — the card-level `[&_*]` rule stays exclusive to the full lock, since everything else on a partially locked card is still editable.

## Impact on existing consumers

None: the prop is opt-in and nothing changes when it is absent.

Worth stating explicitly, since the alternative considered was relaxing `locked` itself so it would stop hiding the menu. Every current consumer of `locked` wants the full freeze:

| Consumer | What is locked | Should delete/duplicate appear? |
| --- | --- | --- |
| Trainings | the score question of the effectiveness / satisfaction surveys | no |
| IT Management | a category's native and inherited fields (via locked sections) | no |
| Procurement | `request_title` | no |

So `locked`'s meaning is left alone and the weaker restriction gets its own opt-in prop.

## Testing

Three tests added to `Form/__tests__/SurveyFormBuilder.blockedQuestions.test.tsx`, beside the ones already covering `locked`:

- the wording is frozen **and the actions menu is still rendered** — the point of the prop;
- only the fields named in `lockedFields` are frozen, checked both ways round;
- `locked` still wins over `lockedFields`.

Each was verified to fail against a deliberately broken implementation (menu hidden on a partial lock; every field frozen regardless of which is named; `locked` no longer winning), so they protect the behaviour rather than merely exercising it.

- `WordingLocked` story added to `BaseQuestion`.
- Survey suite green (105 tests), `tsc` and `oxfmt --check` clean.
- Exercised end to end against the Factorial procurement form editor, through a local `pnpm patch` of this change, over every combination:

| `lockedFields` | title | description | notice on hover | actions menu |
| --- | --- | --- | --- | --- |
| `["title"]` | disabled, blocked cursor | editable | yes, on the title | Required, Duplicate, Delete |
| `["description"]` | editable | disabled, blocked cursor | yes, on the description | Required, Duplicate, Delete |
| `["title", "description"]` | disabled, blocked cursor | disabled, blocked cursor | yes, on either | Required, Duplicate, Delete |

No lock icon in any of them, and a fully `locked` question in the same form still behaves as it always did.

### DIFF
--- a/packages/react/src/kits/surveys/SurveyFormBuilder/Form/__tests__/SurveyFormBuilder.blockedQuestions.test.tsx
+++ b/packages/react/src/kits/surveys/SurveyFormBuilder/Form/__tests__/SurveyFormBuilder.blockedQuestions.test.tsx
@@ -2,7 +2,7 @@ import { describe, expect, it, vi } from "vitest"
 
 import { zeroRender as render, screen } from "@/testing/test-utils"
 
-import { SurveyFormBuilderElement } from "../../types"
+import { LockedFields, SurveyFormBuilderElement } from "../../types"
 import { SurveyFormBuilder } from "../index"
 
 // --- Test fixtures ---
@@ -10,7 +10,11 @@ import { SurveyFormBuilder } from "../index"
 const makeQuestion = (
   id: string,
   title: string,
-  opts: { description?: string } = {}
+  opts: {
+    description?: string
+    locked?: boolean
+    lockedFields?: LockedFields
+  } = {}
 ): SurveyFormBuilderElement => ({
   type: "question",
   question: {
@@ -18,6 +22,8 @@ const makeQuestion = (
     title,
     type: "text" as const,
     ...(opts.description !== undefined && { description: opts.description }),
+    ...(opts.locked !== undefined && { locked: opts.locked }),
+    ...(opts.lockedFields !== undefined && { lockedFields: opts.lockedFields }),
   },
 })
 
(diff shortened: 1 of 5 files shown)


FactSolStiep (approved):
LGTM

One ask: a test in `Form/__tests__/SurveyFormBuilder.blockedQuestions.test.tsx`, which already covers `locked`. With `lockedFields: ["title", "description"]`: both inputs disabled, and the actions menu still rendered. That last part is the whole point of the prop and nothing in CI protects it, the story covers how it looks, not what stays editable

GuillenSanti:
Added, thanks — good catch. Three tests in `SurveyFormBuilder.blockedQuestions.test.tsx`:

- `freezes the wording but keeps the actions menu on a partially locked question` — the one you asked for: both inputs disabled, `Actions` still rendered, no `Locked` button.
- `freezes only the fields named in lockedFields` — checked both ways round. Worth having: an earlier revision of this branch surfaced the notice only on the title, so a `["description"]`-only lock had no way to explain itself.
- `lets locked win over lockedFields` — pins the precedence the prop docs claim.

I checked each one fails against a deliberately broken implementation (menu hidden on a partial lock, every field frozen regardless of which is named, `locked` no longer winning), so they protect the behaviour instead of just exercising it.

sauldom102 (approved):

PR #5408, opened 2026-09-07
feat(F0LocationInput): add F0LocationInput component

AngelMsaa:
## Description

Adds `F0LocationInput`, one structured address input for the design system, so no product surface has to know what Google Places is again. It has two shapes: a searchable address field, or `manualEntry`, which renders country, address line 1 and 2, city, region and postal code as typed fields.

The address field is an `F0Select` whose options are the provider's suggestions, so searching, keyboard handling, dialog portalling and the a11y contract are the select's. The component makes no network call and ships no API key: the consumer injects `searchPlaces(query, { country })` and `resolvePlace(id)`, and without them the address line is a plain `F0TextInput`.

The F0Form `location` field type is stacked on top in #5428.

## Type of change

- [x] New component (aligned in #f0-support, lands in `experimental/`)

- Linked #f0-support thread: pending
- Phase exit criteria met: build phase; stories, tests, play function, snapshot, docs at the good tier and axe enforced

Storybook: `Components / Inputs / Location input`. Picking a suggestion fills country, address, city, region and postal code.

## Implementation details

- feat: `manualEntry` for the typed block, plus `partLabels`, `allowedCountries`, `defaultCountry`, and the shared input vocabulary from `F0PhoneInput`. Defaults to `size="md"`
- feat: structured `F0LocationInputValue` (camelCase, lowercase ISO-2 country) and `onChange(value, { source, isResolved })`; editing a part that moves the pin invalidates `placeId`, coordinates and `timezone`

  <details>
  <summary>Why?</summary>

  `WorkplaceCreateDialog` hand-rolls this today with a `resolvedAddress` ref, because reusing stale coordinates would geofence the wrong site. The rule belongs in the component; the consumer decides whether to geocode a typed address. `addressLine2` is the exception: a floor number stays inside the same building, so it keeps the resolution.
  </details>

- feat: the address list is an `F0Select` with `searchFn` disabled (the provider already ranked the results, and the built-in comparison is accent-sensitive: "Colon" dropped "Colón"), `defaultItem` carrying the current address so the trigger can display a value no option covers, and no create action - only a suggestion can be chosen
- feat: `F0Select` gains `hideArrow`, for a field where the select is an implementation detail rather than the affordance
- feat: `locationInput` i18n namespace, with generic part labels ("Region", "Postal code") overridable through `partLabels`
- test: the component, the value rule, the search hook, and the two `F0Select` changes; stories with play functions and a snapshot; docs at the good tier

### Review fixes in the last commit

- `F0Select` no longer hands `required` to the select primitive. It reached Radix's hidden native `<select>`, whose options exist only while the dropdown is mounted, so a closed select with a value failed constraint validation and the browser blocked the submit - for every consumer, not just this component.
- Controlled-ness is frozen on the first render: the component emits `undefined` for an empty value, and a controlled parent handing that back used to flip the field to uncontrolled, where the address it held reappeared and the field could not be cleared.
- A pick the provider cannot resolve now replaces the value instead of patching the address line over the previous city and postal code.
- Deselecting the suggestion in the open list clears the value, rather than emptying the trigger while the form keeps the old address.

Follow-ups, deliberately not here: monolith adoption in `WorkplaceCreateDialog`, a Places session-token seam in the injected contract, an autofill opt-out for surfaces capturing someone else's address, and making a required `F0Select` announce itself to assistive tech (a pre-existing gap this PR no longer pretends to fix).


AndresRojoFactorial (commented):
## Review notes: two accessibility gaps, and two items for a follow-up PR

Reviewed locally against `origin/main` at `1ccb1e24f`: 31 unit tests run, the Storybook runner reproduced, and the component driven in Chromium with Playwright. Inline comments carry the rest. This comment holds the two accessibility findings that I think belong in this PR, and the two that do not.

The component itself is good. The rule it exists for — a manual edit to a part that moves the pin invalidates `placeId`, the coordinates and the timezone — is correctly implemented, tested in both directions, and fails safe for any part added later, because `keepResolution` defaults to `false`. The `F0Select` change is surgical and provably inert for the other 29 consumers: `hideArrow = false` is destructured before `...props`, and with the default the ternary produces the same `<Arrow>` element as before. Docs land at `gold` with `a11yTier: "enforced"`, above the `good` the description claims.

Note on the four red jobs: `Format` and `Linting` fail because the branch left a `main` from before `experimentalSortImports` and `--type-aware` were turned on, not because of anything in the code. A rebase plus `pnpm format` and `pnpm lint:fix` clears both. `Storybook Tests 4/8` is a real failure — see the inline comment on the play function.

---

## In scope here: two accessibility gaps

The stories set `a11y: { test: "error" }`, which is stricter than the repo default, and axe reports zero violations even with the dropdown open over the whole document. Both findings below are things axe structurally cannot see, so the strict setting is not evidence against them.

### 1. The error message is neither announced nor associated with a field

With `required` and `error` set, the trigger carries no `aria-invalid`, no `aria-describedby` and no `aria-required`, and the message renders as an orphan `<li>` with no `id`, no `role="alert"` and no `aria-live` (`components/F0InputField/components/InputMessages.tsx:36-69`).

Visually it is right: the red border goes to the field and the text to the group, without duplication (`F0LocationInput.tsx:161-164`). Nothing links the two in the accessibility tree.

```mermaid
sequenceDiagram
    participant U as Screen reader user
    participant F as Form
    participant S as Server
    U->>F: fills the address block
    U->>F: submits
    F->>S: POST
    S-->>F: address rejected
    F->>F: renders the red message under the fieldset
    Note over U,F: nothing is announced,<br/>and no field reports aria-invalid
    U->>F: walks the six fields again
    Note over U: no field says "invalid",<br/>none says "required"
    U->>F: cannot tell what to correct
```

WCAG 3.3.1 (Error Identification, level A). About five lines, on the `<fieldset>` that already exists:

```tsx
const messageId = `${useId()}-message`

<fieldset
  aria-label={label}
  aria-describedby={effectiveStatus?.message ? messageId : undefined}
  aria-invalid={effectiveStatus?.type === "error" || undefined}
>
  {/* … */}
  <div id={messageId} role={effectiveStatus?.type === "error" ? "alert" : undefined}>
    <InputMessages status={effectiveStatus} />
  </div>
</fieldset>
```

Per-field `aria-invalid` needs `F0InputField`, which is not in this diff. `F0TextInput` already accepts `aria-describedby` and `aria-invalid` in its type and nobody passes them, so that is a separate ticket.

One related line that *is* in this diff: `required` never reaches assistive tech either. The vendored Radix trigger would emit `aria-required`, but `F0Select` extracts `required` from the spread before building the props that reach the root (`components/F0Select/F0Select.tsx:1113-1123`). Adding `required` there fixes it for every select, in a file this PR already touches.

### 2. The async combobox announces nothing

There is no live region anywhere in the component's tree. The three messages that do exist — searching, type more, no results — only feed the list's empty text, which a screen reader does not read on change. The one `aria-live` in `SelectContent` sits behind a condition that is false here, because `AddressSelect` passes no `children` (`ui/Select/components/SelectContent.tsx:445-455`). `aria-busy` does not help either: `isSearching` is never propagated as `loading` to the select (`F0LocationInput.tsx:182`).

Verified by dumping the DOM: `liveRegions: []` with the list open and populated, `[]` during the search, and `aria-busy: false` on both the trigger and the search box while searching.

```mermaid
sequenceDiagram
    participant U as Screen reader user
    participant T as Trigger
    participant B as Search box
    participant P as Provider
    U->>T: Tab
    T-->>U: "Address, combo box, collapsed"
    U->>T: Enter
    T->>B: opens and moves focus
    B-->>U: "Enter an address, search"
    U->>B: types "Colon"
    B->>P: after the 250 ms debounce
    Note over U: silence: no "searching", no aria-busy
    P-->>B: two suggestions
    Note over U: silence: no "2 results available"
    U->>U: waits at a field that seems unresponsive
```

About eight lines in `AddressSelect.tsx`, plus one i18n key:

```tsx
const status = isSearching
  ? i18n.locationInput.searching
  : query.trim().length < MIN_QUERY_LENGTH
    ? ""
    : suggestions.length
      ? i18n.t("locationInput.resultsCount", { count: suggestions.length })
      : i18n.locationInput.noResults

<span className="sr-only" role="status" aria-live="polite">{status}</span>
```

…and `loading: loading || isSearching` in the shared props, which also adds the loading feedback that is missing today when results are already on screen.

This also mitigates the `role="option"` status message from the inline comment on the play function: the user hears the state from the live region instead of as a phantom option.

---

## Out of scope: separate PR, Workplaces backlog next to the F0Map work

These two are real, and neither is this PR's to fix. Both root causes live in `F0Select` / `ui/Select` and affect all 30 consumers, so a local patch here would be the wrong shape. Proposing they go to a separate PR, tracked in the Workplaces backlog alongside `F0Map`.

### 3. Focus lands on `<body>` when a suggestion is picked

`SelectContent` calls `event.preventDefault()` unconditionally in `onCloseAutoFocus`, with the comment that the browser will return focus naturally to the last focused element. That is not true here: the focused element is the search box, which lives *inside* the portal, so when the portal unmounts, focus falls to `<body>`.

Verified in three independent scenarios: mouse pick gives `focus: BODY`, keyboard pick gives `focus: BODY`, and Escape from the search box gives `focus: BODY` without returning to the trigger.

```mermaid
flowchart TD
  A["Tab reaches the trigger"] --> B["Enter opens the panel"]
  B --> C["Focus moves to the search box,<br/>which lives inside the portal"]
  C --> D["User types, ArrowDown, Enter"]
  D --> E["Panel closes"]
  E --> F["onCloseAutoFocus calls<br/>event.preventDefault()"]
  F --> G["Portal unmounts.<br/>The focused element no longer exists."]
  G --> H["Focus falls to body"]
  H --> I["The next Tab restarts from<br/>the top of the page or dialog"]
  I --> J["In a six-field block the user<br/>walks the whole form again"]
```

Concrete example, detailed mode with the keyboard only: country, then address, then pick a suggestion. The next Tab should reach "Address line 2" and instead reaches the first focusable element of the page. Same with Escape. WCAG 2.4.3 (Focus Order, level A).

Two fixes, at two levels:

- **Local, about four lines**, if this ever has to ship before the structural one. Capture the trigger ref, which `F0Select` forwards as `React.Ref<HTMLButtonElement>`, and refocus it on close:
  ```tsx
  const triggerRef = useRef<HTMLButtonElement>(null)
  // in the shared props:
  ref: triggerRef,
  onOpenChange: (open: boolean) => { if (!open) triggerRef.current?.focus() },
  ```
- **Structural, the one worth doing.** In `SelectContent.onCloseAutoFocus`, call `preventDefault()` only when the previously focused element is still in the document, and otherwise let Radix restore focus to the trigger. Every select in the design system gains correct focus return, and this component needs no code of its own.

### 4. In controlled mode the field shows the value the parent rejected

`F0Select` keeps its own selection cache, and the order it resolves the trigger from puts that cache *before* `value` and `defaultItem` (`components/F0Select/F0Select.tsx:518-552`). Once the select marks the chosen option internally it paints it, whether or not the parent accepted the change.

The rest of the component is correctly controlled: `useControllableState` with no `onChange` of its own, and no sync effect that could overwrite what the user types. The problem is entirely in the select layer.

Reproduced: with `value={{ addressLine1: "Fixed street 1" }}` held constant and a parent that ignores `onChange`, picking a suggestion makes the trigger read "Carrer de Colón 12, Barcelona" while `value` is still "Fixed street 1". The `Controlled` story does not expose it, because its parent always accepts.

```mermaid
sequenceDiagram
    participant U as User
    participant S as F0Select
    participant L as F0LocationInput
    participant P as Parent (controlled)
    P->>L: value = { addressLine1: "A" }
    U->>S: picks suggestion "B"
    S->>S: marks B in its selection cache
    S->>L: onPick(B)
    L->>P: onChange(B)
    P->>P: validates and rejects B
    P->>L: value is still A
    Note over S: getDisplayItemsForSelection reads its<br/>cache before value and defaultItem
    S-->>U: the field shows B
```

Concrete example: a form that only accepts addresses inside a known region. The user picks an out-of-region address, the parent rejects it and keeps the previous one, and the field keeps showing the rejected address. The form saves A and the user reads B.

Two fixes, again at two levels:

- **Local mitigation**: remount the select when the external value does not match the selection, with `key={selectedValue ?? "empty"}` on the `F0Select` in `AddressSelect`. It works, and it throws away the select's internal state on every mismatch.
- **Structural, the one worth doing**: make `F0Select` treat `value` as authoritative in the `field` variant, the way it already does in `variant="inline"` (`components/F0Select/F0Select.tsx:805-819`).

Either way, the test that pins it: a parent that rejects the change must keep seeing its own value.

---

## What this review does not cover

- **No test against the real Google Places API.** Everything verified uses the simulated providers from the stories and the tests. How the component behaves with real responses — points of interest with no breakdown, plus codes, results without complete `address_components` — is exactly the ground of the inline comment on `isLocationValueEmpty`, and there it is only a synthetic reproduction.
- **No real screen reader.** The two accessibility findings above, and the focus one, come from accessibility-tree dumps and from axe in jsdom, not from VoiceOver or NVDA. Whether the trigger exposes the address as the combobox *value* depends on each browser's AX mapping and is still open.
- **Contrast and target size unevaluated.** jsdom has no layout, so those two axe rules cannot be checked. CI does not audit the dropdown either, because it portals outside the root that gets scanned.
- **One browser only.** Everything interactive was checked in Chromium. Safari and Firefox can differ in focus order when a portal unmounts, which is the exact mechanism of finding 3.

packages/react/src/experimental/Forms/F0LocationInput/__stories__/F0LocationInput.stories.tsx
```diff
@@ -0,0 +1,391 @@
...
+    const body = within(document.body)
+
+    await userEvent.click(canvas.getByRole("combobox", { name: "Address" }))
+    const search = await body.findByRole("searchbox")
+    await userEvent.type(search, "Colon")
+
+    await waitFor(() =>
+      expect(body.getAllByRole("option").length).toBeGreaterThan(0)
+    )
+    await userEvent.click(body.getAllByRole("option")[0])
```
AndresRojoFactorial: **The play function cannot pass. `getAllByRole("option")` counts the empty-state message.**

`SelectContent` renders the empty state as `<div role="option" aria-disabled="true">` (`packages/react/src/ui/Select/components/SelectContent.tsx:311`). The list therefore exposes one `option` before the provider returns anything, `length > 0` resolves on the first tick, and the click lands on a disabled node. `onChange` never fires, and `toHaveBeenLastCalledWith` fails after the timeout.

Reproduced locally with the real runner (`test-storybook --url http://localhost:6006`): 1 failed, 12 passed, same message as CI. Instrumenting the same flow, `getAllByRole("option")` returns `["Type an address to search"]` and `onChange` is called 0 times. Filtering the disabled node returns the two expected addresses and `onChange` is called exactly once, with the value this assertion expects.

Fix — wait for the suggestion, not for "some option":

```diff
-    await waitFor(() =>
-      expect(body.getAllByRole("option").length).toBeGreaterThan(0)
-    )
-    await userEvent.click(body.getAllByRole("option")[0])
+    const suggestion = await body.findByRole("option", {
+      name: /Carrer de Colón/,
+    })
+    await userEvent.click(suggestion)
```

The unit tests avoid this by asserting an exact count (`toHaveLength(2)`, `__tests__/F0LocationInput.test.tsx:178`), which waits until both suggestions arrive.

The root cause is in `ui/Select` and affects every select: a status message with `role="option"` also makes a screen reader announce "Type an address to search" as a selectable option, and it falsifies the list's item count. That part is not yours to fix here, but it deserves its own ticket.

packages/react/src/experimental/Forms/F0LocationInput/types.ts
```diff
@@ -0,0 +1,125 @@
...
+   * Suggestion provider. Without it there is no autocomplete and the parts
+   * stand alone as plain fields. Called with the current country so the
+   * consumer can restrict the provider (e.g. Places `componentRestrictions`).
+   */
+  searchPlaces?: (
+    query: string,
+    context: F0LocationSearchContext
+  ) => Promise<F0LocationSuggestion[]>
+  /** Resolves a picked suggestion into a full value. */
+  resolvePlace?: (id: string) => Promise<F0LocationInputValue | undefined>
```
AndresRojoFactorial, line 116: **A resolved place without granular parts clears the field.**

`isLocationValueEmpty` decides emptiness from `country` and the five editable parts only (`lib/format.ts:30-36`). It ignores `formatted`, `placeId` and the coordinates. `emit` applies it to every change, including the `picked` path, so a value that carries `formatted` plus coordinates but no breakdown normalizes to `undefined`. The user picks the right address and the field goes blank.

Reproduced with a `resolvePlace` that returns `{ formatted, placeId, latitude, longitude, timezone }`: `onChange(undefined, { source: "picked", isResolved: false })`, and the trigger falls back to the placeholder. The test fixture hides it, because it always carries `addressLine1`, city, postal code, region **and** country, so it never reaches this branch.

The one-line fix in `lib/format.ts`:

```diff
-  if (value.country) return false
+  if (value.country || value.placeId || trimmed(value.formatted)) return false
```

…plus a `formatted` fallback for the trigger text in `F0LocationInput.tsx:171`, so the user sees what is stored.

### Beyond the fix: tie the resolver to a provider

The bug above is a symptom. The contract leaves the whole mapping to each consumer, so every product writes its own translation from the provider response into `F0LocationInputValue`, and every product can get it wrong the same way. Naming the provider in the props moves that mapping into the component:

```ts
export type F0LocationProvider = "google"

export interface F0LocationInputProps {
  /** Suggestion provider. Defaults to `"google"`. */
  provider?: F0LocationProvider
  // …
}
```

With a declared provider the component owns the mapping, and `searchPlaces` / `resolvePlace` stay as the escape hatch for a consumer on another service. Google is the only value we need today, so the union starts with one member and grows when a second provider appears.

Two things this changes for the API, worth deciding now while there are no consumers:

- `resolvePlace` becomes optional in a stronger sense: with `provider="google"` the consumer supplies credentials and the component resolves, instead of supplying a resolver.
- The `isLocationValueEmpty` guard above is still needed, because a consumer using the escape hatch can still return a value with no breakdown.

packages/react/src/experimental/Forms/F0LocationInput/F0LocationInput.tsx
```diff
@@ -0,0 +1,259 @@
...
+      postalCode: partLabels?.postalCode ?? i18n.locationInput.postalCode,
+      country: partLabels?.country ?? i18n.locationInput.country,
+    }),
+    [partLabels, detailed, label, i18n]
+  )
+
+  const searchCountry =
+    value?.country ??
+    defaultCountry ??
+    (countries?.length === 1 ? countries[0] : undefined)
```
AndresRojoFactorial: **In simple mode, the first picked address locks the search to that country with no way out.**

`searchCountry` reads `value?.country` first, and `usePlaceSearch` passes it to the provider as the country scope (`hooks/usePlaceSearch.ts:61`). In detailed mode that is right: the user sees the country and can change it. In simple mode there is no country selector, and `clearable` defaults to `false`, so nothing on screen explains or undoes the restriction.

Reproduced in simple mode: after picking a Spanish address, the next search runs as `searchPlaces("Downing", { country: "es" })` and the provider returns nothing. The user reads "No addresses found" for an address that exists.

Fix — scope by the value's country only when the selector is visible:

```diff
-  const searchCountry =
-    value?.country ??
-    defaultCountry ??
-    (countries?.length === 1 ? countries[0] : undefined)
+  const searchCountry =
+    (visibleFields.has("country") ? value?.country : undefined) ??
+    defaultCountry ??
+    (countries?.length === 1 ? countries[0] : undefined)
```

`visibleFields` is already computed above at line 73, so nothing needs reordering. Simple mode then searches worldwide, which is what its interface promises, and detailed mode keeps scoping by the country the user can see.

packages/react/src/experimental/Forms/F0LocationInput/hooks/useLocationValue.ts
```diff
@@ -0,0 +1,109 @@
...
+    },
+    [edited]
+  )
+
+  const setCountry = useCallback(
+    (country: CountryCode | undefined) => {
+      edited({ ...valueRef.current, country })
+    },
+    [edited]
+  )
```
AndresRojoFactorial, line 117: **Question for you: what should happen to the address when the country changes?**

`setCountry` replaces the country and leaves `addressLine1`, `city`, `state` and `postalCode` untouched. Only the resolution data is dropped, which is correct, because changing the country moves the pin. But `formatted` is then rebuilt by appending the new country to the old parts.

Measured: starting from a resolved Barcelona value and switching the country to France emits

```
{ formatted: "Carrer de Colón 12, 08001 Barcelona, Catalonia, France",
  city: "Barcelona", state: "Catalonia", postalCode: "08001", country: "fr" }
```

An address that does not exist, and the form accepts it without complaint.

```mermaid
flowchart TD
  A["User picks<br/>Carrer de Colón 12, Barcelona (es)"] --> B["Value is resolved:<br/>placeId + lat/lng + city/state/postalCode"]
  B --> C["User realises the office is the French one<br/>and changes country to fr"]
  C --> D["setCountry:<br/>{ ...valueRef.current, country }"]
  D --> E["invalidateResolution drops<br/>placeId, latitude, longitude, timezone"]
  E --> F["formatted rebuilt as<br/>'…Barcelona, Catalonia, France'"]
  F --> G{"Which behaviour<br/>do we want?"}
  G -->|"Option A: clear"| H["Also clear addressLine1, city,<br/>state and postalCode.<br/>Consistent with the invalidation rule<br/>the component already applies."]
  G -->|"Option B: keep"| I["Leave the parts as they are.<br/>Document it, and the consumer validates.<br/>Better if the country field exists to fix<br/>a border case on a correct address."]
```

Both options are defensible, so this is your call rather than a defect I can settle.

Option A is the more consistent one: the component already argues that a change which moves the pin invalidates what the pin described, and a country change moves it further than a street change. Option B is better if you expect the country selector to be used to correct one wrong field on an otherwise right address.

What is not fine is the current state, because it is surprising behaviour that no TSDoc or MDX section describes. Whichever you pick, please add the test: `setCountry` is not covered today, so nothing would catch a regression either way.

packages/react/src/experimental/Forms/F0LocationInput/hooks/usePlaceSearch.ts
```diff
@@ -0,0 +1,92 @@
...
+            if (currentSearchId !== searchIdRef.current) return
+            setSuggestions(data)
+          })
+          .catch((error: unknown) => {
+            if (currentSearchId !== searchIdRef.current) return
+            setSuggestions([])
+            if (process.env.NODE_ENV !== "production") {
+              console.warn("F0LocationInput: searchPlaces rejected", error)
+            }
+          })
```
AndresRojoFactorial, line 90: **A provider failure is shown to the user as "this address does not exist".**

The `catch` empties the suggestions and warns only outside production. There is no error state anywhere in the component, and `emptyMessage` has three cases only (`components/AddressSelect.tsx:101-105`): searching, type more, no results. A failure inevitably lands on the third one.

Measured with a rejecting `searchPlaces`: the list shows "No addresses found", identical to a provider that legitimately returns zero results. There is no retry, because the component does not know an error happened.

### Proposal: treat "not found" and "failed" as the same fallback

Both cases have the same consequence for the user: the address cannot be resolved, so the structured parts can no longer be derived from a suggestion. In that state the component should fall back to manual entry and require what it can no longer infer.

1. Keep the string the user typed as `addressLine1`, instead of discarding it.
2. Show country, state and city, and make them required.
3. Keep emitting `isResolved: false`, which already carries the right meaning for the consumer.

That needs the hook to distinguish three outcomes instead of two:

```diff
-  const [suggestions, setSuggestions] = useState<F0LocationSuggestion[]>([])
-  const [isSearching, setIsSearching] = useState(false)
+  const [suggestions, setSuggestions] = useState<F0LocationSuggestion[]>([])
+  const [status, setStatus] = useState<"idle" | "searching" | "empty" | "error">("idle")
```

`setStatus("error")` in the `catch`, `setStatus(data.length ? "idle" : "empty")` in the `then`, and `setStatus("searching")` where `setIsSearching(true)` is today. The component then keys the fallback off `status === "error" || status === "empty"` once the query is long enough to have been sent.

Three notes on scope:

- Making the parts required changes the public contract, so it needs a decision from design, not only from code review.
- The error message needs its own copy, in a new `locationInput.searchError` key. "No addresses found" must not be reused for a failure.
- Forcing `fields` from inside the component conflicts with `fields` being the consumer's choice. The narrower version is to require the parts only when the consumer already renders them, and to surface the state through `onChange` meta so the consumer can decide.

packages/react/src/experimental/Forms/F0LocationInput/hooks/usePlaceSearch.ts
```diff
@@ -0,0 +1,92 @@
...
+        setIsSearching(false)
+        return
+      }
+
+      cancelPending()
+      setIsSearching(true)
+      const currentSearchId = searchIdRef.current
+
+      debounceRef.current = setTimeout(() => {
+        searchPlaces(trimmed, { country: countryRef.current })
```
AndresRojoFactorial: **A synchronous throw in `searchPlaces` escapes the promise chain and freezes the list on "Searching addresses".**

`searchPlaces` is called directly inside the `setTimeout` callback, and `then` / `catch` / `finally` are chained onto its return value. If the function throws *before* its first `await` — no API key, `google.maps` not loaded yet — there is no promise to reject. The exception escapes the timer, no error boundary catches it, and `isSearching` stays `true`.

Measured with `searchPlaces = () => { throw new Error("boom") }`: after two seconds the list still reads "Searching addresses", and the runner reports `Unhandled Errors: Error: boom`.

```mermaid
sequenceDiagram
    participant U as User
    participant H as usePlaceSearch
    participant T as setTimeout callback
    participant P as searchPlaces (consumer)
    U->>H: types "Colon"
    H->>H: setIsSearching(true)
    H->>T: schedule after 250 ms
    T->>P: searchPlaces(q, { country })
    P--xT: throws synchronously
    Note over T,P: no promise is returned,<br/>so .catch() is never attached
    Note over H: isSearching stays true;<br/>emptyMessage stays "Searching addresses"
    U->>H: the field never answers
```

Fix — one line, so a synchronous throw becomes a rejection the existing `catch` already handles:

```diff
-        searchPlaces(trimmed, { country: countryRef.current })
+        Promise.resolve()
+          .then(() => searchPlaces(trimmed, { country: countryRef.current }))
           .then((data) => {
```

The same pattern applies to `resolvePlace` in `F0LocationInput.tsx:117-142`, where a synchronous throw leaves `resolving` and `aria-busy` stuck and the field looks locked instead.

packages/react/src/experimental/Forms/F0LocationInput/F0LocationInput.tsx
```diff
@@ -0,0 +1,259 @@
...
+
+  // Picking shows the suggestion right away, but nothing is emitted until the
+  // place resolves: an intermediate "typed" change would make consumers
+  // geocode an address that is about to arrive with coordinates.
+  const [pendingLabel, setPendingLabel] = useState<string | undefined>()
+  const [resolving, setResolving] = useState(false)
+  const pickIdRef = useRef(0)
```
AndresRojoFactorial, line 132: **No unmount guard: closing the form right after picking can still save the address.**

`pickIdRef` correctly stops a stale resolution from overwriting a newer edit, but nothing increments it on unmount. `usePlaceSearch` has that guard (`hooks/usePlaceSearch.ts:89`); the pick path does not.

Measured: pick a suggestion, unmount immediately, and 300 ms later `onChange` fires with `{ source: "picked", isResolved: true }`. A dialog the user just closed can mark the form dirty, or autosave a value the user considered discarded.

Fix — the same pattern the search hook already uses:

```diff
   const pickIdRef = useRef(0)
+
+  // A resolution in flight must not emit after the field is gone
+  useEffect(() => () => { pickIdRef.current += 1 }, [])
```

`useEffect` needs adding to the React import on line 1. Worth a test as well: unmounting with `resolvePlace` in flight must not call `onChange`.

packages/react/src/experimental/Forms/F0LocationInput/components/AddressSelect.tsx
```diff
@@ -0,0 +1,151 @@
...
+    searchBoxPlaceholder: i18n.locationInput.placeholder,
+    searchEmptyMessage: emptyMessage,
+    onSearchChange: search,
+    status,
+    required,
+    disabled: noEdit,
+    loading,
+    size,
+    name,
+    "data-testid": "location-input-address",
```
AndresRojoFactorial: **This test id never reaches the DOM. The prop is `dataTestId`.**

`F0Select` accepts `dataTestId` in camelCase through `WithDataTestIdProps` (`lib/data-testid/index.tsx:34-36`) and renders it with its own wrapper. `"data-testid"` with hyphens is not that prop. TypeScript does not flag it, because hyphenated JSX attributes are exempt from the excess property check, and `F0Select` never spreads its props onto the DOM.

Measured: `document.querySelector('[data-testid="location-input-address"]')` returns `null`. Same for `location-input-country` in `components/CountrySelect.tsx:52`.

Nothing breaks today, because no test and no story uses either id. They are dead hooks. The failure arrives the day someone writes an end-to-end test against them.

Simplest option: delete both lines. `F0PhoneInput` exposes no test ids either, so dropping them keeps the two components in the same folder consistent, and the tests here already query by role. If you want to keep them, rename to `dataTestId` and add a test that asserts they are in the DOM — this is exactly the kind of thing that breaks silently.

packages/react/src/experimental/Forms/F0LocationInput/components/AddressParts.tsx
```diff
@@ -0,0 +1,69 @@
...
+    1: "sm:grid-cols-1",
+    2: "sm:grid-cols-2",
+    3: "sm:grid-cols-3",
+  }[rowParts.length]
+
+  return (
+    <>
+      {fields.has("addressLine2") && part("addressLine2")}
+      {rowParts.length > 0 && (
+        <div className={cn("grid grid-cols-1 gap-3", rowColumns)}>
```
AndresRojoFactorial: **`sm:` is a viewport breakpoint, so this row breaks in the container where detailed mode is most used.**

Detailed mode is meant for side panels, dialogs and form columns. On a desktop screen with the component inside a 280 px panel, `sm:` is active and the three fields get about 85 px each: the labels truncate and the postal code stops being usable. On a 500 px phone the opposite happens, and the row stacks even when two columns would fit.

Neither `InsideDialog` nor `Snapshot` exposes it: the dialog is wide, and the snapshot uses `max-w-md` on a desktop viewport. Chromatic will not catch it either.

Fix — container queries, which the package already uses for this (`components/F0Alert/F0Alert.tsx:83`, `OneEmptyState`, `F0CardHorizontal`), and which the styling conventions name as the pattern to follow:

```diff
   const rowColumns = {
-    1: "sm:grid-cols-1",
-    2: "sm:grid-cols-2",
-    3: "sm:grid-cols-3",
+    1: "@sm:grid-cols-1",
+    2: "@sm:grid-cols-2",
+    3: "@sm:grid-cols-3",
   }[rowParts.length]
```

The `@container` has to sit on the parent, not on the element that reads the query, so the row needs one wrapper:

```diff
       {rowParts.length > 0 && (
-        <div className={cn("grid grid-cols-1 gap-3", rowColumns)}>
-          {rowParts.map(part)}
-        </div>
+        <div className="@container">
+          <div className={cn("grid grid-cols-1 gap-3", rowColumns)}>
+            {rowParts.map(part)}
+          </div>
+        </div>
       )}
```

The literal class map stays as it is: it is the right way to keep Tailwind from purging the classes, and the comment above it is worth keeping. Worth adding a narrow variant to the snapshot as well, so Chromatic watches this from now on.

packages/react/src/experimental/Forms/F0LocationInput/types.ts
```diff
@@ -0,0 +1,125 @@
...
+}
+
+export type F0LocationSearchContext = {
+  /** Country to scope the search to, when one is selected or implied */
+  country?: CountryCode
+}
+
+export type F0LocationInputChangeMeta = {
+  /** `"picked"` right after a suggestion is chosen, `"typed"` once any part is edited */
+  source: "picked" | "typed"
```
AndresRojoFactorial: **`source: "picked"` only holds on the happy path.**

Three of the four pick paths emit `"typed"`: no `resolvePlace`, `resolvePlace` returning `undefined`, and `resolvePlace` rejecting all fall through to `setPart("addressLine1", ...)` (`F0LocationInput.tsx:113-138`). The current test pins that behaviour rather than catching it (`__tests__/F0LocationInput.test.tsx:229-231`).

`isResolved: false` is correct in those three cases. Only `source` is wrong, because it exists to say *how* the value arrived, not whether it can be trusted.

Taking the documentation route rather than changing the emission: keep the behaviour and make the type say what the code does.

```diff
-  /** `"picked"` right after a suggestion is chosen, `"typed"` once any part is edited */
+  /**
+   * `"picked"` when a suggestion was chosen *and* resolved into a full value.
+   * `"typed"` in every other case, which includes a suggestion that could not
+   * be resolved, so a consumer must not read `"typed"` as "the user did not
+   * use the list". Use `isResolved` to decide whether the value is trustworthy.
+   */
   source: "picked" | "typed"
```

That leaves `onChange` untouched and removes the mismatch between the contract and the code. It also keeps the door open: if a consumer later needs to tell a failed pick from typed input, that is a new field rather than a change of meaning on this one.

packages/react/src/experimental/Forms/F0LocationInput/components/AddressSelect.tsx
```diff
@@ -0,0 +1,151 @@
...
+  disabled,
+  readonly,
+  loading,
+  clearable,
+  size,
+  name,
+}: Props) => {
+  const i18n = useI18n()
+  const noEdit = disabled || readonly
+  const { suggestions, isSearching, query, search } = usePlaceSearch({
```
AndresRojoFactorial: **Reopening the dropdown shows the previous search's suggestions, with an empty search box.**

Two pieces of state that never talk to each other. `usePlaceSearch` returns a `reset` that this destructuring does not take and nothing ever calls. And `F0Select` clears its own search box without notifying through `onSearchChange` (`components/F0Select/F0Select.tsx:715-722`), so the hook keeps `query` and `suggestions`. Because local filtering is deliberately off, nothing hides the stale suggestions.

Measured: after picking and closing, reopening leaves the search box at `"Colon"`, and typing on top produced the query `"ColonDowning"`.

Proposal:

```diff
-  const { suggestions, isSearching, query, search } = usePlaceSearch({
+  const { suggestions, isSearching, query, search, reset } = usePlaceSearch({
```

```ts
// in the shared props
onOpenChange: (open: boolean) => {
  if (!open) reset()
},
```

Feel free to discard this one. Keeping the last search is a defensible choice if the user usually reopens to pick a different suggestion from the same query, and resetting costs one extra provider call in that case.

If you keep the current behaviour, the part still worth fixing is the mismatch rather than the caching: the search box shows text while the list says "type an address to search". That half lives in `F0Select`, which clears its search without telling the consumer, so it would be a separate ticket either way.

packages/react/src/experimental/Forms/F0LocationInput/__stories__/F0LocationInput.stories.tsx
AngelMsaa: Fixed in d9d3e9f7f. The play function now waits for `findByRole("option", { name: /Carrer de Colón/ })` and clicks that, with a comment explaining why "some option" is the wrong wait. Agree the `role="option"` on the empty-state message belongs to `ui/Select` and its own ticket.

packages/react/src/experimental/Forms/F0LocationInput/types.ts
AngelMsaa: The emptiness bug is fixed in d9d3e9f7f: `isLocationValueEmpty` treats `placeId` and a non-blank `formatted` as content, and the trigger falls back to `formatted` when there is no `addressLine1`. Covered by a test that picks a POI resolving to `{ formatted, placeId, latitude, longitude }` and checks both the emitted value and the trigger text.

The provider prop is a separate conversation; leaving this thread open for it.

packages/react/src/experimental/Forms/F0LocationInput/F0LocationInput.tsx
AngelMsaa: Fixed in d9d3e9f7f, with a smaller patch than the one proposed because the shape moved under you: since 1ea2630b0 manual entry types the address and never renders the search select, so the search only exists where no country selector can show or undo the scope. `searchCountry` no longer reads `value?.country` at all; `defaultCountry` and a single-entry `countries` keep scoping. Test: a resolved Spanish value followed by a search for "Downing" calls the provider with `country: undefined`.

Side finding while testing: that reset-on-country-change was masking F0Select re-emitting `onChange` for the value it already holds when its `value` prop moves (`undefined` to unlisted to `placeId`). Without the reset the echo re-ran the pick and `onChange` fired twice. `AddressSelect` now ignores an emission equal to `selectedValue`.

packages/react/src/experimental/Forms/F0LocationInput/hooks/useLocationValue.ts
AngelMsaa: Option A, in d9d3e9f7f. `setCountry` emits `{ country }` only, so the parts go with the resolution. The `manualEntry` TSDoc says so, and there is a test: from a resolved Barcelona value, picking France emits `{ country: "fr", formatted: "France" }` with `isResolved: false` and the City field reads empty.

packages/react/src/experimental/Forms/F0LocationInput/hooks/usePlaceSearch.ts
AngelMsaa: Done in d9d3e9f7f, the narrower version. `usePlaceSearch` now reports `idle | searching | empty | error`, and a failure shows a new `locationInput.searchError` ("Couldn't load addresses. Try again.") in the list instead of "No addresses found". Tests pin the four statuses and the copy.

Not taking the fallback to manual entry or making the parts required: Angel's call, the failure stays a message in the search list and the consumer still gets `isResolved: false`.

packages/react/src/experimental/Forms/F0LocationInput/hooks/usePlaceSearch.ts
AngelMsaa: Fixed in d9d3e9f7f for both `searchPlaces` and `resolvePlace`, through a small `lib/invokeAsync` helper (try/call, catch to `Promise.reject`) rather than `Promise.resolve().then(...)`, so the provider is still called synchronously when the debounce fires and the existing stale-response test keeps its timing. Tests: a provider that throws synchronously ends in `status: "error"` with `isSearching: false`; a resolver that throws synchronously falls back to the label with `source: "typed"`.

packages/react/src/experimental/Forms/F0LocationInput/F0LocationInput.tsx
AngelMsaa: Fixed in d9d3e9f7f, same pattern as the search hook: the cleanup bumps `pickIdRef`. Test: pick, unmount, resolve; `onChange` is not called.

packages/react/src/experimental/Forms/F0LocationInput/components/AddressSelect.tsx
AngelMsaa: Both lines deleted in d9d3e9f7f, matching F0PhoneInput.

packages/react/src/experimental/Forms/F0LocationInput/components/AddressParts.tsx
AngelMsaa: Fixed in d9d3e9f7f: `@container` wrapper plus `@sm:grid-cols-3`. The Snapshot story gains a 280px manual-entry instance so Chromatic watches it.

packages/react/src/experimental/Forms/F0LocationInput/types.ts
AngelMsaa: Took the documentation route in d9d3e9f7f: the TSDoc now says `"typed"` also covers a pick that could not be resolved and points consumers at `isResolved`. Emission unchanged.

packages/react/src/experimental/Forms/F0LocationInput/components/AddressSelect.tsx
AngelMsaa: Took it in d9d3e9f7f: `reset()` runs on `onOpenChange(false)`, so reopening shows the hint again. Test covers pick, close, reopen. The "search box shows text while the list says type to search" half will also go away with #5422, which makes F0Select call `onSearchChange("")` when it clears.

sauldom102 (approved):

PR #5411, opened 2026-09-07
perf(OneDataCollection): give rows the definition instead of the live source

factmarc10010:
Follow-up to #5378, now merged and part of `main`.

## Why

#5378 stopped a page append from re-rendering the rows already on screen. It left the other half untouched: **any render of the consumer still re-renders every row.**

`useDataCollectionSource` returns a new object each render and rows take it as a prop, so their memo can never hold.

Pinning that object's identity does not work. `Row` calls `selectable`, `selectionDisabled`, `itemUrl`, `itemOnClick`, `selectionInherited` and `itemsWithChildren` during its own render, so a row that never re-renders never reads a new value — however fresh the object it holds claims to be. `SourceStaleness.integration.test.tsx`, on `main`, is the guard for that.

Memoizing the source does not work either. It holds two things with different lifetimes:

```
return {
  ...rest,          // selectable, itemActions, itemUrl... — the definition
  currentFilters,   // live state; changes on every interaction
  currentSearch,
  currentSortings,
  isLoading,
}
```

Freeze it and the live state freezes with it.

## What this does

Splits them. `useDataCollectionSource` now also exposes `definition` — what the consumer wrote at the call site — memoized on the hook's `deps`. Rows take that.

Rows that render nested children still get the live source, now under `liveSource`. Resolving a child reads `currentFilters` and `currentSortings`, and children arrive too late for the table to have resolved anything on their behalf. Flat rows get `undefined`, which is stable, so nested tables behave exactly as before and flat ones — the large collections — get the memo.

`useItemActions` drops to the definition type too; it only ever read `itemActions`. A source built by hand without a `definition` falls back to the old behaviour rather than breaking, since the interface is public and mocks construct it directly.

## Opt-in, and why

Pinning the definition extends what `deps` governs. `itemActions`, `selectable`, `selectionDisabled`, `itemUrl` and `itemOnClick` ride in `...rest` and are read fresh on every render today, so a missing dep is harmless. Pinned, a missing dep leaves a row calling the closure it mounted with.

On by default, that rule would reach every consumer through a version bump, with nobody reading their `deps` first. The employees list in the monolith is already an example: a careful `deps` array — it even carries `hasAnyBulkAction` with a comment about `selectable` — but `itemActions` closes over an `itemActionsContext` rebuilt each render whose members are not all listed.

So it is off unless the source sets `memoizeDefinition`. A table takes it on in a change where someone can check its `deps`, and everyone else keeps exactly the behaviour they have. `Table.tsx` falls back to the live source when `definition` is absent, which is also what a hand-built source gets.

## Testing

`ConsumerRenderRowCommits.integration.test.tsx` renders a consumer that owns state the table knows nothing about and writes its definition inline, the way every real call site does, then bumps that state. It pins both directions: opted in, that costs no row commits; not opted in, it costs one per row, exactly as before.

The opted-in case is red on `main`:

```
× does not re-render them
  AssertionError: expected 2 to be 1
```

Green here, and `SourceStaleness.integration.test.tsx` still passes — which is the part that matters, because it means the memo holds without anything going stale.

`SelectAllUnderMemo.integration.test.tsx` answers what memoizing rows does to the selection registry, which they fill from an effect. It does not go stale in a way that matters: `collectSelectableEntries` re-filters whatever the registry hands it through the live `selectionDisabled`, so the registry decides which rows *exist* and never which of them may be picked. The test locks a row from the consumer and runs select-all, once with the definition rebuilt on `deps` and once with it pinned to an empty `deps` so the rows never re-register at all. Select-all skips the locked row either way.

Full unit suite: 7677 passing.

## Reading it yourself

`Data Collection/Performance` is a measuring harness, not a showcase. It counts row commits inside a column renderer — work a skipped row never does — and reports them per action:

- **Render the consumer** changes state the collection knows nothing about. Every row commit it reports is waste. Without the opt-in that is one per loaded row; with it, none.
- **Measure the next page** arms the counter, then you load one. Under infinite scroll the count should stay near a page's worth however many rows are already loaded — that is what #5378 fixed, and this is where a regression would show.

Both stories run the same harness, one on `pages` and one on `infinite-scroll`.

## Not in this PR

One caveat is real and worth naming: `Row` now reads `itemsWithChildren` from the pinned definition while the table picks `liveSource` from the live one. Those are the same function unless `deps` is incomplete *and* that predicate depends on consumer state — and if they ever disagree, a row's children silently do not render. It is a stable predicate on a record in every consumer checked, so this is a caveat rather than a finding.

Deriving the selection registry from the loaded data rather than the rendered rows is **not planned**. It would be a cleaner design and would delete N per-row effects, but the correctness argument for it is gone (see above, and windowing is dropped in [FCT-62647](https://factorialmakers.atlassian.net/browse/FCT-62647)), and it needs `NestedDataProvider` lifted above `TableCollection`'s hooks — it is mounted inside its JSX — which means splitting a 1200-line component for a cleanup. Recorded in [FCT-63014](https://factorialmakers.atlassian.net/browse/FCT-63014).

No measurement yet of how often a consumer of a large collection actually re-renders *in the product*. The story measures what a render costs; it cannot say how often the monolith's pages trigger one. If that turns out to be rare, this buys less than it looks.

[FCT-63014]: https://factorialmakers.atlassian.net/browse/FCT-63014?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[FCT-62647]: https://factorialmakers.atlassian.net/browse/FCT-62647?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/react/src/patterns/OneDataCollection/__stories__/performance/row-commits.stories.tsx
+++ b/packages/react/src/patterns/OneDataCollection/__stories__/performance/row-commits.stories.tsx
@@ -29,11 +29,7 @@ const people: Person[] = Array.from({ length: TOTAL }, (_, index) => ({
   department: DEPARTMENTS[index % DEPARTMENTS.length],
 }))
 
-/**
- * Counts how many times a row's cells actually ran. A row that skips its render
- * calls none of its column renderers, so this is the row-commit count, taken
- * from where the work would be paid.
- */
+/** A skipped row calls none of its column renderers, so counting there counts commits. */
 const useRowCommitCounter = () => {
   const total = useRef(0)
   const [readout, setReadout] = useState<
@@ -47,8 +43,7 @@ const useRowCommitCounter = () => {
   const measure = useCallback(
     (label: string, rows: number) => {
       const before = total.current
-      // One frame is enough: the render this triggers is synchronous, and an
-      // append settles before the next paint.
+      // The render is synchronous and an append settles before the next paint.
       requestAnimationFrame(() =>
         requestAnimationFrame(() =>
           setReadout((entries) => [
@@ -73,11 +68,9 @@ type Adapter = DataCollectionDataAdapter<
 const pageOf = (offset: number) => people.slice(offset, offset + PER_PAGE)
 
 /**
- * Split by pagination type rather than parameterised: the adapter picks its
- * response shape from a literal `paginationType`, so one function returning both
- * shapes satisfies neither. `fetchData`'s parameters are left to be inferred
- * from the annotation for the same reason — annotating them explicitly stops
- * the union resolving.
+ * Split, not parameterised: the response shape comes from a literal
+ * `paginationType`, so one function returning both satisfies neither. Leave
+ * `fetchData`'s parameters inferred for the same reason.
  */
 const makeAdapter = (
   paginationType: "pages" | "infinite-scroll",
@@ -142,8 +135,8 @@ const Harness = ({
         loadedRef.current = rows
       }),
     },
-    // Nothing here closes over component state, so the definition never has to
-    // be rebuilt. A real consumer must list whatever its callbacks capture.
+    // Nothing here closes over component state. A real consumer must list
+    // whatever its callbacks capture.
     [paginationType]
   )
 
--- a/packages/react/src/patterns/OneDataCollection/__stories__/performance/row-commits.stories.tsx
+++ b/packages/react/src/patterns/OneDataCollection/__stories__/performance/row-commits.stories.tsx
@@ -1,0 +1,244 @@
+import { Meta, StoryObj } from "@storybook/react-vite"
+import { useCallback, useRef, useState } from "react"
+
+import type { FiltersDefinition } from "@/hooks/datasource"
+
+import { F0Button } from "@/components/F0Button"
+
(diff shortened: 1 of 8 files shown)


albertpmz (commented):
Looks good! I just have one question regarding the `deps` trade:

Currently, `deps` covers `dataAdapter`, `filters` and `summaries`; everything else goes through `...rest` in useDataSource.ts' ln 240 fresh each render,  which is why a forgotten dependency is survivable. After useDataCollectionSource.ts:100-104 it also pins itemActions, selectable, selectionDisabled, itemUrl and itemOnClick, default []. Forget one now and a handler runs with mount-time state, silently. You already found the employees list doing exactly that, and every consumer picks this up on the next version bump.

Can `definition` be opt-in? `Table.tsx`'s ln 251 already falls back with `source.definition ?? source`. Populate it only when the consumer asks, and each heavy table opts in through a PR where someone can check its deps. Slower rollout, no release-day risk.

albertpmz (approved):
LGTM!

PR #5414, opened 2026-09-07
ci(lint): enable type-aware oxlint with sonarjs rules

dmitry-zaets:
## Description

Makes `pnpm lint` in `packages/react` type-aware and adds new rules: the rest of `eslint:recommended`, typescript-eslint strict and stylistic, sonarjs recommended, and eslint-plugin-import. Only rules that already pass are turned on, so no source files change. Rules that still need code fixes are listed as PENDING in `.oxlintrc.json` and land in follow-up PRs. Rules we keep off are listed with the reason.

## Type of change

- [x] Other: lint tooling

## Implementation details

- ci: run oxlint with `--type-aware` through `oxlint-tsgolint`. Needs `@factorialco/f0-core` built first, which CI already does.
- ci: load `eslint-plugin-sonarjs` and `eslint-plugin-import` as oxlint JS plugins. They run in `pnpm lint` and the hooks, not in the editor.
- build: declare the type-aware and JS-plugin dependencies in the React package manifest so frozen installs match the lockfile.
- ci: remove `baseUrl` from the react tsconfigs. tsgolint rejects it, and `paths` were already relative.
- ci: type-aware rules are off in test files, which sit outside `tsconfig.json`.
- ci: same non-type-aware rules for `packages/react-native`.
- docs: `.oxlintrc.json` now has comments, VS Code treats it as jsonc.

`pnpm lint` takes about 35 s for the whole package now, most of it the sonarjs plugin.

---
> - factorial-skills
> - factorial-skill-tracking
> - f0-pr


packages/react/tsconfig.json
```diff
@@ -20,7 +20,6 @@
     /* Shadcn */
     /* */
     "types": ["vite/client", "@testing-library/jest-dom", "vitest/globals"],
-    "baseUrl": ".",
```
sauldom102, line 23: Don't we need this anymore?

sauldom102 (approved):
LGTM!

PR #5415, opened 2026-09-07
feat(F0AiCallout): add the AI callout and deprecate F0Callout

desiree-np:
Adds `F0AiCallout` to `kits/ai/Banners`: what One found about the record on screen, and what it needs from the reader. Figma: [`RiV4xYq8R839J8NzTaBCr8`, frame `2863-49931`](https://www.figma.com/design/RiV4xYq8R839J8NzTaBCr8?node-id=2863-49931).
Screenshot 2026-09-07 at 15 22 15

## One component, four shapes

The content selects the shape — there is no `shape` or `stacked` flag, because the product already knows what it holds.

| Shape | Turned on by | For |
| --- | --- | --- |
| Verdict | `children` | One assessment, self-contained |
| Findings | `findings` | Several issues on one record, each with its own action |
| Verdict with a rationale | `evidence` | One recommendation, plus the reasoning behind it |
| Verdict with a plan | `evidence` + `kind: "steps"` | One assessment whose answer is several things to do, in order |

## Decisions worth reviewing

- **`status` is required, with no default.** A blocking finding sent as `info` still *looks* correct, which is the one mistake that survives review. The status asks whether the **reader** has work, not whether we can offer them a click — so `info` with no action is a normal shape, not an oversight.
- **The byline is not a prop.** How much a message claims is the status's job; a configurable attribution would be a second signal for the same thing. Every callout renders "Suggested by One".
- **`role="status"` + `aria-live="polite"`, always.** This diverges from `F0Alert`, which promotes `warning` and `critical` to `role="alert"`. The reason is authorship: `F0Alert` speaks for the system and a system failure may interrupt; a machine's opinion about a record the reader is already looking at does not get to seize their attention.
- **Folding differs by shape.** `findings` starts open, because a folded finding is a finding nobody read; `evidence` starts folded, because a rationale is read on demand. Folded rows are unmounted rather than hidden, and the deck behind a folded group caps at two layers.
- **The disclosure's noun comes from the product (`evidence.name`) and its verb from the component** — "See the 5 checks" / "Hide the 5 checks". The state shows in one word instead of replacing the label, and the label can only ever *name* what is behind it.
- **No per-item state anywhere** — no checkboxes, no radios. A message holding a value nobody submitted has stopped being a message, so a choice between options belongs in a dialog behind the action.

## Deprecation

`F0Callout` is deprecated: `tags: ["deprecated"]` (which is what the sidebar badge and the component-status API read), a deprecation note on its docs page, and a `@deprecated` JSDoc carrying `@removeIn 7.0.0` and `@migration`. Its Storybook title is deliberately unchanged — a "(deprecated)" suffix changes the page id, and the docs-index check flagged that it would 404 every existing link. It gets `critical` wrong — no glyph and an uncoloured title, so the strongest status is the only one not signalled — and its `ai` variant stacks a gradient on a semantic colour.

**It keeps working and none of its product call sites move.** Its skeleton is split into its own file, though: exporting two components from `CalloutInternal.tsx` made react-docgen fail with "Multiple exported component definitions found", which Storybook renders as an empty docs page — the page someone reads to migrate.

## Verified

- 33 unit tests; `tsc` and `oxlint` clean on the touched paths
- **Contrast measured, not assumed.** The byline originally carried `opacity-50` from Figma, which halves the secondary token to **2.00:1** where AA asks 4.5 — axe reported it as a serious failure on every story. Both bylines drop the opacity; re-measured across `Verdict`, `Statuses` and `Findings`, nothing falls under 4.55 and axe now reports no issues.
- The development warnings run in an effect rather than during render, so one violation logs once instead of once per re-render, and an empty `findings` array renders nothing rather than an empty critical shell
- Docs page renders 8 canvases, 0 empty, no overflow; `AICallout` reports `deprecated` and `F0AiCallout` `experimental` through the status API

`tsc` also reports one pre-existing error in `src/sds/chat/F0Chat/utils/emoji-support.ts`, from a stale `@factorialco/f0-core` in my worktree rather than from this branch.

## Translation keys

`ai.attribution` and `ai.evidence` are added to the dictionary. Additive, but consumers that maintain a full translation object have to add them — flagged by the API-breaking check for that reason, and called out here rather than shipped as a major.

## Worth deciding, not applied here

`kits/ai/exports.ts` never exports `./Banners`, so no banner — `F0AiCallout` included — reaches the `ai` entry point. The only public route is `experimental`, whose re-export is marked `@deprecated` and points at `@/kits/ai/Banners`, which is not a build entry. So this component currently ships only through a deprecated path, and so do `F0AiBanner` and `BaseBanner`.

One line in `kits/ai/exports.ts` fixes it and matches what those deprecation notices already say, but it changes what `@factorialco/f0-react/ai` exposes — happy to add it if the team wants it in this PR.

## Open, and deliberately not in this PR

- Foundations: a neutral **border** token at the 6% step (F0's scale starts at `neutral-20`, 10%, where the design asks for 6%), and a `neutral` **glyph** meaning "done"/"informative" — today `neutral` has to borrow a content icon, which is why it requires `icon`.

### DIFF
--- a/packages/react/src/kits/ai/Banners/F0AiCallout/AiCalloutInternal.tsx
+++ b/packages/react/src/kits/ai/Banners/F0AiCallout/AiCalloutInternal.tsx
@@ -1,7 +1,7 @@
 import { AnimatePresence, motion } from "motion/react"
-import { forwardRef, useEffect, useState } from "react"
+import { forwardRef, useEffect, useState, type ReactNode } from "react"
 import { F0Button } from "@/components/F0Button"
-import { F0Icon } from "@/components/F0Icon"
+import { F0Icon, IconType } from "@/components/F0Icon"
 import { One } from "@/icons/ai"
 import { ChevronDown, ChevronUp, Cross } from "@/icons/app"
 import { useReducedMotion } from "@/lib/a11y"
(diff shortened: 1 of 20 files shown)


packages/react/src/kits/ai/Banners/F0AiCallout/AiCalloutInternal.tsx
```diff
@@ -0,0 +1,570 @@
...
+              { inset: "inset-x-6", offset: "bottom-0" },
+              { inset: "inset-x-3", offset: "bottom-2" },
+            ]
+    const headerIcon = icon ?? statusIcons[status]
+
+    // The rule that keeps `neutral` and `info` apart, enforced where it is used
+    // rather than left in a doc: if there is something to do, the callout is not
+    // neutral. Without this the two rungs collapse into "grey or blue" and
+    // whichever feels calmer wins.
+    if (status === "neutral" && (action || secondaryAction)) {
```
albertpmz: These warnings will fire on every render. For example, a `neutral` callout with no icon logs 2 lines on mount and 9 after 5 parent re-renders; if there's a second violation too, it's 11 warnings on mount and 23 after the 5 re-renders.

A good fix would be wrapping these checks in a `useEffect` with:
- `[status, action, secondaryAction, icon, stacked, headline]` as deps.
- Gated on `dev`.

This also fixes warnings during render.

packages/react/src/kits/ai/Banners/F0AiCallout/AiCalloutInternal.tsx
```diff
@@ -0,0 +1,570 @@
...
+      )
+    }
+
+    if (status === "neutral" && !icon) {
+      console.warn(
+        "F0AiCallout: `neutral` carries no semantic glyph, so it needs an `icon` that describes the content (e.g. `Summary` from @/icons/ai)."
+      )
+    }
+
+    if (stacked && !headline) {
```
albertpmz: This warning indicates "nothing will be rendered instead", but the component behavior doesn't match – the header and the shell render unconditionally (see lines 488-493).

This will render a critical pill reading "Issues to resolve · Suggested by One", no buttons: a user will see something is wrong, with no content.


```suggestion
   if (stacked && !headline) {
       console.warn(
         "F0AiCallout: `findings` is empty. A callout titled like a verdict with nothing under it tells the reader something is wrong and not what — rendering nothing instead."
       )
       return null
     }
```

Also check `/F0AiCallout.test.tsx`, line 363: worth having `expect(container).toBeEmptyDOMElement()` plus `expect(screen.queryByRole("status")).toBeNull()` to stop the empty alarm being announced.

albertpmz (changes requested):
Nice effort! Some things that need attention before merging:

- The deprecation JSDoc is incomplete. `@deprecated` is present, but neither `@removeIn` nor `@migration` are, and we should add them. The body already includes this guidance, so you just need to add the tags for tools to read it.

packages/react/src/kits/ai/Banners/F0AiCallout/AiCalloutInternal.tsx
desiree-np: Applied in 84d220c75. All three checks now live in a single development-gated `useEffect` (`process.env.NODE_ENV === "production"` returns early).

One deviation from your suggestion, deliberately: the deps are `[status, hasAction, icon, stacked, hasHeadline]` rather than the `action` / `secondaryAction` / `headline` values themselves. Products pass those inline, so their identity changes on every render and the effect would re-log exactly what it is there to stop — the deps are the *presence* of an action and of a headline instead.

Pinned by a test that re-renders five times and asserts the call count does not move, so a regression here fails rather than just getting noisier.

packages/react/src/kits/ai/Banners/F0AiCallout/AiCalloutInternal.tsx
desiree-np: You were right, and the wording was the tell: the warning claimed "render nothing instead" while the header and shell were built unconditionally. Applied in 84d220c75 — it returns `null` now, so an empty array cannot put a critical pill on the page reading like a verdict with nothing under it, or have a live region announce it.

Also took the test suggestion. The old assertion was only `queryAllByRole("button")).toHaveLength(0)`, which passed with the shell rendered — which is precisely why this survived. It now asserts `expect(container).toBeEmptyDOMElement()` and `expect(screen.queryByRole("status")).toBeNull()`.

siguenzaraul (approved):

warcos-fact (approved):
LGTM

albertpmz (approved):
Looks good to me too! 🥳

PR #5419, opened 2026-09-07
chore(deps): drop the react-virtuoso patch for 4.18.13

albertcalasanzs:
## Description

`react-virtuoso@4.18.13` is the upstream release of the change f0 was carrying as a `pnpm patch`, so the vendored copy can go. #5394 patched 4.18.11 with the design accepted in [petyosi/react-virtuoso#1493](https://github.com/petyosi/react-virtuoso/pull/1493); that PR merged and shipped in 4.18.13 on 2026-09-05, crediting the same author. This drops `patches/react-virtuoso@4.18.11.patch` and its `patchedDependencies` entry, and raises the range to `^4.18.13` so the floor sits above the fix.

## Type of change

- [x] Refactor / internal change (no API or behavior change)

## Implementation details

- chore: remove `patches/react-virtuoso@4.18.11.patch` and its `patchedDependencies` entry in `pnpm-workspace.yaml`
- chore: raise `react-virtuoso` in `packages/react` from `^4.18.10` to `^4.18.13`

  <details>
  <summary>What the patch did, and why the release replaces it</summary>

  react-virtuoso compensated a prepend in two steps: `deviation` grew the content by pushing the list
  down (`margin-top`), and `scrollBy` cancelled that push **one `requestAnimationFrame` later**. In
  between sat a painted frame displaced by the full size of the page — and because pagination fires
  near `scrollTop` 0, that displacement was the whole viewport.

  The patch added a `deviationCommitted` stream to `domIOSystem`, had `List` publish the committed
  value from a **layout effect** (after React mutates the DOM, before paint), and had
  `upwardScrollFixSystem` hold the pending prepend and compensate on that acknowledgement — keeping
  the deferred `rAF` path as a fallback so the worst case equalled pre-patch behaviour.

  4.18.13 is that same mechanism, and slightly more of it: upstream also wires the acknowledgement in
  the renderer that writes `marginTop` imperatively, which the patch deliberately left on the fallback
  because F0Chat does not use it. The published dist carries the fix in **both** the ESM and the CJS
  build; the patch could only reach the ESM one, since the CJS build is a single minified line.

  The only other release in the range, 4.18.12, keeps a measurement range active when a
  window-scrolling list starts below the viewport, so its estimated height joins the document layout.
  That stream is fed only by `windowViewportRect`, published only under `useWindowScroll` /
  `customScrollParent` — neither appears anywhere in this repo, so nothing here reaches it.
  `dist/index.d.ts` is byte-identical between 4.18.11 and 4.18.13, so the public type surface is
  unchanged.
  </details>

## Verification

Ran in a clean worktree off `origin/main` @ `6c3d7e77b`, after `pnpm install`.

**The installed package is 4.18.13, unpatched**

```
$ pnpm why react-virtuoso -r
react-virtuoso@4.18.13
└── @factorialco/f0-react@6.86.5 (dependencies)
Found 1 version of react-virtuoso

$ readlink packages/react/node_modules/react-virtuoso
../../../node_modules/.pnpm/react-virtuoso@4.18.13_react-dom@18.3.1_react@18.3.1__react@18.3.1/node_modules/react-virtuoso
```

No `patch_hash=` in the store path — before this change it read
`react-virtuoso@4.18.11(patch_hash=569197f3…)`. `patchedDependencies`, `patch_hash` and the string
`4.18.11` now appear **0 times** in `pnpm-lock.yaml`, and nothing else in the repo (docs, CI,
`vendor/skills/`, `AGENTS.md`) references the patch.

**The same-frame compensation is in the installed dist**

Matching the deleted patch's `+` lines against `packages/react/node_modules/react-virtuoso/dist/index.mjs`:

```
418:      deviationCommitted: o,                                     # domIOSystem exposes the stream
2068:    { deviation: t, deviationCommitted: e, scrollBy: n, … }     # upwardScrollFixSystem consumes it
2488:  const x = Tt("deviationCommitted");                           # List publishes it
2854:  const I = wt("deviationCommitted");                           # …and the marginTop renderer too
```

The acknowledgement path and the `rAF` fallback, same shape as the patch:

```
2146:  a = { acknowledgeable: ot(t) !== x, offset: x }, M(t, x), requestAnimationFrame(T);
2123:  a === null || !a.acknowledgeable || x === a.offset && T();
 358:  be = typeof document > "u" ? E.useEffect : E.useLayoutEffect   # the layout effect
```

`deviationCommitted` occurs 4× in `dist/index.mjs` **and** 4× in `dist/index.cjs`; it occurs 0× in
4.18.11 and 0× in 4.18.12.

**Tests, types, lint**

| Check | Command | Result |
| --- | --- | --- |
| F0Chat suites | `pnpm vitest --run src/sds/chat/F0Chat` | 71 files, **704 tests passed** |
| Full unit project | `pnpm vitest --run` | 607 files (1 skipped), **7665 tests passed** (13 skipped) |
| Types | `pnpm tsc` | **0 errors** |
| Lint | `pnpm lint` | **0 warnings, 0 errors** on 3462 files |

`pnpm tsc` needs `pnpm --filter @factorialco/f0-core build` first in a fresh worktree —
`packages/core/dist` is gitignored, and without it 40 errors cascade from
`Cannot find module '@factorialco/f0-core'`. None of them mentioned react-virtuoso, and all 40 went
away once core was built.

**Lockfile delta** — 5 insertions, 10 deletions, every line react-virtuoso or the removed
`patchedDependencies` block. No re-resolution, nothing else added, bumped or dropped, `overrides:`
untouched. Reproduced from the manifests alone with the repo-pinned `pnpm@10.34.5`
(`install --lockfile-only`): byte-identical to the committed lockfile, so it is an install product,
not a hand-edit. The recorded integrity `sha512-M4kP7zbV7JmLrn…8CfQSQ==` matches
`npm view react-virtuoso@4.18.13 dist.integrity`, and sha256 of the installed `dist/index.mjs`,
`dist/index.cjs`, `dist/index.d.ts` and `package.json` match the freshly packed npm tarball
byte-for-byte.

### Deferred to review

- **Visual**: the F0Chat prepend stress story (`F0Chat` → *200k messages (virtualized)*) shows 0
  blank frames. Not re-measured here — this
  branch runs no Storybook and no browser. The method (two f0 builds differing only by the patch,
  `5-ab-frames.js`, 120 Hz, alternating arms) is in `~/code/alberttools/README.md`, last section:
  28 → 0 frames with a ≥40 % hole per arm, twice, zero variance.
- No video is attached because this is dependency alignment with no behaviour change of its own —
  the behaviour was demonstrated in #5394 and in upstream [#1493](https://github.com/petyosi/react-virtuoso/pull/1493).

### Notes for the reviewer

- **`chore:` is deliberate, and it means this rides the next release.** `react-virtuoso` is bundled
  into the published package — `vite.config.ts` externalizes only React, `@copilotkit/*`,
  `@livekit/*` and `livekit-client`, and `@factorialco/f0-react@6.86.5`'s
  `dist/experimental-CbOZz_Mw.js` already contains `deviationCommitted` inlined — so consumers get
  this only on the next release, and release-please has not cut one from a chore-only span in the
  last 400 commits on `main`. `fix(deps):` would cut one, but it is the wrong label twice over:
  `main` already ships this behaviour through the patch, so there is no user-visible fix here; and
  `check-bugfix-red-green.ts` matches `/^fix(\([^)]*\))?!?:/`, so a `fix(deps):` commit would be
  required to add a regression test that fails on `main` — which cannot exist, because `main` is
  already green on it. If Foundations would rather this reach consumers immediately, that is a
  release call, not a commit-type one.
- **Removing the patch also removes an accidental hard pin.** `patchedDependencies:
  react-virtuoso@4.18.11` froze the resolution — any drift would have failed to apply the patch.
  With it gone, `^4.18.13` will float to 4.18.14+ unreviewed. Intended, but worth knowing.
- `patchedDependencies` lives in `pnpm-workspace.yaml`, not root `package.json` — pnpm 10 no longer
  reads the `pnpm` field there for patches (#5394's body explains it). Root `package.json` is
  therefore untouched by this PR.
- The range is `^4.18.13`, not an exact pin: 87 of the 92 dependencies in
  `packages/react/package.json` use a caret (4 exact, 1 `workspace:`). 4.18.13 is the newest
  published version, so the range cannot resolve below the fix.
- Pre-existing, untouched, and out of scope, but visible from this diff: `pnpm install` warns that
  root `package.json`'s `pnpm.overrides` / `pnpm.onlyBuiltDependencies` are no longer read by pnpm 10
  — `pnpm-workspace.yaml` is their new home. The lockfile's `overrides:` block currently survives on
  inertia; a future full re-resolve could drop it and move react/react-dom tree-wide. Also
  pre-existing: since the library inlines virtuoso and neither imports nor re-exports it, the
  `dependencies` entry makes every consumer install a copy nothing uses.
- Related: #5394 (introduced the patch, merged), #4762 (relaxed the range to `^4.18.10`, merged).


dmitry-zaets (approved):

eliseo-juan (approved):
LGTM

PR #5421, opened 2026-09-07
feat(F0Select): search inside the trigger, on by default

sauldom102:
## Description

A select with no filters no longer opens a second search box above its list — the field itself takes the query. Because the field is already on screen, the search now costs nothing, so it is on by default wherever it can live there and `showSearchBox` is only needed to turn it off. Filters keep the search in the dropdown, next to the filter picker; a data source still opts in, since only the consumer's own adapter can answer a query. This implements what Ángel Moreno and I agreed in Slack on 7 Sep ("en los casos en los que no haya filtros, y se ponga la prop de search en el select, en vez de que salga fuera, que salga en el input" → "Deal?" / "Agreed").

## Type of change

## Two things to settle before merge

1. **How far "search always automatic" goes.** The agreement ends with "yo haría que el search lo tuviéramos siempre de forma automática entonces" / "Agreed". I read that as _wherever the search is free_ — i.e. filterless field selects carrying static `options`, which this component filters itself. It deliberately does **not** auto-enable for a `source`: only the consumer's adapter can implement the query, so a box we switched on for them would be a field that filters nothing. If the intent was "on for everything", it is a one-line change to `isSearchEnabled`.
2. **Marcos.** The thread ended with "pregúntale a Marcos incluso si quieres por confirmar" and there is no sign that happened. This changes how every filterless `F0Select` in the product behaves when opened, so it wants his and Ángel's sign-off, not just mine.

## Screenshots

No Chromatic diff: every snapshot story renders its select closed, and closed is unchanged. To see it, open **Components/Select → Search In The Trigger** in Storybook and click the field.

- **Closed** — unchanged: the selection with its dot tag, avatar or status pill.
- **Open** — the field becomes a focused text input, with the selected label carried into the placeholder so it still reads as set until the first keystroke replaces it.

One consequence worth a designer's eye: while open, a single selection's **avatar / tag / icon is not drawn** — the field is a text input at that point, and the label moves into the placeholder. That is the part of this I would most expect Ángel to want to change.

## Implementation details

- feat: search renders inside the trigger for `field` selects with no filters, no `asList` and no custom `children` trigger; everywhere else the box stays at the top of the dropdown, exactly as before

- feat: `showSearchBox` defaults to `true` for those selects when they carry static `options`, and to `false` (opt-in) otherwise

  <details>
  <summary>Why the split</summary>

  Static options are filtered by this component, so a search there always works. A `source` is searched through the consumer's `fetchData`; enabling a search box they never asked for would render a field that silently filters nothing. Selects with filters keep the search in the dropdown because the filter picker and the search are one query — splitting them would put half of it in the trigger and half in the popup.
  </details>

- feat: keyboard contract for the inline search — a character typed on the closed field opens the list and becomes the query, Arrow Up/Down hands focus to the list (the same handover the dropdown's search box makes), Enter picks the first option the query left standing, and Escape drops the query before it closes the list

- fix: the vendored Radix select can keep the trigger in the accessibility tree while its content is open (`retainTrigger`)

  <details>
  <summary>Why</summary>

  An open select calls `hideOthers(content)`, which aria-hides everything else — including the trigger. That is right while the trigger is only a button and wrong once it is where the user is typing: it took both the field and the combobox/listbox pairing out of the tree the screen reader reads. `hideOthers` now retains the trigger as well when the search lives there. The pre-existing debt for the `inline` variant's trigger is untouched.
  </details>

- fix: a pointer-down on the open search field no longer dismisses the list it is driving — clicking into the field is placing the caret; the chevron still closes it

- fix: focus the search input from an effect rather than React's `autoFocus`

  <details>
  <summary>Why</summary>

  `autoFocus` focuses during the commit, and around an open popup `focus` is patched by the instrumentation Storybook installs into something that schedules React work. React threw "Should not already be working" mid-commit and the story rendered nothing at all. Focusing from an effect (the pattern `F0SearchInput` already uses for the same reason) puts the call after the commit.
  </details>

- fix: an option matches on anything its row SHOWS — `label`, `description`, and a `metadata` dial code — not just its label

  <details>
  <summary>Why</summary>

  Reported while reviewing this PR against **Components/Select → With Metadata**: a row reading "Spain +34" could not be found by its dial code, and one with a `description` could not be found by the words in it. The documented contract already claimed descriptions were matched and never did, which is why the `WithSearchBox` story passes its own `searchFn` — consumers were reimplementing behavior they had been promised. Matching is a substring, so "34" finds "+34" without the plus nobody types. Metadata is matched per variant, so a new `F0SelectItemMetadata` variant stays a deliberate decision about whether it is searchable.
  </details>

- fix: the trigger button no longer draws its own native focus ring

  <details>
  <summary>Why</summary>

  `reset.css` clears the native `:focus-visible` outline for `input` and `textarea` and stops at buttons. Nothing focused that button before, but closing the dropdown now hands focus back to it, so Chrome drew its heavy blue ring inside F0's own — two rings, after every Enter. The field draws the ring from `focus-within` on its wrapper; the button just stops competing.
  </details>

- fix: Escape drops the query before it closes the list, through Radix's `onEscapeKeyDown`

  <details>
  <summary>Why</summary>

  This started on the input's `onKeyDown`, where it can never work: the dismiss runs from a listener on `document` in the CAPTURE phase, so it has already fired by the time the event reaches anything inside, and no `stopPropagation` from a descendant gets in front of it. Worth flagging that the first test of this passed anyway — pressing Escape twice closes either way, since the query is dropped on close regardless. It now asserts the two presses separately, and the behavior was confirmed in a browser.
  </details>

- test: 12 new cases — both placements, the default being on, `showSearchBox={false}` leaving a plain button, Enter, ArrowDown, the query being dropped on close, and the trigger staying out of the `aria-hidden` sweep
- test: the tests that drove the dropdown's search box now drive the trigger, in F0Select and in the three consumers that reached for `role="searchbox"` (LocationSelector, ProjectSelector, the DatePickerPopup portal test)
- docs: `showSearchBox` documents the conditional default, in `types.ts` and in the Storybook `argTypes`; new `SearchInTheTrigger`, `SearchBoxWithFilters` and `WithoutSearch` stories

## Verified

`pnpm tsc`, `pnpm lint src`, `pnpm format:check src`, and the full unit suite (7672 passed, 608 files) all green. The inline trigger was also driven by hand in Storybook: input, focus, `role="combobox"` with `aria-expanded` / `aria-controls` → the listbox / `aria-autocomplete="list"`, and typing all behave. The dropdown body does not paint in the headless browser I used (pre-existing — `WithoutSearch` on `main`'s code path is equally blank there), so list rendering is covered by the unit tests rather than by eye.

### DIFF
--- a/packages/react/src/components/F0Select/F0Select.tsx
+++ b/packages/react/src/components/F0Select/F0Select.tsx
@@ -66,14 +66,34 @@ import { SelectItem } from "./components/SelectItem"
 import { SelectTopActions } from "./components/SelectTopActions"
 export * from "./types"
 
+/**
+ * Matches a query against EVERYTHING THE ROW SHOWS, not just its label.
+ *
+ * A row that reads "Spain +34" has to be findable by "34" — a substring, so
+ * the "+" nobody types is optional — and one with a `description` by the words
+ * in it. Matching the label alone made the search quietly ignore half of what
+ * was on screen: the documented contract already claimed descriptions were
+ * matched, and consumers were passing their own `searchFn` to get back
+ * behavior they had been promised.
+ *
+ * Metadata is matched per variant rather than by walking the object, so a new
+ * `F0SelectItemMetadata` variant is a deliberate decision about whether it is
+ * searchable — not something that silently becomes part of the query surface.
+ */
 const defaultSearchFn = (
   option: F0SelectItemProps<string>,
   search?: string
 ) => {
-  return (
-    option.type === "separator" ||
-    !search ||
-    option.label.toLowerCase().includes(search.toLowerCase())
+  if (option.type === "separator" || !search) {
+    return true
+  }
+
+  const query = search.toLowerCase()
+  const metadata =
+    option.metadata?.type === "dialCode" ? option.metadata.dialCode : undefined
+
+  return [option.label, option.description, metadata].some((field) =>
+    field?.toLowerCase().includes(query)
   )
 }
 
@@ -1299,6 +1319,22 @@ const F0SelectComponent = forwardRef(function Select<
       showLoadingIndicator={!!children}
       portalContainer={effectivePortalContainer}
       retainTrigger={searchesInTrigger}
+      onEscapeKeyDown={(event) => {
+        /**
+         * Escape drops the QUERY first, and only closes the list on the second
+         * press, once there is nothing left to undo.
+         *
+         * Through Radix's own hook, never the input's `onKeyDown`: the dismiss
+         * runs from a listener on `document` in the CAPTURE phase, so it has
+         * already fired by the time the event reaches anything inside — no
+         * amount of `stopPropagation` from a descendant can get in front of
+         * it. This hook is the one place that can.
+         */
+        if (searchesInTrigger && currentSearch) {
+          event.preventDefault()
+          onSearchChangeLocal("")
+        }
+      }}
       onPointerDownOutside={(event) => {
         /**
          * With the search inside the trigger, the trigger is part of the thing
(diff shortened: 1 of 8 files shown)


eliseo-juan (approved):
LGTM

PR #5422, opened 2026-09-07
feat(F0Select): search from the trigger when there are no filters

AngelMsaa:
## Description

When a select has search on and no filters, the field itself becomes the search box: you type where the value shows, instead of a second input opening at the top of the dropdown. With filters the dropdown keeps its top row, since that row holds the filter picker and the applied-filter chips. Agreed with @sauldominguez in DM: F0Select is the combobox, so no new component.

`showSearchBox` still turns search on, exactly as on `main`. This PR only changes **where** the search field goes.

### Before
image

### After
image

## Type of change

## Where the search field goes

| Select | Where |
| --- | --- |
| Field trigger, no filters | in the trigger |
| Field trigger, `source` with filters | dropdown's top row, beside the filters |
| `variant="inline"`, `asList`, custom trigger | dropdown's top row |

A grouping selector does not move the search: the row keeps it and the field keeps the query.

## How the field behaves

- **Arriving selects the label.** Clicking in or tabbing in leaves the whole label selected, so the first keystroke is a query and not one appended to the answer already there. Settled in, the field is an ordinary text field again: a second click places a caret to edit part of the label.
- **The selected label is the field's text.** Select it, copy it, double-click it, Shift+Arrow through it: all the browser's own. Editing it turns the text into a query; the selection itself only changes when the user picks or clears, so a required form field never goes empty mid-search. Closing puts the label back, unless the user emptied the field: leaving an empty clearable field clears the selection. A multiple selection shows its count beside the caret instead, since there is no single label.
- **The caret never leaves the field.** Every text key keeps working, IME composition included. The arrows move an active option named by `aria-activedescendant` (disabled options are skipped), hovering a row moves it too, Enter takes it, Escape or the arrow glyph closes. Enter never submits a surrounding form.
- **Tab goes to the popover's controls.** Tab from the field lands on the first action or footer control, Shift+Tab comes back. Focus leaving for anywhere else closes the list.
- **The clear button is for the selection.** What the user typed stays; a single selection's label goes because it was the selection.
- **A selected option always looks selected.** Hover, focus and the arrow highlight no longer swap its colour.
- **Query timing matches the popover search.** Static options filter in the same render; a `source` is asked after the same 400ms pause the popover's box used.
- **Opening is immediate.** Typing skips the select's 100ms open debounce, which exists to resolve a close and an open arriving together. Measured, from the keystroke: list mounted 125ms → 13ms, rows on screen 145ms → 38ms.

## What this touches outside F0Select

| File | Lines | Why |
| --- | --- | --- |
| `F0InputField` | 63 | New `valueSlot` (rich content where the text would be), `canClear`, `clearKeepsText`, `aria-describedby` passthrough, and `onClickContent` now receives the click. The field's internals are private, so the selection cannot be placed before the caret from outside. All default to today's behavior. |
| `ui/Select/radix-ui/select.tsx` | 16 | One flag, default false. The open content calls `hideOthers(content)`, which aria-hides the trigger too. Right for a button, wrong for the field you are typing into. |
| `ui/Select/SelectContent.tsx` | 20 | Passes that flag through, and exposes the "tabbable controls in the popup" query it already had so the field's Tab handling uses the same rule. |
| `hooks/datasource/useData.ts` | 16 | Bug fix. The search value was mirrored into a ref by an effect, one render after the render that read it, so a query change with no render behind it (clearing on close) never refetched. Derived in render now. Every data-layer and DataCollection test passes. |
| `F0NumberInput/internal.tsx` | 2 | `F0InputField.onBlur` now hands over the focus event; the one caller forwards it. |
| 3 consumer test files | 7 | The ClockIn project and location pickers and the calendar year select pass `showSearchBox` themselves, so the placement rule moves their search box. |
| `F0LocationInput` tests and stories | 54 | Its address select passes `showSearchBox` with no filters, so its search moves into the field: the queries go to the trigger, the label is read as a value rather than as text, and the active option is read off `aria-activedescendant`. No component change. |

No consumer call sites change, and the shared option row (`ui/Select/SelectItem.tsx`) is untouched.

## Tests

232 unit tests across F0Select, F0InputField and the Select primitive (full suite: 7722 green), 55 story tests in Chromium, and a real-browser pass over Cmd+A, Shift+Arrow, copy and paste on a selected label, disabled-option skipping, Tab into the footer and back, hover, blur and the label restore on Escape. New coverage: placement per trigger kind, the keyboard contract, editing the label without touching the selection, clearing once, disabled options, Enter inside a form, Tab and blur, IME, option ids for any value, the source debounce, the announced count, and the primitive's aria-hidden exemption.

## Notes for review

- **Deliberately not shipped:** a live region announcing the result count, a pre-existing gap on the popover-search path too. Space types a space rather than picking, by design for a text field.
- **Left at the primitive level:** the trigger sits above the dropdown's click-blocking overlay through `z-50`, and `keepTriggerAccessible` is passed rather than derived from the trigger's tag. Both are the shared primitive's concern and out of this PR.
- **Known limitation:** the dropdown's rows only render once its entrance animation fires, so in a background tab the list is empty. Pre-existing, in the shared dropdown, left alone.
- **Follow-up:** `searchValue` is a dead prop. It is never destructured and ends up spread onto the primitive.
- The sorting and grouping selectors style their trigger with `[&_button]:h-8`. Nothing here breaks that, but it would break silently if a select's trigger ever became an input by default.


sauldom102 (approved):

warcos-fact (approved):
Looks good to me, but let's keep an eye on this to see if the users have trouble understanding this interaction.

PR #5426, opened 2026-09-07
refactor(lint): enable the hand-fixed lint rules

dmitry-zaets:
## Description

Turns on the remaining rules from the PENDING block and fixes their 159 sites by hand (plus 7 in react-native). The PENDING block is now empty; what is left off is listed under DEBT, OXLINT BUGS and OFF ON PURPOSE with a reason each. Behavior is preserved, except where noted in the table.

## Public API note

The "breaking changes" label comes from the API surface check, which flags any retyped member. The three changes it lists are widenings and do not break how these APIs are used:

- `onClick` on `F0DialogPrimaryAction` and `F0DialogSecondaryAction` (and the types built from them: `DialogControls`, `F0DialogActionsProps`, `ProductModal`, `F0CarouselDialogProps`) is now `() => void | Promise<void>`. Callers provide this callback, and every existing `() => void` handler still type-checks. The dialog footer already awaited a returned promise.
- `useChatHistory().refetch` is now typed `() => Promise<void>`. It always returned a promise; the type was wrong.
- `F0VersionHistoryProps.activeVersionId` is now `"current" | (string & {})`. Any string is still accepted; the form only keeps the `"current"` autocomplete hint.

No release bump is needed.

## Type of change

- [x] Refactor / internal change (no API or behavior change)

## Rules turned on

| Rule | Sites | What it enforces / what changed |
|---|---|---|
| `typescript/no-misused-promises` | 16 | No promise where a `void` callback is expected. `F0DialogPrimaryAction.onClick` and `F0DialogSecondaryAction.onClick` now accept `() => void \| Promise<void>`; other sites wrap with `void`. |
| `typescript/no-unnecessary-template-expression` | 14 | No `${"literal"}` inside templates. |
| `sonarjs/no-duplicated-branches` | 11 | Identical branches merged (OneCalendar views, KanbanLane, isDataChartEmpty, SurveyFormBuilder). |
| `typescript/no-dynamic-delete` | 10 | `delete obj[key]` replaced with object rebuilds or a `Map`. |
| `typescript/no-redundant-type-constituents` | 10 | No `never \| undefined`, `unknown \| X`; autocomplete unions use `"literal" \| (string & {})`. |
| `typescript/prefer-for-of` | 9 | Index loops become `for...of`. Media track lists iterate `Array.from(x)` because tests mock them as array-likes. |
| `typescript/prefer-promise-reject-errors` | 8 | `reject(...)` always gets an `Error`. One story mock rejected with a plain object before. |
| `typescript/use-unknown-in-catch-callback-variable` | 7 | `catch` callbacks take `unknown` and narrow. |
| `sonarjs/no-nested-template-literals` | 7 | Inner templates hoisted to a const. |
| `sonarjs/no-trivial-assertions` | 7 | `expect(true).toBe(true)` replaced with real assertions. |
| `sonarjs/no-nested-assignment` | 6 | No assignment inside an expression. |
| `sonarjs/no-all-duplicated-branches`, `public-static-readonly`, `concise-regex`, `no-small-switch`, `no-dead-store`, `no-redundant-jump` | 5, 5, 4, 3, 3, 3 | Straightforward simplifications. |
| `typescript/await-thenable` | 4 | No `await` on non-promises. `ProductModal` `Action.onClick` widened instead. |
| `typescript/no-extraneous-class`, `no-unsafe-enum-comparison`, `restrict-plus-operands` | 2, 2, 1 | Test mocks got a member; `F0GraphEdge` compares with the `Position` enum. |
| `sonarjs/no-skipped-tests` | 2 | Two `it.skip` had no recorded reason. A comment above each says so; un-skip or document them. |
| 12 more sonarjs rules with 1 or 2 sites | 18 | `prefer-native-lodash-alternative` (`lodash/flatten` to `flatMap`), `no-clear-text-protocols`, `link-with-target-blank`, `no-hardcoded-passwords` (story value renamed), `reduce-initial-value`, `prefer-promise-shorthand`, `no-duplicate-test-title` (one duplicate test removed), and others. |

One disable comment: `sonarjs/bitwise-operators` in `SelectContent.tsx`, `compareDocumentPosition` returns a bitmask.

react-native: `ban-ts-comment` (`@ts-ignore` to `@ts-expect-error`), `import/no-duplicates`, `no-dynamic-delete`.

### DIFF
--- a/packages/react/src/components/F0Select/components/SelectItem.tsx
+++ b/packages/react/src/components/F0Select/components/SelectItem.tsx
@@ -13,6 +13,9 @@ const DIAL_CODE_PATTERN = /^\+\d{1,4}$/
 const warnedDialCodes = new Set<string>()
 
 const metadataText = (metadata: F0SelectItemMetadata): string => {
+  if (metadata.type !== "dialCode") {
+    return ""
+  }
   if (
     process.env.NODE_ENV !== "production" &&
     !DIAL_CODE_PATTERN.test(metadata.dialCode) &&
--- a/packages/react/src/components/F0Select/components/SelectItem.tsx
+++ b/packages/react/src/components/F0Select/components/SelectItem.tsx
@@ -13,20 +13,17 @@ const DIAL_CODE_PATTERN = /^\+\d{1,4}$/
 const warnedDialCodes = new Set<string>()
 
 const metadataText = (metadata: F0SelectItemMetadata): string => {
-  switch (metadata.type) {
-    case "dialCode":
-      if (
-        process.env.NODE_ENV !== "production" &&
-        !DIAL_CODE_PATTERN.test(metadata.dialCode) &&
-        !warnedDialCodes.has(metadata.dialCode)
-      ) {
-        warnedDialCodes.add(metadata.dialCode)
-        console.warn(
-          `[F0Select] metadata dialCode "${metadata.dialCode}" is not a valid dial code (expected "+" followed by 1-4 digits).`
-        )
-      }
-      return metadata.dialCode
+  if (
+    process.env.NODE_ENV !== "production" &&
+    !DIAL_CODE_PATTERN.test(metadata.dialCode) &&
+    !warnedDialCodes.has(metadata.dialCode)
+  ) {
+    warnedDialCodes.add(metadata.dialCode)
+    console.warn(
+      `[F0Select] metadata dialCode "${metadata.dialCode}" is not a valid dial code (expected "+" followed by 1-4 digits).`
+    )
   }
+  return metadata.dialCode
 }
 
 export const SelectItem = <T extends string, R>({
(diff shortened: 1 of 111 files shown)


siguenzaraul (approved):

sauldom102 (approved):

siguenzaraul (approved):

packages/react/src/components/F0Select/components/SelectItem.tsx
```diff
@@ -13,20 +13,17 @@ const DIAL_CODE_PATTERN = /^\+\d{1,4}$/
 const warnedDialCodes = new Set<string>()
 
 const metadataText = (metadata: F0SelectItemMetadata): string => {
-  switch (metadata.type) {
-    case "dialCode":
-      if (
```
dmitry-zaets, line 18: Adding this one back

PR #5427, opened 2026-09-07
chore(F0GraphEdge): meet the stable DoD

albertpmz:
## Description

Closes the three stable DoD gaps on `Graph/F0GraphEdge`: 

- adds a play function 
- adds MDX docs at the `good` tier
- axe not enforced.

**Stacked on #5191.** Base branch is `fix/f0graph-accessible-tree-stable-dod`, so review that first. The axe gap cannot be closed without it, and the diff here is four files.

Every F0GraphEdge story mounts a real `<F0Graph>`, so enforcing axe puts F0GraphView's DOM under a blocking run. On `main` that fails:

```
aria-required-children  [critical]  wcag2a wcag131
Element has children which are not allowed: [role=application]
```

React Flow hardcodes `role="application"` on its wrapper, after the props spread, and on `main` that wrapper is a child of the graph's `role="tree"`. #5191 moves the tree to an empty sibling that claims its treeitems through `aria-owns`.

## Type of change

- [x] Documentation only
- [x] Other: clears `Graph/F0GraphEdge` from `stable-dod-debt.json`

Not a promotion. F0GraphEdge is already tagged `stable` and sat on the ratchet because it did not meet the mechanical DoD.

## Implementation details

**Play function.** `NonInteractive` claims that hover is opt-in per edge, so the play function tests exactly that: the panel with `onEdgeClick`/`onEdgeHover` swaps to `var(--f0-graph-edge-hover)`, the panel without handlers stays on the default stroke. Edges only enter the DOM after React Flow measures both nodes and the initial `fitView` settles, so the query sits inside `waitFor`.

**New `PathTypes` story.** It renders a parent with two children rather than one. With a single centred child, source and target fall inside `crossAxisDelta < 2`, where F0GraphEdge snaps every path type straight to avoid a one-pixel jog, and all three panels would have looked identical.

**MDX `stub` to `good`.** The old page had headings Overview, Variants, Props and Usage, none of which the doc scorer counts, so `sectionsCount` was 0 and the tier could not rise above `stub`. Added Anatomy, Guidelines with when-to-use and when-not-to-use, Accessibility, a Do/Don't pair on variant usage, and canvases for `NonInteractive` and `PathTypes`.

Three things the old page got wrong:

- The props table documented `type`. That prop does not exist. The real one is `pathType`, renamed to avoid xyflow's reserved `type` key, so anyone copying the table got silently ignored props.
- `strokeWidth` was documented with a default of `—` and "adjusts with zoom when omitted". Inside `<F0Graph>` the zoom context is always present, so the zoom-derived width short-circuits both the prop and `style.strokeWidth`. The prop only applies to a standalone mount.
- `showDot` is now documented as computed rather than as a knob. `useGraphRenderModel` spreads consumer edge data and then overwrites `showDot`, and it resolves to `false` for any edge whose source is an expanded parent.

**Unit tests, 3 to 20.** The existing file replaced `BaseEdge` with a stub that dropped its props, so nothing covered the stroke, opacity or path that the docs describe. The mock now captures them, and the tests cover the variant-to-token map, the `data.variant` fallback and prop precedence, the dimmed opacity, the sub-2px straight snap, `pathType` selection, standalone stroke width, and the end-dot marker.

## Verification

From `packages/react`:

- `pnpm tsc` clean
- `npx oxlint src/patterns/F0Graph --max-warnings 0` 0 warnings, 0 errors (`pnpm lint` is a no-op in this package)
- `npx oxfmt --check` clean
- `pnpm vitest --project=unit run src/patterns/F0Graph` 39 files, 385 tests
- `axe-probe` against a built Storybook: 4/4 stories clean
- `pnpm test-storybook src/patterns/F0Graph` 4 suites, 30 tests

The plan also flagged `target-size` and `color-contrast` as likely real-browser failures. Neither appears. 11 enabled targets measure under 24x24 CSS px and axe flagged none, because `target-size` is an ANY-of rule that `target-offset` satisfies on its own.

## Notes for the reviewer

`F0Graph.stories.tsx` failed once in a full-suite `test-storybook` run and passed on a rerun and in isolation. It takes 23s isolated and 32s in the run that failed, so it is close to the timeout under parallel load. That file belongs to #5191, not to this PR, and F0GraphEdge passed in every run.

This branch sits 83 commits behind `main`, inherited from #5191's base. It will need restacking when #5191 rebases.

### DIFF
--- a/packages/react/.scripts/stable-dod-debt.json
+++ b/packages/react/.scripts/stable-dod-debt.json
@@ -20,7 +20,6 @@
     "EmptyState",
     "Filters/FilterPicker",
     "Graph/F0GraphControls",
-    "Graph/F0GraphEdge",
     "Graph/F0GraphNode",
     "Heading",
     "Image",
--- a/packages/react/src/patterns/F0Graph/components/F0GraphEdge/__stories__/F0GraphEdge.mdx
+++ b/packages/react/src/patterns/F0Graph/components/F0GraphEdge/__stories__/F0GraphEdge.mdx
@@ -1,25 +1,36 @@
 import { Canvas, Meta } from "@storybook/addon-docs/blocks"
 
+import { DoDonts } from "@/lib/storybook-utils/do-donts"
+
 import * as Stories from "./F0GraphEdge.stories"
 
 <Meta of={Stories} />
 
 # F0GraphEdge
 
-The default edge connector between nodes in F0Graph. Adapts its visual
-treatment based on the relationship context (highlighted, dimmed, or
-default).
+The default connector between nodes in `F0Graph`. It draws the line and nothing
+else: no label, no handlers, no focus. The relationship it represents is carried
+by the tree semantics on the nodes, not by the line.
+
+> `F0Graph` mounts its own `ReactFlowProvider`, so you never wrap it yourself.
+> Every canvas below mounts a small `<F0Graph>` per panel so each edge is shown
+> in the context it actually renders in.
 
 ---
 
-## Overview
+## Anatomy
 
-F0GraphEdge is used internally by `<F0Graph>` for all edges
-unless a custom `renderEdge` is provided. It renders a smooth
-SVG path that stays visually quiet so nodes remain the primary
-focus.
+An edge is one SVG path plus an optional dot at the target end.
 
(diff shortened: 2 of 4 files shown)


eliseo-juan (approved):
LGTM

PR #5428, opened 2026-09-07
feat(F0Form): add the location field type

AngelMsaa:
## Description

Wires [F0LocationInput](https://github.com/factorialco/f0/pull/5408) into F0Form, so a form can declare an address field with `f0FormField.location()` and get the structured value back. #5408 has landed, so this branch is now rebased on `main` and its diff is only the F0Form plumbing.

```tsx
const schema = z.object({
  address: f0FormField.location({
    label: "Office address",
    manualEntry: true,
    allowedCountries: ["es", "pt", "fr"],
    searchPlaces: (query, { country }) => places.autocomplete(query, country),
    resolvePlace: (id) => places.details(id),
  }),
})
```

## Type of change

- [x] New pattern (composition of existing F0 components)

- Linked #f0-support thread: pending
- Phase exit criteria met: follows the `phone` field type precedent one for one

## Implementation details

- feat: `f0FormField.location()` with the same overload shape as `phone` (`optional: true` returns the optional schema), `F0LocationFieldConfig`, `F0LocationField`, `LocationFieldRenderer`, the `"location"` arm in `renderFieldInput` and `useSchemaDefinition`, and the barrel exports
- feat: validation lives in the shortcut, because an all-optional object parses `{}` happily

  <details>
  <summary>Why?</summary>

  A required field must carry a non-empty address (a country-only value counts as filled), and `requireResolved` additionally demands a picked place, for consumers that feed a map or a geofence rather than accepting typed text. Both issues are localized through `zodErrorMap` (`forms.validation.location.empty` / `.unresolved`) and overridable with `emptyMessage` / `unresolvedMessage`. The refinement runs on the optional schema so it also sees the `undefined` the component emits for an empty value; otherwise a cleared required field fell through to the generic "required" copy and a consumer's `emptyMessage` was silently ignored.
  </details>

- test: the config, both validation modes including the `undefined` a cleared field emits, the optional path, the localized messages, and a form that submits a typed address and asserts the structured value reaches `onSubmit`
- docs: a row in the F0Form field-type table and a Location field in the `All field types` story, with an in-memory Places provider so it needs no API key


packages/react/src/patterns/F0Form/fields/location/types.ts
```diff
@@ -0,0 +1,67 @@
...
+  searchPlaces?: (
+    query: string,
+    context: F0LocationSearchContext
+  ) => Promise<F0LocationSuggestion[]>
+  /**
+   * Resolves a picked suggestion into a full value with coordinates. Without
+   * it the suggestion label is kept as the address line and the value stays
+   * unresolved.
+   */
+  resolvePlace?: (id: string) => Promise<F0LocationInputValue | undefined>
```
sauldom102: We should improve this object type so a dev is not allowed to pass both `manualEntry` as true while adding as well `searchPlaces` or `resolvePlace` as it's completely useless.
Additionally, should we always allow inputting it manually when the search returns no results? Maybe it's something we want to restrict with another prop

sauldom102 (approved):

warcos-fact (approved):

packages/react/src/patterns/F0Form/fields/location/types.ts
AngelMsaa: Both done in f98a426. The props are a union now, so `manualEntry: true` sets `searchPlaces` and `resolvePlace` to `never` and the combination no longer compiles; this config reuses that union (`F0LocationInputShapeProps`) instead of re-declaring the props, so the rule holds at both levels. On the second point you're right that it shouldn't always be allowed: the empty-search escape hatch is now `manualEntryFallback` (default `true`), and in F0Form `requireResolved` turns it off on its own, since typing an address can never satisfy it and the offer would be a dead end.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

PR #5441, opened 2026-09-08
chore(charts): deprecate RadarChart in favour of F0DataChart

albertpmz:
## Description

`RadarChart` claimed to be stable and never was. It is the only story in `kits/Charts` carrying a maturity tag at all, and that tag arrived as a side effect of [#4110](https://github.com/factorialco/f0/pull/4110), a PR about the interactive legend, rather than as a decision, six months after we had already hidden the kit from the sidebar for being deprecated.

So this PR does not pay down the debt on the component, it removes a claim that was never true: the three JSDoc tags the deprecation policy requires, a story tag that says `deprecated`, and a migration guide pointing at `F0DataChart`, which is what the rest of F0 already draws its radars with.

## Type of change

- [x] Deprecation (adds `@deprecated` + `@removeIn` + `@migration`)

### The deprecation markers

Both the public `RadarChart` export and the `RadarChartProps` type now carry `@deprecated`, `@removeIn 7.0.0` and `@migration`, the three tags the policy requires together. The export is tagged in the kit barrel because that is the path consumers import through, twice over: `src/f0.ts` for the main entry and `src/experimental.ts` for the deprecated one. Tagging the type as well means the warning fires on a type-only import too.

> [!WARNING]
> The version in the `@removein` tag is provisional, as the ≥90-day SLA must be aligned with this version.

### The story tag

`stable` becomes `deprecated`. `no-sidebar` stays. Every sibling chart has it, [#4812](https://github.com/factorialco/f0/pull/4812) put it there deliberately, and removing it would surface a dead component in the sidebar.

One consequence to know about: `component-status-build.mjs` skips any `no-sidebar` story that is not `stable`, so the component now leaves the component-status dataset entirely, exactly like its nine siblings. 

### The migration guide

The two components model radar data along opposite axes. `RadarChart` keys values by series inside each row, as `data[].values[seriesKey]`. `F0DataChart` keys them by axis position, as `indicators[]` in order with each series carrying a parallel `data: number[]`. The guide's mapping table covers that transpose row by row, with a worked before/after example.

No codemod, deliberately. The transpose is mechanical but three props have no successor:

| Prop | Why it is gone |
| --- | --- |
| `scaleMin` | Radar axes always start at 0. A chart that used a non-zero floor to spread its values apart will read flatter, and the author has to decide whether to rescale the data. |
| `aspect` | Size the chart through its container instead. |
| `defaultHiddenSeries` | The legend starts fully visible. `onLegendSelectionChange` only reports state, it does not set it. |

Each of those needs a judgment call about how the chart should read, so the guide gives steps rather than a command.

## Evidence

Nothing internal depends on the deprecated component. `F0AnalyticsDashboard` already imports its radar indicators, series and skeletons from `@/kits/F0DataChart`. A grep across `src` finds exactly one importer, the kit's own `exports.ts`; every other `RadarChart*` hit is `RadarChartConfig`, `RadarChartSkeleton` or `ChatDashboardRadarChartConfig`, all of which belong to `F0DataChart`.

`pnpm check:stable-dod --verbose` goes from 41 stable-tagged components with 27 in debt, to 40 with 26, and reports no stale entry. `pnpm tsc`, `oxlint`, `oxfmt --check` and `pnpm vitest --project=unit run src/kits` (73 files, 847 tests) are all green, as is the full pre-commit suite.

### DIFF
--- a/packages/react/.scripts/stable-dod-debt.json
+++ b/packages/react/.scripts/stable-dod-debt.json
@@ -14,7 +14,6 @@
     "Avatars/AvatarPerson",
     "Avatars/AvatarTeam",
     "BigNumber",
-    "Charts/RadarChart",
     "Checkbox",
     "DatePicker",
     "EmptyState",
--- a/packages/react/docs/migrations/f0-radarchart-to-f0datachart.md
+++ b/packages/react/docs/migrations/f0-radarchart-to-f0datachart.md
@@ -1,0 +1,125 @@
+# Migrating from `RadarChart` to `F0DataChart`
+
+**Status:** deprecated in `v6.89.0` · removed in `v7.0.0` (no earlier than 90 days after deprecation)
+**Owner:** `@factorialco/f0-devs` · questions in [#f0-support](https://factorialteam.slack.com/archives/C082ZNKS403)
+
+## What changed, in one line
+
+> `RadarChart` — the last tagged component of the deprecated recharts-based `kits/Charts` kit — is replaced by `F0DataChart` with `type="radar"`, the echarts-based chart the rest of F0 already runs on.
+
+## Finding every usage
+
+```bash
+# imports (main entry and the deprecated experimental entry)
+rg "\bRadarChart\b" frontend/src -l
+rg "RadarChartProps" frontend/src -l
+# JSX call sites
+rg "<RadarChart" frontend/src -l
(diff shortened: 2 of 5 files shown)


eliseo-juan (approved):
LGTM

PR #5444, opened 2026-09-08
feat(F0Drawer): add Profile panel prototype story

igorsafonov-gif:
## Description

**What this PR is.** A working prototype of the employee profile drawer that the AI Insight "Employee summary" opens from a report, built as a Storybook story. **The new panel in Factorial should be developed from this design, not from this code:** the stories are the spec for layout, content, states and copy, while the code behind them is mock data and story-local components meant to be thrown away. The two f0 additions below are the only pieces meant for reuse. The `Menu` story puts the three variants behind one menu — opened from a people report, from an absences report, and from a worked hours report — so the shared skeleton (header → verdict → summary tiles → sections → contacts → workplace → teams) can be compared across the three.

The absences panel is the most worked-through one: one period (last 12 months), summary tiles, a "Compared to team average" block, an accordion of leave types that opens to each type's records, upcoming time off and balances in the time off page's accrued / available / taken terms. It uses the Time and Attendance cube's `absenteeism_rate` definition (absence hours over planned hours, all approved leave) rather than an invented planned/unplanned split. Design notes: `~/code/notes/dashboards/absences-profile-panel-shape.md` (local).

**Component changes are limited to what the panels render with.** Two additions, both exercised by the absences panel's leave-type accordion:

- `F0Accordion` items: `summary` (header slot that stays visible while collapsed) and `content` (rich body); `description` becomes optional when `content` carries the body.
- `DataList.RecordItem`: title · description · `detail` tag · item action. A `navigate` row drops its trailing chevron so the tag keeps the right edge. `DataList` is now exported from the package so the item is reachable from Factorial.

An earlier revision also shipped a `drawer` action (DataList + DetailsItem), `action` on the tag items, a `progress` bar on `RecordItem`, `id` on `DetailsItemType` and a `data-accordion-item` hook. None of those are used by the panels, so they were removed rather than fixed (they were the subject of review points 2–6).

## Type of change

- [ ] New component (aligned in #f0-support, lands in `experimental/`)
- [x] New pattern (composition of existing F0 components)
- [ ] Bug fix
- [ ] Documentation only
- [x] Other: Storybook prototype (`experimental` tag) meant as a hand-off spec for the product implementation

- Linked #f0-support thread: —
- Phase exit criteria met: prototype only; the story is the design artefact for the AI Insight employee summary drawer

Story: `Drawer / Profile panel / Menu` (pick "Absences report"). Hosted Storybook for this branch: see the "Visual review for your branch is published" bot comment on this PR.

## Implementation details

- **Story** (`F0Drawer/__stories__/ProfilePanel.stories.tsx`): all data is inline mock data. Summary tiles are plain `<button>`s (a real name and focus ring; `F0Card` cannot be clickable without a title) and scroll to their section. Every list passes `itemUrl` so its rows are links.
- **F0Accordion**: `summary` renders before the actions/chevron and is not part of the toggle target; `content` renders under `description`. `description` going optional is a required→optional change (SAFE per `check-api-surface`); the two Factorial consumers (`employee_lifecycle` TaskCard, `ats` Competencies) never read it as required.
- **DataList**: `RecordItem` plus a `content` slot and `hideChevron` on `ItemContainer`/`NavigateAction` for it. Tests for `RecordItem` added.
- **Definition of Done gate:** the preflight treats the new story title as a new component and asks for play function, snapshot story, docs tier and axe. This is a prototype story (`experimental` tag, no exported API of its own), so the PR carries the `skip-new-component-dod` label. If the panel graduates into a component, it goes through the full DoD then.
- Verified locally: `pnpm lint`, `pnpm format:check` and `pnpm check:untranslated-copy` clean on all changed files, `tsc --noEmit` clean, `vitest` green for F0Accordion and DataList. axe over the whole document (the drawer portals out of `#storybook-root`, so CI's axe never sees it) reports 0 serious violations on the Details, Absences and Worked hours panels; before this revision it reported 8 / 6 / 6 (`aria-command-name`, `scrollable-region-focusable`).

## Review follow-ups not in this PR

- Placement under `sds/Profile/` instead of `F0Drawer/__stories__`, and the DoD `play` for the `Menu` story.
- Promotion verdict computed from `REVIEWS`/`GOALS` instead of typed in (flips the box to its amber state).
- `check:new-component-dod` cwd/pathspec bug: yes please to the separate PR.

---
> - factorial-pr
> - factorial-f0


zygisS22 (commented):
Ran the branch locally (Storybook, unit tests, test-runner, axe, lint, tsc, format, API surface, DoD checks) and clicked through all three panels in Chrome. The story looks great and the three variants share one clean `ProfilePanel` seam. The component additions are typed strictly, use native buttons with correct ARIA, and every automated gate is green. But a few things need fixing before a product team can hook this up, and the green CI is partly vacuous (see the last section).

## Must fix

**1. `DataList` is not exported from the package.**
[`experimental/exports.ts`](https://github.com/factorialco/f0/blob/d1874b686/packages/react/src/experimental/exports.ts#L34) exports `DetailsItem` and `DetailsItemsList` but not `./Lists/DataList`. So `DataList.RecordItem`, `DrawerActionType`, `RecordItemProps`, `RecordDetail`, `RecordProgress` and the new tag-item props are unreachable from Factorial frontend except structurally through `DetailsItemContent`. The F0Accordion MDX this PR adds shows `<DataList><DataList.Item … /></DataList>` under a `@factorialco/f0-react` import, which a consumer cannot compile.

**2. Drawer open state is positional and the content entries have no identity.**
[`DetailsItem/index.tsx#L229`](https://github.com/factorialco/f0/blob/d1874b686/packages/react/src/experimental/Lists/DetailsItem/index.tsx#L229) keeps `Set<number>` keyed on array index; React keys and the `aria-controls` id are positional too. A screen that builds `content` conditionally (`[baseRow, ...(canSeeSalary ? [salaryRow] : []), notesRow]`) shows the wrong row expanded after a permission/filter flip and the consumer cannot fix it from outside. F0Accordion in this same PR keys on `item.id`; same shape here (optional `id` on content entries, `Set<string>`) would solve it.

**3. The `drawer` action is half-controlled.**
- `expanded` without `onToggle` type-checks and yields a row that can never close (the JSDoc says "required", the type does not).
- `onToggle()` carries neither the next value nor the row, so "lazy-load on open" or analytics forces full control.
- No `defaultExpanded`, so "start open" (deep link) also forces full control.

Suggest the Radix/F0Accordion shape: `expanded` / `defaultExpanded` / `onExpandedChange(next: boolean)` on a discriminated union, and `console.warn` on `expanded` without a handler per `AGENTS.md`.

**4. Revealed rows render after the whole list, not under their row.**
[`DetailsItem/index.tsx#L290-L343`](https://github.com/factorialco/f0/blob/d1874b686/packages/react/src/experimental/Lists/DetailsItem/index.tsx#L290-L343): the `AnimatePresence` block is a sibling *after* the `<ul>`. Fine while every DetailsItem has one content entry (all stories/tests), wrong for `content={[rowA(drawer), rowB]}`, where rowA's nested rows appear below rowB. The DataList story does it right (an `<li>` inside the list).

**5. `DetailsItemsList` keys rows by `title`, which now decides drawer state.**
[`DetailsItemsList/index.tsx#L62`](https://github.com/factorialco/f0/blob/d1874b686/packages/react/src/experimental/Lists/DetailsItemsList/index.tsx#L62). Duplicate or empty titles (the PR itself anticipates label-less record rows) collide and share/reset state. The `id` you added is the natural key.

**6. Hardcoded English in an aria-label.**
[`RecordItem.tsx#L61`](https://github.com/factorialco/f0/blob/d1874b686/packages/react/src/experimental/Lists/DataList/items/RecordItem.tsx#L61): `` aria-label={`${title} progress`} ``. `check:untranslated-copy` misses it because it does not read inside template literals. `progress.label ?? title`, or an i18n key.

**7. 20 serious axe violations in the story (`aria-command-name`).**
All from the summary tiles at [`ProfilePanel.stories.tsx#L539`](https://github.com/factorialco/f0/blob/d1874b686/packages/react/src/components/dialog-alike/F0Drawer/__stories__/ProfilePanel.stories.tsx#L539). `F0Card` always renders its title row and puts `role="button"` + `aria-label={title}` on it when `onClick` is set (`CardInternal.tsx:386-398`), so a title-less clickable card is a nameless, non-focusable button. The `-mt-2` only hides the empty row. Details 8, Absences 6, Worked hours 6. CI never sees them because axe is scoped to `#storybook-root` and F0Drawer portals out. Your own `VerdictDisclosure` in the same file is the right pattern (real `<button>`, `focusRing()`, `aria-expanded`).

## Should fix

- **Placement.** A 2540-line domain prototype under `components/dialog-alike/F0Drawer/__stories__` registers a "Drawer/Profile panel" entry in the Foundations-owned zone, and the status scan inherits F0Drawer's `__tests__` and MDX by directory adjacency, so it reads as tested and documented when it has neither. `where-it-goes.mdx` and `f0-component-contribution` route single-screen compositions to `sds/<area>/` or the product repo.
- **The DoD skip was moot and its stated grounds are not the gate's criterion.** Forcing the check yields: play function, snapshot story, docs "good" tier, axe enforced. Phase 3 requires the first and axe passing regardless of exported API.
- **The promotion verdict contradicts its own panel.** `PROMOTION_CONDITIONS` (L226-235) assert "Avg. score: 4.1" and "87%" while the drawer renders 3.6 / 5 and goals 90/65/50 with one review at 2.6 in critical. Everything else in the file is computed from data; this block is typed in.
- **The story never uses the API it motivated.** No `action: { type: "drawer" }` anywhere; `DetailsItemType.id` unused; scrolling is hand-rolled with `getElementById`. The story description says "a Summary card scrolls to its row and opens it", but nothing opens. (The scroll itself works, I verified with real clicks on every card.)
- **`RolesCollection` lacks `itemUrl`** unlike every sibling list, so its scroll container trips `scrollable-region-focusable` (serious).
- **Public API check will label this PR `breaking`** for an api-extractor alias rename in `ChartWidgetEmptyState` (`Props_6` → `Props_5`, byte-identical interfaces) that the PR never touches. Noise, but the label will appear.
- Smaller: `max: 0` renders `NaN%` in `RecordProgressBar`; `memo(DrawerAction)` never hits because `actionFor` builds a fresh closure every render; `BalanceTagItem` was left out of the `action` widening; `F0AccordionItem.description` going optional is a source-breaking type change for anyone reading `item.description.length` (and an item with neither `description` nor `content` renders an empty body silently); `DetailsItem` has no MDX so the drawer action is documented only in a story description; the DataList MDX snippet points `aria-controls` at an id that does not exist in the snippet; story meta uses `Meta<typeof F0Drawer>` instead of `satisfies Meta`, `cn()` skipped at L559/L564, `HOURS = 60` is minutes per hour, `[&_.rounded-xl]:rounded-none` at L728 reaches into another component's internals.

## Repo bug found while checking the gates

`check:new-component-dod` is inert for every PR, not just this one. [`check-new-component-dod.ts#L192`](https://github.com/factorialco/f0/blob/d1874b686/packages/react/.scripts/check-new-component-dod.ts#L192) runs `git diff … -- packages/react/src` with `cwd: PKG_DIR` (= `packages/react`); git pathspecs are cwd-relative, so it matches nothing and always reports "No new components in this PR". `BARREL_PATHSPECS` has the same prefix. Happy to open a separate issue/PR for it.

## What is good

`WithDetailsItemAction<T>` is a correct distributive `Omit` with the reason spelled out; discriminated unions stay exhaustive end to end, no `any`; `aria-controls` is set only while the panel is mounted and that is asserted in a test; `useReducedMotion` wired on both reveals; tests assert behaviour, not snapshots; lint (347 rules incl. the new ones), tsc and format are clean; dark mode renders fine. The three-variant `ProfilePanel` seam with an explicit per-report axis is the actual design deliverable and it reads well. Contacts, Workplace and the figure blocks would port to the frontend as-is.

PR #5449, opened 2026-09-08
feat(F0Checkbox): stabilizing checkbox, adding description prop

AngelMsaa:
Checkbox was tagged `stable` but sat below the stable bar: its label was hand-rolled markup that ignored the type scale, and it had no way to explain an option beyond its title. This brings the label onto `F0Text`, adds a `description` prop, and closes the last two Definition-of-Done requirements, so Checkbox now genuinely clears the bar it claimed.

## Before
image

## After
image
image

## The label is an F0Text label

It was a bare `<label>` rendering its text at whatever weight and colour it inherited, so it read lighter than every other label in the system. It now goes through the same `Text` component that backs `F0Text variant="label"`.

| | before | after |
|---|---|---|
| weight | inherited (400) | `font-medium` (500) |
| size | inherited | `text-base` (14px) |
| colour | `text-current` | `text-f1-foreground` |
| left padding | `pl-2.5` (10px) | `pl-2` (8px) |

The required asterisk now comes from `Text`'s own `required` prop instead of a hand-rolled `<span>`, matching how form labels render it everywhere else.

## New: `description`

A second line under the title, styled as `F0Text variant="description"`.

```tsx
<F0Checkbox
  title="Share usage data"
  description="Helps us decide which reports are worth keeping."
  checked={checked}
  onCheckedChange={setChecked}
/>
```

Three things worth knowing about how it behaves:

- **It is inside the `<label>`**, so clicking the description toggles the checkbox.
- **It reaches assistive tech through `aria-describedby`, not the label text**, so it never leaks into the checkbox's accessible name. A screen reader still announces "Share usage data, checkbox" and only then the description. A consumer-supplied `aria-describedby` wins.
- **`hideLabel` hides it along with the title**, and drops the `aria-describedby` with it.

With a description the row aligns to the top instead of the centre, and the label takes a 2px top offset so the title lines up with the *visible* 20px square rather than the top of the 24px hit area (`after:top-0.5`). Measured in the browser: title box and visible square share the same top and centre to the pixel.

## Two decisions to sanity-check

**`Text` rather than the public `F0Text`.** `F0Text` omits `className`, and the label needs its own layout and disabled classes; `ui/checkbox.tsx` already reaches into `components/F0Icon`, so importing the sibling `ui/Text` is in keeping with the file. Same component and same variants either way, so the two cannot drift. One deliberate difference: `F0Text` defaults `markdown` to `true`, and going through `Text` leaves it off, so an existing checkbox whose title contains `*` or `_` keeps rendering as literal text instead of turning into emphasis.

**`TextProps` gains `htmlFor`.** `as="label"` was already an allowed tag on `Text` (and on the public `F0Text`), but `React.HTMLAttributes` does not carry `htmlFor`, so a `Text` could not actually label a control. Added as a documented optional prop; additive.

## Definition of Done

Two mechanically-checked stable requirements were unmet, and both are now closed:

- **Do's and don'ts** in the docs, one of the criteria for the "good" doc tier. The pair covers the mistake the component invites now that `description` exists: a title phrased as a question, or negated, so ticking the box reads as a double negative.
- **Blocking axe.** The stories move off the global `test: "todo"` default to `a11y: { test: "error" }`, so a WCAG 2.0-2.2 A/AA violation in any Checkbox story fails CI instead of warning into the job summary.

Those were the last two gaps, so Checkbox comes off `.scripts/stable-dod-debt.json`. The ratchet fails if a listed component starts passing, which is what makes that permanent.

## Retroactive impact

Any checkbox with a visible title gets a heavier label. Inside f0 that is a short list, because almost every internal consumer passes `hideLabel` (row, table, filter and select-all checkboxes are all icon-only): `F0Form` checkbox fields and the survey checkbox questions built on them, `F0QuestionCard`, and a few stories.

**The colour change is the one to look at on review.** A checkbox on a non-default background used to inherit its label colour and now pins to `f1-foreground`. I found no such usage in f0, but a consumer could have one.

`F0Switch` has byte-identical bare-label markup and the same `pl-2.5` (`ui/switch.tsx`). Deliberately left alone to keep this scoped to Checkbox, so the two are inconsistent until a follow-up. Happy to extend it here if you would rather they move together.

## Testing

- Full unit suite green. `F0Checkbox.test.tsx` gains 5 description tests: renders, the `aria-describedby` pairing, absent without a description, hidden by `hideLabel`, and click-to-toggle. One existing assertion updated, since the title is now a `<span>` inside the `<label>` and the `for` check reads `getByText(...).closest("label")`.
- `pnpm test-storybook` against a local Storybook: 7 Checkbox stories pass with axe **blocking**, including the new `WithDescription` and its play function asserting the `aria-label` / `aria-describedby` split.
- `pnpm check:stable-dod`: ratchet holds, Checkbox among the components meeting the full DoD.
- `pnpm tsc`, `pnpm lint`, `pnpm format` clean.
- Geometry verified in the browser rather than by eye: 8px gap on every row, and exact top/centre alignment on the description rows.

### DIFF
--- a/packages/react/.scripts/stable-dod-debt.json
+++ b/packages/react/.scripts/stable-dod-debt.json
@@ -15,7 +15,6 @@
     "Avatars/AvatarTeam",
     "BigNumber",
     "Charts/RadarChart",
-    "Checkbox",
     "DatePicker",
     "EmptyState",
     "Filters/FilterPicker",
--- a/packages/react/src/components/F0Checkbox/F0Checkbox.tsx
+++ b/packages/react/src/components/F0Checkbox/F0Checkbox.tsx
@@ -9,6 +9,13 @@ interface CheckboxProps extends DataAttributes {
   title?: string
 
   /**
+   * A secondary line of text rendered under the title, for context the title
+   * cannot carry on its own. Hidden along with the title when `hideLabel` is
+   * set, and exposed to assistive technology as the checkbox's description.
+   */
+  description?: string
+
+  /**
    * The id of the checkbox
    */
   id?: string
@@ -73,6 +80,7 @@ interface CheckboxProps extends DataAttributes {
 
 function _F0Checkbox({
   title,
+  description,
   onCheckedChange,
   id,
   disabled,
@@ -89,6 +97,7 @@ function _F0Checkbox({
   return (
     <CheckboxRoot
       title={title}
+      description={description}
       onCheckedChange={onCheckedChange}
       id={id}
       disabled={disabled}
--- a/packages/react/src/components/F0Checkbox/__stories__/F0Checkbox.mdx
+++ b/packages/react/src/components/F0Checkbox/__stories__/F0Checkbox.mdx
@@ -22,13 +22,26 @@ indeterminate.
 
 ## Anatomy
 
-A checkbox is composed of the selection control itself plus a `title` label.
-Always provide a meaningful `title` so the control is clearly labelled, both
-visually and for assistive technology. The component is controlled — provide
-`checked` and update it from the `onCheckedChange` callback.
+A checkbox is composed of the selection control itself plus a `title` label, and
+optionally a `description` under it. Always provide a meaningful `title` so the
+control is clearly labelled, both visually and for assistive technology. The
+component is controlled — provide `checked` and update it from the
+`onCheckedChange` callback.
 
 <Canvas of={CheckboxStories.Default} />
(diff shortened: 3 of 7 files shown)


warcos-fact (approved):
I think it looks better than before in isolation.
Still don't know if radio buttons have the same margins, but that is something we can review in the future.

sauldom102 (approved):

PR #5450, opened 2026-09-08
test(F0Map): assert the MapAdapter port with one battery per adapter

factmarc10010:
**Stacked on #5447** → #5446 → #5445 → #5440 → #5438 → `main`. Review the parents first; the diff below is this PR's commit alone.

Closes [FCT-63106](https://factorialmakers.atlassian.net/browse/FCT-63106).

## What

One battery of cases asserting the `MapAdapter` port, registered by each adapter's own test file. A second engine implements the port and registers the same cases — it does not bring a second battery.

| | |
|---|---|
| `providers/__tests__/adapterContract.ts` | the cases + the harness they need | +195 |
| `providers/maplibre/__tests__/engineStub.ts` | the MapLibre harness | +213 |
| `providers/maplibre/__tests__/contract.test.ts` | registers them | +66 |

## Three of the ticket's cases are not here, on purpose

The ticket lists eight cases from the spike. Three of them measure the *engine's* geometry, not the adapter's translation of the port, and jsdom has no engine — MapLibre needs WebGL, so the suite runs the real adapter against a stub. Written at this level they could not fail:

- **`project()` pixel stability** — the stub returns whatever pixel it is told to.
- **A camera read mid-flight** — needs an engine that interpolates; a stub jumps.
- **`centerInFreeRegion({ right: N })`** — **no such member on the port.** Framing today is a symmetric `gutter` (`toPadding` puts the same value on all four sides), so the case cannot be written without widening the API first.

They are recorded in the suite's own docblock rather than as green tests that assert nothing. The first two want a browser-level test with real WebGL; the third wants a port change, and I would rather do that when the Google adapter actually needs asymmetric framing than invent the shape now.

## The stub reproduces two structural facts, so the cases can fail

A harness that agrees with itself proves nothing, so the stub mirrors the engine where the contract depends on it:

- **The canvas and the markers are siblings.** MapLibre binds `click` on the canvas and hosts markers beside it, so a click on a pin never reaches the map handler. Flattening this would make the marker-click case pass no matter what the adapter did.
- **`project()` answers only once loaded.** That makes "ready means projecting" a real ordering instead of an assumption.

## Red-green

Every case was watched failing against a deliberately broken adapter, then passing again with it restored (`git diff` on the adapter empty at the end):

| Mutation | Cases that went red |
|---|---|
| `toBounds` drops the first vertex | frames every coordinate; reframes to the next ones |
| `destroy` skips `map.remove()` | removes every node on destroy |
| `camera()` drops the requested zoom | ends at the requested zoom |
| `project()` fakes a point when the engine returns none | ready before projecting; removes every node |
| `addDomMarker` never calls `addTo` | marker click; style swap; destroy |
| `setGlobeProjection` inverts the flag | switches to the globe projection |

## Capability-aware skipping

`splitByCapabilities` reads the capabilities off the adapter under test, so a case gated on one it does not declare is skipped **by name** in the output. The globe-projection case uses it today; on Google (`globeProjection: false`) it will skip visibly rather than silently vanish.

## Not in this PR

- **#5381's engine stub.** The ticket's third AC asks for it to be absorbed or deleted. That PR is not merged, so its +268-line stub is not in this tree — `engineStub.ts` is the place it should live, but pointing 5381 at it is an edit in 5381, alongside the `getMap` → `getNativeMap` rename it already needs.
- **`F0Map.test.tsx`'s own hoisted mock.** It still stubs the engine itself. Folding it onto `engineStub.ts` would cut real duplication, but it is a 471-line file the ticket does not ask about, and it would put the component suite at risk inside a test-only PR.

## Testing

```
vitest run src/patterns/F0Map   →  3 files, 34 tests passed
oxlint --type-aware src/patterns/F0Map --max-warnings 0  →  0 errors
tsc --noEmit  →  no F0Map diagnostics
```

[FCT-63106]: https://factorialmakers.atlassian.net/browse/FCT-63106?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


sauldom102 (approved):
Could you combine next time tests along with its proper code on the same PR?

PR #5455, opened 2026-09-08
feat(F0Map): select the rendering engine through an async provider registry

factmarc10010:
**Stacked on #5450** → #5447 → #5446 → #5445 → #5440 → #5438 → `main`. Review the parents first; the diff below is this PR's commit alone.

Closes [FCT-63107](https://factorialmakers.atlassian.net/browse/FCT-63107).

## What

The engine becomes selectable: a `provider` prop, an app-level default on `F0Provider`, and each adapter behind its own dynamic `import()`.

| | |
|---|---|
| `providers/names.ts` | the engine name, as a leaf module | +12 |
| `providers/registry.ts` | one dynamic import per engine, memoised | +27 |
| `lib/providers/map/` | the `F0Provider` `map` block + `useMapProvider` | +87 |
| `F0Map.tsx` | creation goes async | 161 changed |
| `package.json` | engines move to optional peers | |

Resolution order is `provider` prop → `F0Provider` `map.provider` → `maplibre`. The library default stays MapLibre deliberately: it is keyless, so it cannot depend on someone having registered a billing account, and every Chromatic story render would bill on a metered engine.

## Two consequences worth knowing before review

**The imperative handle is no longer live in the same tick as the render.** The engine arrives a microtask later, so `ref.current.focusMarker(...)` called synchronously after mount now does nothing where it used to work. Six existing tests were doing exactly that and had to wait first, act second. Real callers drive these from user interaction rather than from mount, so I did not add a command queue for it — but it is a behaviour change, not just a test change, and it is the thing to push back on if you disagree.

**The dot's z-order is kept by a different mechanism.** `CurrentLocationLayer` anchors itself under the bottom line layer *when that layer exists*, and otherwise just goes in first. With the engine landing late, the located dot can now win the race against the lines, so the `before:` anchor gives way to plain insertion order. Both orders keep the dot beneath the lines, so the invariant holds; the test now asserts that invariant instead of the mechanism it used to assume.

## A defect I introduced and then fixed

The first cut had `void loadMapAdapterFactory(...).then(...)` with no `catch`, so an engine chunk that fails to load left the map hanging forever with no fallback and no diagnostic — the rejection was swallowed. It now falls back to the list (same outcome as a machine without WebGL) and logs, because a failed chunk is a deployment problem rather than a capability the visitor lacks.

## Red-green

`useMapProvider`'s resolution chain, each path broken separately:

| Mutation | Case that went red |
|---|---|
| hook ignores the context | takes the app-level engine from the provider |
| hook ignores the per-map override | lets one map override the app-level engine |

The provider tests deserve a note: the union has one member until the Google adapter lands, and a single value cannot tell "reads the context" apart from "returns the constant" — the four cases would all have passed against a hook that ignored both inputs. They use a second engine name, cast in that test file and nowhere else, so each path is actually observable.

Also fixed a latent flake the new timing exposed: `MockMarker` captured its own index and removed by `splice(index)`, so removing one marker shifted every later marker's slot out from under it. Now held by reference. Suites run green three times in a row (39 tests).

## What I did not do, and why

- **`provider: "google"` is not in the type.** The ticket writes the prop as `"maplibre" | "google"`, but the Google adapter does not exist until FCT-63108. Exposing a value the registry cannot resolve would turn a compile error into a runtime one, so the union stays at `maplibre` and 63108 widens it.
- **No `apiKey` on the `map` block.** It has no consumer until Google, and I would rather add it beside the code that reads it.
- **The bundle AC is unverified here.** "A consumer using only the table visualization ships neither engine" depends on #5381's `React.lazy` boundary, which is not merged, so I cannot measure it in this tree.
- **The monorepo side is untouched.** Setting `map={{ provider: "google" }}` in `frontend/src/modules/core/lib/f0-provider.tsx` needs a published f0 and an API key, so it follows 63108.

## Testing

```
vitest run src/patterns/F0Map src/lib/providers/map  →  4 files, 39 tests passed (x3)
oxlint --type-aware (F0Map + lib/providers) --max-warnings 0  →  0 errors
tsc --noEmit  →  one error, pre-existing on the parent branch
                 (src/sds/chat/F0Chat/utils/emoji-support.ts, `fontFamily` from f0-core)
```

The lockfile diff carries two unrelated transitive patch bumps (`semver`, `postcss`) that the re-resolve materialises; the map engines are the only intentional change.

[FCT-63107]: https://factorialmakers.atlassian.net/browse/FCT-63107?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/react/src/lib/providers/f0/f0-provider.tsx
```diff
@@ -120,6 +121,11 @@ export const F0Provider: React.FC<{
    * ```
    */
   formComponent?: F0FormLikeComponent
+  /**
+   * App-level map engine, so switching provider is one change at the root
```
albertpmz: Let's rename the variable to `mapEngine`, that way we can delete this comment as the variable name will be self-explanatory.

packages/react/src/lib/providers/map/MapProvider.tsx
```diff
@@ -0,0 +1,24 @@
+import { createContext, useContext, useMemo } from "react"
+import type { F0MapProvider } from "@/patterns/F0Map/providers/names"
+import { DEFAULT_MAP_PROVIDER } from "@/patterns/F0Map/providers/names"
+
+export interface MapContextValue {
+  /** Engine every `F0Map` uses unless it names its own. Keyless by default. */
```
albertpmz: Same as above, we can remove the comment – the var name is self-explanatory

packages/react/src/patterns/F0Map/__tests__/F0Map.test.tsx
```diff
@@ -116,8 +116,11 @@ const mock = vi.hoisted(() => {
     zoomOut() {
       this.calls.zoomOut++
     }
-    project() {
-      return { x: 0, y: 0 }
+    // Coordinate-dependent, so clustering and label collision behave like
```
albertpmz: Too long, consider shortening/removing it

packages/react/src/patterns/F0Map/__tests__/F0Map.test.tsx
```diff
@@ -162,21 +165,30 @@ const mock = vi.hoisted(() => {
     }
   }
   class MockMarker {
+    // By reference, not index: a removal shifted every later marker's slot.
```
albertpmz: Remove the comment

packages/react/src/patterns/F0Map/__tests__/F0Map.test.tsx
```diff
@@ -390,23 +411,29 @@ describe("F0Map", () => {
...
       await waitFor(() =>
         expect(map.layers.has("f0-current-location")).toBe(true)
       )
       expect(map.sources["f0-current-location"]).toBeDefined()
-      // Inserted below the bottom line layer, so lines and pins stay above it.
-      expect(map.layerInsertions.at(-1)).toEqual([
-        "f0-current-location",
-        "f0-map-lines-solid",
-      ])
+      // Below the lines either way: anchored under them when they already
```
albertpmz: Consider removing this comment too

packages/react/src/patterns/F0Map/providers/names.ts
```diff
@@ -0,0 +1,7 @@
+/**
+ * A leaf module on purpose: `F0Provider` and the registry both need the name,
```
albertpmz: Consider removing this one too

packages/react/src/patterns/F0Map/providers/registry.ts
albertpmz, line 1: Consider removing the comments here

packages/react/src/patterns/F0Map/F0Map.tsx
```diff
@@ -157,6 +159,8 @@ export interface F0MapProps extends WithDataTestIdProps {
    * zoom in - best for a world-scale view. Changing it re-projects live.
    */
   projection?: F0MapProjection
+  /** Defaults to the `F0Provider` `map.provider`, then to `maplibre`. */
```
albertpmz: We should document props with what they are/do, not with their default value

packages/react/src/patterns/F0Map/F0Map.tsx
```diff
@@ -418,66 +424,100 @@ const F0MapBase = forwardRef<F0MapHandle, F0MapProps>(function F0Map(
...
-      adapter = createMaplibreAdapter({
-        container,
-        style: styleRef.current,
-        center: viewport.center,
-        zoom: viewport.zoom ?? DEFAULT_VIEWPORT.zoom,
-        minZoom,
-        maxZoom,
-        interactive,
-        cooperativeGestures: gestureHandling === "cooperative",
+    // Behind a dynamic import, so the teardown has to cope with an unmount
```
albertpmz: Remove this one too

packages/react/src/patterns/F0Map/F0Map.tsx
```diff
@@ -418,66 +424,100 @@ const F0MapBase = forwardRef<F0MapHandle, F0MapProps>(function F0Map(
...
+            container,
+            style: styleRef.current,
+            center: viewport.center,
+            zoom: viewport.zoom ?? DEFAULT_VIEWPORT.zoom,
+            minZoom,
+            maxZoom,
+            interactive,
+            cooperativeGestures: gestureHandling === "cooperative",
+          })
+        } catch {
```
albertpmz, line 447: Code is self-explanatory, let's remove this one

packages/react/src/patterns/F0Map/F0Map.tsx
```diff
@@ -418,66 +424,100 @@ const F0MapBase = forwardRef<F0MapHandle, F0MapProps>(function F0Map(
...
+          // The engine could not start (no WebGL): fall back to the list view.
+          setWebglFailed(true)
+          return
+        }
+        wire(adapter)
+      })
+      .catch((error: unknown) => {
+        // A failed chunk is a deployment problem, not a missing capability,
```
albertpmz: Same here, code is self-explanatory

packages/react/src/patterns/F0Map/F0Map.tsx
```diff
@@ -418,66 +424,100 @@ const F0MapBase = forwardRef<F0MapHandle, F0MapProps>(function F0Map(
...
+        setAdapterInstance(null)
+        adapter.destroy()
+      }
     }
-  }, [loading, interactive, gestureHandling, minZoom, maxZoom, shouldFit])
+
+    function wire(created: MapAdapter) {
+      adapterRef.current = created
+      setAdapterInstance(created)
+      // A previous run may have failed (and set the list fallback) with props
```
albertpmz: We can clean this one too

packages/react/src/patterns/F0Map/F0Map.tsx
```diff
@@ -418,66 +424,100 @@ const F0MapBase = forwardRef<F0MapHandle, F0MapProps>(function F0Map(
...
+      // that made creation throw; this run succeeded, so clear it.
+      setWebglFailed(false)
+
+      // Errors before the map is ready mean the style/tiles failed to come up -
```
albertpmz: This one we can also clean it

packages/react/src/patterns/F0Map/F0Map.tsx
```diff
@@ -418,66 +424,100 @@ const F0MapBase = forwardRef<F0MapHandle, F0MapProps>(function F0Map(
...
+            })
+          }
+          created.setGlobeProjection(projectionRef.current === "globe")
+        }),
+        created.on("error", () => {
+          if (!ready) {
+            setTileError(true)
+          }
+        }),
+        // Background click clears the selection (marker clicks are DOM events
```
albertpmz: Same here

albertpmz (changes requested):
Let's check the comments below, they're mostly style–wise

albertpmz (approved):

warcos-fact (approved):
I don't think there's much design to review here.

PR #5460, opened 2026-09-08
feat(F0Map): generate the Google cartography from the f0 tokens

factmarc10010:
> **Stacked on #5458**, which targets `main`; every other parent (#5438 → #5455) is merged. The diff below is this PR's own commits — the cartography — and it cannot merge until #5458 does.
>
> No public surface change: `googleMapStyles` and `defaultMapStyle()` are new exports, and nothing existing changes shape.

https://github.com/user-attachments/assets/69157840-4541-449b-8da6-162774d4666b

## The ticket asked whether this was worth generating at all

It is, for **14 of the flavor's 25 cartographic roles**.

I checked Google's style reference rather than guessing: `featureType` comes from a fixed vocabulary (`administrative`, `landscape`, `poi`, `road`, `transit`, `water` and their documented children), `elementType` from `geometry[.fill|.stroke]` / `labels[...]`, and stylers are flat — `color`, `lightness`, `saturation`, `visibility`, `weight`. **No zoom-dependent or data-driven expressions exist**, and there is no way to add a layer the basemap omits.

So eleven roles have no expression and keep Google's own colours:

| role | why not |
|---|---|
| `worldLand` | the base land is zoom-interpolated on MapLibre; Google has no zoom styler at all, so the near-view colour wins |
| `path` | no `featureType` for footways or cycleways |
| `wood` | none distinct from `landscape.natural.landcover` |
| `residential`, `industrial`, `sand`, `farmland`, `rock`, `glacier` | no landuse `featureType`s exist |
| `building`, `buildingLine` | no building `featureType`; `landscape.man_made` is the nearest approximation and covers more than buildings |

**The two maps will not match.** What they share is the source of truth: both derive from the same `saturatedFlavor(theme)`, so a token change lands as a reviewable diff on both rather than silently drifting on one. That was the whole point of preferring generation over a hand-clicked Cloud style.

A Google map with no `mapStyle` now gets this pair by default.

## A mismatched style now falls back rather than blanking

#5458 made a mismatched provider tag render the map **unstyled**. It now falls back to that engine's own default, still logging the mismatch — a wrong style is a caller error, not a reason to render nothing.

The existing test's assertion changed with it, deliberately: `expect(map.opts.style).toBe(f0MapStyles.light)` where it used to expect `undefined`.

## A wrong diagnosis of mine, corrected in this PR

The first commit here claimed the committed MapLibre pair was no longer reproducible because upstream Bright had moved, and split the emitters on that basis. **That was wrong**, and the follow-up commit says so.

Measured properly:

- the two JSON documents compare **equal** with keys sorted — same 122 layers, same sprite, same values;
- running `oxfmt` over the freshly generated file produces a file **byte-identical** to the committed one.

The whole-file diff is the formatter. The script emits plain `JSON.stringify`, the pre-commit hook reformats every staged `.json`, so the committed form has never matched the emitted one. A regeneration looks enormous and means nothing until it is formatted.

Splitting `--google` from `--maplibre` still earns its place — the Google pair needs only f0 tokens and should not reach for the network — but that is the reason, not a drift that does not exist. **There is no upstream-Bright decision to make and no ticket to open.**

## Testing

- New `defaultMapStyle` suite: each engine gets a pair tagged for itself (the tag is what `F0Map` checks before handing a style over, so a default carrying the wrong one would be rejected by its own consumer), both halves present and distinct, and the Google pair is shaped as Google's model rather than a `StyleSpecification` — every `featureType` root drawn from the documented vocabulary.
- Verified load-bearing: pinning `defaultMapStyle` to `f0MapStyles` fails with `expected 'maplibre' to be 'google'`.
- `F0Map` + map-provider + styles suites: **51 pass, 1 skipped**.
- `pnpm tsc`: 1 error, pre-existing (`F0Chat/utils/emoji-support.ts`).
- `format:check`, `check-cycle-dependencies`, `check-file-sizes` clean.

Regenerating the Google pair twice is idempotent.

**The flake reported on #5458 is fixed here, and it was not timing noise.** The stubbed `project()` returned `{ x: 0, y: 0 }` for every coordinate, so every point landed on the same pixel and the clustering pass grouped all of them on its first frame; the marker test only passed in the window before that. The pre-push hook caught it with the message that explains it — `expected [ [ 2.17, 41.405 ] ] to deeply equal [ [ 2.19, 41.4 ], [ 2.15, 41.41 ] ]`, where `[2.17, 41.405]` is the centroid of the two points. Projecting from the coordinate fixes it, and unlocks the label-collision and selection-collapse branches that were equally unreachable. Twelve consecutive runs, no failures.

## Looked at, and redesigned because of it

Reviewed in the product on both themes, and the first version did not survive it. Mapping MapLibre's roles onto Google's featureTypes one for one carries decisions that only hold when MapLibre's other layers are there:

- **every dark surface collapses onto one step**, because contrast is supposed to live in the buildings and the linework. Google draws neither, so `landscape`, `landcover` and `poi.park` were all `#192231` — a dead plane with no parks;
- **local streets are pure white**, because a darker casing separates them. Google's casing was *lighter* than the ground, so streets vanished;
- **dark water is de-blued** to a whisper of white over the page, which only reads when the whole surface palette is flat. Here it was near-black.

So Google gets its own flavor from the same tokens, composed for what Google actually renders:

```
dark    landscape #0d1626   landcover #26372b   park #364c2f   water #26426d
light   highway   #bac1cd   arterial  #cdd3da   local #ffffff on a #cdd3da casing
```

Also caught an inversion I had introduced: with the arterial a step lighter, local streets read as *more* prominent than the arterials they feed, because a local's casing was darker than an arterial's fill.

The goal was settled explicitly as **on-brand on its own terms**, not a copy of MapLibre — which is what makes composing a separate flavor the right shape rather than a workaround.

Confirmed in the product afterwards.


factmarc10010:
Folded into #5458 rather than reviewed separately.

Both needed the same two approvals, and a Google map without its cartography is not something a designer can meaningfully sign off — split, the design review would have had to approve a render that never ships. The three commits here are unchanged and now sit at the top of #5458, which targets `main` and closes both FCT-63108 and FCT-63109.

Nothing is lost: `git log` on `feat/f0map-google-adapter` carries them verbatim.

PR #5468, opened 2026-09-08
refactor(lint): keep the command palette's state updaters out of nested callbacks

dmitry-zaets:
`pnpm lint` fails on `main` with 3 errors. This fixes them.

## What happened

#5461 turned on `sonarjs/no-nested-functions` at depth 5. The command palette (#5403) merged while #5461 was open and added three call sites past that depth. Neither PR's CI saw the combination, and the quality workflow only runs on `pull_request`, so main itself is never linted. The failure lands on the next PR anyone opens: it is already failing the `[⚛️ REACT] Linting` job on at least two open PRs.

## The fix

The two `setState` updaters inside `.then` and `.catch` sat five callbacks deep. They are pure state transitions, so they move to module level as `withAnswer` and `withFailure`. The effect now reads `setState((current) => withAnswer(current, type, answer))`. No behavior change: same objects, same fields, same order.

## Why `refactor:` and not `fix:`

Nothing behaves differently: the same state objects are built from the same fields in the same order. There is no regression test to write, which is also what the red-green gate says when the commit is typed `fix`. What was broken is the lint gate, not the component.

## Also

That file had a literal NUL character where `join("\0")` was meant. Git treats a file containing NUL as binary, which is why this diff renders as `Bin` instead of lines. It is an escape now, so diffs of this file are readable from here on.


eliseo-juan (approved):
LGTM

siguenzaraul (approved):

PR #5472, opened 2026-09-08
fix(ai-canvas): add BulkResolveInconsistenciesCanvasContent to the canvas union

neil-krichi:
## What

Adds `BulkResolveInconsistenciesCanvasContent` to the canvas `CanvasContent` union, following the exact shape of #5033 (type-only, additive, no renderer by design — the canvas registry is host-supplied, so this only unblocks `openCanvas({ type: "bulkResolveInconsistencies", … })` type-checking on the factorial side).

The content carries scope identity only (`periodStart`, `periodEnd`, optional `scopeInconsistencyIds`) — the proposal rows live in host state, per the write-once canvas-content convention.

Consumer: Factorial's Time domain is moving One's bulk time-tracking-inconsistency resolution from in-chat cards to a canvas Data Collection (monorepo PR to follow, pinned on this PR's alpha build).

### DIFF
--- a/packages/react/src/kits/ai/F0AiChat/index.ts
+++ b/packages/react/src/kits/ai/F0AiChat/index.ts
@@ -24,6 +24,7 @@ export type {
   DashboardCanvasContent,
   DataDownloadCanvasContent,
   AutofillTimesheetCanvasContent,
+  BulkResolveInconsistenciesCanvasContent,
   F0AIMessage,
   F0AiChatWelcomeCard,
   F0Message,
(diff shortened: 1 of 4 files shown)


constaloz (approved):
Looks good. Non-blocking cross-repo follow-up: factorial#113379 currently declares `periodStart`/`periodEnd` as `string | null` and its frontend action parameters are optional, while this public contract requires strings. factorial-agent#2831 always emits `startOn`/`endOn`, so the stricter F0 contract is sound; when removing the temporary consumer cast, please tighten or guard the local type, including legacy replayed messages.

eliseo-juan (approved):
LGTM

PR #5489, opened 2026-09-09
feat(F0ResourceHeader): accept a md as the description

carmatute:
## Description

A resource almost always has context living somewhere else — the doc it is planned in, the dashboard it is measured from, the ticket that opened it — but `description` only took a `string`, so consumers either flattened their markup away (Trainings does exactly this: `DOMPurify.sanitize(description, { ALLOWED_TAGS: [] })`) or gave up and moved the description out of the header into the page body, losing the title/description/metadata reading order the header exists to provide. This widens `description` to `ReactNode` so it can carry a link, and fixes the two accessibility problems that widening exposes.

## Type of change

- Linked #f0-support thread: **not yet raised — this draft is the artifact to bring to that conversation.** `F0ResourceHeader` is stable, so this needs Foundations sign-off before it is more than a draft. Opening it first per the "bring whatever exists: … a draft PR" guidance in the contribution docs.

No new visual states for existing callers — a `string` description renders exactly as before. The new `WithRichDescription` story is included in the Chromatic `Snapshot` story, so the added path is covered there.

## Implementation details

- feat: widen `BaseHeaderProps["description"]` from `string` to `ReactNode`. `F0ResourceHeader` picks the prop through, so its public API widens with it. The component already rendered the value as children, so nothing but the type changed for existing callers.

- fix: expand a clamped description when focus enters it

  <details>
  <summary>Why?</summary>

  The two-line clamp hides overflow with `overflow: hidden`, which clips content visually but leaves it in the tab order and the a11y tree. While `description` was a `string` the clamped-away region held nothing focusable, so this could not happen; a link in it means a keyboard user lands on a control that is nowhere on screen (WCAG 2.4.7 Focus Visible). The clamp box is also a scroll container, so the browser scrolls line 1 out of view with no way to bring it back — the outer wrapper is `overflow-clip` and cannot compensate.

  Guarded on `needsTruncation`, so a description short enough to fit is left alone rather than sprouting a "show less" toggle over nothing.
  </details>

- fix: give the show all/show less toggle disclosure semantics — `aria-expanded`, `aria-controls`, the `type="button"` it always needed (it would submit an enclosing form), and the shared `focusRing()`. Mirrors `ExpandDescriptionButton` in `sds/Home/Communities`. This matters more now the collapsed region can hold content the user needs to reach.

- fix: honour `useReducedMotion()` on the expand/collapse transition, per `packages/react/AGENTS.md`.

- test: four unit tests — a string description still renders, a node description keeps its link in the a11y tree and exactly once (the measure clone stays out), and focus on an unclamped description raises no toggle. Each was checked red before green.

- test: a `RichDescriptionFocusExpands` play story for the focus-expands behaviour. It cannot be a unit test — Vitest stubs `ResizeObserver` with a no-op, so `needsTruncation` is never true under jsdom and the clamp never engages.

- feat: restyle the show all/show less toggle as a label plus a chevron rather than underlined text, at design's request. The chevron is `aria-hidden` (the button is named by its own label) and rotates only when motion is allowed.

- docs: MDX guidance for the new shape, including the constraint below.

## Notes for the reviewer

**`check:api-surface` will flag this as breaking.** `.scripts/check-api-surface.ts` classifies any property whose type changed as BREAKING, so CI will post that comment. It is a widening — `string` is assignable to `ReactNode` — and no consumer needs to change. I traced every one: `F0DialogHeader` passes a `resourceHeader` object through but only reads `.title`; `kits/surveys/SurveyAnsweringForm` `Omit`s `description` and declares its own `string`. Nothing does `.length`, interpolation, or passes it to a string-typed parameter. Please don't cut a major for it.

**Open question — F0 has no inline-prose link.** `actionVariants` gives every variant, `link` included, a base of `inline-flex items-center gap-1 text-base font-medium`, and the `link` variant overrides neither the display nor the size. Inside the description's `text-lg`, an `F0Link` therefore renders a step small (16px in 18px copy) and as a flex box rather than inline, which `-webkit-box` line-clamping does not lay out as part of the text. The story still uses `F0Link` because that is what F0's own markdown renderer uses for inline links (`kits/ai/…/markdownRenderers/components/Link.tsx`), so the snapshot shows the mismatch rather than hiding it. Worth deciding whether the fix is an inline link variant, or size-inheriting link styles in prose contexts. Happy to follow up separately either way.

**Known, left alone deliberately:**

- The description node is rendered twice — once visibly, once in a hidden copy that measures the unclamped height. Harmless for a string; with a node, an `id` inside it would be duplicated and every IDREF would resolve to the hidden copy, which comes first in document order. Documented in the MDX as a constraint for now. Replacing the clone with `scrollHeight > clientHeight` on the single visible node would make the whole hazard class impossible — a good follow-up, but a behaviour change to the truncation detection and beyond this PR.
- The measure clone is safe today because Tailwind `invisible` removes it from focus navigation, not because of `aria-hidden`. That makes the class load-bearing now the slot can hold focusable content. `inert` would make the guarantee explicit and CSS-independent, but React is pinned at 18.3.1 here, where `inert` is not a clean boolean prop.
- When expanded, the description is a `max-h-80 overflow-y-scroll` region that is not keyboard-scrollable (WCAG 2.1.1, axe `scrollable-region-focusable`). Pre-existing, and unrelated to the type.

### DIFF
--- a/packages/react/src/patterns/F0ResourceHeader/index.stories.tsx
+++ b/packages/react/src/patterns/F0ResourceHeader/index.stories.tsx
@@ -601,7 +601,8 @@ export const RichDescriptionFocusExpands: Story = {
     await expect(toggle).toHaveAttribute("aria-expanded", "false")
 
     // Tabbable while the clamp hides it, so focus has to expand (WCAG 2.4.7).
-    canvas.getByRole("link", { name: "interview rubric" }).focus()
+    // F0Link appends an sr-only "(opens in new tab)" to the accessible name.
+    canvas.getByRole("link", { name: /interview rubric/ }).focus()
 
     await waitFor(() => expect(toggle).toHaveAttribute("aria-expanded", "true"))
   },
--- a/packages/react/src/patterns/F0ResourceHeader/index.stories.tsx
+++ b/packages/react/src/patterns/F0ResourceHeader/index.stories.tsx
@@ -1,6 +1,7 @@
 import type { Meta, StoryObj } from "@storybook/react-vite"
 import { ComponentProps } from "react"
-import { expect, fn, userEvent, within } from "storybook/test"
+import { expect, fn, userEvent, waitFor, within } from "storybook/test"
+import { F0Link } from "@/components/F0Link"
 import { PrimaryDropdownAction } from "@/experimental/Information/utils"
 import * as Icon from "@/icons/app"
 import { Archive, Comment, Download, ExternalLink, Pencil } from "@/icons/app"
@@ -20,7 +21,8 @@ const meta: Meta<typeof F0ResourceHeader> = {
       description: "Main heading identifying the resource",
     },
     description: {
-      description: "Supporting text providing additional context",
+      description:
+        "Supporting text providing additional context. Accepts a node, so it can carry inline formatting such as a link",
     },
     status: {
       description: "Visual indicator of the resource's current state",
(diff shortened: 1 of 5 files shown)


packages/react/src/patterns/F0ResourceHeader/index.stories.tsx
```diff
@@ -562,6 +564,50 @@ export const WithLongDescription: Story = {
...
+    ...Default.args,
+    description: (
+      <>
+        Owns the hiring bar for the design org. See the{" "}
+        <F0Link href="https://example.com/rubric" target="_blank">
+          interview rubric
+        </F0Link>{" "}
+        before scheduling a loop.
+      </>
+    ),
```
sauldom102: I would avoid having description not being a `string`, it would open the door to weird use cases and other devs introducing wrong things here, we could accept links with some markdown-formatted text instead

carmatute: Agreed, and done in 0652cf2 — `description` goes back to `string` and is rendered as markdown through `F0RichTextDisplay`, so the public type is now byte-identical to `main` and the only thing a caller can express is inline formatting.

Two things that got better as a result: the API surface check has nothing to flag any more, and the hazard I'd had to document — the description is rendered twice, once in a hidden node that measures its unclamped height, so a node could have smuggled in duplicate `id`s and IDREFs resolving to the hidden copy — is structurally impossible with a string.

The a11y fixes in the PR still apply, since a markdown link inside the clamp is focusable just the same: focus entering a clamped description now expands it (the clamp hides overflow but leaves it in the tab order — WCAG 2.4.7), and the show all/show less toggle gained `aria-expanded`/`aria-controls` plus the `type="button"` it was missing.

One behaviour change worth a second opinion: existing string descriptions now go through a markdown parser, so a description containing `*`, `_` or `[` renders differently than before. Harmless for ordinary prose, and I've documented inline-only in the MDX, but say the word if you'd rather it were opt-in via a prop.

sauldom102 (approved):

desiree-np (approved):
Design approval. Checked the rendered header on the branch: the markdown link renders inline and inherits the description's size, and the show all/show less toggle reads correctly as a disclosure.

PR #5496, opened 2026-09-09
refactor(lint): cut cognitive-complexity findings from 190 to 110

dmitry-zaets:
Takes `sonarjs/cognitive-complexity` from 190 findings to 114, and shrinks the ratchet's baseline with it.

## What is exempt, and why

Three groups are not ours to reshape. Each says so in `.oxlintrc.json`:

| Files | Findings | Why |
| --- | --- | --- |
| Stories, tests, mocks (`**/mocks/**` and `**/__mocks__/**` join the existing override) | 9 | Their score is the size of a data table, not logic. `mockData.tsx` scores 75 because it *is* a fixture table. Same treatment `no-identical-functions` already has there. |
| `**/components/radix-ui/**` | 2 | The select primitive extracted from @radix-ui, kept diffable against the upstream commit its README pins. |
| `**/*.mjs` | 1 | Hand-run generators. `buildStyles.mjs` is one row per map colour, each picking a light or dark value. |

## What was refactored

65 functions: the ones the team edits most, and the whole 16-20 band. Threshold stays at the rule's default of 15.

The shapes that came up, most often first:

- **JSX that carried branches became a component.** Not a `renderX()` helper — a local function returning JSX has no component identity, cannot be memoized, does not appear in the React tree, and re-runs with every render of its parent. 24 new components came out of this: `ToastActions`, `CalendarHeader`, `MetadataHoverCard`, `RailActionGlyph`, `FileDropzone`, `DocumentThumbnail`, `ReorderableItems` and the rest. Where the JSX was already a plain `const x = (<jsx/>)` value it stayed one — a const holding an element is fine; only the function was the problem.
- **A ternary chain dispatching on a discriminant became one `if` per case.** `content === "person" ? … : content === "team" ? … : null` reads better as a component that returns early.
- **Repeated derived expressions hoisted.** `("options" in config && config.options) || ("source" in config && config.source)` appeared twice in `f0Schema`.
- **One switch replaced four sequential `if (kind === …)` blocks** in `partitionChatAttachments`. sonarjs counts a whole switch as 1 and each `case` as 0, so this is the cheapest large reduction available.
- **Key handlers split per gesture**, each returning whether it consumed the event — the convention `handleEmojiAutocompleteKeyDown` already used. `ChatComposer` and the collection `Search`.
- **Four calendar handlers collapsed onto one.** `handleMonthClick`, `handleQuarterClick`, `handleHalfYearClick` and `handleYearClick` were the same algorithm over different period units. They now share `rangeAfterPeriodClick` in the new `granularities/periodClick.ts`, which also took the four copies of `isDateRange`. "The same period" is "the same period start", which is why one function covers all four granularities.

Where a split needed a value AND a subtree, it became two things rather than one awkward one: `listPlaceholderState()` decides which state the filter list is in and `InFilterListPlaceholder` draws it; `lockedSectionItems()` finds a locked section's extent and `LockedSection` draws it.

## Four functions left alone on purpose

Three of them — `Table.tsx`'s summary row, `NestedRow.tsx`'s trailing rows and `WidgetContainer.tsx`'s drag context — each close over a dozen values, and the table pair drags seven generic parameters along with them. Every seam I tried there produced a component whose props were an unreadable list and whose name did not describe anything, which is the signal that the seam is wrong rather than that the props are unavoidable.

The fourth is `EntitySelect/Trigger`'s label, and it is a different reason worth knowing about. **`F0InputField` clones its child** (`F0InputField.tsx:505`, and its own docs say so) to inject the field's props: the label span is supposed to come out as `role="combobox"` with the `aria-controls` / `aria-expanded` / `aria-busy` wiring, a generated id and merged classes. Extract that span into a component and `cloneElement` hands those props to the component, which ignores them — the span keeps its own `role="button"` and the trigger loses its combobox role entirely. So an `asChild` slot or a `cloneElement` consumer needs an element there, not a component, full stop.

All four files are **byte-identical to `main`** and the ratchet records their findings instead (Table 24, WidgetContainer 21, NestedRow 19, Trigger 17). That is what the baseline is for: they get fixed by whoever next opens those components, with the context to pick a real boundary.

## One more, from a commit that landed while this was open

#5471 added `locateMentions` at 16, after #5469 measured its baseline — so `pnpm check:lint-debt` is red on `main` right now. It is in the band this PR clears, so it is cleared here: the grouping and the scanning phases became `groupByPattern` and `findOccurrences`, moved verbatim. Its 279 tests pass.

## Two things found on the way

- **`inferFieldType` had an unreachable branch.** Its `ZodArray` case re-checked `options`/`source`, but the check above it already returns `"select"` for exactly that condition. Removed.
- **`MonthView` and `YearView` each had a `date-fns` import nothing used** once their handlers moved out (`isSameMonth`, `isSameYear`).

## What is left

114 findings: everything above 30, the four files above, and the files nobody has edited in six months. Refactoring code nobody reads is risk without payoff — the ratchet catches those when someone next opens the file. The hot-and-severe ones (`F0Select.tsx:157` at 85, `NewHomeLayout/index.tsx:707` at 129, `Table.tsx:136` at 55, …) want one PR each with the component's owner.

## Verification

`pnpm lint` 0 errors, `tsc` clean, `pnpm format:check` clean, 618 test files / 7831 tests pass, `pnpm check:lint-debt` reports 114 against a baseline of 114.

Two traps worth knowing about, because both cost failing tests:

- **Moving an `await`ing loop into an `async` helper adds a microtask** even when the loop body never awaits. `useWizardNavigation`'s tests call `goToStep` inside a synchronous `act()` and rely on `changeStep` landing in the same tick. The guards were extracted instead and the loop left where it is, with a comment saying why.
- **A component in a `cloneElement` slot silently drops the injected props** — the `EntitySelect/Trigger` case above. The unit suite renders that component and never queried its role; only the Storybook play-test, which asserts a single button in the closed trigger, caught it.

### DIFF
--- a/packages/react/src/patterns/OneDataCollection/visualizations/collection/Table/Table.tsx
+++ b/packages/react/src/patterns/OneDataCollection/visualizations/collection/Table/Table.tsx
@@ -505,176 +505,6 @@ export const TableCollection = <
       ? i18n.status.selected.singular
       : i18n.status.selected.plural
 
-  type SummaryRowData = NonNullable<typeof summaryData>
-
-  const renderSummaryCellValue = (
-    column: (typeof columns)[number],
-    summary: SummaryRowData
-  ) => {
-    const placeholder = getSummaryPlaceholder(column.summaryPlaceholder)
-
-    if (
-      column.summary &&
-      source.summaries &&
-      source.summaries[column.summary]?.type === "sum"
-    ) {
-      const summaryValue = summary.data[column.summary]
-
-      if (isEmptySummaryValue(summaryValue)) {
-        return (
-          <span className="text-f1-foreground-secondary">{placeholder}</span>
-        )
-      }
-
-      return (
-        <div className="flex gap-1">
-          <span className="text-f1-foreground-secondary">
-            {i18n.collections.summaries.types.sum}
-          </span>
-          {`${summaryValue}`}
-        </div>
-      )
-    }
-
-    return <span className="text-f1-foreground-secondary">{placeholder}</span>
-  }
-
-  const renderSummaryRow = (summary: SummaryRowData) => (
-    <TableRow
-      className={cn(
-        summary.sticky &&
-          "sticky bottom-0 z-30 bg-f1-background shadow-[0_-1px_0_0_var(--f1-border-secondary)] hover:bg-f1-background",
-        "font-medium"
-      )}
-    >
-      {source.selectable ? (
-        <TableCell width={checkColumnWidth} sticky={{ left: 0 }}>
-          {summary.label ? (
-            <div className="font-medium text-f1-foreground-secondary">
-              {summary.label}
-            </div>
-          ) : null}
-        </TableCell>
-      ) : null}
-      {columns.map((column, cellIndex) => (
-        <TableCell
-          key={`summary-${String(column.label)}`}
-          firstCell={cellIndex === 0}
-          width={column.width}
-          sticky={getStickyPosition(cellIndex)}
-          highlighted={!!column.highlighted}
-          className={cn(
-            isEditableTable &&
-              (cellIndex !== columns.length - 1 || showItemActions) &&
-              "border-0 border-r-[1px] border-solid border-f1-border-secondary",
-            collapsingCellClasses.get(getColumnId(column))
-          )}
-        >
-          {cellIndex === 0 && !source.selectable && summary.label ? (
-            <div className="font-medium text-f1-foreground-secondary">
-              {summary.label}
-            </div>
-          ) : (
-            <div
-              className={cn(
-                column.align === "right" ? "justify-end" : "",
-                "flex",
-                tableCellContentClassName
-              )}
-            >
-              {renderSummaryCellValue(column, summary)}
-            </div>
-          )}
-        </TableCell>
-      ))}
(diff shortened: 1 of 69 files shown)


packages/react/src/patterns/OneDataCollection/visualizations/collection/Table/Table.tsx
```diff
@@ -505,6 +505,176 @@ export const TableCollection = <
...
+            </div>
+          ) : (
+            <div
+              className={cn(
+                column.align === "right" ? "justify-end" : "",
+                "flex",
+                tableCellContentClassName
+              )}
+            >
+              {renderSummaryCellValue(column, summary)}
```
sauldom102: Do you think we should rather have `<SummaryCellValue column={column} summary={summary} />` with that component defined in a separate file? I'm seeing this pattern repeated across the code

siguenzaraul (approved):

packages/react/src/patterns/OneDataCollection/visualizations/collection/Table/Table.tsx
dmitry-zaets: Very good catch! We should avoid using "render" functions! Gonna fix it, thank you very much!

PR #5499, opened 2026-09-09
feat(F0Chat): mark messages drafted with One

dmitry-zaets:
## Description

Adds `aiAssisted` to `F0ChatMessage` so a message drafted with One and approved by its sender shows a muted "drafted with One" label in the per-message meta cluster ("drafted with One · edited · 22:14"), on every surface that shows the message, and reaches assistive technology through the existing sr-only twin. Needed by Factorial Chat's AI-assisted direct messages (factorialco/factorial#113479): the product rule requires the marker wherever the content renders, never as body text.

## Type of change

Storybook: `Domain specific / Communications / F0Chat → Snapshot`, the "Sender surface palette" own bubble shows the full cluster; `ApplicationFrame → EverythingChannel` has two Marcus messages with the marker, one of them edited. On a narrow bubble the origin folds onto its own line and "edited · time" stays together.

## Implementation details

- feat: add `aiAssisted?: boolean` to `F0ChatMessage`, mapped by hosts from server-set message metadata
- feat: render "drafted with One" in `ChatMessageMeta` as the first part of the meta cluster, hidden on deleted tombstones; i18n key `chat.aiAssisted`, product name untranslated
- fix: render the sr-only meta twin on attachment-only messages, which had none, so the marker and "edited" are announced there too

  <details>
  <summary>Why?</summary>

  The twin lived only in the text bubble. An attachment-only message skips the bubble, so its meta was `aria-hidden` with no readable copy. `ChatMessageAttachments` now renders the twin after the content exactly when the bubble is absent, so nothing is announced twice.
  </details>

- fix: let the meta cluster wrap between the origin and the nowrap "edited · time" tail instead of overflowing a narrow bubble
- test: ChatBubble, ChatMessageMeta, F0Chat, announcement and seed tests for the new field, the ordering, the tombstone, the twin, and the wrap; Snapshot palette message and everything-stress seed carry the marker
- docs: MDX props row, behavior rule and screen-reader note
- 
image

### DIFF
--- a/packages/react/src/lib/providers/i18n/i18n-provider-defaults.ts
+++ b/packages/react/src/lib/providers/i18n/i18n-provider-defaults.ts
@@ -641,6 +641,9 @@ export const defaultTranslations = {
     edit: "Edit",
     editing: "Editing",
     edited: "edited",
+    // Muted label on a message drafted with One and approved by its sender.
+    // "One" is the product name and stays untranslated.
+    aiAssisted: "Drafted with One",
     cancelEdit: "Cancel edit",
     saveEdit: "Save",
     // Shown as the quoted sender's name when the replied-to message is your own.
--- a/packages/react/src/patterns/ApplicationFrame/index.stories.tsx
+++ b/packages/react/src/patterns/ApplicationFrame/index.stories.tsx
@@ -781,7 +781,8 @@ export const Default: Story = {
  * `grp-everything-stress` — a year of history, ten extra pages, every message
  * shape (albums of 1/2/3/4/7, a 1:10 tower, a dimensionless photo, video,
  * voice, location, pdf/sheet/docx/text cards, file chips, link previews,
- * replies, mentions, reactions, edits, deletions, failures, system rows) and a
+ * replies, mentions, reactions, edits, AI-assisted drafts, deletions, failures,
+ * system rows) and a
  * live typing indicator.
  *
  * What to exercise here:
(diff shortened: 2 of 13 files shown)


packages/react/src/lib/providers/i18n/i18n-provider-defaults.ts
```diff
@@ -638,6 +638,9 @@ export const defaultTranslations = {
     edit: "Edit",
     editing: "Editing",
     edited: "edited",
+    // Muted label on a message drafted with One and approved by its sender.
+    // "One" is the product name and stays untranslated.
```
sauldom102: Maybe we could drop some of these comments, I don't see them anywhere in this file for example

sauldom102 (approved):

siguenzaraul (approved):

warcos-fact:
Do we have a design on this? With the example I see it seems as if the text could collide with the text message.
More feedback: that text contrast won't pass the tests, is that the way we display the timestamp in chat messages?

dmitry-zaets:
> Do we have a design on this? With the example I see it seems as if the text could collide with the text message

We don't have a design on this, neither we have a designer yet😅
Here are some more screens:
| Screen | Screen |
|---|---|
| image | image |
| image | image |

Edge cases on the tiny screens:

| Screen | Screen |
|---|---|
| image | image |


> More feedback: that text contrast won't pass the tests, is that the way we display the timestamp in chat messages?

Yep, that's the current design of the timestamps and meta labels.

warcos-fact:
All of them have a different distance between text and the "drafted with..." line.

I would suggest to have it always under the message, at a consistent distance.
The first examples on screen are strange, we shouldn't make the message pill longer because of that clarifying line.

As for the text, we should capitalise when necessary "Drafted with One - Edited...".

I don't have time to help with the design but I recommend you share the results over the foundations channel so more designers can help out with this.

dmitry-zaets:
Thanks for the feedback.
The lowercase in the meta line is done on purpose (same as telegram and slack, etc).
i'll check about the width of one-line messages

PR #5506, opened 2026-09-09
feat(EditableTable): forward select actions to the cell

rubenmoya:
## Why

In the new One Members table, where users select the plan the want to apply, we need an option to cancel a scheduled plan change:

image

`F0Select` already has `actions` which renders buttons below the options via `SelectBottomActions`. The editable table just never passed them through, so the only way to offer one from a table was a separate row action sitting outside the cell it belongs to.

## What

`SelectCellConfig` gains an `actions` prop, forwarded to the `F0Select` the cell renders:

```ts
actions?: SelectAction[] | ((item: R) => SelectAction[] | undefined)
```

It takes a function as well as an array so the actions can be decided per row — in the case that prompted this, only members with a pending change get one.

## Changes

- **`SelectCellConfig`** — new optional `actions`, typed against `F0Select`'s own `Action`.
- **`SelectCell`** — resolves it the same way it already resolves `options`, and passes it to both the static-options and `source` variants.
- **Tests** — the action fires when clicked, and the function form receives the row item.
- **Story** — the existing status-select example now shows a per-row "Withdraw request" action on pending rows.

## Verification

`oxlint --type-aware` clean, `tsc --project tsconfig-build.json` clean, 111 tests pass across the EditableTable suite.

### DIFF
--- a/packages/react/src/patterns/OneDataCollection/__stories__/visualizations/editable-table/editable-table.stories.tsx
+++ b/packages/react/src/patterns/OneDataCollection/__stories__/visualizations/editable-table/editable-table.stories.tsx
@@ -1892,6 +1892,17 @@ export const EditableTableWithStatusPillSelect: Story = {
                   selectConfig: {
                     placeholder: "Status",
                     showSearchBox: false,
+                    actions: (item: MockUser) =>
+                      item.status === "pending"
+                        ? [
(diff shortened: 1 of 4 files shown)


eliseo-juan (approved):
LGTM

PR #5532, opened 2026-09-14
fix(OneDataCollection): report a selection-aware total

lluis-pujalte:
## Description

A tree collection paginates its parent rows while selecting the children rendered underneath them, so `paginationInfo.total` undercounts the selection. The in-table banner already compensated for this, but the `onLoadData` effect reported the raw total — and that is what reaches the action bar, so the two disagreed: a "Select all 10 items" button that then claimed "All 2 items selected". This reports a separate `selectableTotal` and feeds the action bar from it.

## Implementation details

- fix: feed the action bar a selection-aware total so it agrees with the "Select all N items" banner it came from

- feat: add an optional `selectableTotal` to the `onLoadData` payload

  <details>
  <summary>Why a second field instead of redefining <code>totalItems</code>?</summary>

  `totalItems` is not only a selection input — it also backs `totalItemSummary`
  ("Total items: N") and `resultCount`, where the paginated total is the correct
  value. Changing it in place would have fixed selection and broken the summary.

  The two fields answer different questions: `totalItems` is how many items the
  collection holds, `selectableTotal` is how many bulk selection acts on. They
  only diverge for collections that paginate something other than their
  selectable rows. The field is optional, so Card, List and Kanban are unchanged
  and fall back to `totalItems`.

  Flagging this as the main design decision to review — if Foundations would
  rather derive this inside the action bar, the change is easy to reshape.
  </details>

- refactor: compute `currentPageSelectableIds` and `selectableTotal` once, beside the selection registry

  <details>
  <summary>Why the move?</summary>

  The banner and `onLoadData` need the same number, and the duplicate derivation
  is how they drifted apart in the first place. Both now read one computation.

  They must stay above the `isInitialLoading` early return — placing the effect
  below it makes the hook conditional and fails 70 of the 73 Table tests with
  "Rendered more hooks than during the previous render". There is a comment on
  the effect recording that invariant.
  </details>

- test: add a regression test asserting a tree collection reports `totalItems: 2` and `selectableTotal: 5`

## Verification

- 745 tests pass across all 78 `OneDataCollection` suites
- The new test was confirmed red before the fix (`selectableTotal` undefined, while the `totalItems` assertion already passed — precisely the split being fixed)
- `tsc` clean; `oxlint` 0 errors, no new lint debt

Not visually verified — I had no working browser automation in this environment, so the rendered action bar has not been checked. The downstream symptom and both numbers are reproduced from the source and covered by the test. Screenshots welcome from anyone who can run it.

## Context

Found while fixing [FCT-59748](https://factorialmakers.atlassian.net/browse/FCT-59748). The consumer is worked around separately in factorialco/factorial#114127 (the headcount employee sheet now reports an honest item total), but any future tree collection reporting a parent-row total would hit this again — hence the fix here.

[FCT-59748]: https://factorialmakers.atlassian.net/browse/FCT-59748?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/react/src/patterns/OneDataCollection/OneDatacollection.tsx
```diff
@@ -894,6 +894,13 @@ const OneDataCollectionComp = <
   }
 
   const [totalItems, setTotalItems] = useState<undefined | number>(undefined)
+  // Kept apart from `totalItems` because the two answer different questions:
```
albertpmz: We don't need this comment here, the variable name already carries its meaning

packages/react/src/patterns/OneDataCollection/types.ts
```diff
@@ -62,6 +62,13 @@ export type OnLoadDataCallback<
   Filters extends FiltersDefinition,
 > = (data: {
   totalItems: number | undefined
+  /**
+   * How many items selection acts on, when that differs from `totalItems`. A
```
albertpmz: Same goes here,  we don't need this prose

albertpmz (changes requested):
Good job! Let's just remove these comments and we'll be good to go 👍🏼

albertpmz (approved):
LGTM!

PR #5535, opened 2026-09-14
feat(F0Select): shape the inline variant for detail rows

albertpmz:
## Description

The employee-details prototype edits the manager from the detail row itself, so the value sits where a read-only row would print it and the chevron only appears once the row is hovered. The inline F0Select variant now matches that, and a disabled inline select renders its value as plain text instead of a greyed-out trigger.

Four states, from `Components / Select / Inline / Detail Row` and `Detail Row Disabled` in Storybook:

| State | What it looks like |
| --- | --- |
| Rest | No border, no background, no chevron. Reads like the read-only rows above and below it. |
| Hover | Background and chevron fade in. The chevron holds its space at rest, so the label never reflows. |
| Open | Same background and chevron, dropdown sized to the value column, with the search box. |
| Disabled | Avatar and label only. No button, no chevron, no popup. |

image

[Link to prototype](https://factorial-composer-employee-details.vercel.app/p/employee-details?preview=true&viewer=admin)

### Trigger

The trigger was `w-fit`, which hugs the label and leaves the chevron floating mid-row. It now fills its container, so in a detail table the chevron lands at the right edge of the value column. The variant has no consumers outside its own tests, so no existing layout moves. If the header-chip shape is wanted later, it can come back as a prop.

The chevron sits at `opacity-0` and fades in on hover, focus-visible and open. It holds its space at rest, so revealing it never reflows the label.

Box styling moved from `pl-3 pr-2 gap-1 rounded` to `px-2 gap-2 rounded-md` to match the prototype.

### Disabled state

A disabled inline select now returns before the popup mounts, so there is no button, no chevron, and no data source fetch. It renders whatever the option carries, avatar and icon and label, as plain text. An option with no avatar renders its label alone.

The trigger and the disabled value share one box class. Both measure 32px, so a column of rows does not shift when one of them turns read-only.

### Tests

| Covers | Where |
| --- | --- |
| Disabled render, with and without an avatar | `F0Select.test.tsx` |
| Placeholder fallback when disabled with no value | `F0Select.test.tsx` |
| Chevron reveal classes and full-width trigger | `F0Select.test.tsx`, `DetailRow` play function |
| Opening, searching and selecting | `F0Select.test.tsx` |

The `DetailRow` play function asserts only synchronous state. Driving the dropdown from it made the story depend on when the popup's rows resolve, which was flaky.

### Stories

`DetailRow` and `DetailRowDisabled` reproduce the prototype's label and value columns. Both are folded into the snapshot story, along with a disabled row holding no value.


albertpmz:
Superseded by `feat/form-inline-field-row`, which puts the detail row in the form layer instead of in each component.

The reason: built per component, the row is one copy per field type. `DateDisplay` on the DatePicker branch was 347 lines, two of them date-specific. `F0Form` and `F0FormField` already know what type every field is, so the row belongs there once — `inline` on any field definition, rendered through one private `InlineFieldRow`. The DatePicker branch reverts its variant in the same PR.

F0Select keeps `variant="inline"` exactly as `main` has it: it is public and documented, and nothing here changes that. What comes out is only the detail-row shaping this branch added on top of it — the full-width box, the hover-revealed chevron and the disabled static value. Nothing was merged, so there is nothing to revert.

albertpmz:
Closing: the detail row ships from the form layer instead.

PR #5562, opened 2026-09-17
fix(F0DatePicker): report the close a selection causes and paint the value on the first frame

albertpmz:
## Description

Fixing two independent bugs in `F0DatePicker`: 
- Picking a day closed the calendar without firing `onOpenChange`.
- The picker mounted with a value rendered empty for one frame. 

Both have been found while planning the new read-edit pattern; in the designed architecture, an external layer owns the open state and the value has to read as text from the first paint.


eliseo-juan (approved):
LGTM

PR #5563, opened 2026-09-17
feat(F0TextInput): add Inline compound component

albertpmz:
## Description

Add `F0TextInput.Inline` for detail rows whose value reads as text until the parent activates the editor. It shares the existing input implementation while keeping `F0TextInputProps` object-shaped, so existing field wrappers can continue extending it. Inline editing controls belong to `.Inline` and are rejected on standalone fields.

## Type of change

- [x] Component enhancement / variant

## Implementation details

- feat: expose the detail-row presentation through `F0TextInput.Inline`, with its own `F0TextInputInlineProps`. Consumers do not pass a `variant` prop.
- feat: let the parent control `editing` and receive `onDismiss("commit" | "escape" | "blur")`. Blur dismissal happens only when focus leaves the whole editor, so clear and password visibility controls remain usable by mouse and keyboard.
- fix: mask resting password and private values, including their title text, and give hidden-label resting values a named group for assistive technology.
- test: retain inline interaction and accessibility evaluations and add compiler-enforced checks for field compatibility and the separation of inline-only props.

The component retains local input-value state. Parents that need commit or rollback behavior should control `value` through `onChange`. Private text uses the focus-to-reveal behavior introduced in this PR; this also applies to standalone private fields.

The props aliases preserve their pre-PR object shape. Adding `.Inline` still adds a required static member to the component value, which the API checker reports for `F0TextInput` and its deprecated `Input` alias. Code assigning a replacement component to `typeof F0TextInput` must also supply that member; ordinary JSX and interfaces extending the field props remain compatible.

## Verification

- 177 unit tests passed, with one existing skip, across TextInput, InputField, and Select.
- 27 Chromium Storybook tests passed, including axe checks and mouse/keyboard interactions.
- Compiler contract test checks field compatibility, direct and spread prop rejection, refs, dismiss reason inference, and deprecated aliases.
- Full TypeScript, formatting, and lint checks passed. Lint retains 114 existing warnings and no errors.


eliseo-juan (approved):
LGTM

warcos-fact (approved):

PR #5572, opened 2026-09-17
feat(F0NumberInput): add the inline variant that reads as text

albertpmz:
## Description

`variant="inline"` gives F0NumberInput the detail-row presentation: the number prints as text and becomes its own editor when the row activates it. Stacked on #5563, which put the inline box and inset in `F0InputField`; base branch is `feat/f0textinput-inline-variant`.

## The contract

The variant is presentational. It holds no state: the parent passes `editing` and the component draws the matching presentation.

| Prop | Accepted when | Behaviour |
| --- | --- | --- |
| `variant` | always | `"field"` (default) or `"inline"` |
| `editing` | `variant="inline"` only | `false` (default) draws the number as text; `true` draws the input and focuses it |
| `onDismiss` | `variant="inline"` only | `"commit"` on Enter, `"escape"` on Escape, `"blur"` when focus leaves |

`editing` and `onDismiss` sit on a discriminated union, matching `F0TextInput`, so they are a type error on the field variant.

## What the number reads as at rest

One formatter, so no glyph moves between modes. At rest the text is the string the resting editor shows — `locale` grouping, `maxDecimals` — plus the `units` suffix the editor renders as a trailing tag. `45,000 EUR` at rest, `45,000` + an `EUR` tag in the editor, `45000` once the input takes focus and drops the separators, which is the component's existing focused-editing behaviour.

A `null` or empty value prints the `placeholder`. The stepper is not drawn at rest: the arrows belong to the editor.

Carrying the units suffix needed one addition outside F0NumberInput — `inlineText` on `F0InputField`, the string the inline variant prints when the value alone is not what the editor shows. It defaults to the value, so `F0TextInput` is untouched.

## Breaking-change bot

The API-surface bot will flag `F0NumberInputProps` as `object → union`. That is expected and accepted: a grep across the Factorial monorepo found zero type-level consumers of these props.

## Verification

- 19 new unit tests for the variant; 156 pass across F0NumberInput, F0InputField and F0TextInput.
- 9 stories in `F0NumberInput.inline.stories.tsx` with `a11y: { test: "error" }`, including the glyph-position story (read text and input start at the same x, both 40px tall) and a 40×320 fill story. All 71 story tests across the three components pass against a local Storybook.
- `pnpm tsc`, `oxfmt --check src`, `npx oxlint src` at the 114-warning baseline with no new lint debt.
- Opened in Chrome, not only under the runner.


eliseo-juan (approved):
LGTM!

PR #5579, opened 2026-09-17
feat(F0Form): inline detail rows

albertpmz:
## Description

`inline` turns an F0Form into a record: every field becomes a row with its label on the left and its value on the right, printed as text. Activating a row swaps that text for the field's own editor, in the same box, at the same inset, so nothing moves under the pointer. Leaving the row puts the text back and keeps the edit in the form state, unsaved — nothing commits per row, and the existing floating action bar is what writes the record. Text, number, select, date, switch and checkbox read as rows; any other type keeps its standard field inside the row and logs one development warning.

This PR carries the whole stack, not only the last slice, so its alpha can be installed and tested in the Factorial monorepo in one go. Each slice is also open on its own for review:

| PR | What it adds |
| --- | --- |
| [#5562](https://github.com/factorialco/f0/pull/5562) | F0DatePicker reports the close a selection causes, and paints its value on the first frame |
| [#5563](https://github.com/factorialco/f0/pull/5563) | The inline variant of F0TextInput, and with it the box and inset every `F0InputField` inherits |
| [#5572](https://github.com/factorialco/f0/pull/5572) | The inline variant of F0NumberInput, money and percentage included |
| [#5576](https://github.com/factorialco/f0/pull/5576) | The inline variant of F0DatePicker: the calendar opens in place, Escape is read from the document |
| [#5577](https://github.com/factorialco/f0/pull/5577) | F0Select reads as text at rest — avatar and icon included, no chevron, no button |
| [#5574](https://github.com/factorialco/f0/pull/5574) | The inline variant of Switch and F0Checkbox, which are their own editor |
| [#5573](https://github.com/factorialco/f0/pull/5573) | `InlineFieldRow`, the row chrome: the activator, the hover reveal, the action strip, copy |

The last two slices have no PR of their own and land here: the field layer (`useInlineField`, `InlineFieldRenderer`, the `inline` prop on F0Form, the `editable` and `copyable` field properties), and the profile stories with the three fixes below.

## The consumer API

```tsx
<F0Form
  name="employee-profile"
  inline
  schema={profileSchema}
  defaultValues={profile}
  onSubmit={save}
  submitConfig={{ type: "action-bar", discardable: true }}
/>
```

`inline` is form-wide: either every field reads as a row or none does. Two field properties are read only in this mode.

```tsx
const profileSchema = z.object({
  employeeNumber: f0FormField(z.string(), {
    label: "Employee number",
    editable: false, // reads, never activates; a toggle is drawn disabled
    copyable: true, // adds a copy action writing the value as plain text
  }),
})
```

`renderIf` already existed and does the third tier. Together the two cover the prototype's `hidden | read | edit` access map, so one schema builder serves every viewer:

```tsx
const tier = (id: string) => ({
  renderIf: () => access(id) !== "hidden",
  editable: access(id) === "edit",
})
```

The three profile stories prove it: the same component, built from one `buildProfileSchema(viewer)`, renders the HR admin's, the manager's and the employee's view of the same record.

## What lands here beyond the field layer

**The row says why it refused a value.** A failing field only drew a critical border. `InlineFieldRow` gains a `message` slot under the value box, inside the value column, so the row grows downwards and the value keeps its width. The field layer resolves the text the way `FormMessage` does and passes it down as a node — the row still knows nothing about validation.

**The inline select reads at the weight of its neighbours.** It printed through `textVariants({ variant: "label" })`, which adds `font-medium`, while `F0InputField`'s `InlineValue` prints at the ambient weight. Both the at-rest text and the open trigger now use `textVariants({ variant: "body" })`.

**A hidden row no longer leaves a divider.** A field `renderIf` hides keeps its Controller mounted, so it stays in the card as a hidden `<span>`. Rows dropped their bottom border with `last:`, a DOM sibling rule, so that span counted as the last child and the row above it kept a divider running to nothing. The rule is now `last-of-type`: rows are `div`s, so it counts rows.

**Enter no longer saves the record.** Enter in an inline editor means "I am done with this field". In a browser the same keystroke is also a form's implicit submission, so a profile row saved everything the moment the user pressed it. jsdom never submits, so every unit test agreed with the intent and only a real browser disagreed. The three inline editors that take a keystroke now call `preventDefault` on Enter, and only while inline.

## Deliberate API shape

Every inline variant is a discriminated union on `variant`: `editing` and `onDismiss` exist on the inline member and are `never` on the field member. That is what makes "pass `editing` without `variant="inline"`" a compile error instead of a silent no-op, and it is why the API-surface bot flags the change as breaking. No consumer in the Factorial monorepo depends on these props at the type level — the union only narrows what was previously unconstrained — so the flag is accepted rather than worked around with optional props that admit nonsense.

## Known limitations

- Single-schema forms only. The per-section mode does not read `inline`.
- Inline rows carry no anchor ids, so `styling.showSectionsSidepanel` has nothing to jump to and error scroll-to-field does not reach a row.
- Request-change — the ask-for-an-edit tier, its dialog and the pending banner — is iteration 2. `editable: false` is the whole read-only story today.
- Dates print what `DateInput` prints (`10 Apr 2026`), which is not locale-aware until `DateInput` itself is.
- A source-backed select resolves its options asynchronously inside F0Select, so the row cannot reach the label and offers no copy action even with `copyable`.
- A date row carries a calendar glyph at rest, so its value starts further right than a text or select row in the same card.
- A section whose every row is hidden still renders its title above an empty card.

## Storybook

- `Forms/F0Form/Inline` — nine stories: every type at rest, activating a row, the action bar after an edit, Escape reverting, `editable: false`, `copyable`, the fallback for an unsupported type, a hidden last row, sections.
- `Forms/F0Form/Inline profile` — the employee profile as admin, as the employee, as the manager.
- `Forms/InlineFieldRow` — the row on its own, with a `<span>` for a value and fake actions.
- `Inputs/Text input/Inline`, `Inputs/Number input/Inline`, `DatePicker/Inline`, `Select/Inline`, `Switch/Inline`, `Checkbox/Inline` — one per component.
- The pattern is documented under "Inline detail rows" on the F0Form docs page.


albertpmz:
build

PR #5582, opened 2026-09-18
feat(react): add itemActionsOnHover to the editable table

omedale:
### Default action
Screenshot 2026-09-18 at 9 55 42 AM

### Opt-in on hover action
https://github.com/user-attachments/assets/85c2e278-4ae5-4eb4-9635-2a02812b74a4

## Description

The editable table paints the row-actions button on **every** row, while the plain table reveals its actions on hover via `ItemActionsRowContainer`. On a long sheet that reads as a column of buttons running down the table, competing with the data.

This adds **`itemActionsOnHover`** to the editable table's visualization options. It is **off by default**, so every existing table keeps the behaviour it has today; callers opt in where the quieter row suits them better.

```tsx
visualizations={[
  {
    type: "editableTable",
    options: { columns, onCellChange, itemActionsOnHover: true },
  },
]}
```

## Type of change

## Implementation details

`itemActionsOnHover` threads from the visualization options through `Table.tsx` to `Row.tsx`, alongside `referenceRowType`. In `Row.tsx` the `editableTable` branch composes its class through `editableActionsClass(onHover, dropDownOpen)`:

- **Default (off)** — unchanged: `flex flex-nowrap justify-center`, actions always visible.
- **On** — adds `opacity-0 transition-opacity group-hover:opacity-100 focus-within:opacity-100`, driven by the `group` class the row already carries.

The actions **cell** stays mounted either way, so the column keeps its width and a row's content cannot shift sideways as the pointer enters it. An open dropdown outlives the hover, and `focus-within` reveals the actions for keyboard users.

Note this gates the whole actions container, so primary action buttons hide along with the `⋮`.

## Docs and stories

- New story: **Editable Table → `EditableTableWithItemActionsOnHover`**.
- `editable-table.mdx` gains a section under **Item Actions**. That page previously stated flatly that "The column is always visible — no hover required", which now needs the caveat.

## Testing

`vitest --project=unit src/patterns/OneDataCollection` — **780 passed** (was 778).

Two paired tests in `EditableTableItemActions.test.tsx`:

- `paints the row actions on every row by default` — guards the existing behaviour against regression.
- `reveals the row actions on hover when itemActionsOnHover is set` — covers the new option.

jsdom computes no Tailwind styles, so both assert on the class the row's `group` hover drives rather than on computed visibility. Also corrected a now-stale comment in that file claiming the editable table renders its actions always-visible.

`pnpm tsc` — 0 errors (needs `pnpm --filter @factorialco/f0-core build` first, otherwise unrelated files report `never` types). `oxlint`/`oxfmt` clean; lint-debt reports no new debt.

## Context

Found while working the salary review employee sheet QA — [FCT-63802](https://factorialmakers.atlassian.net/browse/FCT-63802), "Row actions button only on row hover".

Consuming PR: factorialco/factorial#114954. That ticket closes only once this ships: f0 merge → release → Factorial bumps `@factorialco/f0-react` off `6.98.0` → the sheet passes `itemActionsOnHover: true`. The flag is deliberately not in the Factorial PR yet, since the option does not exist in the pinned version.

## History

First pushed as a `fix:` that made hover the unconditional behaviour. Reworked to an opt-in option so existing consumers are untouched, and force-pushed to keep a single coherent commit — release-please reads commit types, and a `fix:` claiming behaviour that is no longer the default would have been wrong in the changelog.

[FCT-63802]: https://factorialmakers.atlassian.net/browse/FCT-63802?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


arnaub (approved):
LGTM

warcos-fact (approved):
Looks consistent with the other hovers we have in the Data Collection.

PR #5589, opened 2026-09-18
chore(react): skip aria surface on nondeterministic stories

sauldom102:
## Description

The aria-surface check compares every story's role + accessible-name pairs against the baseline from the latest `main` run, but two stories render content that changes on its own — so they report a rename on every run of every PR, whatever the diff actually touched. This opts both out through the `ariaSnapshot: { skip: true }` escape hatch the test-runner already reads, so the check's output is about the PR under review rather than about the clock and the RNG.

## Implementation details

- chore: add a `withSkipAriaSnapshot` parameter helper next to the existing `withSkipA11y`, so opt-outs stay greppable and carry their reasoning

- chore: opt `Dashboard` out of the aria surface

  <details>
  <summary>Why?</summary>

  The meta decorator mounts a widget on every load with `id: \`widget-${Math.random()}\`` and `title: \`Title ${Math.random()}\``, so each run renders a different `heading "Title 0.123…"`. Both `Default` and `With Data Test Id` reported that as a rename on the last run.
  </details>

- chore: opt `VersionHistory` out of the aria surface

  <details>
  <summary>Why?</summary>

  Two of the four `mockVersions` fixtures are built from `Date.now()` ("7 days ago", "Yesterday"), and a version button's accessible name is its formatted timestamp — so the names track today's date and all three stories report renames every day.
  </details>

## Notes for the reviewer

Pinning the fixtures would be the better fix: it keeps these stories' accessible names under the check instead of dropping them from it entirely. It moves their Chromatic baselines, though, so it is left for a follow-up rather than bundled into a noise-reduction change.

Spotted while triaging the aria-surface comment on #5499, which reported 157 breaking changes across 14 stories for a purely additive PR. These eight rows are the provably nondeterministic share. The larger share — whole `table` / `dialog` subtrees disappearing from one story while an equivalent subtree appears in a sibling story in the same run — is a settle race in `recordAriaSurface`, which waits two `requestAnimationFrame` ticks before snapshotting; that is worth a separate fix before the check is flipped to `--fail-on-breaking`.


sauldom102:
Superseded by #5590, which fixes the fixtures instead of opting the stories out of the check — the stories keep their aria-surface coverage that way.

PR #5592, opened 2026-09-18
fix(react): clip and reduce-motion the height reveals

sauldom102:
## Description

Five reveals animate height from zero to their intrinsic size without clipping themselves and without honoring reduced motion. Unclipped, the content keeps its full height while the box is still growing, so it paints outside that box over whatever sits below. Unreduced, it animates for people who asked their system for no animation. This fixes both, and deliberately leaves the height animations themselves alone.

## Why the height animations stay

This came out of a pass over the 78 `react-doctor/no-layout-property-animation` findings, which are 34 distinct animations once the initial/animate/exit triples are collapsed. Following React Doctor's own guidance for the rule, most are the case it names as correct rather than a defect:

- **22 are intrinsic reveals.** An element mounts or unmounts and animates height or width between zero and `auto`. The guidance calls this "the least-distorting option" and says to keep it. The `layout` prop is not a substitute, because layout projection animates between two measured states of an element that persists, and these have no previous geometry to project from. CSS cannot transition to `auto` either. What the guidance does ask for is containment and reduced motion, which is what this PR adds where they were missing.
- **3 are in ApplicationFrame**, whose width and padding animations carry a comment explaining that a projection-based approach was tried and removed, because a FLIP over a box whose width changes is a scale.
- **2 animate width to `100%`**, one of which documents that it cannot set overflow without clipping a shadow it needs to escape.
- **The rest** need a visual decision that a static review cannot make: replacing a size change with a transform changes whether siblings reflow, and the guidance is explicit that this must not be done mechanically.

Two are left as follow-ups rather than fixed here:

- The Kanban group collapse has no containment, but it is a drag surface and clipping could catch drag previews. It needs a visual check first.
- The Sidebar animates an 8px float through `top` and `left`. Folding that into the existing `x` transform is possible, but `x` is deliberately on a slower curve to stay in sync with the frame, so the float would inherit the wrong timing and the two axes would move at different speeds. The large 260px movement it also performs is already a transform, so the layout cost being flagged is an 8px inset on a state toggle.


albertpmz (commented):
Let's do a cleanup pass of Claude's comments and we'll be good to go 👍🏼

PR #5594, opened 2026-09-18
fix(react): reclaim the actions column when itemActionsOnHover is set

omedale:
### Before
Screenshot 2026-09-18 at 3 26 00 PM

### AFTER with hover

https://github.com/user-attachments/assets/dde2931a-eec2-4a3e-8c7d-78f5e75a20b5

### AFTER without hover

https://github.com/user-attachments/assets/6cf51f90-80a8-402f-b62d-1c48a63921fe

## Description

Follow-up to #5582, which shipped in 7.7.0.

`itemActionsOnHover` faded the row actions out until the row was hovered — but the editable table gives those actions a **column of their own**, and that column stayed. Its width went with it, so a consuming table reads with an empty strip down its right edge: the actions are hidden, the space they occupied is not.

| | before | after |
| --- | --- | --- |
| actions column | rendered, contents faded | not rendered |
| width reserved | yes | none |

With the option set, the actions now take the same overlay the **regular** table has always used — absolutely positioned over the last column, reserving no width. The cell is removed rather than emptied, so nothing is left to occupy space, and a row's content still cannot shift as the pointer enters it.

**Nothing changes for tables that don't set the option.** `actionsAsColumn` is `isEditableTable && !itemActionsOnHover`, so with the flag unset it is the same expression as before and the dedicated column renders exactly as it always has. The deleted `editableActionsClass` returned a bare `"flex flex-nowrap justify-center"` in that case, which is now the literal className in the same branch — byte-identical output.

## Type of change

- [x] Bug fix

## Implementation details

`actionsAsColumn` replaces the bare `isEditableTable` check at all four sites — `Row` plus the header, sub-header and summary rows in `Table`. All four have to agree or the three header rows and the body disagree on how many cells a row has.

`editableActionsClass` is gone: `ItemActionsRowContainer` already carries `opacity-0 … group-hover:opacity-100` and the dropdown-open override, so routing to the overlay made the helper redundant.

**Keyboard focus.** The helper also carried `focus-within:opacity-100`, which the overlay container did not — so routing to it would have dropped focus reveal. `ItemActionsRowContainer` now answers `focus-within` as well. Every view using the overlay (table, list, card) gains it; previously a keyboard user could tab into buttons that never became visible.

## Why this slipped past #5582

**The stories it was verified against don't have enough columns to scroll.** With no horizontal overflow the actions column collapses to roughly the width of its contents, so once those were faded there was nothing visibly left — the reserved width only reads as an empty strip on a wide sheet that scrolls, which is exactly where the consuming table needed it. The unit test had the same blind spot: it asserted the *contents* were hidden, which stayed true, and never asked whether the column was still there.

The test now asserts structure rather than styling — that the actions hang off the zero-width sticky overlay rather than sitting in a cell of their own.

## Testing

`vitest --project=unit src/patterns/OneDataCollection` — **781 passed**.

Red-green verified: reverting `actionsAsColumn` to the old `isEditableTable` check fails 2 of 7 in that file, passes 5. The default-behaviour test (`paints the row actions on every row by default`) stays green throughout.

`pnpm tsc` — 0 errors (after `pnpm --filter @factorialco/f0-core build`). `oxlint`/`oxfmt` clean; lint-debt reports no new debt.

## Context

[FCT-63802](https://factorialmakers.atlassian.net/browse/FCT-63802). Consuming PR: factorialco/factorial#114954, which is on 7.7.0 and already passes `itemActionsOnHover: true` — it picks this up on the next release with no code change.

[FCT-63802]: https://factorialmakers.atlassian.net/browse/FCT-63802?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


sauldom102 (approved):
I think this should be the behavior by default though, actions shouldn't be hidden in some far away column because of the viewport of the user

PR #5598, opened 2026-09-18
feat(filters): let the panel carry a way of filtering that is not a filter

ibrahimkamal-factorial:
## Description

Adds `quickFilter` to `OneFilterPicker`: an entry above the filter list with a pane of its own, for a way of narrowing a list that is not one filter.

The design and the original implementation are **Pili's**, from `feat/data-collection-search-ui`. This is that slot extracted onto current `main`, on its own — without the assisted-search presentation it was built to carry, which is opinionated in ways only that feature needs.

## Type of change

## Implementation details

```ts
quickFilter?: {
  label: string
  icon?: IconType
  render: (api: { stage: (filters: Record<string, unknown>) => void }) => ReactNode
  chip?: { label: string; onRemove: () => void }
  active?: boolean
}
```

f0 draws the entry; the pane is the consumer's. That is deliberate — it is the only way one slot can serve a natural-language question, a saved view and a team picker without f0 knowing about any of them.

Four behaviours worth reviewing:

- **`stage()` writes into the draft, not the collection.** Inside the popover everything waits for the apply button; a pane that quietly bypassed it would be the odd one out. Staging merges into the same local filter state the list renders and Apply emits, so the list reflects what was staged, anything ticked afterwards survives Apply, and a key staged as `undefined` clears. Nothing is held aside for Apply to merge later.
- **`chip` gives the pane a presence outside the popover.** A quick filter can narrow the list in a way no filter chip shows — a row set resolved elsewhere, a lookup still running — and once the popover closes nothing else on the page would say so. When the consumer supplies `chip`, it is drawn first in the applied row, alone if need be; its × calls `onRemove`, clicking it opens the panel, and both Clear buttons — the chips row's and the panel's Clear filters — remove it along with the filters, since to the reader the chip is one of them.
- **The two selections exclude each other.** The entry is not one of the filters, so it cannot live in `selectedFilterKey` — picking either clears the other.
- **The panel opens on the entry when nothing is applied**, and on the filter carrying a value when something is: whoever opened a filtered panel opened it to see or change that. Unless the consumer marks the entry `active` — what is applied is the quick filter's own answer — in which case the reader comes back to the question they asked rather than to a filter it set.
- **Both modes are wired** — the default popover and the compact drawer (`F0FilterPickerContent`), where the entry gets the same chevron the filters have.

`icon` is the one thing not in the original: our own first use puts One's mark on the entry, and an entry that cannot carry an icon is a weaker primitive. Happy to drop it if you would rather keep the slot minimal.

One change outside the picker rides along, because the first consumer needs it in the same release: **`F0SearchInput` gains `icon?: IconType`**, replacing the magnifier and defaulting to it. A field that puts a question to an assistant finds rows by other means than matching text, and the magnifier promises that it does.

## Screenshots

`Filters/FilterPicker → WithQuickFilter` in Storybook is a working example: a text field that stages a `department` filter, applied only when Apply is pressed. `WithQuickFilterChip` shows the chip standing in the applied row on its own.
image

## Testing

- `pnpm --filter @factorialco/f0-react run tsc` — clean
- `pnpm --filter @factorialco/f0-react run lint` — 0 errors, no new lint debt (3 warnings, baseline 475)
- `pnpm exec vitest run src/patterns/OneFilterPicker src/patterns/F0FilterPickerContent` — 109 passed; `FiltersChipsList.test.tsx` covers the chip (alone, ordering, ×, Clear, click, `hideChips`)
- Story for staging-then-applying, which reads better as something you can click.

## Context

First consumer is the AI filter assistant on Organization → People, which needs to ask for a list in words from inside the filters panel. Nothing about that is in this PR.


warcos-fact (approved):
Approving, I recommend reviewing the way we establish that the field is AI enabled, atm there is nothing telling me One is going to generate filters based on my search.

Review with the designer, perhaps we need clarifying text under the input, or the icon of One to signal this.

packages/react/src/patterns/OneDataCollection/__tests__/quickFilterViews.test.tsx
```diff
@@ -0,0 +1,223 @@
...
+      <button type="button" onClick={() => setAsked("who is away")}>
+        Ask
+      </button>
+      <OneDataCollection
+        id="quick-filter-views/v1"
+        source={source}
+        visualizations={visualizations}
+        quickFilter={{
+          label: "Ask ONE",
+          render: () => <div>pane</div>,
```
sauldom102, line 74: Please, let's avoid props that are this flexible, we shouldn't allow to render custom components from here, it should be more semantic and with primitive types

ibrahimkamal-factorial: Agreed. I have reworked it so the panel draws the pane and the consumer only declares it. Before I finish migrating the consumer, does this shape look right to you?

```ts
input: {
  placeholder: string
  icon?: IconType
  value?: string
  onSubmit: (value: string) => void
}
messages?: string[]
suggestions?: {
  label: string
  items: ReadonlyArray<{ id: string; label: string }>
  onSelect: (id: string) => void
}
empty?: { title: string; description?: string; icon?: IconType }
staged?: Record<string, unknown>
```

Two things went with the render prop. `stage` was a callback the pane called to merge filters into the draft; it is now `staged`, a map the panel merges when it changes, so nothing calls back into consumer code. And `onApply` is gone: it only existed because the consumer owned the field, and now that the panel owns it the apply button reads it directly, so that is one less thing for a consumer to wire.

The one judgement call worth flagging is `messages`. What sits there today is eight product states (a criterion dropped, a lookup that failed or was cut short, an allowance spent, a link that went nowhere), each with its own copy. Strings keep that copy in the product rather than in f0, at the cost of f0 not knowing what any line means. If you would rather have a shape here, say so and I will model it.

ibrahimkamal-factorial:
@warcos-fact on the field not saying that ONE answers it: that is done now, both ways you suggested.

The field carries ONE's own mark instead of the AI magnifier, and turns into a spinner while a question is being answered. The placeholder is now "Ask ONE in your own words" rather than the collection's example, which used to read as text already typed in. The example moved to an empty state under the field, so someone who has never used it still sees what the list can be asked: `Like "Hired in the last 6 months". ONE turns it into filters.`

It landed in the monorepo rather than here, because the pane was the consumer's. That is changing with the rework in the thread above, which moves the pane into f0, so the next version of this PR will carry the field and its copy.

PR #5599, opened 2026-09-19
feat(OneDataCollection): let a list row tint its icon glyph

marcbaque:
## Description

A `OneDataCollection` list row can now draw the same tinted icon glyph the Home's `list` slot draws, so a feed and a full list of the same items look alike. The tint already existed inside the Home kit; this moves it to a shared internal module and gives the list visualization an optional `color` on an icon avatar. Nothing is added to `F0AvatarIcon`.

## Type of change

New story: **Data Collection/Visualizations/List → ListVisualizationWithIconColors**, documented under Examples → Icon colors in the List MDX. It cycles four palette hues and puts a custom hex on the first row. Tagged `withSnapshot({})`, so Chromatic covers both themes.

Verified in Storybook rather than by eye, since the tile is 32px:

| | light | dark |
| --- | --- | --- |
| tile | hue at `0.1` | hue at `0.24` |
| icon | full hue | full hue |
| border | none | none |

The Home's glyphs render byte-identical class strings at the same 32px / 10px radius / 0 border, which is the point of the change.

## Implementation details

- refactor: move the list icon tint out of the Home kit into `src/lib/ListIcon/`

  <details>
  <summary>Why, and why the public surface is unchanged</summary>

  `sds/Home/slotRenderers.tsx:272-276` recorded a deliberate decision: `F0AvatarIcon` is neutral on purpose (a white tile with a border) and colouring it is a surface's decision, not the design system's. **This PR keeps that decision** — a tinted glyph stays its own component. What changed is only that it stops living inside the Home kit, because a second surface now draws it. A reviewer expecting `color` on `F0AvatarIcon` should read that comment first.

  The first commit is a pure move: lines 139-299 lifted verbatim, three previously-private declarations exported, and one doc comment reworded because it named the wrong owner. `listIconColors`, `ListIconColor` and `ListIconPaletteColor` are re-exported from `slotRenderers.tsx`, so they still resolve from `experimental` and no consumer import path changes. Confirmed against the built `dist/experimental.d.ts`.
  </details>

- feat: let a list row tint its icon glyph via `avatar.color`

  <details>
  <summary>Scope of the type change</summary>

  Only the `icon` avatar kind takes a colour — every other kind carries its own identity already. `ListItemAvatar` splits the union rather than widening `AvatarVariant`, so no other avatar kind is affected. A hex that cannot be parsed falls back to the plain glyph rather than drawing nothing.
  </details>

- fix: forward `aria-label`, `aria-labelledby` and `state` through the tinted glyph

  <details>
  <summary>This was a real defect, caught in review</summary>

  `IconAvatarVariant` carries all three and `F0AvatarIcon` forwards them. The tinted path rendered `ListIconGlyph`, which accepted only `icon`/`tint`/`size` — so adding a colour silently deleted an accessible name a consumer had set through the public `ItemDefinition.avatar` API.
  </details>

- docs: document `color` under Examples → Icon colors in the List visualization MDX

  <details>
  <summary>Also fixes pre-existing drift</summary>

  The `itemDefinition` type block claimed `avatar?: AvatarVariant & { badge?: AvatarBadge }`, but `ItemDefinition` has never had a `badge` field. Corrected while adding the new prop.
  </details>

- test: cover the tint resolution, which had no test before (4 tests on `listIconTint`)

- chore: move the `no-inline-styles` lint-debt entry to follow the code

  <details>
  <summary>Why lint-debt.json is in this diff</summary>

  The baseline is keyed per file, and the single existing `f0-styles/no-inline-styles` warning moved from `sds/Home/slotRenderers.tsx` to `lib/ListIcon/ListIcon.tsx`. The pre-commit hook blocks on this and instructs you to run `pnpm check:lint-debt --update` and explain it here. The diff is a clean one-for-one swap; the total is unchanged at 475.
  </details>

## Reviewers

Two teams are required: `@factorialco/foundations` owns `packages/react/src/sds/Home/`, `@factorialco/f0-devs` owns everything else this branch touches.

## Known issues, deliberately not fixed here

**Contrast.** Six of the eleven palette hues put the icon below a 3:1 ratio against their own tile in light mode (flubber 1.84, yellow 1.88 are the worst); two fail in dark mode. This is the **moved** palette — identical to what Home ships today — so fixing it would change Home's rendering and break the "pure move" property above. It is not a WCAG AA failure as used, because a row always renders its title and fields in text alongside the glyph, but it becomes one if the tint is ever made the only carrier of meaning. The MDX now warns against that. The palette fix belongs in a foundations-owned follow-up.

**Untyped callers.** `listIconTint` returns `{ className: undefined }` for an unknown palette name, which is truthy, so a JS consumer gets an unstyled tile instead of the documented fallback. Also moved code; a two-line guard would fix it.

**Test coverage.** `ItemTeaser` had no test file before this PR and does not gain one, per repo preference. The tinted, untinted, bad-hex-fallback and aria-forwarding paths are therefore exercised only through the story and Chromatic, not by unit tests.

## Verification

- `pnpm vitest run src/patterns/OneDataCollection src/lib/ListIcon src/sds/Home` — 1215 passed
- `pnpm tsc`, `pnpm format:check`, `pnpm lint` (475 warnings = exact baseline, 0 errors), `pnpm preflight` — all clean
- `pnpm build` — all 11 palette tints plus the custom-hex one emit into `dist/styles.css` (12 light + 12 dark rules), confirming Tailwind still scans the class literals from their new `src/lib/` home


sauldom102 (approved):
Looks perfect! Only thing worth noticing is that we have the avatar being 32x32 here vs 40x40 in the needs you widget in the home, however, that 32x32 is consistent compared to how we have it the rest of list views in the data collection, so maybe we could think of tweaking it in some other PR if we want to

PR #5601, opened 2026-09-21
feat(react): add F0CandidateSuggestionCard and the recruitment area

ReddyWish:
## What

Adds `sds/recruitment/` (first recruitment area in f0, owned by `@factorialco/talent-recruitment`) and `F0CandidateSuggestionCard`: the card a recruiter decides on in the talent-pool swipe review.

The card is a composition over existing f0 pieces rather than a new surface:

- `F0Card` with neither `link` nor `onClick` — header (avatar, name, headline) and `metadata` (fit tag, applied line)
- `F0AiCallout` for the "Why this candidate" reasoning, tinted from the fit level
- `F0TagStatus` per competency signal, in a wrapping row

It renders data only: no queries, no decision buttons, no state. Elevation is deliberately absent — the deck that stacks the cards owns the shadow (FCT-63896).

## Why

FCT-63900, part of the candidate suggestion swipe review (FCT-59234). The constraint for the whole initiative is that no styling ships to the monorepo: `frontend/` only queries and maps props.

Three decisions worth reviewing:

- **`F0Card` is safe to drag on.** The earlier plan avoided it over its full-card link overlay. That overlay and the body click emulation only exist when `link` or `onClick` is passed (`isInteractive` in `CardInternal.tsx`). Without them the card is inert, and the `children` wrapper is already `pointer-events-none` except on interactive descendants — exactly what a swipe card needs. A test locks this in.
- **Signals wrap instead of using `F0TagList`.** `F0TagList` is a single row that collapses the overflow behind a `+N` counter. On this card the competencies are the evidence, so hiding them defeats the point; they wrap onto as many rows as they need.
- **The reasoning renders with `markdown={false}`.** `F0Text` parses markdown by default, which silently deletes anything shaped like a tag: `"Knows <React> and C++"` rendered as `"Knows  and C++"`. The reasoning is model-written prose the card cannot sanitise, so parsing is off and a test covers angle brackets.

The score ring the Figma showed is gone from the design, so FCT-63897 was discarded and this ticket has no blockers left.

## Testing

- `pnpm vitest src/sds/recruitment` — 11 tests: content, fit-level and signal-state variant mapping, the applied line appearing and disappearing, all signals rendering, repeated signal labels, verbatim reasoning, the group's `aria-labelledby`, and no overlay link
- `pnpm lint src/sds/recruitment` — clean, no new lint debt (baseline 475)
- `pnpm tsc` — clean
- Six stories with a play function on `Default`, two snapshot stories, axe enforced (`a11y: { test: "error" }`), MDX at the "gold" doc tier
- Pre-push preflight and the ownership checks pass

## Follow-up at bump time

The new `candidateSuggestionCard` block in `i18n-provider-defaults.ts` needs matching keys in the monorepo's `frontend/src/modules/core/lib/f0-provider.tsx`. The provider merges over the defaults, so nothing breaks without them, but the three strings would ship English-only.


ReddyWish:
Closing rationale, for the record: the card moves to the monorepo.

Decided 2026-09-21. The ATS application side panel and full page are already unified in three layers that all live in `factorial/frontend`: one fetch hook, one judgment-to-visuals mapping, and one pair of section renderers for competencies and signals. The swipe card is a third presentation on those same layers, so building it in f0 would mean duplicating all three across a package boundary. The redesign that replaced the competency chips with bar rows and sentence signals made that duplication concrete rather than theoretical.

f0 keeps what is generic: `F0SwipeDeck` (#5580), the stack and controls, a generic progress bar, and `F0Confetti` (the monorepo cannot resolve `canvas-confetti`). The `sds/recruitment` area this PR created is no longer needed.

Tracked as FCT-63900, now retitled to [ats].

PR #5605, opened 2026-09-21
feat(OneDataCollection): configurable primary actions variant

carlosgarciavilar:
## Before & After
Current view

## Description

A collection's primary actions always render with the primary (default) button style, but a collection is not always the surface's primary action — on Factorial's goal detail the page-level primary is "Update progress" (header), and the sub-goals table's "Add subgoal" must be secondary per the one-primary-per-page guideline. Because the split button's variant is not configurable, ended up rendering a standalone `F0ButtonDropdown` OUTSIDE the collection (factorialco/factorial#114179), which orphaned the table's settings control onto its own toolbar row; This adds `primaryActionsVariant` to the data source definition so the existing toolbar split button can render in any `F0ButtonDropdown` variant, defaulting to `default` — no change for existing callers.

## Type of change

- Linked #f0-support thread: **not yet raised — this draft is the artifact to bring to that conversation**, per the "bring whatever exists: … a draft PR" guidance in the contribution docs.

## Implementation details

- feat: add `primaryActionsVariant?: ButtonDropdownVariant` to `DataCollectionSourceDefinition` (sibling of `primaryActionsLabel`), threaded through `OneDatacollection` into `CollectionActions` and applied to the three primary-action renderings (single `F0Button`, split `F0ButtonDropdown`, and description-driven dropdown mode).
- test: `CollectionActions` renders the variant on each mode and keeps `default` when the option is absent.
- docs: `OutlinePrimaryActionsExample` story added to the collection-actions stories.


sauldom102 (changes requested):
We can already pass `secondaryActions` via some other prop, it doesn't make sense to me that we have a "primary action variant" while we also have secondary actions prop

carlosgarciavilar:
Fair question — the reason `secondaryActions` doesn't cover this case: they render as flat outline `F0Button`s plus an "···" overflow menu, so there is no way to get a **split button** (main route + chevron menu of related routes) in a subdued style. The consumer need (goal detail's sub-goals table, factorialco/factorial#115057) is exactly a split: one promoted action ("Add subgoal") with its sibling route ("Add existing subgoals") behind the attached chevron — the collection's `primaryActions` shape — but the page's primary action lives in the header, so the split must not carry the primary style. We tried the `secondaryActions` + overflow rendering first and design rejected it precisely because the split/chevron affordance disappears.

Semantically the action is still the *collection's* primary action (promoted + menu); only its visual weight changes because the surface's primary is elsewhere — hence a variant on the existing rendering rather than a new action slot.

That said, happy to reshape if you'd rather grow the other side — e.g. letting a `secondaryActions` group render as a split, or a `variant` on the primary definition itself. Whichever API you prefer, the underlying need is "split button, non-primary style, inside the collection toolbar".

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

PR #5608, opened 2026-09-21
feat(react): dropdown-mode secondary header actions

phm-factorial:
## Description

The AI Analytics dashboard canvas header needs a single outline "Share" dropdown (Share access / Schedule delivery / Export) with self-explaining disabled items while the report is unsaved. Today `ResourceHeader` can only render split-mode dropdown secondary actions, and `ButtonDropdownItem` can't carry a disabled-state tooltip. This PR adds both as purely additive passthroughs — split mode stays the default and no existing consumer changes behavior.

https://f0.factorial.dev/iframe.html?id=components-button-buttondropdown--dropdown-mode-example&viewMode=story

## Implementation details

- feat: add `disabledTooltip` to `ButtonDropdownItem` — type-only, since both split and dropdown modes already spread items into `DropdownInternal`, which renders it
- feat: add `mode`/`trigger` to `HeaderSecondaryDropdownAction` so a `ResourceHeader` secondary action can render as a single labeled dropdown trigger instead of a split button

  <details>
  <summary>Why a branched helper instead of spreading?</summary>

  `F0ButtonDropdownProps` is a discriminated union (`trigger` only exists with `mode: "dropdown"`), so `BaseHeader` uses a small shared `SecondaryDropdown` helper that branches on `action.mode` at both secondary-action render sites (mobile `lg`, desktop `md`). The two `primaryAction` dropdown sites are untouched.
  </details>

- fix: model `mode`/`trigger` on `HeaderSecondaryDropdownAction` as a discriminated union mirroring `F0ButtonDropdownProps`, so `trigger` without `mode: "dropdown"` is a compile error instead of a silent no-op
- fix: honor `mode`/`trigger` in `F0NotesTextEditor`'s header too — it re-exports the same secondary-action type, so its `SecondaryActionControl` now reuses the shared `SecondaryDropdown` instead of advertising a capability it dropped
- test: passthrough test for `disabledTooltip` on dropdown-mode items; `F0ResourceHeader` test asserting a dropdown-mode secondary action renders a single labeled trigger (2 responsive copies), opens the menu, and fires `onClick` with the clicked item
- test: add the new dropdown-mode + disabled-item variant to both components' Chromatic `Snapshot` stories
- docs: stories (`DropdownModeWithDisabledItem`, `WithDropdownModeSecondaryAction`) and MDX updates for both components


igorsafonov-gif (approved):
Works great. Aligned with F0.

PR #5610, opened 2026-09-21
fix(OneDataCollection): stop the selection banner overlapping group rows

lluis-pujalte:
## Description

In a grouped collection with `selectable` and `allPagesSelection`, selecting rows and scrolling made the bulk-selection banner and the sticky group row share the same band, leaving the group's label and summary values unreadable. Two separate defects produced that one symptom: sticky rows pinned themselves at a hardcoded offset that assumed a one-row header, and the banner painted itself with a near-transparent background. Both are fixed here. Reported as FCT-63930 against Headcount planning, but nothing about it is Planning-specific — it reproduces in any F0 data collection with those three features.

Before and after are reproducible from the new story — `Patterns/Data Collection/Grouping/Selectable Grouping With All Pages Selection`. Select all rows in the header, then scroll the body.

## Implementation details

- fix: pin sticky rows below the sticky header's measured height instead of a hardcoded `40`

  <details>
  <summary>Why?</summary>

  `TABLE_ROW_STICKY_TOP_OFFSET` is the height of a header holding a single row. Rendering the selection banner into the header adds a second row and takes it to 88px, so group rows pinned at 40 landed inside the banner's band. `TableHeader` now measures itself and publishes its height through `TableContext`; `TableRow` pins at that value. The constant remains as the fallback for a header that has not laid out yet and for a row rendered outside a measuring table.
  </details>

- fix: paint the selection banner on an opaque background

  <details>
  <summary>Why?</summary>

  `bg-f1-background-secondary` is `hsl(var(--neutral-10))`, a 6%-alpha tint — correct for tinting the page behind an element, wrong for a surface in the sticky header with rows scrolling underneath it, which read straight through. The opaque colour now sits underneath with the tint over it as a background-image layer, matching how a sticky row already paints its own.
  </details>

- fix: keep nested parent-row sentinel geometry aligned with the measured header height

  <details>
  <summary>Why?</summary>

  `useStickyParentRow` computed its sticky bottom from the same hardcoded 40, so an expandable parent row unsticked a banner's height too late. It now reads the measured value. Its `stickyTopOffset` option is removed — no caller passed it, and it could only desync the geometry from the `top` the row actually applies.
  </details>

- refactor: memoize the `TableContext` value so a header resize does not re-render every cell
- test: cover the resize path (the banner arrives after first paint), sub-pixel header heights, the context branch in `useStickyParentRow`, and the banner's opacity
- docs: add a story for grouping + `selectable` + `allPagesSelection` in a scrolling container, plus MDX notes


packages/react/src/experimental/OneTable/Table/index.tsx
```diff
@@ -45,10 +48,21 @@ function TableBase({ children, loading = false }: TableProps) {
...
+  const contextValue = useMemo(
+    () => ({
+      isScrolled,
+      setIsScrolled,
+      isScrolledRight,
+      setIsScrolledRight,
+      stickyHeaderHeight,
+      setStickyHeaderHeight,
+    }),
+    [isScrolled, isScrolledRight, stickyHeaderHeight]
```
albertpmz, line 65: The description says:

> memoize the TableContext value so a header resize does not re-render every cell

but here `stickyHeaderHeight` is in the dep array, which means that a heaeder resize will in fact make the table re-render.

packages/react/src/experimental/OneTable/TableHeader/index.tsx
```diff
@@ -1,14 +1,65 @@
...
+      // jsdom and a not-yet-laid-out header both report 0; leaving the default
+      // in place beats pinning every sticky row to the top of the container.
+      if (height > 0) {
+        setStickyHeaderHeight(height)
+      }
+    }
+
+    publishHeight()
+
+    const observer = new ResizeObserver(publishHeight)
```
albertpmz: `getBoundingClientRect().height` won't trigger a new measurement when only a transform changes (when resizing the header).

I'd suggest giving the observer its own callback which reads the entry:

```suggestion
    const observer = new ResizeObserver(([entry]) => {
  const height = entry.borderBoxSize[0]?.blockSize ?? 0
  if (height > 0) setStickyHeaderHeight(height)
})
```

albertpmz (changes requested):
Good job! Let's work on these two little issues

albertpmz (approved):
LGTM!

PR #5624, opened 2026-09-22
fix(F0AnalyticsDashboard): remove the chart's keyboard point menu

igorsafonov-gif:
## Description

Charts in `F0AnalyticsDashboard` carried two Ask One triggers. One is the action anchored to a clicked mark. The other was an invisible button parked in the widget's bottom-left corner, revealed by focus, that opened a menu of every data point for keyboard users.

Answering an ask handed focus back to that hidden button, so it appeared for readers who had only ever used the mouse: click a bar, choose **Ask One**, and a button they never asked for stayed on the chart until they clicked elsewhere. Dismissing the anchored action with <kbd>Escape</kbd> or by scrolling did the same.

This PR removes the keyboard point menu. One route per input is left: a pointer picks a single mark and gets the anchored action; everyone else asks about the widget through **Ask One** in its actions menu, which is keyboard-reachable and quotes the whole chart.

## Type of change

- [x] Bug fix

## Screenshots

Clicking a mark shows the anchored action — unchanged by this PR:

**Before** — the `Ask One from a chart point` story with a host `onAskAi` handler (Factorial's setup), right after choosing Ask One with the mouse. The keyboard trigger has taken focus and shown itself in the bottom-left corner:

**After** — same story, same handler, same clicks. The ask reaches the host and the chart stays clean:

And the same bug as reported in Factorial, which is where this started:

## Implementation details

- Deleted `AccessiblePointActions` and everything that only fed it: `buildAccessibleChartPoints`, `hasAccessibleChartPoint`, the `keyboardPointTriggerRef` focus restores in `ChartItem`, and the legend-selection subscription that existed to keep the menu aligned with visible series.
- `PointActionPopover.onDismiss` no longer reports a reason — the reason only existed to decide whether to move focus back to the removed trigger.
- `F0AnalyticsDashboardPointClick` is now an alias of `F0DataChartPointClick`, which narrows `source` from `"pointer" | "keyboard"` to `"pointer"`. This is what the API-surface bot flags, and it is intentional: a handler typed for the wider union still type-checks, and only a fixture that writes `source: "keyboard"` would break — Factorial main (f0-react 7.15.1, checked 2026-09-22) has none. The pattern is experimental, so no major bump.
- No code in f0 or Factorial branches on `point.source`; the F0DataChart PointClick story displays it and is unchanged.
- The two dashboard-level Ask One tests drove the keyboard menu because jsdom has no canvas. They now go through the pointer path: the mocked chart renders a plain button that reports a mark, which covers `onAskAi` and `onAskAiTarget` end to end.
- Regression test: after a host answers a point ask, no Ask One button exists in the chart. It fails on `main` because the keyboard trigger was always in the DOM, not because focus reveals it.
- Docs and the two point stories drop the keyboard steps, and the WHEN/THEN contract now states the keyboard route: the widget's ⋯ menu offers Ask One and quotes the whole chart.
- The two point stories lose their play functions: driving the pointer path needs a real canvas hit. Interaction coverage lives in the jsdom tests, which mock the chart with a button that reports a mark.
- Follow-up commit: the anchored action focuses its own button on open, so closing it used to drop focus to `<body>`. It now returns focus to the chart container (`tabIndex={-1}`, no ring, `preventScroll`), and only from `<body>`, so dismissing by clicking something else leaves that thing focused.
- New tests for the ⋯ menu ask on a chart — the keyboard route this PR leans on — which nothing covered for a chart item: the host handler gets the widget with no `point`, and the built-in chat quotes the whole chart.

## Testing

- `vitest run --project=unit src/patterns/F0AnalyticsDashboard` — 16 files, 186 tests, green (4 added: two for the ⋯ menu ask on a chart, two for where focus lands when the anchored action closes).
- `tsc --project ./tsconfig-build.json --noEmit` — clean.
- `oxlint --type-aware src/patterns/F0AnalyticsDashboard` — no new debt (10 pre-existing warnings, none in the touched files).
- Manually, in Storybook, on both sides of the change — the screenshots above.

## Follow-up

If a keyboard route to a single mark is wanted later, it should be designed as one deliberate surface rather than a hidden button that focus can reveal by accident.


zygisS22 (approved):
Approved for the intended scope: removing the keyboard point menu is deliberate. I verified the real chart click → anchored Ask One → formatted point quote → focused composer flow.

Two non-blocking follow-ups:

- **Focus after Escape:** the action returns focus to the chart wrapper, which has `tabIndex={-1}` and `outline-none`. In the browser it matches `:focus-visible` but has no visible indicator. Consider using a visible focus treatment or returning focus to an existing control. See [ChartItem.tsx:930–935](https://github.com/factorialco/f0/blob/6a8b121aaae609a10581bffe38421b0556305b00/packages/react/src/patterns/F0AnalyticsDashboard/components/ChartItem/ChartItem.tsx#L930-L935).
- **Browser interaction coverage:** `PointTargetObserver` and `ChartPointFlow` lose their play functions. The mocked chart-button tests cover callback handling, but do not replace real chart-to-composer coverage. Please add a pointer-driven browser regression for the retained flow when following up. See [AskOne.stories.tsx:242–261](https://github.com/factorialco/f0/blob/6a8b121aaae609a10581bffe38421b0556305b00/packages/react/src/patterns/F0AnalyticsDashboard/__stories__/AskOne.stories.tsx#L242-L261).

Validation at this head: 366 local tests passed, React typecheck passed, and the real pointer flow, keyboard whole-widget menu route, Escape dismissal, and preservation of composer focus on outside click were checked in local Storybook.

igorsafonov-gif:
Fixed both issues from @zygisS22 review

eliseo-juan (approved):
LGTM

zygisS22 (approved):
Re-checked at `341a2f520` after the two follow-ups from my earlier review. Both are addressed:

- **Focus after Escape** (`c8d292a20`): the chart container keeps `tabIndex={-1}` but now carries the shared `focusRing` (inset, no offset), so a keyboard dismissal shows where focus went, while a scroll or resize under the mouse, which doesn't match `:focus-visible`, stays quiet.
- **Browser coverage** (`341a2f520`): `PointTargetObserver` and `ChartPointFlow` have play functions again that press a real ECharts bar (`pressChartAt` sends the coordinates zrender hit-tests on), pick Ask One from the portalled anchored action, and assert the formatted quote, the raw target and composer focus. That covers the real chart-to-composer seam the mocked unit tests have to assume.

CI is green on this head (50 passed, 8 skipped), including all 8 Storybook test shards that run these plays. Approving.

